PR #5, opened 2015-04-30
Hold trait notification when setting multiple attributes from js

SylvainCorlay:
This is holding notifications when setting multiple trait attributes from the front-end

Currently, when doing something like the following in a widget view:

```JavaScript
this.model.set('x', x);
this.model.set('y', y);
this.touch();
```
trait assignments and notifications are made and sent in the (arbitrary) order in which the attributes are set from the bulk updates in the backend. If the first set is `'x'` and the change handler for `x` uses `'y'`, it would use the old value...

This should solve this issue using the new `hold_trait_notifications` context manager.

### DIFF
--- a/ipython_widgets/widgets/widget.py
+++ b/ipython_widgets/widgets/widget.py
@@ -282,13 +282,15 @@ class Widget(LoggingConfigurable):
 
     def set_state(self, sync_data):
         """Called when a state is received from the front-end."""
-        for name in self.keys:
-            if name in sync_data:
-                json_value = sync_data[name]
-                from_json = self.trait_metadata(name, 'from_json', self._trait_from_json)
-                with self._lock_property(name, json_value):
-                    setattr(self, name, from_json(json_value))
-    
+        with self.hold_trait_notifications():
+            for name in self.keys:
+                if name in sync_data:
+                    json_value = sync_data[name]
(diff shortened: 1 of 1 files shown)


jdfreder:
Does the same apply back-end to front-end?  i.e. if I use hold sync, set a couple traits in python, do the Javascript backbone model change events fire once all of the values are set, or just when one is set.  I'm not sure these behaviors should be inconsistent.

SylvainCorlay:
Wow, instead of replying, I started editing your comment (involuntarily). I think that it is once everything is set.

jasongrout:
This makes a lot of sense to do, as far as I'm concerned.

jasongrout:
@ssunkara1 found a subtle issue with this and the property lock in set_state.  Sylvain is looking into it.

SylvainCorlay:
@jasongrout I addressed your comments.

SylvainCorlay:
Hum Python 3 tests failing.

jasongrout:
This looks good to me.  I'll test it and merge it if it still looks okay.

jasongrout:
Actually, @SylvainCorlay, could you write a short test for this?  Maybe a sync message that sets two values equal, with a validation check that the two values are always equal?

jasongrout:
Thanks!

PR #6, opened 2015-04-30
Remove jquery calls in widget manager and base widget class

SylvainCorlay:
*Reopening the earlier PR on jupyter_notebook: removing dependency on jquery in Widget.js and manager.js (dumped the previous code there, will fix/rebase later.)*

I would like to eventually allow users to write widgets using [backbone native views](https://github.com/akre54/Backbone.NativeView) or [d3 views](https://github.com/akre54/Backbone.D3View) or other drop-in replacements for `Backbone.View` using the DOM manipulation library of their choice.

These replacements for backbone views share the main `View.el` attribute, which is a native DOM element. Hence it should be used to insert the widget view in the page generically rather than `View.$el` which is specific to jquery.

Besides inserting the widget in the page, jquery is only used for some basic css styling and `$.proxy`.

In this PR, I Widget.js I removed the jquery dependency in manager.js, and Widget.js. 

In the very end, I do:
```JavaScript
var WidgetView = Backbone.View.extend(WidgetInterface);
var DOMWidgetView = WidgetView.extend(DOMWidgetInterface);
```
which I would like to move to a different file. The idea, is that anyone could define their own version of DOMWidgetView based on NativeView:
 
```JavaScript
var NativeWidgetView = Backbone.NativeView.extend(WidgetInterface);
var NativeDOMWidgetView = NativeWidgetView.extend(DOMWidgetInterface);
```

### DIFF
--- a/ipython_widgets/static/widgets/js/widget.js
+++ b/ipython_widgets/static/widgets/js/widget.js
@@ -441,7 +441,7 @@ define(["nbextensions/widgets/widgets/js/manager",
     widgetmanager.WidgetManager.register_widget_model('WidgetModel', WidgetModel);
 
 
-    var WidgetInterface = {
+    var WidgetViewMixin = {
         initialize: function(parameters) {
             /**
              * Public constructor.
@@ -525,12 +525,12 @@ define(["nbextensions/widgets/widgets/js/manager",
     };
 
 
-    var DOMWidgetInterface = {
+    var DOMWidgetViewMixin = {
         initialize: function (parameters) {
             /**
              * Public constructor
              */
-            DOMWidgetView.__super__.initialize.apply(this, [parameters]);
+            WidgetViewMixin.initialize.apply(this, [parameters]);
             this.model.on('change:visible', this.update_visible, this);
             this.model.on('change:_css', this.update_css, this);
 
@@ -802,15 +802,20 @@ define(["nbextensions/widgets/widgets/js/manager",
         },
     });
 
-    var WidgetView = Backbone.View.extend(WidgetInterface);
-    var DOMWidgetView = WidgetView.extend(DOMWidgetInterface);
+    // For backwards compatibility.
+    var WidgetView = Backbone.View.extend(WidgetViewMixin);
+    var DOMWidgetView = WidgetView.extend(DOMWidgetViewMixin);
 
     var widget = {
         'unpack_models': unpack_models,
         'WidgetModel': WidgetModel,
+        'WidgetViewMixin': WidgetViewMixin,
+        'DOMWidgetViewMixin': DOMWidgetViewMixin,
+        'ViewList': ViewList,
+
+        // For backwards compatibility.
         'WidgetView': WidgetView,
         'DOMWidgetView': DOMWidgetView,
-        'ViewList': ViewList,
     };
 
     // For backwards compatability.
--- a/ipython_widgets/static/widgets/js/widget.js
+++ b/ipython_widgets/static/widgets/js/widget.js
@@ -807,8 +807,5 @@ define(["nbextensions/widgets/widgets/js/manager",
         'DOMWidgetView': DOMWidgetView,
     };
 
-    // For backwards compatability.
-    _.extend(IPython, widget);
-
     return widget;
 });
--- a/ipython_widgets/static/widgets/js/widget.js
+++ b/ipython_widgets/static/widgets/js/widget.js
@@ -4,10 +4,9 @@
 define(["nbextensions/widgets/widgets/js/manager",
         "underscore",
         "backbone",
-        "jquery",
         "base/js/utils",
         "base/js/namespace",
-], function(widgetmanager, _, Backbone, $, utils, IPython){
+], function(widgetmanager, _, Backbone, utils, IPython){
     "use strict";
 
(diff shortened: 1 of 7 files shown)


ipython_widgets/static/widgets/js/widget.js
```diff
@@ -443,7 +441,7 @@ define(["nbextensions/widgets/widgets/js/manager",
     widgetmanager.WidgetManager.register_widget_model('WidgetModel', WidgetModel);
 
 
-    var WidgetView = Backbone.View.extend({
+    var WidgetInterface = {
```
minrk: Do you mean to be changing the API and inheritance here? Why is that related to removing jquery?

ipython_widgets/static/widgets/js/widget.js
```diff
@@ -524,10 +522,10 @@ define(["nbextensions/widgets/widgets/js/manager",
             WidgetView.__super__.remove.apply(this, arguments);
             this.trigger('remove');
         }
-    });
+    };
 
 
-    var DOMWidgetView = WidgetView.extend({
```
jdfreder: Same here, this shouldn't be changed.

ipython_widgets/static/widgets/js/widget.js
SylvainCorlay: It is not only about removing jquery but allowing alternate drop-in replacements to Backbone.View like [D3View](https://github.com/akre54/Backbone.D3View) or [NativeView](https://github.com/akre54/Backbone.NativeView). 
I edited the description.

ipython_widgets/static/widgets/js/widget.js
SylvainCorlay: Arg, you beat me to it.

jdfreder:
I think you still want to define WidgetView. i.e.
```js
var WidgetViewMixin = { /* ... blah blah ... */ };
var WidgetView = Backbone.View.extend(WidgetViewMixin);
```

SylvainCorlay:
WidgetView is devined in the end.

jdfreder:
Ah sorry, I missed that! Since this it's not really an interface, maybe s/`WidgetInterface`/`WidgetViewMixin` ?  And the same with the DOM one.

SylvainCorlay:
OK, makes sense. 
Sorry I just dumped things here from the earlier PR that I did in jupyter_notebook. I need to do some testing before asking people to look at it thoroughly. I will add an example with a D3View.

ipython_widgets/static/widgets/js/widget.js
jdfreder: Sorry, see my comment below.  I thought you were missing the DOMWidgetView definition.

jdfreder:
Once you get this figured out, I see no problem with merging it for 4.0, as long as it doesn't introduce an backwards incompatibility.

ipython_widgets/static/widgets/js/widget.js
```diff
@@ -810,7 +814,7 @@ define(["nbextensions/widgets/widgets/js/manager",
     };
 
     // For backwards compatability.
-    $.extend(IPython, widget);
```
SylvainCorlay: By the way, @jdfreder is this also something that we could remove?

jdfreder: I **think** so.  That too has been deprecated since 3.0, but unlink the others, doesn't have a deprecation warning.  Once we started using requirejs is when that got placed there.

SylvainCorlay:
Got the tests to pass. I still had to use this.$el because `register_kernel_events` expects a JQuery selector. In the original PR in jupyter notebook, I had changed register_kernel_events. Hence to no use view.$el, I will need to open a PR in jupyter_notebook.

jdfreder:
That's unfortunate.  The `register_kernel_events` method in the jupyter notebook can just wrap the parameter in `$()`s because `$($(el))[0] === $(el)[0]` (the selector instances are probably different, but they point to the same element!).  This will allow that method to accept either.

SylvainCorlay:
Yeah, this is what I did in the original PR: https://github.com/jupyter/jupyter_notebook/pull/3/files

SylvainCorlay:
This should be looking better.

minrk:
@jdfreder want to have another look at this one?

ipython_widgets/static/widgets/js/widget.js
```diff
@@ -280,9 +279,9 @@ define(["nbextensions/widgets/widgets/js/manager",
             var return_value = WidgetModel.__super__.set.apply(this, arguments);
 
             // Backbone only remembers the diff of the most recent set()
-            // operation.  Calling set multiple times in a row results in a 
-            // loss of diff information.  Here we keep our own running diff.
-            this._buffered_state_diff = $.extend(this._buffered_state_diff, this.changedAttributes() || {});
+            // operation.  Calling set multiple times in a row results in a
+            // loss of diff information .Here we keep our own running diff.
```
jdfreder: lol, strange comment change here, accident?

SylvainCorlay: yes

minrk:
@jdfreder any last review to do, here?

jdfreder:
No, this is good to go!

minrk:
Thanks, guys!

PR #12, opened 2015-05-05
Trigger an event when a widget area is created

jasongrout:
This allows customization of the widget area and other actions when a widget area is created.


jdfreder:
Seems like a good idea to me, thanks!

PR #20, opened 2015-05-06
Use the less file in the less directory to generate css

### LINKED ISSUE #19: display: flex missing
It seems that a fluke in the less means that generated CSS is:

```css
.hbox,.vbox,.widget-area,.widget-area .widget-subarea,.widget-hbox,.widget-hslider,.widget-radio-box,.widget-vbox,.widget-vslider,.widget-vslider .ui-slider {
    display: -webkit-box
}
```
and the display for the widget-subarea is set to `-webkit-box`, rather than `flex`.

jasongrout:
This also seems to maybe fix #19.


jdfreder:
> maybe fix #19.

LOL, what does "maybe fix" mean?

jasongrout:
As in, I don't notice the problem anymore.

jdfreder:
I can verify that this solves the problem, thanks @jasongrout !

PR #23, opened 2015-05-06
Fix bugs in the accordion and tab widget styling example notebook.

### LINKED ISSUE #10: Bugs in tab/accordion widget examples in styling example notebook
~~Even though the examples are not in this repository (they are in notebook right now).~~

jdfreder:
closes #10 

Padding was missing from all of the contents in the tab widget.
The accordion widget needed a fixed width so it wouldn't change size when collapsed.


minrk:
This problem affects 3.x?

jdfreder:
Yes.

PR #29, opened 2015-05-08
disable clean-css restructuring

minrk:
this is a buggy optimization that breaks some css


jdfreder:
Interesting, I wonder if this was the root cause of the problem @jasongrout saw earlier with display: webkit-box .

minrk:
Quite possible. It seems to be super buggy for nontrivial CSS.

PR #34, opened 2015-05-14
Stab at making the tests less painful.

### LINKED ISSUE #15: Test framework improvements or total redesign.
The tests should
- always show the notebook Python output / error on failure
- javascript console output / error on failure
- test error on failure
- should use promises if written in Javascript
- should be clearer about the three contexts (kernel, browser, and tests)
- if written in javascript, ES6 template strings or something better should be used for embedding python
- should have an easy way to expose/export the test notebook, so the dev can interact with the tests to see why they are failing
- should have nice mechanisms for debugging

jdfreder:
Hi guys,

As you all may know, working with the Notebook front-end tests is a painful experience.  Widget authors are often unenthusiastic about writing regression tests because of this.  I decided to set out on an adventure to try to alleviate this pain.

I identified a few problems with out current system:

1. Tests are hard to write because of three contexts: tests, front-end, and kernel.
2. Tests are hard to write because of a miss mash of asynchronous Javascript schemes.
3. Tests are hard to write because of the lack of true multiline strings in vanilla ES5.
4. The tests alone, unmodified, are only useful for debugging 20% of the time (or less).
5. Tests are hard to reproduce in a live notebook.

The solutions in this PR:

1. The three contexts are clearly identified in the test output.  front-end and kernel contexts are enclosed in a large yellow box (it's pretty hard to miss).
2. All of our custom methods now use the same promise scheme that casperjs uses.  Understanding the flow of asynchronous logic is easier.
3. Typescript is used for template strings and additional syntactic sugar.
4. The cause of the failure is clearly identifiable, regardless of which of the three contexts it came from (because of 1.).
5. The usage of template strings allows one to copy and paste the python with little modification.

There are still a few more things I need to finish in this PR, like making sure the tests are all working again and general code clean-up.  I underestimated the effort required to do this, so I wanted to open a PR now for feedback!

## Candy

A failed test because of bad test code.  Note how the extra long comment in the Python code cell is wrapped in the box.
screen shot 2015-05-13 at 5 18 29 pm

A test that failed because of a kernel failure.
screen shot 2015-05-13 at 5 18 49 pm

A test that failed because of a front-end Javascript failure.
screen shot 2015-05-13 at 5 19 16 pm

Expect more details soon! : )

closes #15


jdfreder:
Also, I forgot to mention, this PR makes the tests independent of the notebook JS tests, which means after this is merged, a PR removing all the widget stuff from the notebook stuff can be made, further decoupling the two, woohoo!

**AND** minimal example of a new notebook test:
```js
/// <reference path="../typings/notebook_test.d.ts" />
import base = require('../base');
base.tester

// Test widget button class
.start_notebook_then()
.cell(`
    print('some stuff...')
    print('some more stuff...')
    `,
    function(index){
        // test some things
        // this.test.assert()
    }
)
.stop_notebook_then();

```

jdfreder:
(@ellisonbg here's what I was showing you last week, but quite a bit further along)

rgbkrk:
Oooh, nice.

minrk:
@jdfreder great! Do you think with improvements along these lines that it makes sense to stick with casper and keep improving our use of it, or do you think we should still be looking to abandon casper?

jdfreder:
@minrk for functional tests, casper is probably still our best bet.  I think a lot of potential exist for improvement, and we have the tools to do so (which this PR shows).  It wasn't a significant achievement to display outputs for the different contexts, all of the code it get the information already existed, it was just a matter spending some time to sort and cache it by cell.  I'd still like to try adding an export to notebook feature, the tricky part will be reading the casper.evaluate and thenEvaluate JS.

Of course, this work can propagate upstream once it's cleaned up a bit.

Effort to improve our casper test << effort to rewrite tests using a different framework.

minrk:
@jdfreder great. And what do you think about HTML/CSS tests? With casper, we can't test websockets properly, and the browser behavior is frequently problematic due to the use of phantom. Do you see us sticking with casper for a smaller amount of encapsulated tests and picking up something else for things that cover browser behavior? Or just wait for phantom 2 and call it close enough?

jdfreder:
Honestly, I think we *have* to try something else for high level tests - mouse clicks, keyboard input, HTML&CSS layout etc.  I just don't see casper or phantom solving the pseudo human input problems I've seen- pseudo input it detectably different, on the page, than human input.  I think a higher level, external to the browser, solution is needed for that.

I'm still troubled by the websocket problems though.  I feel like those types of tests belong in the regime of casper, but the fact that casper/phantom is technologically incapable of testing them is troublesome.  I'm inclined to wait to see if Phantom2 really fixes this.

So what I'm thinking now calls for an ASCII chart:

```
+----------------------------------------------------+
| Tests                                              |
|  +--------------------+   +--------------------+   |
|  | User level         |   | Functional         |   |
|  |  +-------------+   |   |  +-------------+   |   |
|  |  | Browser     |   |   |  | JS          |   |   |
|  |  | interaction |   |   |  +-------------+   |   |
|  |  +-------------+   |   |                    |   |
|  |                    |   |  +-------------+   |   |
|  |                    |   |  | Python      |   |   |
|  |                    |   |  +-------------+   |   |
|  +--------------------+   +--------------------+   |
+----------------------------------------------------+
```

minrk:
Phantom2 will definitely fix the binary websocket problem, since it's based on Qt5-webkit. It will inevitably fall well behind over time just like Qt4-webkit did, but it should be less archaic for a while, at least.

rgbkrk:
Not sure if people have adapted to using it this way, but I'll bet we could test using electron somewhat reliably. There's a high degree of control over the whole browser.

jdfreder:
If you guys are interesting in seeing what this looks like on Travis, [look at this](https://travis-ci.org/ipython/ipywidgets/jobs/62750139).

The only output that still isn't captured is the Python output from the **web server**, the kernel, front-end, and tests outputs are all visible.  Web server output may be helpful for 404s and other server errors that may not show as a front-end javascript error, does anyone have an opinion as to whether it's worth the effort to implement?

jasongrout:
@jdfreder: this is seriously awesome.  I'm really looking forward to this.

jasongrout:
Do you need to commit the typings files, or can we just say that tsd is a dependency of the testing system?

rgbkrk:
The typings files should probably be in the gitignore and that tsd is a dependency. /cc @Carreau

jdfreder:
> typings files should probably be in the gitignore

One of the typing files is handwritten and specific to IPython, but I can move that out of the "typings" directory and then gitignore the whole directory, making tsd a dependency like @jasongrout suggests.

jdfreder:
The last commit excludes tsd downloaded definitions, and downloads them on `gulp tests`.

jdfreder:
With the last commit, the tests are all working, but I still want to pull a lot of the methods that are currently being appended to the casper object into their own proper Typescript classes.  The base.ts file is larger than it needs to be, time to put it on a diet... :leaves:

ipywidget-travis:
The last commit adds a GH token associated with this account for Travis.

jdfreder:
This is ready to be reviewed and/or merged.

ipywidgets/tests/.tsdrc
```diff
@@ -0,0 +1,3 @@
+{
+	"token": "94e9b622a27d97abbd3775256c684f39b940a2c0"
```
minrk: what does this token do?

jdfreder: It increases the Github checkout rate limit from a low 60ish to 5000 or so per hour.  This is for TSD, alternatively I can check the .d.ts files in (typescript type definitions).

package.json
```diff
@@ -17,6 +16,9 @@
     "gulp-rename": "^1.2.2",
     "gulp-sourcemaps": "^1.5.1",
     "less": "~2",
```
minrk: need to remove this trailing comma for the JSON to be valid.

package.json
```diff
@@ -17,6 +16,9 @@
...
-    "source-map": "^0.4.2"
+    "source-map": "^0.4.2",
+    "typescript": "~1.5.0-beta",
+    "gulp-typescript": "^2.7",
+    "merge2": "~0.3",
```
minrk: Need to remove this comma for the JSON to be valid.

PR #36, opened 2015-05-27
Allow specifying a Python interpreter for gulp

SylvainCorlay:
I always work with multiple python installs.

The current gulp file for ipywidgets uses the Python interpreter to fetch the notebook static directory. It assumes that we use the python interpreter which is in PATH.

This change allows specifying which Python interpreter to use when running `gulp css`.


jdfreder:
Thanks @SylvainCorlay

PR #38, opened 2015-05-27
Hardcode widgets

jdfreder:
sibling PRs:
https://github.com/ipython/ipykernel/pull/19
https://github.com/jupyter/notebook/pull/120

### DIFF
--- a/ipywidgets/__init__.py
+++ b/ipywidgets/__init__.py
@@ -1,7 +1,11 @@
 from .widgets import *
 
 # Register a comm target for Javascript initialized widgets..
-from IPython import get_ipython
-ip = get_ipython()
-if ip is not None and hasattr(ip, 'kernel') and hasattr(ip.kernel, 'comm_manager'):
-    ip.kernel.comm_manager.register_target('ipython.widget', Widget.handle_comm_opened)
+def handle_kernel(kernel):
+    kernel.comm_manager.register_target('ipython.widget', Widget.handle_comm_opened)
+
+# Return the static assets path.
+def find_static_assets():
+    import os
+    path = os.path.abspath(__file__)
+    return os.path.join(os.path.dirname(path), 'static')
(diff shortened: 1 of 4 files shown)


jdfreder:
@minrk ah, better - from my repo.

SylvainCorlay:
@jdfreder is the goal to skip the nbextension installation stage?

jdfreder:
Yeah, the goal is just to make it pip install-able for 4.0.  The packaging discussion, targeted for 5.0, is here: https://github.com/jupyter/notebook/issues/116

ipywidgets/__init__.py
```diff
@@ -1,7 +1,11 @@
 from .widgets import *
 
 # Register a comm target for Javascript initialized widgets..
-from IPython import get_ipython
-ip = get_ipython()
-if ip is not None and hasattr(ip, 'kernel') and hasattr(ip.kernel, 'comm_manager'):
-    ip.kernel.comm_manager.register_target('ipython.widget', Widget.handle_comm_opened)
+def handle_kernel(kernel):
+    kernel.comm_manager.register_target('ipython.widget', Widget.handle_comm_opened)
```
minrk: I think relying on the shell calling handle_kernel should be a fallback. We should probably still have this logic at import, since we want it to be the way of the future, anyway.

So leave `handle_kernel`, but call also call it on import.

minrk:
@jdfreder can you finish this one?

jdfreder:
Yup

jdfreder:
So I know that the Javascript failure look unrelated, but it actually is.  Running registration on init causes the import to happen before the user actually imports ipywidgets, so the warning text is lost in a cell-less context.  I think, since we've split the repositories, it's probably okay to just remove the warning.  @minrk thoughts?

minrk:
I think it's okay to remove it.

jdfreder:
Ok, this is passing

SylvainCorlay:
:+1: to remove the warning.

PR #42, opened 2015-06-02
No promise required for set_state

SylvainCorlay:
Simply removing the unnecessary Promise in set_state.


jdfreder:
What happened here?

SylvainCorlay:
I will get back to this later. Now I understand why there is a promise. It is the empty `state_change` promise which guarantees order.

PR #43, opened 2015-06-05
add_trait -> add_traits

SylvainCorlay:
This reflects the change in traitlets https://github.com/ipython/traitlets/pull/27: `HasTraits.add_trait` is replaced with `HasTraits.add_traits`, which can add multiple traits at once. The benefit is that it  avoid creating too many types when adding multiple traits.


jdfreder:
The test failure was unrelated to this PR, but it's concerning because it looks like a race condition in the tests.  It's possible that I messed up some of the asynchronicity somewhere.

SylvainCorlay:
OK, seems to pass now.

SylvainCorlay:
Thanks

PR #45, opened 2015-06-10
Pass widget instance to trait serializer

SylvainCorlay:
Adding the widget instance to the signature of custom widget attribute [de]serializers. 
This is meant to enable things like 7e6bb27, where the way things are deserialized depend on the widget instance.

### DIFF
--- a/ipywidgets/tests/tests/widget.ts
+++ b/ipywidgets/tests/tests/widget.ts
@@ -210,7 +210,7 @@ base.tester
     from traitlets import Unicode, Instance, List
     from IPython.display import display
     from array import array
-    def _array_to_memoryview(x):
+    def _array_to_memoryview(obj, x):
         if x is None: return None
         try:
             y = memoryview(x)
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
@jdfreder

SylvainCorlay:
Rebased after #48 got merged.  
Since custom serialization was introduced post 3.0, there is not much incremental backward incompatibility there compared to 3.0. However, if this only gets in after 4.0, it will bring backward incompatibility between 4.0 and 5.0.

SylvainCorlay:
@jasongrout @jdfreder what do you think?

SylvainCorlay:
@jasongrout @jdfreder what do you think?

jasongrout:
At first glance, this seems okay to me.

SylvainCorlay:
Actually, if we pass `obj` as *second* argument, it will make easier to make widgets compatible with both IPython 3.* and 4.0.

SylvainCorlay:
I changed the order between the widget model and the attribute value arguments to make it
 - the same as on the js side
 - easier to create custom serializers that are compatible with both IPython 3.* and Jupyter/IPython 4.0.

jasongrout:
Your reasons make sense.  It'd be nice to have the self-as-first-arg pattern common in python, but you've convinced me the other way makes more sense now.

SylvainCorlay:
@jdfreder any chance we get this one in before 4.0?

jasongrout:
I'm ready to merge this.  @jdfreder - any last words before I merge?

PR #50, opened 2015-06-13
Fix serialization of javascript link

SylvainCorlay:
The attributes of the js link were not being serialized / deserialized correctly since the custom serialization PR.


jdfreder:
Thanks

PR #51, opened 2015-06-14
Read-only widget attributes

SylvainCorlay:
This enables read-only widget attributes, for widget attributes that are only set from the JavaScript side.

```Python
from ipywidgets import *
from traitlets import Int

class Foo(Widget):
    bar = Int(read_only=True)
    
foo = Foo()

foo.bar = 3  # Raises a TraitError
```

Although, one can still use the following context manager, which is used internally when setting state from the JavaScript side.
```Python
with foo._disable_read_only_ness():
    foo.bar = 3
```

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -289,7 +289,7 @@ class Widget(LoggingConfigurable):
         # The order of these context managers is important. Properties must
         # be locked when the hold_trait_notification context manager is
         # released and notifications are fired.
-        with self._disable_read_only_ness(),\
+        with self._allow_write(),\
              self._lock_property(**sync_data),\
              self.hold_trait_notifications():
             for name in sync_data:
@@ -366,7 +366,7 @@ class Widget(LoggingConfigurable):
             self._property_lock = {}
 
     @contextmanager
-    def _disable_read_only_ness(self):
+    def _allow_write(self):
         if self._read_only_enabled is False:
             yield
         else:
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -219,7 +219,8 @@ class Widget(LoggingConfigurable):
     def __setattr__(self, name, value):
         """Overload of HasTraits.__setattr__to handle read-only-ness of widget
         attributes """
-        if self._read_only_enabled and name in self.traits(read_only=True):
+        if (self._read_only_enabled and self.has_trait(name) and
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@jdfreder let me know what you think of this.

jdfreder:
I think this is a cool idea.  I'm a bit hesitant, in that 
1. I'm surprised there isn't already a read-only property of traits, i.e. config=true stuff may benifit from something like this.
2. It's unclear to me how this would be done in Javascript.  I definitely don't think that needs to be done here, but it would probably be in our best interest to do it in Javascript too eventually.

ipywidgets/widgets/widget.py
```diff
@@ -112,6 +112,7 @@ class Widget(LoggingConfigurable):
     # Class attributes
     #-------------------------------------------------------------------------
     _widget_construction_callback = None
+    _read_only_enabled = True
```
jdfreder: Hmm, what's the argument for this?  Seems redundant to me, like the `readonly` kwargs on the traits should be enough, no?

ipywidgets/widgets/widget.py
```diff
@@ -353,6 +366,17 @@ def _lock_property(self, **properties):
             self._property_lock = {}
 
     @contextmanager
+    def _disable_read_only_ness(self):
```
jdfreder: \*grumbles\* - don't like this name.

jdfreder: Maybe just `_disable_readonly`?

SylvainCorlay:
I need this for values that are set by the JavaScript side.

Besides, I guess that we are too close to 4.0 to ship this in traitlets, but it should be possible to move it to the traitlets in the future, in a backward compatible way, if it is useful there.

ipywidgets/widgets/widget.py
SylvainCorlay: you are the native speaker. I let you pick the one you prefer. `_allow_write`, `_chmod_777`, `_rw` :)

jdfreder: `_force_write` ?

jdfreder:
No, I don't think it's necessary to move it upstream, I was just surprised something like it didn't already exist.  It's okay here, for now.

SylvainCorlay:
Renamed to `_allow_write`.

jdfreder:
This is great, thanks @SylvainCorlay .

PR #54, opened 2015-06-16
Vertical progress bar

SylvainCorlay:
Adding an `orientation` attribute to the `FloatProgress` and `IntProgress` widgets.

### DIFF
--- a/ipywidgets/static/widgets/js/widget_int.js
+++ b/ipywidgets/static/widgets/js/widget_int.js
@@ -409,19 +409,20 @@ define([
             /**
              * Called when view is rendered.
              */
-            this.$el
-                .addClass('widget-hbox widget-progress');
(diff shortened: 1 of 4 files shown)


jdfreder:
This PR doesn't work for me.  After comparing against master, not forgetting to run `gulp css` and dump my cache, I realized that it's the contents here.  Although I wasn't quite sure why, the correct CSS isn't being applied.

SylvainCorlay:
Weird, I did check that everything works ok for me, recompiling the css to the master branch and back to this, with multiple browsers.

jdfreder:
I'll verify that all of my other repos are on master, and try again.

jdfreder:
Ah, I had two ipywidgets copies.  Deleted the other, and its conflicting files, and everything works.

PR #55, opened 2015-06-16
Decompose manager.create_model

SylvainCorlay:
When creating multiple widget models from the JavaScript side, I find myself chaining a lot of promises because `create_model` returns a promise, while I would only need to asynchronously load the widget module, and on resolution of this promise, synchronously create a list of models of that type.


jasongrout:
@SylvainCorlay says he doesn't need this functionality anymore, so I'm closing this.

PR #57, opened 2015-06-18
Undefined default value for the color trait

SylvainCorlay:
Since the latest improvement in traitlets default value validation by @takluyver , we need to set the default value of `Color` to `Undefined` for the following code to not raise a `TraitError`. 

Otherwise, the default value ` ''`  for the base Unicode trait type will be instantiated and validation will raise a trait error.

```Python
from ipywidgets import Widget, Color
from traitlets import List

class Foo(Widget):
    bar = List(trait=Color())

Foo()
```


jasongrout:
Looks good to me.

PR #58, opened 2015-06-19
Custom widget persistence

SylvainCorlay:
- Currently, there is the following bug in the widget persistence.

If `Foobar`is a custom widget that has a model defined in a requirejs module, specified in the `_model_module` widget attribute, then something like `HBox([Foobar()])` won't survive a page reload.
However, `Foobar()` alone does persist when it is not a child of a container.

In short, *custom widgets having custom models don't persist when they are children other widgets*.

The symptom in the js console is the `load_class` complaining that `undefined is not in the registry`.

- The reason is that that the creation of child widget views is done by the parent widget, at a time when their state, including `_view_name` and `_view_module` has not yet been set in the model.

Hence I changed the widget persistence mechanism, to first fetch the states of *all* the persisted models before instantiating any view.

Besides, the code is quite simpler, and it significantly reduces the line count.

- Finally, this should probably be backported to 3.x.

### DIFF
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -447,22 +447,17 @@ define([
             // Display all the views
             return all_models.then(function(models) {
                 var all_views = [];
-                var model;
-                for (var i = 0; i < models.length; i++) {
-                    model = Promise.resolve(models[i]);
-                    var views = model.then(function(model) {
-                        // Display the views of the model.
-                        var views = [];
-                        var model_views = state[model.id].views;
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@jasongrout @jdfreder

jasongrout:
FYI, the first three hunks and the last hunk are just whitespace changes.  Real changes are just lines 417-465.

jasongrout:
At a glance, this looks okay.  @jdfreder, since you wrote the original code, can you take a look at this too?  Thanks!

jdfreder:
Thanks @SylvainCorlay !

PR #60, opened 2015-06-20
Save changes should only save if comm is live

SylvainCorlay:
In a notebook, run 
```python
IntSlider()
```
then save the notebook and restart the kernel. The red broken link appears in the corner. However, when moving the slider, the back-end still receives messages and complains that "no such comm exists".

With this change, `save_changes` becomes a no-op in case of a dead comm.

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -114,7 +114,9 @@ define(["nbextensions/widgets/widgets/js/manager",
                 return;
             }
 
-            var msg_id = this.comm.send({method: 'request_state'}, callbacks || this.widget_manager.callbacks());
+            var msg_id = this.comm.send({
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense.

ipywidgets/static/widgets/js/widget.js
```diff
@@ -420,7 +430,9 @@ define(["nbextensions/widgets/widgets/js/manager",
              *
              * This invokes a Backbone.Sync.
              */
-            this.save(this._buffered_state_diff, {patch: true, callbacks: callbacks});
+            if (this.comm_live) {
+                this.save(this._buffered_state_diff, {patch: true, callbacks: callbacks});
+            }
```
SylvainCorlay: Sorry for the white space changes. This is the relevant change.

PR #61, opened 2015-06-22
Move calls to resolve_promise_dict

SylvainCorlay:
`model._deserialize_state` currently returns a dictionary containing values and promises (in the case of a deserializers). Hence, it is always used with `utils.resolve_promises_dict`. I just moved `resolve_promises_dict` to the function so that it always returns a promise.


SylvainCorlay:
@jasongrout

jasongrout:
Looks good to me.

PR #62, opened 2015-06-22
Widget persistence: only create comm for widgets having valid widget ids

SylvainCorlay:
This PR has two part:

 - it implements a workaround only based on comms to the absence of `comm_info_[request/reply]` message. The workaround works as follow: instead of sending a `comm_info_request`, the front-end opens a comm whose target is a widget that sends back the list of live comms and commits suicide.

*This initially required the following PRs related to the `comm_info_[request/reply]` messages: https://github.com/jupyter/jupyter_client/pull/34, https://github.com/ipython/ipykernel/pull/25, https://github.com/jupyter/notebook/pull/166. The code using this proper version is committed and commented out.*

- then it deals with persistence: 
    - comms are only created for valid ids
    - request_state is not called for invalid ids.
    - create_model is decomposed into the creation of the comm and new_model, which does not create a comm when it is not provided with one.

### DIFF
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -504,10 +504,28 @@ define([
         /**
          * Gets a promise for the open comms in the backend
          */
+
+        // Version using the comm_list_[request/reply] shell message.
+        /*var that = this;
         return new Promise(function(resolve, reject) {
              kernel.comm_info(function(msg) {
                  resolve(msg['content']['comms']);
              });
+        });*/
+
+        // Workaround for absence of comm_list_[request/reply] shell message.
+        // Create a new widget that gives the comm list and commits suicide.
+        var that = this;
+        return new Promise(function(resolve, reject) {
+            var comm = that.comm_manager.new_comm('ipython.widget',
+                                {'widget_class': 'ipywidgets.CommInfo'},
+                                'comm_info');
+            comm.on_msg(function(msg) {
+                var data = msg.content.data;
+                if (data.content && data.method === 'custom') {
+                    resolve(data.content.comms);
+                }
+            });
         });
     };
 
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -320,7 +320,7 @@ define([
          *      widget_class: (optional) string
          *          Target name of the widget in the back-end.
          *      comm: (optional) Comm
-         *
+         *          Comm of the widget. If not provided, a new Comm is created.
          * Create a comm if it wasn't provided.
          */
         var comm = options.comm;
@@ -446,7 +446,7 @@ define([
                         });
                     });
                 });
-            }, this));
+            }));
 
             // Display all the views
             return all_models.then(function(models) {
@@ -477,6 +477,17 @@ define([
                     resolve(data.kernel);
                 });
             }
+        });
+    };
+
+    WidgetManager.prototype._get_comm_info = function(kernel) {
+        /**
+         * Gets a promise for the open comms in the backend
+         */
+        return new Promise(function(resolve, reject) {
+             kernel.comm_info(function(msg) {
+                 resolve(msg['content']['comms']);
+             });
         });
     };
 
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -294,6 +294,22 @@ define([
 
     WidgetManager.prototype.create_model = function (options) {
         /**
+         * For backward compatibility. Custom widgets may be relying on the fact
+         * that create_model was creating a comm if none was provided in options.
+         * Unlike the old version of create_model, if no comm is passed,
+         * options.model_id is used to create the new comm.
+         */
+        console.warn('WidgetManager.create_model is deprecated. Use WidgetManager.new_model');
+        if (!options.comm) {
+            options.comm = this.comm_manager.new_comm('ipython.widget',
+                                                      {'widget_class': options.widget_class},
+                                                       options.model_id);
+        }
+        return this.new_model(options);
+    }
+
+    WidgetManager.prototype.new_model = function (options) {
+        /**
          * Create and return a promise for a new widget model
          *
          * Minimally, one must provide the model_name and widget_class
@@ -302,7 +318,7 @@ define([
          * Example
          * --------
          * JS:
-         * IPython.notebook.kernel.widget_manager.create_model({
+         * IPython.notebook.kernel.widget_manager.new_model({
          *      model_name: 'WidgetModel',
          *      widget_class: 'ipywidgets.IntSlider'
          *  })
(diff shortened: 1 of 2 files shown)


minrk:
I know the multi-PR nature of changes like this is annoying, but I think it appropriately reflects the significance of such changes, since there are many consumers of these interfaces that aren't us, now.

SylvainCorlay:
Until last week, widget persistence did for containers like `HBox([Button(), Button()])`, or for custom models. (cf #49 and #58), hence I am not too concerned about breaking backward compatibility of the persistence API.

jdfreder:
> The comm:dead event is not triggered anymore when creating a comm-less model. This event is meant to the case where the comm dies (like for a kernel restart).

Is the thinking here that a widget model starts comm-less, until a comm is connected?

SylvainCorlay:
> > The comm:dead event is not triggered anymore when creating a comm-less model. This event is meant to the case where the comm dies (like for a kernel restart).

> Is the thinking here that a widget model starts comm-less, until a comm is connected?

The idea is to have a version of `create_model` that takes a `model_id` but does not create a comm for it. In the persistence, widget models with model ids that are not in the comm list for the comm_list request will not get a comm.

SylvainCorlay:
- To me, this PR is more of a bug fix. The current behavior in case of reload, is that all models are created with the assumption that they have a valid comm, which is not true. Then after they are constructed, comm_live is forcibly set to `false`, and back to `true` after the state request... 

It creates a lot of bugs for things widgets that create children in their constructors, because thy recreate new widgets, then receive the old children from the backend state request etc...

The `comm_list_[request/reply]` thing is really required for widgets is really required to correctly implement persistence.

jdfreder:
I was thinking from an aesthetic stand point.  There will need to be a convenient way for someone to programmatically detect if the widget is connected to the back-end, for the purpose of the broken link icon that indicates a severed connection.

SylvainCorlay:
No worries, this feature will remain.

SylvainCorlay:
Besides, I think this should be in 4.0.

jdfreder:
4.0 is okay with me.  The other devs will probably have an opinion WRT to the other PRs too.

SylvainCorlay:
I think that Min was in favor of the new message. It is a minor bump in the protocol, and getting a list of open comms from the backend is really needed for anything that would use the comm API to be persistent.

SylvainCorlay:
Finished with the items listed.

ipywidgets/static/widgets/js/manager.js
```diff
@@ -418,30 +430,34 @@ define([
...
+                        return model._deserialize_state(state[model.id].state).then(function(state) {
+                            model.set_state(state);
+                            // Only request state for live comms
+                            if (live) {
+                                return model.request_state().then(function() {
+                                    return model;
+                                });
+                            } else {
+                               return model;
+                            }
```
SylvainCorlay: Regarding this section, one way to shorten would be to have `set_state` return the model, and `request_state` directly return the model if comm is dead, or return a promise for the model.

This would give 

`return model.set_state(state).request_state();`

SylvainCorlay:
@jdfreder I commited a workaround for the comm_list. I could open a PR with the first two commits, which are very simple to remove some noise.

SylvainCorlay:
I updated the workaround to replicate the new schema of the `comm_info_[request/reply]` message.

SylvainCorlay:
rebased.

SylvainCorlay:
I added the model `ready` event to this one. It only really makes sense with the persistence.

SylvainCorlay:
@jdfreder @jasongrout since I have a (not-so-nice) workaround for the absence of  `comm_info_[request/reply]`, would you guys like to have a look at this one?

ipywidgets/static/widgets/js/manager.js
```diff
@@ -306,6 +306,22 @@ define([
 
     WidgetManager.prototype.create_model = function (options) {
         /**
+         * For backward compatibility. Custom widgets may be relying on the fact
+         * that create_model was creating a comm if none was provided in options.
+         * Unlike the old version of create_model, if no comm is passed,
+         * options.model_id is used to create the new comm.
+         */
+        console.warn('WidgetManager.create_model is deprecated. User WidgetManager.new_model');
```
jdfreder: s/User/Use

ipywidgets/static/widgets/js/manager.js
```diff
@@ -306,6 +306,22 @@ define([
...
+        if (!options.comm) {
+            options.comm = this.comm_manager.new_comm('ipython.widget',
+                                                      {'widget_class': options.widget_class},
+                                                       options.model_id);
+        }
+        return this.new_model(options);
+    }
+
+    WidgetManager.prototype.new_model = function (options) {
```
jdfreder: Why create a new method?  Why not change `create_model`?

ipywidgets/static/widgets/js/manager.js
```diff
@@ -306,6 +306,22 @@ define([
 
     WidgetManager.prototype.create_model = function (options) {
         /**
+         * For backward compatibility. Custom widgets may be relying on the fact
+         * that create_model was creating a comm if none was provided in options.
+         * Unlike the old version of create_model, if no comm is passed,
```
jdfreder: Backwards incompatibility

SylvainCorlay: It is not backward incompatible because we could not pass model id in the options before.
The new create_model only *allows* setting the model id.

ipywidgets/static/widgets/js/manager.js
SylvainCorlay: because `create_model` creates a new comm when it is not passed one. Hence it does not allow comm-less models.

ipywidgets/static/widgets/js/manager.js
jdfreder: Ah gotcha, read too quick.  Saw "Unlike the old version of create_model"

ipywidgets/static/widgets/js/manager.js
```diff
@@ -332,19 +348,19 @@ define([
          *      widget_class: (optional) string
          *          Target name of the widget in the back-end.
          *      comm: (optional) Comm
+         *           Comm object associated with the widget.
+         *      model_id: (optional) string
+         *           If not provided, the comm id is used.
          *
-         * Create a comm if it wasn't provided.
+         * Either a comm or a model_id must be provided.
```
jdfreder: I'm not sure that it makes sense to require a comm or model id.  Thinking symmetrically, you don't need to specify a comm or model id when instantiating a widget from the backend.

If not in this method, it makes sense to have a method that can do this, for the case when the user just wants to make a new widget, and doesn't care about comms or model ids.

I do not object to doing this in another PR.

SylvainCorlay:
Fixed the typo you found.

ipywidgets/static/widgets/js/manager.js
jdfreder: Or just renaming and keeping `create_model`.  I do think having both `create_model` and `new_model` would be confusing.

SylvainCorlay: create_model does not need a comm or a comm_id.

`new_model` is meant for 
 - creating models for a given comm (in the context of persistence)
 - creating comm-less models (for dead comm widgets with broken links)
In the latter case, a model id is required.

SylvainCorlay:
ok, there are four situations:

1. Creation from the back-end: comm_open message handler *creates*, *registers* a comm. Pass it to the widget model constructor.

2. Creation from the front-end: *creates* , *registers*, and *opens* a new comm. Pass it to the widget model constructor. (uses widget_manager.comm_open for that)

3. Persistence, case of a live model:  *creates*  and *registers*, use the old id. Pass it to the widget model constructor.

4. Persistence, case of a dead:  Create a *comm-less* model, but pass the id to the widget model constructor.

Besides, we probably want to keep `create_model` for backward compatibility.

SylvainCorlay:
@jdfreder @jasongrout I rebased the PR and squashed a few commits.

SylvainCorlay:
One big issue IMO is that even when the new message gets in, we will need to keep using workarounds to be backward compatible with 4.0...

minrk:
This doesn't merge anymore. Can you rebase? @jdfreder once it's mergeable, can you have another review pass?

jdfreder:
Yes, @SylvainCorlay ping me when it's rebased.

SylvainCorlay:
@jdfreder rebased

jdfreder:
Thanks!

SylvainCorlay:
@jdfreder you were not sure about the names for new_model / create_model. If you have suggestions, let me know. create_model is the only one we must keep for backward compat.

jdfreder:
Yeah it still bothers me, but I figured we should get the PR in so the code doesn't go stale.

PR #48, opened 2015-06-11
Make set_state and get_state symmetrical

SylvainCorlay:
This is meant to solve #47.

PR #63, opened 2015-06-23
Simplify set_state

SylvainCorlay:
@jasongrout there is no need for this function to create a Promise.


jdfreder:
This is possible because of the changes you made in #48, right?  No unpacking logic happens here or in any of the subsequent function calls.

SylvainCorlay:
Yes, possibly. I am not sure why there was ever a promise here :)

jasongrout:
line 205 now doesn't need a return, then: https://github.com/ipython/ipywidgets/pull/63/files#diff-6b2b6bc22cc17b1c445538ce9510948fR205

SylvainCorlay:
Indeed, 6 more characters removed.

jasongrout:
This looks good to me now.

PR #64, opened 2015-06-23
No need for a counter in hold_sync

SylvainCorlay:
It is simpler to keep a `_holding_sync` boolean than counting the number of nested context managers.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -159,7 +159,7 @@ class Widget(LoggingConfigurable):
         return [name for name in self.traits(sync=True)]
     
     _property_lock = Dict()
-    _send_state_lock = Int(0)
+    _holding_sync = False
     _states_to_send = Set()
     _display_callbacks = Instance(CallbackDispatcher, ())
     _msg_callbacks = Instance(CallbackDispatcher, ())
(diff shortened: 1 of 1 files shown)


ipywidgets/widgets/widget.py
```diff
@@ -380,12 +380,14 @@ def hold_sync(self):
         """Hold syncing any state until the context manager is released"""
         # We increment a value so that this can be nested.  Syncing will happen when
```
jasongrout: This looks like a smart way to do this.  This comment should be updated, though.

ipywidgets/widgets/widget.py
```diff
@@ -378,14 +378,14 @@ def _allow_write(self):
     @contextmanager
     def hold_sync(self):
         """Hold syncing any state until the context manager is released"""
```
jasongrout: Can we change this description to "until the outermost context manager exits" to give some clue to the code below and more precisely document what is going on?

jasongrout:
This looks good to me if you change the docs just slightly, as noted above.

SylvainCorlay:
@jasongrout done.

jasongrout:
Looks good.  Thanks!

PR #66, opened 2015-06-24
Remove double call to base ctor

SylvainCorlay:
As per gitter discussion with @jdfreder , the base ctor was already called Line 51.


jdfreder:
Thanks @SylvainCorlay !

jasongrout:
ah, "ctor" == "constructor"

PR #67, opened 2015-06-25
No need of comm:dead event on widget creation

SylvainCorlay:
There is no need to call set_comm_live(false) which triggers an event if the comm was never live.


SylvainCorlay:
Merge!

jasongrout:
ping @jdfreder - do you want to look at this?

jdfreder:
Looks good, thanks @SylvainCorlay

PR #68, opened 2015-06-26
Display return value in interact

takluyver:
As discussed at the dev meeting on Tuesday, you need to write rather odd functions for interact because it doesn't display the return value. No-one at the time could think of a good reason why it shouldn't.

This will probably create some unexpected results for existing code, but we've said widgets are unstable.

Ping @fperez

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -218,6 +218,7 @@ def interactive(__interact_f, **kwargs):
             manual_button.disabled = True
         try:
             container.result = f(**container.kwargs)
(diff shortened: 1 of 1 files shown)


fperez:
Also ping @ellisonbg.

Thanks! Unless there's a very good reason for the old, odd behavior, this
looks great.

ellisonbg:
+1 on this.

minrk:
:+1: this just came up at Southampton, where people expected the returned object to be displayed.

fperez:
Happy camper, thanks!

PR #69, opened 2015-06-26
Zombie models, avoid memory leaks because of references in callbacks

SylvainCorlay:
#### `ListenTo` vs `on`
Backbonejs has two methods to register callbacks: `on` and `listenTo`. (There are also the `once` and `listenToOnce` conterparts).

- When one does `emitter.on('some_event', listener.foo, listener)`, the emitter has a reference to the listener, and unless we call `off` explicitly, it will keep it when the listener is destroyed, preventing the garbage collector to remove it completely.

- When one does `listener.listenTo(emitter, 'some_event', listener.foo, listener)`, the listener has a reference to the emitter. When the listener is destroyed, backbone calls `stopListening` and all these references disappear. 

Hence, in the case of a listener having a shorter lifespan than the emitter, like a backbone view with respect to its model, we should use `listenTo`. The models should not have to call `off` when a view dies.

#### The case of `jslink` and `jsdlink`
However, in the case of the jslink, the link listens to changes in two models that may or may not be destroyed first. 

- When the link is destroyed, linked models should not keep references to the link. Hence `ListenTo` is the right choice.
- However, one of the two  models dies, we should remove that reference the we have on it due to the previous listenTo, to allow it to be collected by the garbage collector. Thus, we must listen to the linked models' 'destroy' events.
- Furthermore, as soon as we receive a destroy event from one of the two linked model, we should also stop listening to it in order to not keep a reference to it.

### DIFF
--- a/ipywidgets/static/widgets/js/widget_link.js
+++ b/ipywidgets/static/widgets/js/widget_link.js
@@ -20,33 +20,38 @@ define([
             } finally {
                 this.updating = false;
             }
-        }
+        },
+
+        cleanup: function() {
+            // Unregister from change and destroy events of source and target.
+            if (this.source) {
+                this.source[0].off("change:" + this.source[1], null, this);
+                this.source[0].off("destroy", null, this);
+            }
+            if (this.target) {
+                this.target[0].off("change:" + this.target[1], null, this);
+                this.target[0].off("destroy", null, this);
+            }
+        },
+
     }, {
+
         serializers: _.extend({
             target: {deserialize: widget.unpack_models},
             source: {deserialize: widget.unpack_models},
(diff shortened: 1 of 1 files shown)


jasongrout:
When you call `listener.listenTo(emitter, 'some_event', listener.foo)`, the emitter also has a reference to the listener, right?  The only thing that happens is that the listener automatically calls `off` for you when it is destroyed.

SylvainCorlay:
That's right, but the control is given to the listenerin terms of public API. There is a common underlying implementation of `off`.

jasongrout:
This looks good.  I'll test it out.  @jdfreder - if you want to comment before I merge, now's the time.

PR #72, opened 2015-06-30
Use `listenTo` rather than `on` in widget views.

SylvainCorlay:
This should fix some views failing to be garbage collected.

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -460,13 +460,13 @@ define(["nbextensions/widgets/widgets/js/manager",
             /**
              * Public constructor.
              */
-            this.model.on('change',this.update,this);
+            this.listenTo(this.model, 'change', this.update, this);
 
             // Bubble the comm live events.
-            this.model.on('comm:live', function() {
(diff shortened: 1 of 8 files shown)


minrk:
Forgive my ignorance, but can you give a small statement about why this fixes it?

SylvainCorlay:
I am sorry I did not detail more. I gave some details in the previous PR on the jslink:

Backbonejs has two methods to register callbacks: `on` and `listenTo`. (There are also the `once` and `listenToOnce` conterparts).

- When one does `emitter.on('some_event', listener.foo, listener)`, the emitter has a reference to the listener, and unless we call `off` explicitly, it will keep it when the listener is destroyed, preventing the garbage collector to remove it completely.

- When we use `listener.listenTo(emitter, 'some_event', listener.foo, listener)`, the listener has a reference to the emitter. When the listener is destroyed, backbone calls `stopListening` and all these references are removed. 

Hence, in the case of a listener having a shorter lifespan than the emitter, like a backbone view with respect to its model, we should use `listenTo`. We should not need to call the `off` method on the model when a view dies.

minrk:
Thanks!

PR #74, opened 2015-07-01
Use a promise rather than after_displayed

SylvainCorlay:
We currently have a `WidgetView.after_displayed` method, which executes the specified callback right away is the view is already inserted in the DOM, or registers the callback to the `displayed` event otherwise.

This adds a new `displayed` Promise to the View, which resolves when the view is inserted into the DOM.

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -471,12 +471,17 @@ define(["nbextensions/widgets/widgets/js/manager",
             }, this);
 
             this.options = parameters.options;
-            this.on('displayed', function() { 
-                this.is_displayed = true; 
-            }, this);
+            /**
+             * this.displayed is a promise that resolves when the view is
+             * inserted in the DOM.
+             */
+            var that = this;
+            this.displayed = new Promise(function(resolve, reject) {
+                that.once('displayed', resolve);
+            });
         },
 
-        update: function(){
+        update: function() {
             /**
              * Triggered on model change.
              *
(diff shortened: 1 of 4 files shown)


ipywidgets/static/widgets/js/widget.js
```diff
@@ -524,7 +532,8 @@ define(["nbextensions/widgets/widgets/js/manager",
              * Calls the callback right away is the view is already displayed
              * otherwise, register the callback to the 'displayed' event.
              */
-            if (this.is_displayed) {
+            console.log('`WidgetView.after_displayed` is deprecated. Use the WidgetView.displayed promise instead.');
+            if (this._is_displayed) {
```
jasongrout: this should probably just use the Promise now: `this.displayed.then(function() {callback.apply(context)})`

ipywidgets/static/widgets/js/widget.js
```diff
@@ -471,12 +471,20 @@ define(["nbextensions/widgets/widgets/js/manager",
...
             this.options = parameters.options;
+            var that = this;
+            /*
+             * this._displayed is a promise that resolves when the view is
+             * inserted in the DOM.
+             */
+            this.displayed = new Promise(function(resolve, reject) {
+                that.once('displayed', resolve);
+            });
             this.on('displayed', function() {
```
jasongrout: Can we use the Promise here? `this.displayed.then(function() {that._is_displayed = true;})`

SylvainCorlay:
Comments addressed.

jasongrout:
Looks good reading through it.  I'll give a whirl too.  @jdfreder - speak now, or hold your peace unless you have something to say, in which case you should say it anyway.

jdfreder:
Looks good!

PR #79, opened 2015-07-08
Fix css ui-slider

### LINKED ISSUE #77: Css issue with the slider.
The `background : none` css property of the `ui-slider` div is not being applied (see [line 118](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/static/widgets/less/widgets.less#L118) in widgets.less).
This causes this div, which is supposed to be invisible to partly hide the slider track.

The reason is that the css from hbox is overriding this property.

SylvainCorlay:
Fixes #77. (Order matters in less.)

Before,
before-slider

After,
after-slider

### DIFF
--- a/ipywidgets/static/widgets/less/widgets.less
+++ b/ipywidgets/static/widgets/less/widgets.less
@@ -114,11 +114,11 @@
 
     .ui-slider {
         /* Inner, invisible slide div */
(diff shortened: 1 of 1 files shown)


blink1073:
Go to bed Sylvain. 



—

PR #80, opened 2015-07-08
Unregister all instances of the callback

SylvainCorlay:
When setting `remove` to True, all the instances of the callback get unregistered.


ipywidgets/widgets/widget.py
```diff
@@ -77,7 +80,7 @@ def register_callback(self, callback, remove=False):
         
         # (Un)Register the callback.
         if remove and callback in self.callbacks:
-            self.callbacks.remove(callback)
+            self.callbacks = [cb for cb in self.callbacks if cb!=callback]
```
minrk: Maybe `is not`? Equality seems iffy for functions, even though it works.

PR #82, opened 2015-07-08
Unregister all instances of the callback

SylvainCorlay:
*WIP* This is a new version of #80, which for some reason I cannot reopen... I took into account the comment of @minrk.

The goal was to get callback_dispatcher to behave like phosphor signaling in that you can register a callback (connect a slot to a signal) multiple times, but when un-registering a callback (disconnecting a slot), all instances get disconnected even if it was connected multiple times.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -68,6 +68,9 @@ class CallbackDispatcher(LoggingConfigurable):
     def register_callback(self, callback, remove=False):
         """(Un)Register a callback
 
+        If remove is True, this will completely unregister the callback, even
+        if it was registered multiple times.
+
         Parameters
         ----------
         callback: method handle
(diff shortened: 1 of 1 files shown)


ipywidgets/widgets/widget.py
```diff
@@ -77,8 +80,8 @@ def register_callback(self, callback, remove=False):
         
         # (Un)Register the callback.
         if remove and callback in self.callbacks:
-            self.callbacks.remove(callback)
-        elif not remove and callback not in self.callbacks:
```
jdfreder: I think, like we talked about in person, I'd rather keep this bit- which I think you said you agreed.

SylvainCorlay: Absolutely.

SylvainCorlay:
Will come back to this later.

PR #83, opened 2015-07-08
change default msg_throttle to 1

minrk:
Even for simple cases like plotting, having more than one outstanding update seems like too much, so let's give 1 a try.


jdfreder:
Strange, you must be working with an old master.  Rebase?

jdfreder:
: ( Didn't realize we have the constant specified in two different places, yuck.

minrk:
rebased.

PR #84, opened 2015-07-11
Push state in comm open msg

jdfreder:


minrk:
js tests appear to be failing.

jdfreder:
Hmmm strange, I'm really not sure why this is failing.  I'll debug locally.

jdfreder:
Ahhh forgot `sync=True` on the model information traitlets!  Should be good now!

ipywidgets/static/widgets/js/manager.js
```diff
@@ -280,11 +280,26 @@ define([
...
          * Handle when a comm is opened.
          */
-        return this.create_model({
-            model_name: msg.content.data.model_name,
-            model_module: msg.content.data.model_module,
-            comm: comm,
-        }).catch(utils.reject("Couldn't create a model.", true));
+        // For < 4.x
+        var model;
+        if (msg.content.data.model_name) {
```
SylvainCorlay: In which case could this happen? Is this to handle an older version of the Python side?

SylvainCorlay:
@jdfreder this seems to be breaking a simple example such as 

```Python
from ipywidgets import *
HBox([Button()])
```

SylvainCorlay:
Fixed in #86.

ipywidgets/static/widgets/js/manager.js
jdfreder: Yes

SylvainCorlay: How do you see this occurring ?

shashi: this did break Interact.jl fwiw. probably good to have a deprecation for such things before breaking changes.

jdfreder: I think we broke it in a later commit, sorry @shashi !  @SylvainCorlay and I will try to be better about things like this.  It was removed in https://github.com/ipython/ipywidgets/commit/474f08cbcdc47a573297447af8474ec0333c9687

shashi: No worries :) that will be great!

SylvainCorlay: It is means that we would need to support different versions of the back-end in the frontend of ipywidgets. I don't think it is the way to go in the long run... It brings again the question of packaging of nbextensions that depend on kernel code: have one version of the js for each kernel.

shashi: About different versions of the back-end in the front end: You definitely don't need the code to sit around for all releases. I think deprecating (printing a message to the console) in the upcoming release and then removing the thing in the next release is good. At least, even a changelog / NEWS entry will be very helpful.

SylvainCorlay: The reasoning is the following: there are two types of nbextensions, the ones that are independent of the kernel, like some nbconvert button in the toolbar, and the ones that have a piece of code on the kernel side.
The latter is more complicated because you may have multiple kernel registered having different versions of the same nbextension like ipywidgets.

I think that for each registered kernel, we should have the Javascript in a different directory, (one per installed kernelspec). The requirejs path would be automatically prefixed with a string corresponding to the current kernel.

SylvainCorlay: Another benefit would be to enable testing two different implementations of nbextensions, including the front-end in the same notebook server by simply switching kernels.

shashi: I don't quite understand the whole process but that sounds like an awful lot of things to support. But yeah, if you want to do it, I won't say no :) I'd be happy if I had the good stack traces back, though.

jdfreder: > I think that for each registered kernel, we should have the Javascript in a different directory, (one per installed kernelspec). The requirejs path would be automatically prefixed with a string corresponding to the current kernel.

I think this is a good idea and I'm +1 for it - it should be addressed with the other packaging problems https://jupyter.hackpad.com/Packaging-crate-PbIgxnC71or .  @shashi with the  feature @SylvainCorlay is describing, packaging would be more like npm.  You, in the Julia widgets, would depend on a specific widget JS version X.  That would allow us to continue chugging along without affecting you, and when you were ready to upgrade, you could.

@SylvainCorlay despite that, I still think for things like this, which are "widget msg spec" changing, we need deprecation warnings.  Julia (@shashi) and Haskell (@sumitsahrawat) implementations both exist now, and we should do our best to help them transition to our latest Javascript.

jdfreder: I know deprecation warnings can be a PITA to keep track of, but I think they are much less of a PITA for us than it is for our friends to migrate without them.

shashi: Nice! Great to hear!

SylvainCorlay: @jdfreder thanks, I will look at the hackpad.

PR #86, opened 2015-07-13
Deserialize initial state

SylvainCorlay:
We must use custom de serializers for initial state push.


minrk:
Ran tests locally, and the failures are fixed by #87, unrelated to this PR.

jdfreder:
Whoops!  Thanks!

PR #87, opened 2015-07-13
add theme.less

minrk:
fixes css builds due to missing variables


minrk:
Merging since this is blocking other tests from making sense.

jdfreder:
Thanks!

PR #88, opened 2015-07-13
fix css widget area

SylvainCorlay:
`.vbox()` has to be called last to be applied correctly, otherwise `.box-flex2` overrides it.


SylvainCorlay:
With @jasongrout  it seems the difference comes from a wrong? behavior of the minifier

Here is the css generated with the old (master) minifier for widget-area
```css
.widget-area .widget-subarea {
    padding: .44em .4em .4em 1px;
    margin-left: 6px;
    box-sizing: border-box;
    -moz-box-sizing: border-box;
    -webkit-box-sizing: border-box;
    display: -webkit-box;
    display: -moz-box;
    display: box;
    box-align: stretch;
    flex-direction: column;
    align-items: stretch;
    -webkit-box-flex: 2;
    -moz-box-flex: 2;
    box-flex: 2;
    flex: 2;
    -webkit-box-align: start;
    -moz-box-align: start;
    box-align: start;
    align-items: flex-start
}
```

display: flex is missing.

SylvainCorlay:
Modified the gulpfile to to disable advanced optimizations - selector & property merging, reduction, etc in the minification of css.

jasongrout:
Disabling the advanced optimizations fixes the issue (apparently the minifier was incorrectly getting rid of `display: flex`, so things reverted back to `display:box`).  

However, Sylvain's change also changes the widget-subarea div to have align-items to be `stretch` instead of `flex-start`.  I agree with that change.  @jdfreder, what do you think of changing widget-subarea to be `align-items: stretch` instead of `align-items: flex-start`?

SylvainCorlay:
This is due to the order in which css properties are overridden - and I believe the new version is the desired behavior for widgets of variable width. The widget subarea should behave like any vbox.

jasongrout:
I'll merge tomorrow unless @jdfreder wants to discuss this further.

jdfreder:
Two questions:
Does align-stretch make the width/height css ignored?
How does interact render with this change?

SylvainCorlay:
> Two questions:
> Does align-stretch make the width/height css ignored?

No. But if these are not set, it is going to stretch to the max width.

> How does interact render with this change?

I don't think it should change anything since core widgets have fixed sizes, but I have not made advanced use of interact.

jdfreder:
Great!

PR #81, opened 2015-07-08
Run npm in pip install, remove gulp

jdfreder:
@minrk I think this is what you were asking for earlier..  Mostly stolen from NB.

PR #91, opened 2015-07-15
revert theme.less addition

SylvainCorlay:
Since the file was removed in notebook.

PR #93, opened 2015-07-16
remove gulp, trigger js/css in setup.py

minrk:
rebase, finish #81


jdfreder:
So the test failure I was seeing earlier with the missing CSS... Magically gone?  Must have been related to another bug.

jdfreder:
Ah, must have been
#87 and #91 (and corresponding notebook PRs)

minrk:
Yup, I think it didn't have anything to do with your PR, but the real error was getting hidden somewhere.

SylvainCorlay:
The README should also be updated.

jdfreder:
I'm a bit confused.  I don't know why this passed the tests without https://github.com/jupyter/notebook/pull/120 .  I don't see where `python -m ipywidgets.install` is being called, therefor it should work without https://github.com/jupyter/notebook/pull/120 because the static assets are unavailbable to the front-end.

PR #95, opened 2015-07-20
Use instances of traits instead of trait classes

jasongrout:
See ipython/traitlets#51

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -157,7 +157,7 @@ class FloatProgress(_BoundedFloat):
 
 
 class _FloatRange(_Float):
-    value = Tuple(CFloat, CFloat, default_value=(0.0, 1.0), help="Tuple of (lower, upper) bounds", sync=True)
+    value = Tuple(CFloat(), CFloat(), default_value=(0.0, 1.0), help="Tuple of (lower, upper) bounds", sync=True)
     lower = CFloat(0.0, help="Lower bound", sync=False)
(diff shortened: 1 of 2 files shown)


jasongrout:
I'm not convinced this is a real error.  I don't see a traceback, just a timeout.  Can someone kick this (or show me how to kick it, since I have permissions to this repo now)?

minrk:
Even if the failure is real, it's likely not due to this PR. Will check locally.

jdfreder:
@jasongrout you should be able to kick it yourself.  All you have to do is authenticate using your GitHub with Travis.  Then a restart icon will appear in the top right that you can click.

jdfreder:
The last PR that was merged was passing the tests - https://github.com/ipython/ipywidgets/pull/93 .  Something must have changed elsewhere in the IPython/Jupyter world that's causing this to break..

jasongrout:
I'm testing Travis by reverting my changes, so that this PR is the same as master.

jdfreder:
My guess is it will fail, see my PR that changes the README #97

jasongrout:
Yep, it did.

jasongrout:
I deleted my revert, so now this PR is back to the original commit.  I tested it manually with the intrange slider.  I'm fine with merging this (but as a matter of policy, I shouldn't merge it :).

jdfreder:
Looks to me like the widget JS isn't getting loaded.

jdfreder:
Ah yes, I think it's because https://github.com/jupyter/notebook/pull/120 hasn't been merged.  Although.. it's concerning that #93 passed.

PR #98, opened 2015-07-21
Install nbextension in setup.py

jdfreder:
Same concept as jupyter-pip

Alternative to jupyter/notebook#120


jdfreder:
Hmm strange... I thought this would have fixed the tests. :question: :question: :confused: :question: :question:

jdfreder:
Ah the test that's failing is the one that checks if the convenience object is defined...  Maybe that's been the problem all along.

PR #102, opened 2015-07-23
Adapt to upstream read-only trait attributes

SylvainCorlay:
Since upstream traitlets now have read-only attributes, I remove the potentially conflicting widget-level read-only attributes.


SylvainCorlay:
@minrk, since ipywidget already have `read-only` trait metadata attribute, it will break with traitlets 4.1. That is why I marked this as 4.0.

minrk:
Ah, my understanding was backwards - I thought this needed to wait for a traitlets release, not the other way around.

PR #104, opened 2015-07-28
prepare for release

minrk:
- add js-releated files to package_data, manifest
- `npm run build` as part of build
- skip npm when run from sdists
- build_css should fail when its suprocesses fail
- fetch less dependencies for building CSS (rather than extracting from the notebook, which doesn't install LESS sources)

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory": "less_include/components"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ docs/gh-pages
 ipywidgets/static/components
 ipywidgets/static/widgets/css
 ipywidgets/tests/bin
+less_include
 node_modules
 *.py[co]
 __pycache__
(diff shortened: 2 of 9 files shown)


jdfreder:
Looking at the notebook LESS fetching step, yikes!  Makes me think `ipywidgets` should not depend on notebook less or that the notebook less should be available via bower and/or npm.

minrk:
@jdfreder I absolutely agree that it shouldn't depend on the notebook LESS, and started going down that road, but there were a few things that got in the way. I can draft an example of that, if you want to see it.

The other option is the make the notebook include its less sources in an install, which is perhaps the simplest of all.

minrk:
#105 is the same, but instead of fetching the notebook LESS, it copies the flexbox mixins, and stops using the few variables it was using. I think it's probably the way to go, now that I see it.

jdfreder:
Ah yes, thanks.  I just saw that.  I agree, we can close this one in favor of that.

PR #105, opened 2015-07-29
prepare for release

minrk:
- add js-related files to package_data, manifest
- `npm run build` as part of build
- skip npm when run from sdists
- build_css should fail when its subprocesses fail
- fetch less dependencies for building CSS (bootstrap, font-awesome)
- remove dependency on notebook LESS by shipping flexbox mixins,
  and removing the few variables that were being used.

alternative to #104

main difference: instead of fetching notebook LESS, ship subset we actually use (a few flexbox mixins).

This may result in some divergence, since variables and bower configuration are not shared,
but it isolates the widget styling and is much simpler.

cc @jdfreder

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory": "less_include/components"
+}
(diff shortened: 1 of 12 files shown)


jdfreder:
I prefer this :)

jdfreder:
Looking good.  Is there anything else you wanted to do here?

minrk:
No, I think this is a good start, at least.

PR #76, opened 2015-07-01
Proxy widget

SylvainCorlay:
The `Proxy` widget is a simple container holding a DOMWidget or None. 

It is generally used as a base class handling the life time of a child widget which we mostly use for layout and positioning.

The commit authored by @jasongrout is a simple specialization of the Proxy widget. The `Place` widget allows you to render the widget at any place specified with a css selector. This is especially useful for rendering things in the middle of rendered Markdown.

**Example:**
In Markdown
```Markdown
Some slider next to a rich text <div class='foobar'> </div>
```
In Python
```Python
Place(IntSlider(), selector='.foobar')
```

Other specializations of the Proxy to come in different PRs.

PR #109, opened 2015-07-30
Proxy

SylvainCorlay:
@jasongrout #76


jasongrout:
#76 had an inadvertant file committed, so we reset master, and this remerges that PR without the file.

PR #111, opened 2015-07-31
The `request_state` promise now resolves to the model itself.

SylvainCorlay:
This simplifies the client code since we were always doing 

```Python
return model.request_state().then(function() {
    return model;		
});
```

This becomes:

```Python
return model.request_state();
```

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -115,13 +115,14 @@ define(["nbextensions/widgets/widgets/js/manager",
                 method: 'request_state'
             }, callbacks || this.widget_manager.callbacks());
 
-            // Promise that is resolved when a state is received
+            // Promise that resolves to the model when the state is received
             // from the back-end.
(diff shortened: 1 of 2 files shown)


ipywidgets/static/widgets/js/widget.js
```diff
@@ -204,7 +203,7 @@ define(["nbextensions/widgets/widgets/js/manager",
                         .then(function() {
                             var parent_id = msg.parent_header.msg_id;
                             if (that._resolve_received_state[parent_id] !== undefined) {
-                                that._resolve_received_state[parent_id].call();
+                                that._resolve_received_state[parent_id](that);
```
jdfreder: Why pass `that` in here?

jdfreder: Ah nvm, I see

jdfreder:
Tested against your gamepad widget, and everything seems to work as expected! Thanks @SylvainCorlay

PR #124, opened 2015-08-03
Fix `.visible` resetting CSS display.

jdfreder:


ipywidgets/static/widgets/js/widget.js
```diff
@@ -642,12 +642,17 @@ define(["nbextensions/widgets/widgets/js/manager",
...
+                    }
                     this.el.style.visibility = 'hidden'; break;
                 case false:
+                    this._old_display = this.el.style.display;
                     this.el.style.display = 'none'; break;
                 case true:
-                    this.el.style.display = '';
+                    if (this._old_display) {
+                        this.el.style.display = this._old_display;
+                    }
```
SylvainCorlay: maybe we should set it to 'null' if `_old_display` is undefined?

jdfreder: No, I don't think so.  display is only modified when `_old_display` gets set, so there's no point in modifying it, since the default is to display the element and any overrides are user explicit.

SylvainCorlay: If display was inherited to start with, and you hide it, would _old_display be set to 'block'?

jdfreder:
Sweet, thanks

PR #129, opened 2015-08-03
Remove link to ipynb

SylvainCorlay:
Use a code block instead of a link to the example notebooks.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,4 +14,7 @@ Interactive HTML widgets for Jupyter notebooks and the IPython kernel.
 Usage
 =====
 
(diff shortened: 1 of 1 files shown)


jdfreder:
I think this is okay for now, until we figure out a better way to link to the correct version of the examples, without the version being embedded in version.rst.  Maybe it can be done in a release script...

minrk:
The contents of PyPI description are populated by the `long_description` field in setup. This can be templated in setup.py to always use the correct version on release, without any hardcoded values anywhere or release scripting.

PR #130, opened 2015-08-03
Use README.rst in manifest

jdfreder:


minrk:
Wait, why are there two READMEs now?

minrk:
I'm pretty sure this isn't the right thing to do - MANIFEST determines what files are included in a source release, which shouldn't exclude the README. I presume only one of README.md or README.rst should exist at all.

SylvainCorlay:
@minrk we remove the rst and included a long description in setup.py.

jasongrout:
See #132, where we did what Sylvain says.

minrk:
Thanks!

PR #132, opened 2015-08-04
Make long description part of setup.py

SylvainCorlay:
So that we can add templating in the long description.
@minrk


jdfreder:
This gets picked up by PyPI ?

jdfreder:
- just read @minrk 's comments on the other PR, answered my question.

SylvainCorlay:
I think it would, but it is difficult to check without a release...

jasongrout:
This sort of thing gets picked up in my pythreejs library: https://github.com/jovyan/pythreejs/blob/master/setup.py#L8

See the PyPi page: https://pypi.python.org/pypi/pythreejs/

minrk:
@jdfreder thanks. If it gets long enough that people would prefer to have .rst highlighting on it, a `pypi_description.rst` that gets read in setup.py would be fine, too.

PR #139, opened 2015-08-07
WIP

SylvainCorlay:


jdfreder:
A mysterious PR appears

SylvainCorlay:
Sorry, I had more there but I reset to the first commit because most of it had to be canned anyway.

PR #144, opened 2015-08-13
Call correct superclass in AccordionView

mwcraig:
I assume `AccordionView` is supposed to call its `__super__`, not that of `TabView`, on `update`.

I'm not sure this actually causes any issues, since `TabView` and `AccordionView` both subclass the same thing, but noticed it when I tried cutting out the `TabView` part of the code to backport accordion changes for reducer.


jdfreder:
Thanks again @mwcraig !

PR #147, opened 2015-08-20
Add __version__ to ipywidgets package

jorisvandenbossche:
Just noticed that `ipywidgets__version__` does not work. Has this a particular reason?

So added it to `__init__.py` like this is also done in `ipykernel`


jasongrout:
Thanks!

PR #149, opened 2015-08-23
Add an on_change callback to Selection widgets

blink1073:
This is similar to the `on_submit` callback for `Text` widgets.

screen shot 2015-08-23 at 11 09 28 am

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -9,7 +9,7 @@ Represents an enumeration using a widget.
 from collections import OrderedDict
 from threading import Lock
 
-from .widget import DOMWidget, register
+from .widget import DOMWidget, register, CallbackDispatcher
 from traitlets import (
     Unicode, Bool, Any, Dict, TraitError, CaselessStrEnum, Tuple, List
 )
@@ -50,6 +50,7 @@ class _Selection(DOMWidget):
     def __init__(self, *args, **kwargs):
         self.value_lock = Lock()
         self.options_lock = Lock()
+        self._onchange_callbacks = CallbackDispatcher()
         self.equals = kwargs.pop('equals', lambda x, y: x == y)
         self.on_trait_change(self._options_readonly_changed, ['_options_dict', '_options_labels', '_options_values', '_options'])
         if 'options' in kwargs:
@@ -73,6 +74,19 @@ class _Selection(DOMWidget):
         
         # Value is already in the correct format.
         return x
+
+    def on_change(self, callback, remove=False):
+        """(Un)Register a callback to handle selection change.
+
+        Triggered when the selection changes.
+
+        Parameters
+        ----------
+        callback: callable
+            Will be called with exactly one argument: the Widget instance
+        remove: bool (optional)
+            Whether to unregister the callback"""
+        self._onchange_callbacks.register_callback(callback, remove=remove)
 
     def _options_changed(self, name, old, new):
         """Handles when the options tuple has been changed.
@@ -104,6 +118,7 @@ class _Selection(DOMWidget):
 
     def _value_changed(self, name, old, new):
         """Called when value has been changed"""
+        self._onchange_callbacks(self)
         if self.value_lock.acquire(False):
             try:
                 # Reverse dictionary lookup for the value name


jasongrout:
Can't you trigger the callback from the change in the value attribute?

```python
from ipywidgets import Dropdown
def callback(trait, value):
    print("***changed", value)
dd = Dropdown(value='hello', options=['hello', 'goodbye'])
dd.on_trait_change(callback, 'value')
display(dd)
```

blink1073:
I tried that, but I wanted a closure around the name of the variable being changed.  I used `functools.partial`, but it was interfering with the argspec logic of `on_trait_change`.  I tried all different numbers of arguments, but none would respect my closure.

jasongrout:
https://github.com/ipython/traitlets/pull/43 gives you a reference to the HasTraits object that contains the Trait that is being changed.  Is that what you needed?  If so (after updating the most recent traitlets), make your callback take (name, old, new, obj) as arguments.

Or were you trying to have a closure for something else?

blink1073:
I wanted a callback that had the name of the variable being edited by the `Dropdown`.  The object in the `on_traits_change` would be the `DropDown`, not my variable, no?

blink1073:
Context: I was trying to set up an auto-editor for a `HasTraits` class.  For an Enum, I was creating a `Dropdown`, where a change in the value would propagate back to my `HasTraits` instance.

jasongrout:
Like this?

```python

from ipywidgets import Dropdown
from traitlets import HasTraits, Enum
class myclass(HasTraits):
    x = Enum(['hello', 'goodbye'], default_value='hello')
def make_gui(obj, attr, trait):
    def callback(name, new):
        print("**changed", attr, new)
    dd = Dropdown(value=getattr(obj, attr), options=trait.values)
    dd.on_trait_change(callback, 'value')
    return dd

def make_editor(obj):    
    for attr, trait in obj.traits().items():
        display(make_gui(obj, attr, trait))
```

blink1073:
Since you asked, here's what I'm using:

```python

def edit_traits(obj):
    """Create an IPython widget editor for a Traits object"""
    names = sorted(obj.trait_names())
    names.remove('name')
    names = ['name'] + names
    items = [Text('', disabled=True)]
    for name in names:
        if name.startswith('_'):
            continue
        try:
            value = getattr(obj, name)
        except TraitError:
            value = None
        trait = obj.traits()[name]
        if (trait.get_metadata('readonly') or
                isinstance(trait, Instance) or value is None):
            items.append(Text(str(value), disabled=True))

        elif isinstance(trait, Enum):
            # create a closure around "name" for the on_trait_change
            # callback
            def create_dropdown(name):
                dd = Dropdown(value=value, options=trait.values)

                def callback(dummy, value):
                    setattr(obj, name, value)
                dd.on_trait_change(callback, 'value')
                items.append(dd)

            create_dropdown(name)
        else:
            def callback(name, textbox):
                try:
                    setattr(obj, name, textbox.value)
                except Exception:
                    textbox.color = 'red'

            text = Text(str(value))
            text.on_submit(functools.partial(callback, name))
            items.append(text)
    labels = [Text(name, disabled=True) for name in names]
    labels = [Text(obj.__class__.__name__, disabled=True)] + labels
    display(HBox(children=[VBox(children=labels), VBox(children=items)]))
```

SylvainCorlay:
Do you still have interference with the argspec when using the new proposed API for traitlets?
https://github.com/ipython/traitlets/pull/61/

blink1073:
@SylvainCorlay, good call, this works:

```python
def callback(tname, change):
                print(tname, change)
                
dd = Dropdown(value=value, options=trait.values)
dd.observe(partial(callback, name), 'value')
items.append(dd)
```

I guess the question is whether this PR is still a valid addition, providing symmetry with other widgets that have callbacks?

SylvainCorlay:
Although we may still want to figure out where this other PR will go
because we also considered calling the `observe` method `on_change`...

Besides, the other non-traitlet-related callbacks in widgets are usually
meant to register to events that are not related to attribute state changes
and the PR on signaling is meant to address this. It deprecates `on_submit`.

Actually it also depends on the trailets PR because I need to know the
final decision for the signature of change handlers. (I have a state_change
signal in every widget)...

blink1073:
Fair enough, closing this one.

PR #151, opened 2015-08-26
Decouple the notebook from the widget manager.

jdfreder:
Separate the parts of the widget manager than need comm_manager and notebook instances to work from the base.  Allows people to write there own widget managers, without a notebook or comm_manager.

### DIFF
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -169,17 +169,17 @@ define([
      * @return {Promise<WidgetModel>}
      */
     ManagerBase.prototype.new_widget = function(options) {
-        var comm;
-        if (!options.comm) {
-            comm = Promise.resolve(options.comm);
+        var commPromise;
+        if (options.comm) {
+            commPromise = Promise.resolve(options.comm);
         } else {
-            comm = this._create_comm('ipython.widget', options.model_id,
+            commPromise = this._create_comm('ipython.widget', options.model_id,
                                     {'widget_class': options.widget_class});
         }
         
         var options_clone = _.clone(options);
         var that = this;
-        return comm.then(function(comm) {
+        return commPromise.then(function(comm) {
             options_clone.comm = comm;
             return that.new_model(options_clone).then(function(model) {
                 // Requesting the state to populate default values.
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -156,14 +156,6 @@ define([
     };
 
     /**
-     * Deprecated, use `new_widget` instead.
-     */
-    ManagerBase.prototype.create_model = function (options) {
-        console.warn('ManagerBase.create_model is deprecated. Use ManagerBase.new_widget');
-        return this.new_widget(options);
-    };
-
-    /**
      * Create a comm and new widget model.
      * @param  {Object} options - see new_model
      * @return {Promise<WidgetModel>}
@@ -491,6 +483,14 @@ define([
                 }
             }
         }
+    };
+
+    /**
+     * Deprecated, use `new_widget` instead.
+     */
+    WidgetManager.prototype.create_model = function (options) {
+        console.warn('WidgetManager.create_model is deprecated. Use ManagerBase.new_widget');
+        return this.new_widget(options);
     };
     
     WidgetManager.prototype.display_model = function(msg, model, options) {
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -178,9 +178,10 @@ define([
         }
         
         var options_clone = _.clone(options);
+        var that = this;
         return comm.then(function(comm) {
             options_clone.comm = comm;
-            return this.new_model(options_clone).then(function(model) {
+            return that.new_model(options_clone).then(function(model) {
                 // Requesting the state to populate default values.
                 return model.request_state();
             });
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -223,7 +223,14 @@ define([
          * Either a comm or a model_id must be provided.
          */
         var that = this;
-        var model_id = options.model_id || options.comm.comm_id;
+        var model_id;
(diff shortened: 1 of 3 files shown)


jdfreder:
@rgbkrk is interested in this

rgbkrk:
Sweet, thanks for the ping. I'll watch this closely, use it in the notebook, and review where I can.

rgbkrk:
From Jon in chat:

> We'll also need to make sure `ipywidgets` can be `npm` installed, in a follow on PR.
> In case you want to experiment while I'm gone, you'll need to inherit from the manager base to create your own manager then, there are a couple methods that need to be overridden
> 
> I'm saying all this in case you want to experiment next week while I'm gone

jdfreder:
@rgbkrk @SylvainCorlay this is ready for review : )

rgbkrk:
`/me drools`

ipywidgets/static/widgets/js/manager.js
```diff
@@ -461,54 +335,248 @@ define([
...
+                    return that.display_model(undefined, model, options);
                 }));
-            });
+            }));
         }).catch(utils.reject('Could not set widget manager state.', true));
     };
 
-    WidgetManager.prototype._get_connected_kernel = function() {
+    ManagerBase.prototype._create_comm = function(comm_target_name, model_id, metadata) {
+        throw new Error("Manager ._create_comm not implemented");
```
rgbkrk: I'm assuming you didn't mean to put a space after `Manager` and before the `.` here or below.

package.json
```diff
@@ -15,7 +15,7 @@
     "css": "node build_css.js",
     "prebuildtests": "rimraf ./ipywidgets/tests/bin/tests",
     "buildtests": "tsc ./ipywidgets/tests/**/*.ts --outDir ./ipywidgets/tests/bin -d -m commonjs -t ES5",
-    "postinstall": "npm run bower"
+    "prepublish": "npm run bower && npm run build"
```
rgbkrk: :+1:

rgbkrk:
Is it too much to ask that since you created new classes and functions that you could write (at the very least) some small `mocha` tests on the side to run through `node`?

jdfreder:
@rgbkrk I just tried, and mocha tests in this PR are a no-go because the widgets use AMD.  I do not object to node-ifying the widgets in another PR and then adding mocha tests to that.

ipywidgets/static/widgets/js/manager.js
```diff
@@ -295,8 +158,8 @@ define([
     /**
      * Deprecated, use `new_widget` instead.
      */
-    WidgetManager.prototype.create_model = function (options) {
-        console.warn('WidgetManager.create_model is deprecated. Use WidgetManager.new_widget');
+    ManagerBase.prototype.create_model = function (options) {
+        console.warn('ManagerBase.create_model is deprecated. Use ManagerBase.new_widget');
```
SylvainCorlay: Could we move this to WidgetManager, so that we don't carry deprecated classes to the cleaner future?

SylvainCorlay:
A big :+1:  to this. @ChakriCherukuri could be interested to look at this as well.

ipywidgets/static/widgets/js/manager.js
jdfreder: :heavy_check_mark:

JanSchulz:
I was looking into what needs to be done to build a shiny clone (like a notebook, which is executed in the backend and only the output is sent to the frontend -> static content and widgets; only widgets can communicate with the backend via the widget protocol) and one thing I found that the widgets have / need direct access to the kernel. IMO this would prevent such a usage, as this means that a potential public interface can (via the browser js console) submit python code to execute on the kernel.

Is it planned to also decouple the kernel connection from the widgets in this PR?

jdfreder:
Hi @JanSchulz ,

This PR decouples the notebook, kernel, etc from the WidgetManager.   My next step, in another PR, is to do the same for the Widget base class.  Once that's done, all that is required to display widgets outside of the notebook is an implementation of the ManagerBase class and a dummy comm.

I did this work a couple of times in the past, never to be merged.  You can see the last go, almost a year and a half ago [here](https://github.com/ipython/ipython/pull/5900).

SylvainCorlay:
Currently running on this branch and things seem to work.

jdfreder:
Awesome, well it's ready to merge when you are.  I'd be happy to continue my decoupling in more PRs, to follow.

rgbkrk:
Awesome! Once this is shipped if there's more decoupling, I'll try it out in jupyter-sidecar.

/cc @parente

lbustelo:
Basic question... If you take the Comm out of the widget manager... what does it do then? From the perspective of [jupyter-incubator/declarativewidgets](https://github.com/jupyter-incubator/declarativewidgets) we've identified WidgetManager as a class that is core to interactive widgets. What are the use case that is requiring both Comm and Notebook to be decoupled?

rgbkrk:
> What are the use case that is requiring both Comm and Notebook to be decoupled?

* [Thebe](https://github.com/oreillymedia/thebe) isn't really the notebook and should work with only kernel access
* Dashboards could work without a notebook setup
* Dashboards that aren't tied to a live backend
* Atom packages that would like to feature the widgets as well [Hydrogen](https://github.com/willwhitney/hydrogen), [Atom-Notebook](https://github.com/jupyter/atom-notebook)
* Electron apps that are *not* the notebook, e.g. [sidecar](https://github.com/nteract/jupyter-sidecar)

lbustelo:
Thanks @rgbkrk! I think many of these PRs need to start with some documentation of WHY. 

It is hard to keep up and almost feel like unnecessary effort until the dust settles.

jdfreder:
@lbustelo sorry, in the next PR I will try to better explain the "why".  In the following PRs I'll embrace duck typing for the comm.

rgbkrk:
Thank you @lbustelo for asking! I definitely don't want to hide any of this away.

SylvainCorlay:
hop!

rgbkrk:
:tada:

SylvainCorlay:
Can't wait to be able to use pythreejs and bqplot in other front-ends like the sidecar.

PR #152, opened 2015-08-26
Fixes flickering on interact

### LINKED ISSUE #145: Interact with simple plot function now flickers on update
The commit 68f771f on `interact` now displays the functions return value, which has the side-effect of causing "flickering" when interacting with a simple matplotlib plot.

Here's an example that used to be a smooth interactive plot with sliders:

    %matplotlib inline
    import numpy as np
    import matplotlib.pyplot as plt
    from ipywidgets import interact, HTML, FloatSlider
    from IPython.display import clear_output, display

    def make_plot(magnitude=5.0, phase=0.0):
        x = np.arange(0, 10, 0.1)
        y = magnitude*np.sin(x+phase)
        plt.plot(x, y)
        plt.ylim(-10, 10)
        
    interact(make_plot, magnitude=(0, 10, 0.1), phase=(0, 4*np.pi, 0.1))

Unfortunately, because make_plot returns `None`, None is printed, which somehow breaks the smooth updating that interact previously exhibited.  Now the plot is erased, the rest of the web interface jumps up to fill the gap, then the plot is replaced... repeatedly and rapidly while the slider is being interacted with.  This drastically reduces the interactivity and aesthetic appeal of simple tools like this.

Here is an example without interact which currently runs fine and demonstrates the smooth behavior that interact previously had:

    %matplotlib inline
    import numpy as np
    import matplotlib.pyplot as plt
    from ipywidgets import interact, HTML, FloatSlider
    from IPython.display import clear_output, display
    
    slider1 = FloatSlider(description='magnitude', min=0, max=10, step=0.1)
    slider2 = FloatSlider(description='phase', min=0, max=4*np.pi, step=0.1)
    
    def update_plot():
        magnitude = slider1.value
        phase = slider2.value
        x = np.arange(0, 10, 0.1)
        y = magnitude*np.sin(x+phase)
        fig = plt.figure()
        plt.plot(x, y)
        plt.ylim(-10, 10)
        
        clear_output(True)
        
    slider1.on_trait_change(update_plot, 'value')
    slider2.on_trait_change(update_plot, 'value')

    display(slider1, slider2)
    update_plot()

Interestingly, adding `display(None)` to the end of update_plot() brings back the flickering from the interact example, demonstrating that the display of text is causing the flicker.  Hmmm... apparently displaying an empty HTML object removes the flicker... I guess that would be the workaround.  If I add `return HTML()` to the first example's make_plot() function, it updates smoothly now.

I don't know if there is a better solution to this.  I wish text and figures could both be updated smoothly without the "flickering effect," but until that is implemented, I'd at least like the option to turn off display of function output for interact or perhaps not display output when output is `None` so that the smooth behavior isn't disrupted simply to display `None`.  Documenting the workaround of returning `HTML()` in an example might be useful in lieu of other solutions (I haven't searched thoroughly to see if this has been done/documented before...).

edit: fixed 2nd example - had left out the slider definitions

alope107:
This change makes it so that the return value of a callback function is only displayed if it is not None.  This solves the problem of flickering output due to the display of multiple items.  Closes #145

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -218,7 +218,8 @@ def interactive(__interact_f, **kwargs):
             manual_button.disabled = True
         try:
             container.result = f(**container.kwargs)
-            display(container.result)
+            if container.result is not None:
+                display(container.result)
         except Exception as e:
             ip = get_ipython()
             if ip is None:
@@ -316,7 +317,8 @@ def interact(__interact_f=None, **kwargs):
             # so wrap in a lambda
             f = lambda *args, **kwargs: __interact_f(*args, **kwargs)
             f.widget = w
-        display(w)
+        if w is not None:
+            display(w)
         return f
     else:
         # This branch handles the case 3
--- a/ipywidgets/widgets/tests/test_interaction.py
+++ b/ipywidgets/widgets/tests/test_interaction.py
@@ -411,6 +411,22 @@ def test_call_interact():
     )
 
 @nt.with_setup(clear_display)
+def test_call_interact_on_trait_changed_none_return():
+    def foo(a='default'):
+        pass
+    with patch.object(interaction, 'display', record_display):
+        ifoo = interact(foo)
+    nt.assert_equal(len(displayed), 1)
+    w = displayed[0].children[0]
+    check_widget(w,
+        cls=widgets.Text,
+        value='default',
+    )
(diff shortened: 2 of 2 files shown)


jdfreder:
@alope107 you should setup your git and github so you get credit for your commits.  Make sure the email you set in your git client is registered on github.

jdfreder:
This works!  Thanks for the tests too!

PR #161, opened 2015-09-15
Fix pip install -e npm invokation for latest pip...

jdfreder:
... which does not work right now.

I think this is following @minrk 's lead in jupyter/notebook.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -266,7 +266,14 @@ setup_args = dict(
     },
 )
 
-if 'develop' in sys.argv or any(a.startswith('bdist') for a in sys.argv):
+# For some commands, use setuptools.  Note that we do NOT list install here!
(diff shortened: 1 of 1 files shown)


minrk:
I'm surprised this is needed. When I do `pip install -e .` it does work. I don't see why it would be different, since `pip install -e` does `setup.py develop`, which is a setuptools command.

minrk:
I'd also be sad to re-add the big list of mostly deprecated setuptools commands.

jdfreder:
Sorry for the noise, I don't think it's needed.  I found out that pip was eating my print statements, which was confusing me.  `rm -rf`ing `node_modules` and rerunning `pip install -e .` lead to `node_modules` reappearing, despite a lack of npm output in the console.

PR #165, opened 2015-09-22
Better handling of nested widgets

egentry:
I've run into situations where it'd be nice to start with one widget, and use the value of that widget to dynamically generate another widget. Unfortunately, when the outer widget changes, it leaves all of the previous widgets it created, which quickly becomes unmanageable.  Basically I'd like something similar to the `clear_output` keyword, but which also clears any widgets within the result of a widget.

(For a basic example of what I mean, check out: https://gist.github.com/egentry/f3f28130162d60c97b59)

In the past I got around this in hackish ways, but others might appreciate a clean fix.

Some thoughts / questions:
- Should this be triggered by a keyword (in the same way `interactive` uses `clear_output`)?
- Would this be better fixed in `clear_output()` (although it'd need to know which widgets to save and which to clear)
- This results in flickering, since it doesn't wait for a new widget to be created. I could look into that, but no promises that I'll find a good solution
- Does this break anyone's own use cases? Are there behaviors I'm not thinking about?
  - Will linked widgets (e.g. via `traitlets.jslink`) be a problem if one of the linked widgets disappears and is replaced by a new widget?  (If this PR seems useful, I can test some of those situations)

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -14,7 +14,7 @@ from inspect import getcallargs
 from IPython.core.getipython import get_ipython
 from . import (Widget, Text,
     FloatSlider, IntSlider, Checkbox, Dropdown,
-    Box, Button, DOMWidget, widget_box)
+    Box, Button, DOMWidget)
 from IPython.display import display, clear_output
 from ipython_genutils.py3compat import string_types, unicode_type
 from traitlets import HasTraits, Any, Unicode
@@ -218,7 +218,7 @@ def interactive(__interact_f, **kwargs):
             manual_button.disabled = True
         try:
             def close_result_widgets(w):
-                if type(w.result) is widget_box.Box:
+                if type(w.result) is Box:
                     close_result_widgets(w.result)
                     w.result.close()
             close_result_widgets(container)
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -14,7 +14,7 @@ from inspect import getcallargs
 from IPython.core.getipython import get_ipython
 from . import (Widget, Text,
     FloatSlider, IntSlider, Checkbox, Dropdown,
-    Box, Button, DOMWidget)
+    Box, Button, DOMWidget, widget_box)
 from IPython.display import display, clear_output
 from ipython_genutils.py3compat import string_types, unicode_type
 from traitlets import HasTraits, Any, Unicode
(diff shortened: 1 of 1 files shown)


ipywidgets/widgets/interaction.py
```diff
@@ -217,6 +217,11 @@ def call_f(name=None, old=None, new=None):
         if manual:
             manual_button.disabled = True
         try:
+            def close_result_widgets(w):
+                if type(w.result) is Box:
+                    close_result_widgets(w.result)
```
jdfreder: Why the recursion?  The line below should close everything within.

ipywidgets/widgets/interaction.py
```diff
@@ -217,6 +217,11 @@ def call_f(name=None, old=None, new=None):
...
+                    w.result.close()
+            close_result_widgets(container)
```
jdfreder: This won't work if people don't display their widgets within the interact container and I haven't seen anyone display additional widgets inside the interact container yet, so my guess is that this won't do anything for most people.

jdfreder:
Have you thought about instead enabling the widget area to be cleared programatically?  I think that would be the most similar to `clear_output`.

ipywidgets/widgets/interaction.py
egentry: That's what I initially expected, but when I tested it on the gist, `.close()` seemed to recursively close the `children` but not the `result` of a widget.

ipywidgets/widgets/interaction.py
egentry: Ahh, you're right. I had misread `interact()` and thought it was standard behavior to attach the result to the container, but that is not the case.  I agree this approach probably wouldn't do much for most people (and it would require some pretty ugly code to work at all for `interact()` rather than `interactive()`). 

(Also, I realized if `result` is just a plain `Box()`, the code currently generates an `AttributeError` when it tries to access `result`, but that's a more minor fix)

egentry:
Oh! I just learned that you can attach attributes to functions and they act like static variables.  That'll allow me to keep any widgets created the last time the function was called, and close them the next time the function is called.  That lets me do exactly what I want, without needing any change to the ipywidgets source.

(If anyone finds this later and wants an example, I made a new gist: https://gist.github.com/egentry/ea5964c1fb755abc6cd1)

This can be closed without merging.

jdfreder:
Thanks @egentry

PR #169, opened 2015-10-08
Adding ColorPicker widget

SylvainCorlay:
Using the native browser color picker.

### DIFF
--- a/ipywidgets/static/widgets/js/widget_color.js
+++ b/ipywidgets/static/widgets/js/widget_color.js
@@ -13,13 +13,6 @@ define([
                 .addClass("widget-label")
                 .appendTo(this.$el);
 
-            var description = this.model.get("description");
-            if (description) {
-                this.$label.text(this.model.get("description"));
-            } else {
-                this.$label.hide();
-            }
-
             this.$color_container = $("<div />")
                 .addClass("hbox")
                 .appendTo(this.$el);
@@ -34,15 +27,26 @@ define([
                 .appendTo(this.$color_container);
 
             this.listenTo(this.model, "change:value", this._update_value, this);
+            this.listenTo(this.model, "change:description", this._update_description, this);
             this.$colorpicker.on("change", this._picker_change.bind(this));
             this.$textbox.on("change", this._text_change.bind(this));
 
             this._update_value();
+            this._update_description();
         },
         _update_value: function() {
             var value = this.model.get("value");
             this.$colorpicker.val(color2hex(value));
             this.$textbox.val(value);
+        },
+        _update_description: function() {
+            var description = this.model.get('description');
+            if (description.length === 0) {
+                this.$label.hide();
+            } else {
+                this.typeset(this.$label, description);
+                this.$label.show();
+            }
         },
         _picker_change: function() {
             this.model.set("value", this.$colorpicker.val());
--- a/ipywidgets/static/widgets/js/widget_color.js
+++ b/ipywidgets/static/widgets/js/widget_color.js
@@ -1,0 +1,86 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+define([
+    "nbextensions/widgets/widgets/js/widget",
+], function(widget) {
+
+    var ColorPicker = widget.DOMWidgetView.extend({
+        render: function() {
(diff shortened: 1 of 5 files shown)


minrk:
Travis is showing a SyntaxError. Typos?

SylvainCorlay:
I see the same error on master.

minrk:
Thanks. Seems strange that a commit that changes no javascript would introduce a Javascript SyntaxError, but I will investigate. @jdfreder any hints for finding *where* the SyntaxError occurs, or what the error actually is?

jdfreder:
Darn, just looked at the Travis output, that's weird.  My guess is that it's a problem with one of the requirejs imports.  @minrk printing the error correctly is one of those N-contexts issues, somewhere I'm probably doing `String(Error instance)` which is not providing enough information.  The best way to go about this one is to reproduce the error as a user.  With so many asynchronous things happening at once the order of events isn't always reliable; however, it does look like the error is getting thrown at the beginningish of page/widgets load.

minrk:
jupyter/notebook#472 is not related, since these tests are running on Travis with current stable of the notebook (4.0.5), prior to that change. I can also verify that the tests pass when using current master of the notebook. The problem is that one of the patches on 4.0.5 was incomplete. I just cut 4.0.6 with the fixed patch and everything's green here.

SylvainCorlay:
Sorry for being misleading here - and thanks for solving the issue.

ipywidgets/static/widgets/js/widget_color.js
```diff
@@ -0,0 +1,89 @@
...
+     * From a valid html color (named color, 6-digits or 3-digits hex format)
+     * return a 6-digits hexadecimal color #rrggbb.
+     */
+    function color2hex(color) {
+        return named_colors[color.toLowerCase()] || rgb2hex(color);
+    };
+
+    function rgb2hex(rgb) {
+        if (rgb.search("rgb") == -1) {
+            return rgb;
```
SylvainCorlay: I still need to handle the convertion `#rgb` to `#rrggbb`

ipywidgets/static/widgets/js/widget_color.js
```diff
@@ -0,0 +1,89 @@
...
+            function hex(x) {
+                return ("0" + parseInt(x).toString(16)).slice(-2);
+            }
+            return "#" + hex(rgb[1]) + hex(rgb[2]) + hex(rgb[3]);
+        }
+    };
+
+    return {
+        ColorPicker: ColorPicker,
+        color2hex: color2hex,
```
jdfreder: Want to export rgb2hex too?

ipywidgets/static/widgets/js/widget_color.js
jdfreder: What color types does the `Color` trait validate?

ipywidgets/static/widgets/js/widget_color.js
SylvainCorlay: I should only export color2hex. And have a hex3/hex6 distinction.

ipywidgets/static/widgets/js/widget_color.js
SylvainCorlay: I think both. ( I need to check. ) plus the named colors.

ipywidgets/static/widgets/js/widget_color.js
jdfreder: I'm confused, what makes color2hex > rgb2hex ?  Why export one and not the other?   Or instead, not export anything but the widget.

ipywidgets/static/widgets/js/widget_color.js
jdfreder: Yeah, whatever `Color` validates, this widget should definitely support.  No more, no less.

ipywidgets/static/widgets/js/widget_color.js
SylvainCorlay: color2hex knows about named colors.

jdfreder: Ah sorry! I see now, color2hex is a superset of rgb2hex.  That makes sense

SylvainCorlay:
This is ready.

jdfreder:
Thanks!

PR #174, opened 2015-10-13
Use the _get_comm_info method to reconstruct all widget models

### LINKED ISSUE #133: Persistence issues with widgets holding other widgets
#### Re-instantiating all valid widget models
Let' s say that we have a container widget with id `id1` holding two widgets of id `id2` and `id3`. (For example, `HBox([Button(), Button()]))`. We save the notebook and the states of the three widgets are saved to the local storage.

``` Python
id1
 |
 +--- id2
 |
 +--- id3
```

Then we change the content initial container wigdet, so that we are in the situation

``` Python
id1
 |
 +--- id4
 |
 +--- id5
```

then, without saving the notebook, we reload the page.

Since `id1` is a valid comm, we request it' s state, but when de-serializing its state, we fail at instantiating the widget models `id4` , `id5` because they were not recreated. (more precisely, get_model will return undefined.) Errors are displayed in the console.

This shows that we should re-instantiate all the valid widget models regardless of whether they were in the local storage. 

#### External state request

The issue is that we need to retrieve the `_model_name` and the `_model_module`, which are part of the state to instantiate the model and make a state request.

Hence, the state request should probably be external to the widget, and sent via the bare comm object.

### LINKED ISSUE #155: reconnect to widget
currently, when starting a long-running process i’d like to close my laptop lid, open it later, and see my progressbar jump to its current position.

the notebook itself handles reconnection well, but the widgets simply vanish.

jdfreder:
Seems to fix a lot of bugs with persistence in general.

@SylvainCorlay woohoo!

closes #133
closes #155

### DIFF
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -400,7 +400,7 @@ define([
                     var update_promise = new Promise(function(resolve, reject) {
                         comm.on_msg(function (msg) {
                             
-                            // A suspected response was recieved, check to see if
+                            // A suspected response was received, check to see if
                             // it's a state update.  If so, resolve.
                             if (msg.content.data.method === 'update') {
                                 resolve({
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -388,7 +388,7 @@ define([
         var that = this;
         var backed_widgets_loaded = this._get_comm_info().then(function(comm_ids) {
             
-            // Filter out non-widget comm ids
+            // Create comm class instances from comm ids.
             var comm_promises = Object.keys(comm_ids).map(function(comm_id) {
                 return that._create_comm(that.comm_target_name, comm_id);
             });
--- a/ipywidgets/static/widgets/js/manager.js
+++ b/ipywidgets/static/widgets/js/manager.js
@@ -165,7 +165,7 @@ define([
         if (options.comm) {
             commPromise = Promise.resolve(options.comm);
         } else {
-            commPromise = this._create_comm('ipython.widget', options.model_id,
+            commPromise = this._create_comm(this.comm_target_name, options.model_id,
                                     {'widget_class': options.widget_class});
         }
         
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
@jdfreder woohoo. I will look at it tonight.

ipywidgets/static/widgets/js/manager.js
```diff
@@ -376,17 +382,69 @@ define([
...
+            });
+            
+            // Send a state request message out for each widget comm and wait
+            // for the responses.
+            return Promise.all(comm_promises).then(function(comms) {
+                return Promise.all(comms.map(function(comm) {
+                    var update_promise = new Promise(function(resolve, reject) {
+                        comm.on_msg(function (msg) {
+                            
+                            // A suspected response was recieved, check to see if
```
SylvainCorlay: typo

ipywidgets/static/widgets/js/manager.js
```diff
@@ -317,6 +317,12 @@ define([
         // Recreate all the widget models for the given notebook state.
         var all_models = that._get_comm_info().then(function(live_comms) {
             return Promise.all(_.map(Object.keys(state), function (model_id) {
+                
+                // If the model has already been created, return it.
+                if (that._models[model_id]) {
+                    return that._models[model_id];
+                }
+
```
SylvainCorlay: In which situation could the model already be created?

SylvainCorlay:
@jdfreder the `comm_info_[request/reply]` shell message allows you to only request comms corresponding to a given target name. (You can see this in #75.)  
It would make sense to make the workaround that we have here do the same thing so that it will be easier to replace when the notebook 4.1 is released like here: https://github.com/SylvainCorlay/ipython_widgets/commit/78984210255f71f1e01c01f331fb978a75feebb6

ipywidgets/static/widgets/js/manager.js
jdfreder: In line 425 of the block below.  Before I call this function I create models based on `_get_comm_info` return info.

SylvainCorlay: oh, ok

jdfreder:
@SylvainCorlay comments addressed

SylvainCorlay:
:+1: :tada: :balloon: 
I am all good with this. I can merge later this afternoon, just to let the time to other people to comment. Any input @jasongrout?

SylvainCorlay:
hop?

PR #175, opened 2015-10-14
Remove workaround for comm_info message

SylvainCorlay:
This depends on versions of jupyter_client, ipykernel and notebook including the following changes

jupyter/jupyter_client#34 *merged*
ipython/ipykernel#25 *merged*
jupyter/notebook#166, *merged*

This now only requires the notebook 4.1 to be released.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -279,10 +279,10 @@ if 'setuptools' in sys.modules:
 
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
-    'ipython>=4.0.0dev0',
-    'ipykernel',
+    'ipython>=4.0.0',
+    'ipykernel>=4.2.0',
     'traitlets',
-    'notebook',
+    'notebook>=4.1.0b1',
 ]
 
 extras_require = setuptools_args['extras_require'] = {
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
I probably closed the earlier PR accidentally.

minrk:
Just noting that this doesn't actually require the jupyter_client patch, since the messages go directly to ipykernel. If you wanted it to work on the notebook today, you could add a backward-compat shim to define `Kernel.comm_info` if it's not already available on Kernel. But waiting for notebook-4.1 is fine, too.

SylvainCorlay:
@jdfreder I rebased an now require notebook 4.1beta.

setup.py
```diff
@@ -279,10 +279,10 @@ def run(self):
 
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
-    'ipython>=4.0.0dev0',
-    'ipykernel',
+    'ipython>=4.0.0',
+    'ipykernel>=4.2.0',
     'traitlets',
-    'notebook',
+    'notebook>=4.1.0b1',
```
jdfreder: My python packaging knowledge to too weak -- will we have to change this when 4.1.0 final is released?

SylvainCorlay: I don't think so. If you look just above, we had `ipython>=4.0.0dev0` for a long time.

jdfreder: Sweet

PR #178, opened 2015-10-14
Stricter LESS and Bower versions,

### LINKED ISSUE #143: Widget-subarea does not enlarge if widget does not have a fixed weight
With IPython 3 the `widget-subarea` div would enlarge to use up the whole available width. With IPython 4 this doesn't happen and the width is super narrow. Only by setting a fixed width on the widget itself it will be displayed correctly.

IPython used to have the CSS property `flex: 2 1 0%;` on `widget-subarea` which is gone now. If I re-add temporarily in the browser's developer tools, the sizing works again as expected.

jdfreder:
and make sure bower is ran silently

closes #143 

@SylvainCorlay @minrk I think this may fix some install problems that people were having.  Bower now prompts for anonymous data collection, which will halt automated build scripts, and halted my `npm install`.  The ` --config.interactive=false` skips this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "bower": "^1.5.3",
-    "clean-css": "*",
+    "clean-css": "^3.4.6",
     "less": "^2.5.3",
     "rimraf": "^2.4.1",
     "spawn-sync": "*",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "build": "npm run css && npm run buildtests",
-    "bower": "bower install",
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Since the problem came from the minification, shouldn't we also specify the version for cleancss? Which one are you using?

jdfreder:
Whoops, didn't realize we were using clean-css to do minification, I thought it was something LESS was doing for us.

:heavy_check_mark: should be fixed now

SylvainCorlay:
AFK. I will check that it works for me before tomorrow.

SylvainCorlay:
thx!

PR #181, opened 2015-10-14
Only update slider description on description change

### LINKED ISSUE #166: Latex labels are redrawn when interacting with widget
I see that ipython/ipython#5937 introduced latex labels for widgets, which would greatly improve the readability of the RISE/live-reveal slides which I am producing.

However, when dragging a slider for example, the widget label also seems to be redrawn at every step, thus triggering a constant series of mathjax renderings, which results in horrible flickering. The following code is sufficient to reproduce the problem:

````
from ipywidgets import *
display(FloatSlider(description='$x^2$'))
````

For my application this makes the latex markup unusable, so it would be great if there was a way that updating of the label could be suppressed when the widget is redrawn.

jdfreder:
Action shot!

demo

closes #166

### DIFF
--- a/ipywidgets/static/widgets/js/widget_int.js
+++ b/ipywidgets/static/widgets/js/widget_int.js
@@ -34,14 +34,19 @@ define([
                 .addClass('widget-readout')
                 .attr('contentEditable', true)
                 .hide();
-
+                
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
:+1:

PR #183, opened 2015-10-15
Add `ready` event which fires when the widget receives its first state

### LINKED ISSUE #71: Creating a new "model_ready" event in widget.js
I encountered the following problem with widgets that create sub-widgets from the JavaScript side in their JavaScript model constructor: 

For example the gamepad widget creates subwidgets for the buttons and axes in its constructor. Once the `connected` trait attribute is set to True, the user may wire things to the gamepad buttons...

However, in the case of a page reload, I would like these widgets **not** to be replaced by new ones, when the constructor of the JavaScript side is executed. The part of the constructor containing the creation of the children should only occur after the state request resolves.

One way to solve this issue would be to create a `model-ready` event for that, of have a certain promise that gets resolved right away in the case of a new model, and after the state request resolves in case of page reload.

Suggestions welcome.

@jdfreder @jasongrout

jdfreder:
@SylvainCorlay can you verify that this is the event you want?  It should fire on all paths **exept**, and by-design, `new_model`.  The higher level `new_widget` function will fire it, once the state has been received.  Additionally, Python instantiated widgets will fire it.

closes #71

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -56,6 +56,7 @@ define(["nbextensions/widgets/widgets/js/manager",
             this.msg_buffer = null;
             this.state_lock = null;
             this.id = model_id;
+            this._first_state = true;
 
             // Force backbone to think that the model has already been
             // synced with the server.  As of backbone 1.1, backbone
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I would like to make use of it right away in the Controller widget to be sure that this solves our issue.

jdfreder:
Ah good point, I forgot that was inside this repo now!  Do you want to make a PR to my PR, or shall I do it in this PR?

SylvainCorlay:
Yes, I was looking at it at the moment.

jdfreder:
You know what, looking at it, I think the event code need to move after the super call.

jdfreder:
Ok my last commit moves it into the right place

jdfreder:
Closing in favor of #189

PR #185, opened 2015-10-16
Fix version in py file

jdfreder:
@SylvainCorlay


SylvainCorlay:
Unless we (properly) release 4.1.0 right away, we will need to merge this, or make it `4.1.0dev` so that other package requiring 4.0.3 don't replace their dev install with the pypi version.

jdfreder:
#187

PR #186, opened 2015-10-16
Add the ability to change the readout logic for sliders easily.

### LINKED ISSUE #172: Setting IntRangeSlider readout to pick elements from an array-like object.
I've seen a few widget applications using the IntRangeSlider as a date selector, for example. It would be cool if the readout display values could be set dynamically.

jdfreder:
With this PR you can change the readout logic by overriding the new `valueToString` and `stringToValue` methods in Javascript.

closes #172

### DIFF
--- a/ipywidgets/static/widgets/js/widget_int.js
+++ b/ipywidgets/static/widgets/js/widget_int.js
@@ -35,7 +35,7 @@ define([
                 .attr('contentEditable', true)
                 .hide();
                 
-            this.listenTo(this.model, 'change:slider_color', function(sender, value) {
+            this.listenTo(this.model, 'change:slider_handleTextChangecolor', function(sender, value) {
                 this.$slider.find('a').css('background', value);
             }, this);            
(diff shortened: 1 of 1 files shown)


jdfreder:
Broken, do not merge yet

jdfreder:
Ok this is working now. @SylvainCorlay

SylvainCorlay:
@jdfreder can you explain a bit what this does?

jdfreder:
@jgoad desired the ability to change the readout format from numbers to dates.  Specifically he wants to use the IntRangeSlider to allow the user to select a range of dates.  This requires him to subclass the JS which is fine; however, the readout logic is tightly coupled to the other properties of the widgets, so he'd have to re-implement a lot of functionality he shouldn't have to.  This PR just decouples the readout parsing and formatting code, so it can be easily replaced.

SylvainCorlay:
This is related to the Date trait type work then.

PR #189, opened 2015-10-19
Controller ready

### LINKED ISSUE #71: Creating a new "model_ready" event in widget.js

SylvainCorlay:
Supersedes #183 but rebased on master. -- closes #71.
Comes on the top of #190.

### DIFF
--- a/ipywidgets/static/widgets/js/widget_controller.js
+++ b/ipywidgets/static/widgets/js/widget_controller.js
@@ -2,8 +2,8 @@
 // Distributed under the terms of the Modified BSD License.
 
 define([
-    "nbextensions/widgets/widgets/js/widget", 
-    "base/js/utils", 
+    "nbextensions/widgets/widgets/js/widget",
+    "base/js/utils",
 ], function(widget, utils) {
     'use strict';
 
@@ -92,7 +92,18 @@ define([
                 // Start the wait loop, and listen to updates of the only
                 // user-provided attribute, the gamepad index.
                 this.readout = 'Connect gamepad and press any button.';
-                this.wait_loop();
+                // Wait for the state to be provided by the backend.
+                this.on('ready', function() {
+                    if (this.get('connected')) {
+                        // No need to re-create Button and Axis widgets, re-use
+                        // the models provided by the backend which may already
+                        // be wired to other things.
+                        this.update_loop();
+                    } else {
+                        // Wait for a gamepad to be connected.
+                        this.wait_loop();
+                    }
+                }, this);
             }
         },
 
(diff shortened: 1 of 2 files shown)


jdfreder:
closes https://github.com/ipython/ipywidgets/issues/71

ipywidgets/static/widgets/js/widget_controller.js
```diff
@@ -144,12 +155,13 @@ define([
         },
 
         update_loop: function() {
-            /* Populates axes and button values, until the gamepad is disconnected.
+            /* Upadte axes and buttons values, until the gamepad is disconnected.
```
jdfreder: typo `update`

jdfreder:
One small comment typo, otherwise this is ready to go!

SylvainCorlay:
Except that it does not work. Working on it.

jdfreder:
Oh, missed the "WIP" note in the title

SylvainCorlay:
Lol. I think I found a problem in ipykernel. Opened a related PR in ipykernel.

jdfreder:
Oh okay, good.  Sorry you had to dig that deep, sounds like a pain.

jdfreder:
In the current state, does this PR break anything in master more than it's already broken, without https://github.com/ipython/ipykernel/pull/67 merged?  I'm wondering if we can merge this for 4.1 or if we should push on without it.

SylvainCorlay:
I don't think that it breaks anything but persistence is not quite there yet for the controller. I think that there is a problem with the beastmode code.

SylvainCorlay:
I don't think it breaks anything but persistence is not there yet for the controller widget.
There is a problem with the beastmode code. Now looking into it.

jdfreder:
@SylvainCorlay does your latest commit (ee6d590) fix the problem?

SylvainCorlay:
This seems to be ok now.

SylvainCorlay:
Now it seems to be working fine.

jdfreder:
Woohoo!

PR #190, opened 2015-10-19
Comm info dummy widget

SylvainCorlay:
@jdfreder this should solve the issue in master.


jdfreder:
I tested this, it works well!  Thanks @SylvainCorlay !

PR #191, opened 2015-10-20
Update the RELEASE instructions

jdfreder:
@jasongrout @SylvainCorlay so we can remember.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,9 +1,15 @@
 To release a new version of the widgets on PyPI:
 
-1. Update \_version.py
+1. Update \_version.py (set release version, remove 'dev')
 2. `git add` and `git commit`
 3. `python setup.py sdist upload` 
 4. `python setup.py bdist_wheel upload`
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -1,9 +1,15 @@
 To release a new version of the widgets on PyPI:
 
-1. Update \_version.py
+1. Update \_version.py (set release version, remove 'dev')
 2. `git add` and `git commit`
 3. `python setup.py sdist upload` 
 4. `python setup.py bdist_wheel upload`
```
SylvainCorlay: Is it ok when it is on two different lines?

jdfreder: It's what I've been doing.  Is there a better way to do it?  I'm a PyPI noob

blink1073: Yeah, you can upload as many times as you want (as long as it is a different file)

SylvainCorlay: Me too, just that it is different from the wheels tutorial, hence the question. If it worked, you answered my question :)

PR #198, opened 2015-10-26
fix version_info

minrk:
dev needs to be in it


jdfreder:
Good to know, thanks!

PR #201, opened 2015-10-26
Fix the margin and padding of the DropDown,

### LINKED ISSUE #192: Dropdown widget does not set margin/padding correctly
Whenever I try to set a custom margin (or padding) for a dropdown widget, I get this messy shift between the label and the dropdwon menu:

widget

The code for the given example:
```python
drpdwn = widgets.Dropdown(options={'1': 1, '2': 2}, description='numbers:', margin=10)
sel = widgets.Select(options={'1': 1, '2': 2}, description='numbers:', margin=10)

display(drpdwn, sel)
```

The select widget looks fine, though. 
I looked into it and the problem seems to be that the margin is not set properly. For all other widgets the margin is pushed to the surrounding FlexBox. The dropdown widget sets the margin only for the `widget-item btn-group`:

html

I think I found the problem in the [widget_selection.js](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/static/widgets/js/widget_selection.js). The `update_attr` method for the dropdown widget looks like this:

```javascript
...
} else if (name == 'padding') {
  this.$droplist.css(name, value);
  this.$buttongroup.css(name, value);
} else if (name == 'margin') {
  this.$buttongroup.css(name, value);
} else if (name == 'height') {
  this.$droplabel.css(name, value);
  this.$dropbutton.css(name, value);
} else if (name == 'padding' || name == 'margin') {
  this.$el.css(name, value);
...
```
Is it intended to add the margin/padding only to the `btn-group` or is the `else if` part for margin and padding wrong? The last part, which sets the margin for the FlexBox (`this.$el`), will never be excuted or am I wrong?

jdfreder:
also override weird bootstrap margin set on special screen
width for the Listbox

closes #192

### DIFF
--- a/ipywidgets/static/widgets/less/widgets.less
+++ b/ipywidgets/static/widgets/less/widgets.less
@@ -188,6 +188,10 @@
     /* Listbox */
     width         : @widget-width;
     margin-bottom : 0px;
+    /* Bootstrap does some weird stuff with the margins of select boxes.  The 
+    specifity of the media selector needs to be overwritten using important */
+    margin-left   : 0px !important; 
+    margin-right  : 0px !important;
 }
 
 .widget-numeric-text {
--- a/ipywidgets/static/widgets/js/widget_selection.js
+++ b/ipywidgets/static/widgets/js/widget_selection.js
@@ -128,15 +128,10 @@ define([
             } else if (name == 'width') {
                 this.$droplist.css(name, value);
                 this.$droplabel.css(name, value);
-            } else if (name == 'padding') {
-                this.$droplist.css(name, value);
-                this.$buttongroup.css(name, value);
-            } else if (name == 'margin') {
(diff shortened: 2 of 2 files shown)


ipywidgets/static/widgets/less/widgets.less
```diff
@@ -188,6 +188,10 @@
     /* Listbox */
     width         : @widget-width;
     margin-bottom : 0px;
+    /* Bootstrap does some weird stuff with the margins of select boxes.  The 
+    specifity of the media selector needs to be overwritten using important */
+    margin-left   : 0px !important; 
+    margin-right  : 0px !important;
```
jdfreder: @SylvainCorlay this may look alarming, but it's actually getting applied to the inner element, not the element that gets margin-ed by the traitlet, which is the outer element.

PR #208, opened 2015-10-29
Fix button width issue

SylvainCorlay:
- add a semantic css class to the button
- style the button to use widget-width-short by default
- as suggested by Brian make widget-width 300px so that the short width is half the default width. (These two values will have to be adjusted when we add margin to non-container widgets.)


jdfreder:
Looks good!

PR #219, opened 2015-10-30
Add a Style widget

jdfreder:
Which, when displayed, is capable of changing style attributes on the parent.

TODO
- [ ] Use union for all trait definitions in Style??
- [x] Add display logic to DOMWidgetView for Style child widget
- [x] Test it


ipywidgets/static/widgets/js/widget.js
```diff
@@ -628,8 +636,29 @@ define(["nbextensions/widgets/widgets/js/utils",
...
+                
+                this.setStyle(this.model.get('style'));
             }, this));
         },
+        
+        serializers: _.extend({
+            style: {deserialize: unpack_models},
+        }, WidgetModel.serializers),
+        
+        setStyle: function(style, oldStyle) {
```
ellisonbg: Do we need to undo the styles applied by the oldStyle when a new one is set?

ipywidgets/widgets/domwidget.py
```diff
@@ -0,0 +1,76 @@
...
+
+class DOMWidget(Widget):
+    """Widget that can be inserted into the DOM"""
+    
+    visible = Bool(True, allow_none=True, help="Whether the widget is visible.  False collapses the empty space, while None preserves the empty space.", sync=True)
+    _css = Tuple(sync=True, help="CSS property list: (selector, key, value)")
+    _dom_classes = Tuple(sync=True, help="DOM classes applied to widget.$el.")
+
+    style = Instance(Style, allow_none=True, sync=True, **widget_serialization)
+    def _style_default():
```
ellisonbg: Would it make more sense for the detail to by None so Python code can test to see if any Style is applied more easily?

jdfreder: > can test to see if any Style is applied more easily?

I don't understand this comment, sorry.

The reason why Style is instantiated by default is that it allows the user to do this:

```python
w = IntSlider()
w.style.width = 500
```

ipywidgets/static/widgets/js/widget.js
jdfreder: I probably should try to do this, but it gets tricky.  I'll try doing this and ping back.

jdfreder:
Action shot:

style

@SylvainCorlay ready for review

jdfreder:
@SylvainCorlay a couple of questions:

- Should I add a deprecation message to the horde of style traits in DOMWidget?  If so, both on the Python and JS sides?  
- Do you think using the CUnicode trait in Style is the right thing to do?  Or instead should I be using a Union(Unicode, Int) and Union(Unicode, Float) where possible?

I'm sure you see the design of this is a little different than we talked about.  This is basically a large subset of CSS keys, all that I could find which were cross platform.  I think it makes sense to do this, instead of shooting ourselves in the foot by trying to abstract CSS into simpler keys.

jdfreder:
> Funny, I use bind for the controller widget, but there is no browser supporting controllers that does not support bind.

IKR, we really need to move away from Phantom.  I'll make that a priority for 6.0

ipywidgets/static/widgets/js/widget_selectioncontainer.js
```diff
@@ -160,7 +160,7 @@ define([
 
                 // Trigger the displayed event of the child view.
                 that.displayed.then(function() {
-                    view.trigger('displayed');
+                    view.trigger('displayed', that);
```
SylvainCorlay: Why you need to add this argument?

jdfreder: Widgets and WidgetModels don't have a notion of parent, this is as-designed.  However, DOM nodes have a notion of parent and children yet DOMWidgetViews do not.  By adding that, I'm adding the idea of a parent view to the DOMWidgetView.  I though this made sense, because the alternative would be to make the StyleView have an invisible el which gets attached to the DOM, and then use that to get the parent.  All of this is because the Style applies to the parent.

SylvainCorlay:
What do you mean regarding the Union?

jdfreder:
I was wondering if, for things like `.left`, I should use `Union(Int, Unicode)` or if `Any` is okay.

jdfreder:
I talked to @SylvainCorlay before he left for the airport and he said it was okay to merge this.

PR #221, opened 2015-10-31
Resize Dropdown and Colopicker to their full width

SylvainCorlay:
Colopicker and Dropdown now use all the available width. 

- The dropdown button and color-picker buttons now have the same width (28px).
- A new boolean attribute `short` is added to the color picker specifying whether the text box should be displayed or not.
- If the text box is displayed, the widget width is `@widget-width`. If not, it is `@widget-width-short`.
- In the short mode, if there is a description, the color picker button is justified to the right, and if there is only the color picker button to displayed, it is on the left.


SylvainCorlay:
@jdfreder green button?

jdfreder:
Sounds good! Thanks Sylvain!

PR #230, opened 2015-11-06
remove bower and build from prepublish step

minrk:
prepublish runs before install, so can't include bower or build. Renamed to `really-prepublish` to run when it's supposed to, and added the bower step as postinstall.

See npm/npm#3059 for discussion. This is deliberate behavior for some reason. Essentially, "prepublish" is not at all the step you would think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,14 +10,15 @@
     "url": "https://github.com/ipython/ipywidgets.git"
   },
   "scripts": {
+    "postinstall": "npm run bower",
     "build": "npm run css && npm run buildtests",
     "bower": "bower install --config.interactive=false",
     "precss": "rimraf ./ipywidgets/static/widgets/css/",
     "css": "node ipywidgets/build_css.js",
     "prebuildtests": "rimraf ./ipywidgets/tests/bin/tests",
     "buildtests": "tsc ./ipywidgets/tests/**/*.ts --outDir ./ipywidgets/tests/bin -d -m commonjs -t ES5",
-    "prepublish": "npm run bower && npm run build",
-    "publish": "npm run publish:pypi",
+    "really-prepublish": "npm run bower && npm run build",
+    "publish": "npm run really-prepublish && npm run publish:pypi",
     "publish:pypi": "python setup.py sdist upload && python setup.py bdist_wheel upload",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,14 +10,15 @@
...
   "scripts": {
+    "postinstall": "npm run bower",
     "build": "npm run css && npm run buildtests",
     "bower": "bower install --config.interactive=false",
     "precss": "rimraf ./ipywidgets/static/widgets/css/",
     "css": "node ipywidgets/build_css.js",
     "prebuildtests": "rimraf ./ipywidgets/tests/bin/tests",
     "buildtests": "tsc ./ipywidgets/tests/**/*.ts --outDir ./ipywidgets/tests/bin -d -m commonjs -t ES5",
-    "prepublish": "npm run bower && npm run build",
-    "publish": "npm run publish:pypi",
```
jdfreder: I understood prepublish was ran at install time for local installs.  I thought it ran after the dev-dependencies were installed, so executing bower and build is okay.  Did you find otherwise?

With the new setup, if I `git clone`, `cd`, and then `npm install`, the build step no longer is ran for me.

package.json
```diff
@@ -10,14 +10,15 @@
     "url": "https://github.com/ipython/ipywidgets.git"
   },
   "scripts": {
+    "postinstall": "npm run bower",
```
jdfreder: I agree this is needed, because we don't package bower contents, but I don't think the changes to prepublish and publish are needed.

jdfreder: Wait, now that I think some more, I *don't* think this is needed.  We package the compiled CSS which is all the bower dependencies are used for, no?

jdfreder:
I'm a little confused about whether this PR is actually needed or not, see comments above.  Did you encounter a problem that this solves?

package.json
minrk: Hm, I opened this PR because `npm install` failed with `no such command: bower`, but I re-ran it now after a clean and it's okay. Not sure what I had messed up.

minrk:
Today, I ran `npm install`, which failed due to missing bower. This fixed it, but when I more diligently clean out the repo and reinstall with master it's fine, so chalk it up to an unclean dev environment.

PR #231, opened 2015-11-06
fixes for binary state

minrk:
- avoid changing dict in-place when pulling out buffers
- workaround inability to send buffers in open message on Comm init (ipython/ipykernel#73)

There appears to be a test for this, but I cannot see how it would ever pass, since sending binary state can never work with master due to mutating the state dict while iterating through it.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -185,11 +185,19 @@ class Widget(LoggingConfigurable):
     def open(self):
         """Open a comm to the frontend if one isn't already open."""
         if self.comm is None:
+            state, buffer_keys, buffers = self._split_state_buffers(self.get_state())
+            
             args = dict(target_name='ipython.widget',
-                        data=self.get_state())
+                        data=state,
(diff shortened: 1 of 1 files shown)


minrk:
I'll make a separate PR that always serializes Bytes.

ipywidgets/widgets/widget.py
```diff
@@ -185,11 +185,19 @@ def __del__(self):
...
+                        data=state,
+            )
             if self._model_id is not None:
                 args['comm_id'] = self._model_id
+            
             self.comm = Comm(**args)
+            if buffers:
+                # FIXME: workaround ipykernel missing binary message support in open-on-init
+                # send state with binary elements as second message
+                self.send_state()
```
jdfreder: Eventually we will need to add support for binary messages in open on init, because sending state on comm open fixed a state race condition (of course, I failed to link to any issues in the original PR, *sigh* #84)

jdfreder:
Thanks!

PR #232, opened 2015-11-06
Always serialize Bytes with binary messages

minrk:
Requires a little extra checking to work on Python 2, due to str being bytes,
but should work well on Python 3.

Based on #231


jdfreder:
Thanks min!

PR #236, opened 2015-11-12
Use numbers abstract base class

jdemeyer:
Be compliant with [PEP 3141](https://www.python.org/dev/peps/pep-3141): replace code like `isinstance(foo, (int, float))` by `isinstance(foo, numbers.Real)`. This is in particular needed for better SageMath support.

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -37,14 +37,11 @@ def _get_min_max_value(min, max, value=None, step=None):
     if value is None:
         if not max > min:
             raise ValueError('max must be greater than min: (min={0}, max={1})'.format(min, max))
-        t = type(min)
-        try:
-            half = t((max - min) / 2)
-        except TypeError:
-            # Support for types which might raise TypeError when
-            # converting to type(min), such as SageMath's Integer.
-            half = t((max - min) // 2)
-        value = min + half
+        diff = max - min
+        value = min + (diff / 2)
+        # Ensure that value has the same type as diff
+        if not isinstance(value, type(diff)):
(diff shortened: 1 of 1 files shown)


jdfreder:
Looks like the tests themselves are failing at a lower level, in the widgets themselves.  We use traitlets to strictly type min and max to `int` and `float`, so I think the SageMath numbers are conflicting with that.  You'll either need to implement custom trait types that does the type checking how you propose here and then sprinkle those throughout the widgets, or probably better, open a PR against ipython/traitlets fixing the validation logic of Int and Float accordingly.

jdemeyer:
> Looks like the tests themselves are failing at a lower level, in the widgets themselves

Can you please tell me how to run the testsuite? I have no idea...

jdemeyer:
I spoke too quickly. How to run the tests is in `README.md`, not in `CONTRIBUTING.md` (which I would find more logical).

jdemeyer:
> I think the SageMath numbers are conflicting with that.

Why do you think that? I tested the `@interact` in SageMath and it works.

jdfreder:
> Can you please tell me how to run the testsuite? I have no idea...


> I spoke too quickly. How to run the tests is in README.md, not in CONTRIBUTING.md (which I would find more logical).

Another way is to peak at how our CI (Travis) does it, defined in  .travis.yml in the repo root.
`nosetests --with-coverage --cover-package=ipywidgets ipywidgets` runs the Python tests, which were failing.

> Why do you think that? I tested the @interact in SageMath and it works.

Awesome, glad it's working.  It was an erroneous conclusion on my part after a quick look at the test output.

PR #239, opened 2015-11-13
Download widget manager state

SylvainCorlay:
For now, this simply adds a button to download the widget manager state, but this should probably be in a menu. 

Looking at it, I realize that the widget manager has redundant information

```json
    "f7f010b9ef5f4352857009e34122f6e4": {
        "model_name": "LinearScaleModel",
        "model_module": "nbextensions/bqplot/LinearScaleModel",
        "state": {
            "_view_name": "LinearScale",
            "reverse": false,
            "msg_throttle": 3,
            "_model_name": "LinearScaleModel",
            "max": 1,
            "_view_module": "nbextensions/bqplot/LinearScale",
            "version": 0,
            "allow_padding": true,
            "min": 0,
            "_model_module": "nbextensions/bqplot/LinearScaleModel"
        },
        "views": []
    },
```

- The `views` property probably does not need to be there. Especially since it only contains the top-level views in cell output areas. 
- The `model_name` and `model_module`are both at the top level and in the `state` property.

### DIFF
--- a/ipywidgets/static/widgets/js/widget_state.js
+++ b/ipywidgets/static/widgets/js/widget_state.js
@@ -1,0 +1,31 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+// npm compatibility
+if (typeof define !== 'function') { var define = require('./requirejs-shim')(module); }
+
+define(["base/js/namespace"], function(Jupyter) {
+    "use strict";
+
+    var save_state = function() {
+        Jupyter.WidgetManager._managers[0].get_state().then(function(state) {
+            var data = "text/json;charset=utf-8," + encodeURIComponent(JSON.stringify(state, null, "    "));
+            var a = document.createElement("a");
+            a.download = "state.json";
+            a.href = "data:" + data;
+            a.click();
+        });
+    };
+
+    var load_extension = function() {
+        Jupyter.toolbar.add_buttons_group([{
+            id : 'widget_manager_state',
+            label : 'Download Widget State',
+            icon : 'fa-sliders',
+            callback : save_state
+        }]);
+    };
+
+    load_extension();
+
+});
--- a/ipywidgets/static/widgets/js/widget_state.js
+++ b/ipywidgets/static/widgets/js/widget_state.js
@@ -1,9 +1,6 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-// npm compatibility
-if (typeof define !== 'function') { var define = require('./requirejs-shim')(module); }
-
 define(["base/js/namespace"], function(Jupyter) {
     "use strict";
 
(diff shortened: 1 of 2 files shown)


minrk:
Like you said, this should go in a menu. The toolbar real-estate is much too precious to add something like this. You can start by defining the Action, then it's easy to add UI for it in various places, which can be done in another PR.

SylvainCorlay:
That's right.  I mostly that I am working on using custom widgets outside of the notebook and I needed a convenience thing to get the serialized state. I guess that in the end, we will have some sort of `embed` button which does that and more...

ipywidgets/static/widgets/js/widget_state.js
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+// npm compatibility
+if (typeof define !== 'function') { var define = require('./requirejs-shim')(module); }
+
+define(["base/js/namespace"], function(Jupyter) {
+    "use strict";
+
+    var save_state = function() {
```
jdfreder: Like @minrk suggests, you should probably implement this as an action.  Don't add it to the toolbar, and if you want, you don't even have to work out how to add it to the menu.  If implemented as an action, you'll be able to access it through the command palette.

ipywidgets/static/widgets/js/widget_state.js
```diff
@@ -0,0 +1,31 @@
...
+        Jupyter.WidgetManager._managers[0].get_state().then(function(state) {
+            var data = "text/json;charset=utf-8," + encodeURIComponent(JSON.stringify(state, null, "    "));
+            var a = document.createElement("a");
+            a.download = "state.json";
+            a.href = "data:" + data;
+            a.click();
+        });
+    };
+
+    var load_extension = function() {
```
jdfreder: Then you can get rid of this

ipywidgets/static/widgets/js/widget_state.js
```diff
@@ -0,0 +1,31 @@
...
+        Jupyter.toolbar.add_buttons_group([{
+            id : 'widget_manager_state',
+            label : 'Download Widget State',
+            icon : 'fa-sliders',
+            callback : save_state
+        }]);
+    };
+
+    load_extension();
```
jdfreder: And this

ipywidgets/static/widgets/js/widget_state.js
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+// npm compatibility
+if (typeof define !== 'function') { var define = require('./requirejs-shim')(module); }
```
jdfreder: If you want a class to be npm loadable you also have to add a reference to it inside index.js in addition to init.js.  However, I think it doesn't make sense to load this in a non-notebook context, so I suggest just removing these lines.

SylvainCorlay:
@jdfreder I created an action instead of a toolbar menu. Sorry if you spent time reviewing this that early. It was just a placeholder thing for me to get some content for my experiments with embedding bqplot in other web pages.

jdfreder:
This is great, thanks!  If it makes sense, we should remember to add something to the menubar

jdfreder:
@SylvainCorlay in #314 I add a "widgets" menu.  Would you object to me adding this action to that menu, so the user can click on it there?

SylvainCorlay:
Sure no problem. We need to merge the last PR for the action to work though.
S.

PR #240, opened 2015-11-13
Create loadClass method in widget manager to allow overloading

SylvainCorlay:
Since using custom widgets in static pages will require more shims, this allows overloading loadClass in specific implementations of the widget manager. 

All these should probably disappear when we switch to stealjs for everything, but for now, this allows me to transform module path strings...


jdfreder:
Thanks!

PR #243, opened 2015-11-19
Apply Python .style to `el` only

jdfreder:
@SylvainCorlay the styling will no longer rely on the kludgey update_attr method


SylvainCorlay:
Sorry I thought I had merged that already.

PR #246, opened 2015-11-20
Use backbone 1.2.0 to avoid backward-incompatible change between 1.2.0 and 1.2.3

SylvainCorlay:
The problem with 1.2.3 is that it deletes model ids by replacing them with the model state's `idAttribute`. This seems to have been solved in the dev version of backbone but the fix was not released yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "amdefine": "^1.0.0",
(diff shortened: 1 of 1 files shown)


jdfreder:
It is extremely annoying that backbone does not choose to follow semver strictly.

jdfreder:
Thanks!

PR #247, opened 2015-11-20
Use require-css for widget styling

SylvainCorlay:
- I use `require-css` for the styling of widgets and make the less file accessible under static so that other libraries may rely on `require-less`.
- I added an example with a jsdlink.


.npmignore
```diff
@@ -24,9 +23,6 @@ __pycache__
 *.py
 ipywidgets/widgets
 ipywidgets/tests
-.bowerrc
 .travis.yml
-bower.json
-build_css.js
```
jdfreder: I'm torn, I don't think we should be uploading these to npm

.npmignore
```diff
@@ -6,7 +6,6 @@ docs/man/*.gz
 docs/source/api/generated
 docs/gh-pages
 ipywidgets/static/components
```
jdfreder: I'm thinking, maybe we should upload this instead

package.json
```diff
@@ -10,10 +10,10 @@
     "url": "https://github.com/ipython/ipywidgets.git"
   },
   "scripts": {
-    "build": "npm run css && npm run buildtests",
+    "build": "npm run buildtests",
     "bower": "bower install --config.interactive=false",
+    "postinstall": "npm run bower",
```
jdfreder: hmm idem

.npmignore
SylvainCorlay: except for build_css which is completely removed since we don't build css anymore.

jdfreder:
I'm not sure how I feel about having people run bower on install from npm.

I don't really like the alternative though, where our package becomes bloated because we include the other package's source.

I'm not sure which is worse.

jdfreder:
Another question, what does this solve?  Is it just so that you have access to the less from within the browser?

SylvainCorlay:
Accessing the less from within other packages is one advantage.
Another is being able to include the css when loading the widget module - but this could be done with require-css too. It simplifies embedding the css with the web example IMO.

jdfreder:
> Accessing the less from within other packages is one advantage.

Hmm, can't that be done the same way that ipywidgets currently accesses less of other projects?

SylvainCorlay:
> >Accessing the less from within other packages is one advantage.
> Hmm, can't that be done the same way that ipywidgets currently accesses less of other projects?

Except for custom widget using require-less. Although the less source just needs to be available but not necessarily compiled on the fly.

SylvainCorlay:
@jdfreder I just added a new development example. This example requires a dumped `widget_state.json` which is typically created with the new notebook action downloading the widget manager state.

This example works with pretty much any widget manager state that is based only on ipywidgets, but here I just used two sliders and a jsdlink.

SylvainCorlay:
No idea why the tests are not passing. The only thing that is remaining otherwise is loading boostrap's css.

jdfreder:
> No idea why the tests are not passing. The only thing that is remaining otherwise is loading boostrap's css.

Looks like the less parser is incompatible with Phantom - it uses .bind which Phantom does not support.

SylvainCorlay:
Arg this is really too bad for Phantom... In any case, another solution is to use require-css, and pre-build the less. 

Besides, as per the conversation on gitter, I realized that we depend on jquery-ui's css for the styling.

ellisonbg:
😓

minrk:
Time to abandon phantom once and for all...

SylvainCorlay:
OK, the tests are now passing and the static widgets have styling! Let me know what you think.

SylvainCorlay:
I added a codemirror text editor with the static page example.

SylvainCorlay:
staticwidgets

SylvainCorlay:
This is ready to get in. #253 comes on the top of it.

minrk:
Awesome

PR #248, opened 2015-11-20
Make _css stateful!

jdfreder:
@ellisonbg @SylvainCorlay @jasongrout @minrk I'm very excited about the functionality introduced by this PR!  This PR does two things:

1. It solves the problem of `_css` not being stateful by making it stateful!  This means we can once again have nice helper methods for allowing users to use all of CSS.
2. It allows you to use pseudo selectors with css

Note, this PR makes no changes to the Python API.  I figured we could iterate on that in another PR.  I'm just excited that this may end the "I need to be able to style X, but I don't want to subclass widgets" issue.

This is marked as backwards incompatible because the fixed behavior is not the same as the broken behavior, and may change people's layouts who are dependent on the broken behavior.

## Action shot!!
(LICEcap is broken on my machine right now, probably because of my OReilly screen casting setup)

## Demo notebook
https://gist.github.com/jdfreder/d0ec16a8f699b86895f1

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -558,6 +558,22 @@ define(["nbextensions/widgets/widgets/js/utils",
              * Public constructor
              */
             WidgetViewMixin.initialize.apply(this, [parameters]);
+            this.id = utils.uuid();
+            
+            // Create and apply a unique style class name that can be used to
+            // style this view directly.
+            this.styleClassName = 'widget' + '-' + this.model.id + '-' + this.id;
+            this.el.className += ' ' + this.styleClassName;
+            
+            // Find or create a style tag for this widget view
+            this.styleNode = document.querySelectorAll('style.' + this.styleClassName);
+            if (this.styleNode && this.styleNode.length > 0) {
+                this.styleNode = this.styleNode[0];
+            } else {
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
:+1: this looks good to me. Not sure we should consider this backward incompatible.

SylvainCorlay:
Will merge in a bit unless someone comments before.

jdfreder:
Awesome!
I'm super excited about this getting merged.  I think we can build a cool API on top of this, especially once your observable dict traitlet is ready.

ipywidgets/static/widgets/js/widget.js
```diff
@@ -558,6 +558,22 @@ define(["nbextensions/widgets/widgets/js/utils",
              * Public constructor
              */
             WidgetViewMixin.initialize.apply(this, [parameters]);
+            this.id = utils.uuid();
+            
+            // Create and apply a unique style class name that can be used to
+            // style this view directly.
+            this.styleClassName = 'widget' + '-' + this.model.id + '-' + this.id;
+            this.el.className += ' ' + this.styleClassName;
```
SylvainCorlay: Arg @jdfreder this actually broke bqplot. If `el` is an svg element, we cannot set the className.

SylvainCorlay: Fixed in #260 .

PR #250, opened 2015-11-23
Deserialize Style widget for containers too

SylvainCorlay:
@jdfreder this solves some issues with container models. Besides, WidgetModel.serializers was always undefined because the serializers were added to the prototype...


jdfreder:
:+1:

PR #255, opened 2015-11-26
WIP Use deamdify instead of amdefine

SylvainCorlay:
deamdify is a browserify transform, which should make our browserify trickeries unnecessary.


SylvainCorlay:
well, it seems that it does not play well with AMD extensions like require-css.

jdfreder:
I also tried this at one point. I don't remember what the problem was

PR #256, opened 2015-11-27
DOM widget controller

SylvainCorlay:
Bug fix for controller widget.

Detail: The Controller widget's view is a `DOMWidgetView` but its model was only a `WidgetModel` instead of a `DOMWidgetModel`. With the recent changes with the `_css` attribute, the `DOMWidgetView` tried to access the unexistant `_css` attribute in its constructor, which caused JS errors.


jdfreder:
:+1:

PR #260, opened 2015-12-05
Fix SVG className access

SylvainCorlay:
@jdfreder #248 broke bqplot because we cannot set `className` on a SVG element (oddly).


jdfreder:
Thanks! :+1:

PR #261, opened 2015-12-07
Move package.json into ipywidgets

jdfreder:
This makes the JS portion of ipywidgets look more like a JS project's structure.

Suggested by @SylvainCorlay in a conversation last week.

PR #264, opened 2015-12-08
Create basic comm API shim for jupyter-js-services

jdfreder:
Here's a simple shim that allows use of jupyter-js-services via the old comm API.

- [x] shim
- [x] test it / create an example

@parente @blink1073


parente:
Sweet. Thanks for jumping on this. Definitely looks like what I was envisioning. 

So on our end we'd have to instantiate and pass the shim to the WidgetManager, plus alias the `services/kernels/comm` module import here (https://github.com/ipython/ipywidgets/blob/master/ipywidgets/static/widgets/js/manager.js#L8) so that the shim gets imported. I think.

/cc @lbustelo, @dalogsdon @jhpedemonte who are all closer to widgets / dahsboards JS than I have been

jdfreder:
> So on our end we'd have to instantiate and pass the shim to the WidgetManager, plus alias the services/kernels/comm module import here (https://github.com/ipython/ipywidgets/blob/master/ipywidgets/static/widgets/js/manager.js#L8) so that the shim gets imported. I think.
> 
> /cc @lbustelo, @dalogsdon @jhpedemonte who are all closer to widgets / dahsboards JS than I have bee

Actually, that's not what I was intending.  The WidgetManager you reference is intended for notebook use only, it'd surely break outside of the notebook.  Instead my suggestion is to look at examples/development/web2.  This will allow projects like that to have live widgets using jupyter-js-services.  Right now that example only shows widgets w/o comms.

I'll add another example to this PR clarifying what I mean and demonstrating jupyter-js-services powered widgets outside of the notebook.

jdfreder:
This is ready to go now!  See `ipywidgets/examples/development/web3`

cc @SylvainCorlay

rgbkrk:
yeah this is awesome

blink1073:
Looks like you had to poke into the private API, feel free to push back on what the API needs to look like.

lbustelo:
Pretty awesome indeed. 

Gino B.

parente:
@lbustelo and I poked around earlier while you were working up the new demo. Totally get it now. Implement a subclass of the base widget manager. Instantiate the shim. Override the necessary methods. Away you go. Looks great.

jdfreder:
> Looks like you had to poke into the private API, feel free to push back on what the API needs to look like.

Thanks, I know that ipywidgets actually uses some of the private methods of the old API too.  The private methods used in the shim may not be essential.  I'll have to dig deeper into the code to remember what the reasoning was for creating comms that are unregistered and unknown to the manager (the kernel, in jupyter-js-services's case).  It may be that no changes are needed to the new API when I drop the old comm API completely.

jdfreder:
@SylvainCorlay this PR also fixes some hiccups introduced in the examples by #261

SylvainCorlay:
@jdfreder thank you for fixing the example after #261.

If we use private APIs, this may be an indication that we need to do things otherwise!

jdfreder:
Actually, I just looked, we aren't using any private APIs in master right now, so we must have fixed that at some point.

SylvainCorlay:
Why moving the `examples` directory?

jdfreder:
@blink1073 the only lines that are causing me to use private APIs ATM are these: https://github.com/ipython/ipywidgets/blob/master/ipywidgets/static/widgets/js/manager.js#L260-L262  They create a comm and don't send a comm open message.  Looks like the new API supports this, as `connectToComm`, right?

jdfreder:
> Why moving the examples directory?

I had trouble with it complaining about missing dev-dependencies of ipywidgets otherwise.  Basically, in order for the funky relative dependency that we have to work, the sub projects have to be inside the parent project's directory.  

My theory is that npm walks outwards for node_modules folders looking for binary dependencies.  I found this out because I don't have `tsc` installed globally on my machine, and yet it's a local installed dependency of ipywidgets.  If the examples were outside of the parent folder, I'd get `tsc` not found errors.

SylvainCorlay:
> > Why moving the examples directory?
> I had trouble with it complaining about missing dev-dependencies of ipywidgets otherwise. Basically, in order for the funky relative dependency that we have to work, the sub projects have to be inside the parent project's directory.

> My theory is that npm walks outwards for node_modules folders looking for binary dependencies. I found this out because I don't have tsc installed globally on my machine, and yet it's a local installed dependency of ipywidgets. If the examples were outside of the parent folder, I'd get tsc not found errors.

It is weird because I did not have such issues with bqplot while requiring ipywidgets which was not in a parent project directory.

blink1073:
@jdfreder, `connectToComm` gives you a comm object, it is up to you whether to call `open` on it or not.

jdfreder:
> It is weird because I did not have such issues with bqplot while requiring ipywidgets which was not in a parent project directory.

I just double checked and it looks like I can move them back.  Now I think what I was seeing earlier was caused by a separate problem, which was that I hadn't ran `npm install` in the widgets directory yet.

jdfreder:
> @jdfreder, connectToComm gives you a comm object, it is up to you whether to call open on it or not.

Awesome, I shouldn't need any private members when I make the full transition to jupyter-js-services.  Also, I just checked, and the old API methods I'm emulating, that do need private members of jupyter-js-services, aren't even being used by the example, so I went ahead and removed them.  There shouldn't be any more private member usage now.

blink1073:
Great, glad to hear it.

ipywidgets/static/widgets/js/services-shim.js
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * Hooks callback object up with jupyter-js-services IKernelFuture
+     * @param  {IKernelFuture} future - jupyter-js-services IKernelFuture instance
+     * @param  {object} callbacks
+     */
+    Comm.prototype._hookupCallbacks = function(future, callbacks) {
+        if (callbacks) {
+            future.onReply = function(msg) {
+                if (callbacks.shell && callbacks.shell.reply) callbacks.shell.reply(msg);
+                // TODO: Handle payloads.  See https://github.com/jupyter/notebook/blob/master/notebook/static/services/kernels/kernel.js#L923-L947
```
blink1073: Do you think we should update KernelFuture to handle this?

SylvainCorlay:
Thanks for moving examples back. I have build scripts that fetch this directory and I wanted to be sure it is not moving.

SylvainCorlay:
LGTM!

jdfreder:
:tada:

blink1073:
Nicely done!

parente:
:boom: yay!

ipywidgets/static/widgets/js/services-shim.js
jdfreder: Honestly, I don't think so.  As you know, the channels are `shell`, `iopub`, and `stdin`.  The payload handler acts on shell messages.  In the spirit of minimal message processing at the kernel level and the 'hand the full message to the handler and allow it to decide what to do with it' thinking, it makes sense to *not* handle this in the kernel.

blink1073: Great, that was my reasoning as well.

PR #265, opened 2015-12-11
Remove skipStateRequest attribute and switch to comm-less models

SylvainCorlay:
Besides, the default_implementation of `_create_comm` now simply returns a rejected promise, so that we can always expect a promise.

### DIFF
--- a/examples/development/web/index.js
+++ b/examples/development/web/index.js
@@ -1,10 +1,11 @@
 
 // Create a widget manager instance.
 var WidgetManager = require('./manager').WidgetManager;
+var uuid = require("jupyter-js-widgets").uuid;
 
-document.addEventListener("DOMContentLoaded", function(event) { 
+document.addEventListener("DOMContentLoaded", function(event) {
     var manager = new WidgetManager(document.body);
-    
+
     /**
      * Helper function for creating and displaying widgets.
      * @return {Promise<WidgetView>}
@@ -15,19 +16,18 @@ document.addEventListener("DOMContentLoaded", function(event) {
         return manager.new_widget({
             model_name: 'WidgetModel',
             widget_class: 'ipywidgets.' + widgetType,
-            skipStateRequest: true
-            
+            model_id: uuid()
         // Create a view for the model.
-        }).then(function(model) { 
+        }).then(function(model) {
             console.log(widgetType + ' model created', model);
-            
+
             model.set({
                 _view_name: widgetType + 'View',
                 description: description || '',
                 value: value || null,
                 visible: true
             });
-            
+
             return  manager.create_view(model);
         }, console.error.bind(console)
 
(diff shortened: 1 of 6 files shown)


SylvainCorlay:
@jdfreder this removes `skipTestRequest` - and mostly removes code actually.

examples/development/web/index.js
```diff
@@ -15,19 +16,18 @@ document.addEventListener("DOMContentLoaded", function(event) {
...
             model_name: 'WidgetModel',
             widget_class: 'ipywidgets.' + widgetType,
-            skipStateRequest: true
-            
+            model_id: uuid()
         // Create a view for the model.
-        }).then(function(model) { 
+        }).then(function(model) {
             console.log(widgetType + ' model created', model);
-
```
jdfreder: If you're using atom, make sure to disable the whitespace package

SylvainCorlay: I am using `vi` and it was a voluntary change :)

ipywidgets/static/widgets/js/manager-base.js
```diff
@@ -386,13 +386,13 @@ define([
     };
 
     ManagerBase.prototype._create_comm = function(comm_target_name, model_id, metadata) {
-        throw new Error("Manager._create_comm not implemented");
+        return Promise.reject("No backend.");
     };
 
     ManagerBase.prototype._get_comm_info = function() {
-        throw new Error("Manager._get_comm_info not implemented");
+        return Promise.resolve({});
```
jdfreder: Shouldn't this be a Promise.reject ?

ipywidgets/static/widgets/js/manager-base.js
```diff
@@ -179,19 +178,21 @@ define([
...
-                // Requesting the state to populate default values.
-                if (options.skipStateRequest) {
-                    return model;
-                } else {                    
-                    return model.request_state();
-                }
+                return model.request_state();
             });
+        }, function() {
+            // Comme Promise Rejected.
```
jdfreder: s/Comme/Comm

ipywidgets/static/widgets/js/manager-base.js
```diff
@@ -179,19 +178,21 @@ define([
...
+            if (!options_clone.model_id) {
+                options_clone.model_id = utils.uuid();
```
jdfreder: Curious, should this be done inside new_model ?

SylvainCorlay: Maybe. 
I have not needed it in `new_model` yet...

ipywidgets/static/widgets/js/manager-base.js
SylvainCorlay: `ManagerBase.set_state` does everything in the resolve handler of the comm_info...

SylvainCorlay:
Done.

ipywidgets/static/widgets/js/manager-base.js
jdfreder: That's okay, usually when I suffix a class with "Base" I mean this *must* be inherited to work.

SylvainCorlay:
Rebased.

ipywidgets/static/widgets/js/manager-base.js
jdfreder: I think this needs to be `return Promise.reject("No backend.");` because `_get_comm_info` is a method that returns a list of the existing comms from the backend.

If your problem with this is that you feel `set_state` should work without `_get_comm_info` being implemented, then `set_state` in the base should be modified to work without, and the one in the notebook WidgetManager should be implement so that it does.

SylvainCorlay: Could we make this a different PR?

jdfreder: Sure, but I would either leave the error that was there before, or use return `Promise.reject("No backend.");` here.

SylvainCorlay: OK, will add this right away.

SylvainCorlay:
@jdfreder done!

jdfreder:
Thanks @SylvainCorlay !

PR #266, opened 2015-12-11
 WidgetModel.__super__.constructor should be called last

SylvainCorlay:
`__super__.constructor` is the function call `Model.initialize`. We need everything that is in constructor to be done before initialize otherwise, we cannot call model.set in initialize.

The reason is that we overload `Model.set` and our overloaded method uses attributes set in `constructor`.

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -52,7 +52,6 @@ define(["./utils",
              *      An ID unique to this model.
              * comm : Comm instance (optional)
              */
-            WidgetModel.__super__.constructor.apply(this);
             this.widget_manager = widget_manager;
             this.state_change = Promise.resolve();
             this._buffered_state_diff = {};
@@ -99,6 +98,7 @@ define(["./utils",
                 widget_manager.notebook.events.on('kernel_restarting.Kernel', died);
                 widget_manager.notebook.events.on('kernel_dead.Kernel', died);
             }
+            WidgetModel.__super__.constructor.apply(this);
         },
 
         send: function (content, callbacks, buffers) {


jdfreder:
Shoot, we may have problems when doing this in Babel and/or Typescript.  IIRC both require you to call super before anything else.

SylvainCorlay:
How come?

jasongrout:
Typescript doesn't insist that super is called first, only that it is called: http://www.typescriptlang.org/Playground#src=class%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%201%3B%0A%09%7D%0A%09a%3A%20number%3B%0A%7D%0A%0Aclass%20B%20extends%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%202%3B%0A%09%09super()%3B%0A%09%7D%0A%7D

jasongrout:
Besides, that is only for classes written in typescript.  This is about classes written using Babel's class structure, which may be very different.

jdfreder:
@jasongrout yeah, babel doesn't allow that - https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=class%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%201%3B%0A%09%7D%0A%7D%0A%0Aclass%20B%20extends%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%202%3B%0A%09%09super()%3B%0A%09%7D%0A%7D

jdfreder:
Regardless, we should just keep this in mind.  We can find a workaround if we need to when we need to.  :+1: to merge as-is.

SylvainCorlay:
@jdfreder isn't ES6 class `constructor` a completely different thing from backbone model's constructor method?

jdfreder:
> @jdfreder isn't ES6 class constructor a completely different thing from backbone model's constructor method?

Honestly, I don't know how backbone interprets its `constructor` key, but babel compiles constructor to what you'd expect: https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=class%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%201%3B%0A%09%7D%0A%7D%0A .  I'd assume backbone did the same, but you're saying it doesn't?

SylvainCorlay:
I just thought that it was a coincidental thing that both were called `constructor`.

jdfreder:
Possibly

jasongrout:
fyi, typescript does something similar to Babel: http://www.typescriptlang.org/Playground#src=class%20A%20%7B%0A%09constructor()%20%7B%0A%09%09this.a%20%3D%201%3B%0A%09%7D%0A%7D%0A

Actually, it does look like Backbone does something similar: http://backbonejs.org/docs/backbone.html#section-248

PR #267, opened 2015-12-11
Some changes to be able to run examples

lbustelo:
Basically using `npm link`

The only bad thing is that the dependency to ipywidgets is not explicit  in the examples json. I only did web3.

### DIFF
--- a/examples/development/web3/README.md
+++ b/examples/development/web3/README.md
@@ -5,8 +5,10 @@ a context other than the notebook.
 
 ## Try it
 1. Open a command line in the `ipywidgets/ipywidgets` subdirectory and run `npm install`.
-2. Cd into this directory and run `npm install`.
-3. Now open the `index.html` file.
+2. Do `npm link`
+3. Cd into this directory and run `npm link jupyter-js-widgets`.
+4. Run `npm install`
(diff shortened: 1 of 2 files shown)


lbustelo:
info about `npm link` https://docs.npmjs.com/cli/link

SylvainCorlay:
Why renaming the npm package to ipywidgets?

lbustelo:
Cause when doing the npm link it used the name in package.json. Since in code you guys do require('ipywidgets'), that was the easiest thing to change n

SylvainCorlay:
Ah, I see the remaining `require("ipywidget")` in web3, but it should be changed into
require("jupyter-js-widgets"), which will be the name on npm.

SylvainCorlay:
@lbustelo I opened #268 fixing the naming issue and checked that web3 works.

lbustelo:
After much trial and error... my issues were resolved by upgrading `npm`. I'm now at `3.5.2` and it worked. Previously I was running `2.7.4`,

jdfreder:
Thanks @lbustelo !  Sorry for the pain this caused :(

lbustelo:
no problem. Good excuse to dig around and learn

PR #268, opened 2015-12-12
Remaining `require("ipywidget")`

SylvainCorlay:
As reported by @lbustelo, there was a remaining occurence of `require("ipywidgets")` which was causing the web3 example to fail.

cc @jdfreder


jdfreder:
:+1:

PR #270, opened 2015-12-12
Notify front-end first

SylvainCorlay:
In the case where a trait change handler raises an exception, the front-end was not being notified of the changes since the front-end messages were sent after the trait notifications. This caused front-end and backend to be in inconsistent states.

The inline comments about validation were incorrect since cross validation should occur before `_notify_trait` is called anyway, using the `_[name]_validate` magic methods or the new `@validate` decorator in traitlets 4.1.

cc @dmadeka @rmenegaux @jdfreder.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -231,7 +231,7 @@ class Widget(LoggingConfigurable):
             Widget.widgets.pop(self.model_id, None)
             self.comm.close()
             self.comm = None
-    
+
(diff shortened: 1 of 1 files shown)


jdfreder:
So one of the downsides to the new karma based tests as-is, is that I haven't written support for round trip tests.  I think once I do, it will be really important to write tests for these kinds of things, because they are small behaviors that are really easy to get wrong!  Thanks for fixing this!

PR #276, opened 2015-12-22
some docstrings

### LINKED ISSUE #275: `__init__` needs a docstring
So that when users type `ipywidgets?`, at least *something* is printed...

We should *never* have top-level modules with zero documentation, never.

minrk:
Add docstring to top-level ipywidgets module and the integer widgets.

I think there are a *lot* more docstrings we need to add/improve, but this is a start.

closes #275

### DIFF
--- a/ipywidgets/__init__.py
+++ b/ipywidgets/__init__.py
@@ -1,3 +1,21 @@
+"""IPython widgets for Jupyter notebooks.
+
+IPython Widgets provide simple interactive controls in the notebook.
+Widgets provide a mechanism for creating intera
+Each Widget corresponds to an object in Python and Javascript,
+with controls on the page.
+
+To put a Widget on the page, you can display it with IPython's display machinery::
+
+    from ipywidgets import IntSlider
+    from IPython.display import display
+    slider = IntSlider(min=1, max=10)
+    display(slider)
+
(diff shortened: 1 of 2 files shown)


jdfreder:
Thanks min!

ipywidgets/__init__.py
```diff
@@ -1,3 +1,21 @@
+"""IPython widgets for Jupyter notebooks.
+
+IPython Widgets provide simple interactive controls in the notebook.
+Widgets provide a mechanism for creating intera
```
jdfreder: Whoops, forgot to delete this line?

ipywidgets/__init__.py
```diff
@@ -1,3 +1,21 @@
...
+Each Widget corresponds to an object in Python and Javascript,
+with controls on the page.
+
+To put a Widget on the page, you can display it with IPython's display machinery::
```
jdfreder: Double ::

jdfreder: Ah, rst

jdfreder:
Sent a PR - https://github.com/minrk/ipywidgets/pull/2

minrk:
Thanks, @jdfreder!

fperez:
Thanks so much, folks!!

PR #278, opened 2015-12-31
Alleviate silent install failure problems

### LINKED ISSUE #277: Install fails silently
The steps describe in the readme, specifically the `pip install -e .` step fails silently on some machines.

Reported by @fperez

### LINKED ISSUE #279: Issues with ipywidgets developer install and npm
So I'm trying to install ipywidgets in dev mode and I'm experiencing a bit of trouble. I started off by running `pip install -e .` and recieved the following image.

```
rebuilding js and css failed
missing files: ['/Users/safia/dev/ipywidgets/ipywidgets/static/widgets/css/widgets.min.css']
error: [Errno 2] No such file or directory: '/Users/safia/dev/ipywidgets/package.json'
```

Hm. So that's strange. It failed to build and it appears it is looking for `package.json` in the wrong place. So I go into `setup.py` and update line 199 to the following.

```
 return mtime(self.node_modules) < mtime(pjoin(repo_root, 'ipywidgets', 'package.json'))
```

By the way, why is it not this in the first place?

Then I run `pip install -e .` again and get the following error.

```
subprocess.CalledProcessError: Command '['npm', 'run', 'build']' returned non-zero exit status 1
```

Alright, so it looks like `npm run build` didn't do its job well. So I do `cd ipywidgets && npm run build`` and get the following error.

```
> node build_css.js

> lessc --include-path=./static/components ./static/widgets/less/widgets.less ./static/widgets/css/widgets.css
/Users/safia/dev/ipywidgets/ipywidgets/build_css.js:19
    process.stdout.write(p.stdout.toString());
                                 ^

TypeError: Cannot read property 'toString' of null
    at run (/Users/safia/dev/ipywidgets/ipywidgets/build_css.js:19:34)
    at Object.<anonymous> (/Users/safia/dev/ipywidgets/ipywidgets/build_css.js:28:1)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:475:10)
    at startup (node.js:117:18)
    at node.js:951:3
```

Alright, so it looks like there is something up with `build_css.js`, so I dive into the file. I add two `console.logs` to print out the value of `p`, run `node build_css.js`, and start investigating the output.

```
cleancss --source-map --skip-restructuring  -o ./static/widgets/css/widgets.min.css ./static/widgets/css/widgets.css
{ error: 
   { [Error: spawnSync cleancss ENOENT]
     code: 'ENOENT',
     errno: 'ENOENT',
     syscall: 'spawnSync cleancss',
     path: 'cleancss',
     spawnargs: 
      [ '--source-map',
        '--skip-restructuring ',
        '-o',
        './static/widgets/css/widgets.min.css',
        './static/widgets/css/widgets.css' ] },
  status: null,
  signal: null,
  output: null,
  pid: 4846,
  stdout: null,
  stderr: null,
```

So it looks like the troublesome command is the `cleancss` command. Hm. That's weird. `npm install` should have taken care of it. I run `cleancss` on the commandline and get the usual command not found error. So I run `npm install clean-css` and run `cleancss` and it doesn't work. Then I run `npm install -g clean-css` and rerun `node build_css.js` and it works well. Woot!

Then I go back and run `pip install -e .` and the install works well! Woot.

But I'm curious and have destructive tendencies so I removed the `ipywidgets` that I added to the `package.json` path earlier and I rerun the install and it fails. So I'm curious as to why this happens for me but persumably not for anyone else.

This isn't really an issue since I ended up solving the problem in the process of writing up the issue and going through the above steps, but I figured I should leave it here in case other users experience similar issues.

So long story short:
* Is there a way to ensure that `npm install cleancss` installs the command line program too?
* Why is the path for `package.json` in `setup.py` seemingly incorrect?

cc: @jdfreder

jdfreder:
Clarify development install requirements, and suggest a verbose install, so the installee can see what is going on behind the curtains.

closes #277
closes #279

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -135,7 +135,9 @@ class NPM(Command):
             return False
 
     def should_run_npm_install(self):
-        return self.has_npm() and time(self.node_modules) < mtime(pjoin(repo_root, 'package.json'))
+        package_json = pjoin(repo_root, 'ipywidgets', 'package.json')
+        node_modules_exists = os.path.exists(self.node_modules)
+        return self.has_npm() and (not node_modules_exists or mtime(self.node_modules) < mtime(package_json))
     
     def run(self):
         has_npm = self.has_npm()
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,9 @@ from glob import glob
 from os.path import join as pjoin
 from subprocess import check_call
 
+log.set_verbosity(log.DEBUG)
+log.info('setup.py entered')
+log.info('$PATH=%s' % os.environ['PATH'])
 
 repo_root = os.path.dirname(os.path.abspath(__file__))
 is_repo = os.path.exists(pjoin(repo_root, '.git'))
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -189,30 +126,37 @@ def initialize_options(self):
...
+    def has_npm(self):
+        try:
+            check_call(['npm', '--version'])
             return True
-        return mtime(self.node_modules) < mtime(pjoin(repo_root, 'package.json'))
+        except:
+            return False
+
+    def should_run_npm_install(self):
+        return self.has_npm() and time(self.node_modules) < mtime(pjoin(repo_root, 'package.json'))
```
minrk: s/time/mtime/

jdfreder: :heavy_check_mark:

jdfreder:
This is ready for review.

@captainsafia this contains the fixed path you found.

jdfreder:
Hey @minrk I talked to @ellisonbg about this, and he suggested, since it's only a development install, to ditch the logic that runs npm install inside pip install, and instead have the installees run npm install separate to pip install.  It sounded like a good idea to me, but I fear I may be overlooking something.  Does it sound okay to you too?  AFAIK non development installs will still be pip only, since the static files will be packed with the Python content.

cc @SylvainCorlay

jdfreder:
I should add, I *think* everything works right now, as-is.  Alternatively we could take the "it works, why change it" stance.

minrk:
Running npm install after pip can't work because then pup is installing an incomplete Python package.

minrk:
We have to switch over to completely separate installation (js never goes in the Python package dir) before we can do that.

jdfreder:
>  (js never goes in the Python package dir) 

We are definitely uploading most of the Javascript to PYPI, I just double checked.  However, we don't upload the `components` bower'ed directory.  I think what you say is best though, we'll just wait till the frontend is moved to jupyter-js-widgets.

minrk:
I didn't notice that `node_modules` had moved inside the Python package. That's certainly wrong, because those are *build* dependencies, not dependencies needed by the actual resulting package.

jdfreder:
@SylvainCorlay suggested that, because the structure of the npm package was atypical otherwise.  Also I think he had other reasons, but I'll let him comment on that.

minrk:
If we ditch the logic to run npm as part of pip, then `pip install git+...widgets` will no longer work, because it is necessary to build the js/css for a complete installation, so we can't do that until/unless installation of the js becomes a completely separate operation, writing to a different part of the filesystem.

PR #284, opened 2016-01-05
Call base ctor

SylvainCorlay:
`Flexbox`, `VBox` and `HBox` did not work anymore because the new `__init__` including the deprecation warning did not call the base class' `__init__`.


jdfreder:
Whoops, thanks!

PR #285, opened 2016-01-05
Hyphen-separated css attribute names and flexbox

SylvainCorlay:
This contains two bug fixes: 
- currently, css attributes specified in layout were not taken into account if specified before the widget is displayed
- hyphen-separated css attribute names did not work

Besides, I re-add the `VBox` and `HBox` helper functions, but use the layout widget in their implementation.

### DIFF
--- a/ipywidgets/widgets/widget_box.py
+++ b/ipywidgets/widgets/widget_box.py
@@ -75,6 +75,22 @@ class PlaceProxy(Proxy):
     selector = Unicode(sync=True)
 
 
+def VBox(*pargs, **kwargs):
+    """Displays multiple widgets vertically using the flexible box model."""
+    box = Box(*pargs, **kwargs)
+    box.layout.display = 'flex'
+    box.layout.flex_direction = 'column'
+    return box
+
+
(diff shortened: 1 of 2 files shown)


ipywidgets/widgets/widget_box.py
```diff
@@ -75,6 +75,22 @@ class PlaceProxy(Proxy):
...
 
+def VBox(*pargs, **kwargs):
+    """Displays multiple widgets vertically using the flexible box model."""
+    box = Box(*pargs, **kwargs)
+    box.layout.display = 'flex'
+    box.layout.flex_direction = 'column'
+    return box
+
+
+def HBox(*pargs, **kwargs):
```
jdfreder: These should still be deprecated, no?

SylvainCorlay: It is not using `FlexBox` anymore.

These helper functions are much more used than `Box` or `FlexBox` around me (which are almost never used directly). I don't think that we should deprecate it.

jdfreder: :+1:

PR #286, opened 2016-01-05
Fix issues with shorthand css properties

SylvainCorlay:
The `flex-flow` attribute is a shortcut containing the `flex-direction` and `flex-wrap`. 

When setting `flex-flow` to an empty string, this resets `flex-direction` to its default, that is `row`, even if the trait attribute for `flex-direction` was `column` which is a bug.

We actually don't need `flex-flow` which is only a shortcut to the two other properties so I am removing it.

### DIFF
--- a/ipywidgets/static/widgets/js/layout.js
+++ b/ipywidgets/static/widgets/js/layout.js
@@ -38,16 +38,12 @@ define([
                 'bottom',
                 'display',
                 'flex',
-                'flex_basis',
-                'flex_direction',
                 'flex_flow',
-                'flex_grow',
-                'flex_shrink',
-                'flex_wrap',
                 'height',
                 'justify_content',
                 'left',
                 'margin',
+                'overflow',
                 'padding',
                 'right',
                 'top',
@@ -88,9 +84,9 @@ define([
         },
 
         /**
-         * Get the the name of the style attribute from the trait name
+         * Get the the name of the css property from the trait name
          * @param  {string} model attribute name
-         * @return {string} css attribute name.
+         * @return {string} css property name.
          */
         css_name: function(trait) {
             return trait.replace('_', '-');
@@ -104,7 +100,7 @@ define([
          */
         handleChange: function(trait, value) {
             this.displayed.then(_.bind(function(parent) {
-                if (parent && value) {
+                if (parent) {
                     parent.el.style[this.css_name(trait)] = value;
                 } else {
                     console.warn("Style not applied because a parent view doesn't exist");
@@ -128,5 +124,7 @@ define([
         }
     });
 
-    return {LayoutView: LayoutView};
+    return {
+        LayoutView: LayoutView
+    };
 });
--- a/ipywidgets/widgets/widget_layout.py
+++ b/ipywidgets/widgets/widget_layout.py
@@ -6,14 +6,22 @@
 from .widget import Widget, register
 from traitlets import Unicode, CUnicode
 
-@register('IPython.Button')
+
 class Layout(Widget):
     """Layout specification
-    
+
     Defines a layout that can be expressed using CSS.  Supports a subset of
     https://developer.mozilla.org/en-US/docs/Web/CSS/Reference
+
+    When a property is also accessible via a shorthand property, we only
+    expose the shorthand.
+
+    For example:
+    - ``flex-grow``, ``flex-shrink`` and ``flex-basis`` are bound to ``flex``.
+    - ``flex-wrap`` and ``flex-direction`` are bound to ``flex-flow``.
+    - ``margin-[top/bottom/left/right]`` values are bound to ``margin``, etc.
     """
-    
+
     _view_name = Unicode('LayoutView', sync=True)
 
     # Keys
@@ -23,19 +31,15 @@ class Layout(Widget):
     bottom = CUnicode(sync=True, allow_none=True)
     display = CUnicode(sync=True, allow_none=True)
     flex = CUnicode(sync=True, allow_none=True)
-    flex_basis = CUnicode(sync=True, allow_none=True)
-    flex_direction = CUnicode(sync=True, allow_none=True)
     flex_flow = CUnicode(sync=True, allow_none=True)
-    flex_grow = CUnicode(sync=True, allow_none=True)
-    flex_shrink = CUnicode(sync=True, allow_none=True)
-    flex_wrap = CUnicode(sync=True, allow_none=True)
     height = CUnicode(sync=True, allow_none=True)
     justify_content = CUnicode(sync=True, allow_none=True)
     left = CUnicode(sync=True, allow_none=True)
     margin = CUnicode(sync=True, allow_none=True)
+    overflow = CUnicode(sync=True, allow_none=True)
     padding = CUnicode(sync=True, allow_none=True)
     right = CUnicode(sync=True, allow_none=True)
     top = CUnicode(sync=True, allow_none=True)
     visibility = CUnicode(sync=True, allow_none=True)
     width = CUnicode(sync=True, allow_none=True)
-    
+
--- a/examples/notebooks/Flexbox Layout.ipynb
+++ b/examples/notebooks/Flexbox Layout.ipynb
@@ -1,0 +1,187 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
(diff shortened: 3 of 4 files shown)


jdfreder:
I don't think we need to remove this trait.  Also, this heads us down the path where we are choosing a subset of the CSS layout arbitrarily.  How about checking the current value of the trait before setting it to `""`.  If the trait is undefined or null (w/e it is by default), don't reset to `""`.

SylvainCorlay:
@jdfreder consider the following case. 

```Python
VBox([Button(), Button()])
```
The initial value for the trait `flex-direction` is then `'column'` and for `flex-flow` it is `''`. 

In the view constructor, we loop over these properties: We first set `style['flex-direction']` to `'column'`, which immediately changes the `flex-flow`attribute.
In the next iteration of the loop, we meet the `flex-flow` attribute. Setting it to `''`, we effectively replace the value of the `flex-direction` from `column` to `''`, that is `row nowrap`. I don't think that we can work around that in a simple way.

SylvainCorlay:
For this reason, we cannot be stateful with css properties that are shorthand properties.

jdfreder:
Ah yes, I see.  This is the only shorthand property?

SylvainCorlay:
The only one that I noticed. There are others for which we don't expose the bound properties like `margin`.

SylvainCorlay:
We need to be consistent about shorthand properties: either we only expose shorthands or we only expose bound properties - everywhere.

In the flexbox css guide, regarding the flex property, they recommend using the shorthand.

ipywidgets/static/widgets/js/layout.js
```diff
@@ -104,7 +103,7 @@ define([
          */
         handleChange: function(trait, value) {
             this.displayed.then(_.bind(function(parent) {
-                if (parent && value) {
```
jdfreder: Why did you change this behavior?

ipywidgets/widgets/widget_layout.py
```diff
@@ -6,14 +6,14 @@
 from .widget import Widget, register
 from traitlets import Unicode, CUnicode
 
-@register('IPython.Button')
```
jdfreder: You can see what file I used as a template ;)

ipywidgets/static/widgets/js/layout.js
SylvainCorlay: I am reverting a change I did in #285. Otherwise, one cannot set a property back to `''`.

jdfreder: Actually, I think this should be `(value !== undefined || value !== null)`

jdfreder: If you make this change, and make sure the traits default to `None`, I think you don't have to remove the shorthand trait

SylvainCorlay:
@jdfreder this is ready for rereview.

ipywidgets/static/widgets/js/layout.js
```diff
@@ -36,18 +36,15 @@ define([
                 'align_items',
                 'align_self',
                 'bottom',
+                'border',
```
jdfreder: Do not add this, it's not layout

ipywidgets/widgets/widget_layout.py
```diff
@@ -6,36 +6,41 @@
...
     https://developer.mozilla.org/en-US/docs/Web/CSS/Reference
+
+    When a property is also accessible via a shorthand property, we only
+    expose the shorthand.
+
+    For example:
+    - ``flex-grow``, ``flex-shrink`` and ``flex-basis`` are bound to ``flex``.
+    - ``flex-wrap`` and ``flex-direction`` are bound to ``flex-flow``.
+    - ``margin-[top/bottom/left/right]`` values are bound to ``margin``, etc.
     """
```
jdfreder: Nice, I like this doc string :)

ipywidgets/widgets/widget_layout.py
```diff
@@ -6,36 +6,41 @@
...
-    
+
     _view_name = Unicode('LayoutView', sync=True)
 
     # Keys
     align_content = CUnicode(sync=True, allow_none=True)
     align_items = CUnicode(sync=True, allow_none=True)
     align_self = CUnicode(sync=True, allow_none=True)
+    border = CUnicode(sync=True, allow_none=True)
```
jdfreder: -1

ipywidgets/static/widgets/js/layout.js
SylvainCorlay: Well `border-width-[top/bottom/left/right]` is...

jdfreder: That's like arguing font size is layout because it can affect the width and height of the object depending on the contents.  I would not add this, it's primary function is to set border style.  Nor would I add border-width-top/bottom/left/right.  All of that stuff should be set by CSS and not accessible in `Layout`

SylvainCorlay: I included it because it is part of the [css box model](http://www.w3.org/TR/CSS21/images/boxdim.png).

jdfreder: Just because the box model accounts for it in its calculations, doesn't mean it should be used to conduct layout.  Instead padding or margin should be used.

SylvainCorlay: The nice demo notebook uses it for the wrapping example :(

SylvainCorlay: Besides, I like the idea of being able to group items together in a widget dashboard using a border, like this.

sliders
)

jdfreder: And I like the idea of being able to set background color, but that's not Layout. ;)

But seriously, you can achieve the same affect fairly easily by using CSS and `add_class`.

```
%%html
<style>.bordered { border: 2px solid grey; }</style>
```

and

```python
orient.add_class('bordered')
speeds.add_class('bordered')
```

SylvainCorlay: Yeah but it is not a python API anymore. It requires a magic...

jdfreder: I don't think we'll agree on this- we should probably try to get a second opinion: @lbustelo , @ellisonbg , or @jasongrout  , what are your opinions on adding `border` to the mutable `Layout` object?

SylvainCorlay: @all it is not only about adding border, but about only using shorthand css properties when there is one in order to avoid the issues discussed in this issue. border is the top-most shorthand property covering the border widths, which concern layout in my opinion.

jdfreder: Here's an idea, just use `border-width`, since it doesn't conflate the styling with the layout, and it itselft is a top-level shorthand: `https://developer.mozilla.org/en-US/docs/Web/CSS/border-width

SylvainCorlay: It was my initial idea, but the `border-width` is only used if `border-style` is not `none`. That is why I ended up using the `border` shorthand.

jdfreder: > It was my initial idea, but the border-width is only used if border-style is not none. That is why I ended up using the border shorthand.

So in order to use the layout part of border, you also need to set the styling part of border?  Now I definitely don't think border or border-width should be included.

SylvainCorlay: > > It was my initial idea, but the border-width is only used if border-style is not none. That is why I ended up using the border shorthand.

> So in order to use the layout part of border, you also need to set the styling part of border? 

That is right.

SylvainCorlay:
After one on one discussion with @jdfreder I removed the `border` attribute so that we can merge the bug fix and discuss this later.

jdfreder:
Thanks! :)

jdfreder:
Test failure looks unrelated, and actually a problem with karma.

SylvainCorlay:
Thanks, I was looking into the failures.

ellisonbg:
The case of border width is a subtle one. If it is part of the box model,
then we may want to include it in layout.

On Tue, Jan 5, 2016 at 4:35 PM, Sylvain Corlay <notifications@github.com>
wrote:

> Thanks, I was looking into the failures.
>
> —
> <https://github.com/ipython/ipywidgets/pull/286#issuecomment-169178709>.
>



-- 
Brian E. Granger

PR #289, opened 2016-01-06
interact - continuous update

denfromufa:
from gitter chat: https://gitter.im/ipython/ipywidgets

### DIFF
--- a/examples/notebooks/Using Interact.ipynb
+++ b/examples/notebooks/Using Interact.ipynb
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false
    },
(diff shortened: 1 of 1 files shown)


jdfreder:
Awesome, thanks!

SylvainCorlay:
It would be nice to strip the output of notebooks before committing it.

denfromufa:
Why output should be removed? Do you having testing for these notebooks?

SylvainCorlay:
Output it adds a lot of noise on git since it may change from execution to execution.

jdfreder:
@SylvainCorlay it's okay, I'm going to add a PR that will add a few more things to these docs, I'll strip the output there.

denfromufa:
OK, I'm adding the nested example as well. Continuous update as a second
button with True and False options.

SylvainCorlay:
@denfromufa @jdfreder thanks!

PR #290, opened 2016-01-06
Serializers should only be added to the model prototype

SylvainCorlay:
@jdfreder regarding the second commit, I don't know why we ever added serializers to the instance.


jdfreder:
I think that was my fault, in #219 , because of incorrect implementation https://github.com/ipython/ipywidgets/commit/974fadaf5822ae92271148d4e8511e1b93cc28b1

SylvainCorlay:
ok, but I was wondering why we look at `constructor.serializers` at all...

jdfreder:
I was trying to say I think it was an accident as a byproduct of aforementioned PR.

PR #292, opened 2016-01-06
Update Using Interact - traitlet link for continuous update as option

denfromufa:
outputs removed

### DIFF
--- a/examples/notebooks/Using Interact.ipynb
+++ b/examples/notebooks/Using Interact.ipynb
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "collapsed": false
    },
@@ -53,7 +53,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {
     "collapsed": false
    },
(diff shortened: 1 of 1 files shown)


jdfreder:
I merged this into 291

denfromufa:
BTW, the example I gave is different from the one you merged, my example continues on the topic of continuous updates:

```
def long_running_function(i,j):
    print(int(i),list(x for x in range(int(i)) if 
                str(x)==str(x)[::-1] and 
                str(x**2)==str(x**2)[::-1]))
    return

from ipywidgets import Checkbox
from traitlets import link
j_widget = Checkbox(description='Continuous Update',options=[False,True])
i_widget=FloatSlider(min=1e5, max=1e7, step=1e5)

def update_i_options():
    i_widget.continuous_update=j_widget.value
j_widget.on_trait_change(update_i_options, 'value')

interact(long_running_function,i=i_widget, j=j_widget)
```

jdfreder:
> I gave is different from the one you merged

Ah yes, sorry, I meant to comment.  It does not make sense to add a parameter to a function just to get interact to render a widget for it.  Instead it makes more sense to construct and display the additional widget separate to interact, using `from IPython.display import display; display(mywidget)`.  Since that is out of scope of interact, and gets into the widgets themselves, I thought it better to have a function that uses both of the parameters.

Additionally, I wanted to make sure a transform was applied to the value, otherwise it makes more sense to use a traitlet link, which neither of the examples actually used.

Lastly, because master is version 5.0dev, it should use the new API of traitlets, which is `.observe` instead of `.on_trait_change` which is deprecated.

jdfreder:
> from the one you merged

Also, I should mention, I merged you example as-is, and in a later commit changed it.  See [here](https://github.com/ipython/ipywidgets/commit/c17a2bf33c512720c4026df87c4254ce1f2556d0#diff-83f12380225702e4e15cd5c63ea681ec) is where your commit lives in master.  [And here](https://github.com/ipython/ipywidgets/commit/cacd95d1f2e87362327cba1e97f9028c95cfa249#diff-83f12380225702e4e15cd5c63ea681ec) is the commit I made that changes it.

PR #293, opened 2016-01-06
Only use prototype serializers

SylvainCorlay:


SylvainCorlay:
Hum, I think this broke a bunch of stuffs.

jdfreder:
Yay for breaking stuff :boom:

PR #295, opened 2016-01-06
No prototype serializers

SylvainCorlay:
@jdfreder this should solve the issues that I was seeing. Sorry for the noise.


jdfreder:
No prob, glad you figured it out!

PR #299, opened 2016-01-06
Use Jupyter namespace for widget registry

SylvainCorlay:
This registry was meant to replace the class name in the front-end `create_widget` function, in order to be language agnostic. Hence it makes sense for it to be name-spaced with Jupyter rather than IPython.


jdfreder:
Thanks!

PR #303, opened 2016-01-07
Cross validate frontend and backend widget versions at runtime

### LINKED ISSUE #103: Could not open comm
I installed Jupyter Notebook according to the [latest README](https://github.com/jupyter/notebook/blob/2b5ba10ce980912de1c456c665aee7af2a2e6211/README.md) then followed by installing ipywidgets  

When opening the [Widget Basics example](https://github.com/ipython/ipywidgets/blob/master/examples/Widget%20Basics.ipynb) notebook, the import cell worked fine, but the cells creating widgets failed.  These errors were shown in the browser console from executing `IntSlider()`:

```
Error: Could not open comm
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)
Error: Couldn't process kernel message
    at new Error (native)
    at Error.WrappedError (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12706:25)
    at reject (http://192.168.100.100:8888/static/notebook/js/main.min.js?v=5481c6cbbfa3491e70ffb5a7c79f89f7:12785:33)

```
@lbustelo

### LINKED ISSUE #148: Interact in notebook hangs if ipywidgets is not installed in python which runs the notebook
When you have the following situation:

- using a non-default kernel (eg a conda env) in jupyter notebook
- you installed `ipywidgets` in this environment (but not in the python which runs jupyter)

then you can import `ipywidgets` (since it is installed in that kernel), but when using `interact` the notebook just hangs (kernel is busy but does not do anything, restarting the kernel does not solve it).

In the webconsole of firefox, I can see that there is a message like 'ipywidgets is not installed'. When I installed `ipywidgets` also in the python which runs jupyter notebook, the problem is solved (and you get a message like "Loaded extension: widgets/notebook/js/extension. loaded widgets" in the webconsole)

The problem is that it seems that things should work (as you can import ipywidgets), but when using interact it just hangs. And it is not really obvious what the reason is. 
Would it be possible to detect this and eg raise a warning?

### LINKED ISSUE #235: Validate that static content version matches backend version on import
I see a lot of problems where people with envs that differ slightly from the simplest possible have trouble using ipywidgets because the front-end content doesn't align with the back-end content.  I think we should investigate checking/validating the versions of the two on import, and error/warn on mismatch.

jdfreder:
I expect this will save a lot of installation head scratching and come in handy for when we split the widgets repo into two pieces.

This also finishes implementing the backend widget registry.

closes #103
closes #148
closes #235

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -13,7 +13,7 @@ from IPython.core.getipython import get_ipython
 from ipykernel.comm import Comm
 from traitlets.config import LoggingConfigurable
 from traitlets.utils.importstring import import_item
-from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes, Bool
+from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes
 from ipython_genutils.py3compat import string_types, PY3
 
 
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -463,13 +463,13 @@ class Widget(LoggingConfigurable):
         """Called when `IPython.display.display` is called on the widget."""
         # Show view.
         if self._view_name is not None:
-            version_widget = get_version_widget()
+            validated = validate_version()
             
             # Before the user tries to display a widget.  Validate that the
             # widget front-end is what is expected.
-            if version_widget is None:
+            if validated is None:
                 self.log.warn('Widget Javascript not detected.  It may not be installed properly.')
-            elif not version_widget.validated:
+            elif not validated:
                 self.log.warn('The installed widget Javascript is the wrong version.')
             
             self._send({"method": "display"})
@@ -479,12 +479,15 @@ class Widget(LoggingConfigurable):
         """Sends a message to the model in the front-end."""
         self.comm.send(data=msg, buffers=buffers)
 
-@register('Jupyter.__version')
-class VersionWidget(Widget):
-    version = Unicode('4.1.0dev', help="Expected front-end version", sync=True)
-    validated = Bool(False, help="Version validated by the front-end", sync=True)
+
+_version_validated = None
+def handle_version_comm_opened(comm, msg):
+    """Called when version comm is opened, because the front-end wants to 
+    validate the version."""
+    def handle_version_message(msg):
+        _version_validated = msg['content']['data']['validated']
+    comm.on_msg(handle_version_message)
+    comm.send({'version': '4.1.0dev'})
     
-def get_version_widget():
-    for widget in Widget.widgets.values():
-        if isinstance(widget, VersionWidget):
-            return widget
+def validate_version():
+    return _version_validated
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -160,7 +160,6 @@ class Widget(LoggingConfigurable):
     msg_throttle = Int(3, sync=True, help="""Maximum number of msgs the
         front-end can send before receiving an idle msg from the back-end.""")
 
-    version = Int(0, sync=True, help="""Widget's version""")
     keys = List()
     def _keys_default(self):
         return [name for name in self.traits(sync=True)]
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -13,7 +13,7 @@ from IPython.core.getipython import get_ipython
 from ipykernel.comm import Comm
 from traitlets.config import LoggingConfigurable
 from traitlets.utils.importstring import import_item
-from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes
+from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes, Bool
 from ipython_genutils.py3compat import string_types, PY3
 
 
@@ -464,9 +464,28 @@ class Widget(LoggingConfigurable):
         """Called when `IPython.display.display` is called on the widget."""
         # Show view.
         if self._view_name is not None:
+            version_widget = get_version_widget()
+            
+            # Before the user tries to display a widget.  Validate that the
+            # widget front-end is what is expected.
+            if version_widget is None:
+                self.log.warn('Widget Javascript not detected.  It may not be installed properly.')
+            elif not version_widget.validated:
+                self.log.warn('The installed widget Javascript is the wrong version.')
+            
             self._send({"method": "display"})
             self._handle_displayed(**kwargs)
 
     def _send(self, msg, buffers=None):
         """Sends a message to the model in the front-end."""
         self.comm.send(data=msg, buffers=buffers)
+
+@register('Jupyter.__version')
+class VersionWidget(Widget):
+    version = Unicode('4.1.0dev', help="Expected front-end version", sync=True)
+    validated = Bool(False, help="Version validated by the front-end", sync=True)
+    
+def get_version_widget():
+    for widget in Widget.widgets.values():
+        if isinstance(widget, VersionWidget):
+            return widget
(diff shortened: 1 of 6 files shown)


jdfreder:
Marking as "behavior change" since other kernel authors will need to know about this new API

jdfreder:
Ready for review @SylvainCorlay

SylvainCorlay:
There is a `version` model attribute attribute in the base widget class. Do we still need this?

jdfreder:
That actually was supposed  to have an entirely different purpose.   I  added that to allow us to distinguish persisted state versions and versions of state for individual widgets.  It was a neat idea in principle, but we don't ever use it.   I'll remove it and we can re-add it when it becomes necessary.

jdfreder:
Ok, it's removed.

ipywidgets/widgets/widget.py
```diff
@@ -134,7 +134,11 @@ def _call_widget_constructed(widget):
     @staticmethod
     def handle_comm_opened(comm, msg):
         """Static method, called when a widget is constructed."""
-        widget_class = import_item(str(msg['content']['data']['widget_class']))
+        class_name = str(msg['content']['data']['widget_class'])
+        if class_name in Widget.widget_types:
+            widget_class = Widget.widget_types[class_name]
+        else:
+            widget_class = import_item(class_name)
```
SylvainCorlay: :+1:

SylvainCorlay: If you do this in this PR, I guess you could also change the widget controller to instantiate buttons and axis using the registry.

SylvainCorlay:
How does this behave when there is no kernel?

SylvainCorlay:
This makes me think of the suicidal widget workaround we had for the comm info. Somehow, it feels weird that we need to create a widget model to get info about the backend.

jdfreder:
> How does this behave when there is no kernel?

It's up to the manager to do the validation. So widget managers that are built to work without a backend, i.e. the web examples, simply don't call the validate method

jdfreder:
> This makes me think of the suicidal widget workaround we had for the comm info. Somehow, it feels wrong that we need to create a widget model to get info about the backend.

Alternatively I could create a comm, but as you said on Gitter, it's still a little awkward since there is a widget manager in the frontend, where this logic belongs, but no equivalent in the backend.

jdfreder:
Okay @SylvainCorlay , I'm using a comm for this now.  Do you think the comm target name of `jupyter.widget.version` is appropriate?  Or should we use a more general name, like `jupyter.widget.manager` in anticipation of more manager level messages in the future?  Or rename `jupyter.widget` to `jupyter.widgets.widget` and this to `jupyter.widgets.manager`?

ipywidgets/widgets/widget.py
```diff
@@ -13,7 +13,7 @@
 from ipykernel.comm import Comm
 from traitlets.config import LoggingConfigurable
 from traitlets.utils.importstring import import_item
-from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes
+from traitlets import Unicode, Dict, Instance, List, Int, Set, Bytes, Bool
```
SylvainCorlay: Why do you need to import Bool?

ipywidgets/widgets/widget.py
```diff
@@ -464,9 +463,31 @@ def _ipython_display_(self, **kwargs):
...
+
+
+_version_validated = None
+def handle_version_comm_opened(comm, msg):
+    """Called when version comm is opened, because the front-end wants to 
+    validate the version."""
+    def handle_version_message(msg):
+        _version_validated = msg['content']['data']['validated']
+    comm.on_msg(handle_version_message)
+    comm.send({'version': '4.1.0dev'})
```
SylvainCorlay: So this is the required js version for this backend?

SylvainCorlay:
Besides a small inline comment, this looks good to me. I don't really have an opinion about the naming.

ipywidgets/widgets/widget.py
jdfreder: Whoops, that was accidentally left over from when I used a widget implementation to check version.  It's removed now.

jdfreder:
Okay, I'll leave the naming as-is for this PR.

jdfreder:
:tada:

bmharsha:
I'm on windows, and trying to use interact on Firefox. I installed jupyter and ipywidgets using `pip install jupyter[all]` and also enabled ipywidgets as specified in the README. 

```
Could not cross validate the widget frontend and backend versions. Error: Timeout while trying to cross validate the widget frontend and backend versions.
Stack trace:
ManagerBase.prototype.validateVersion/</</<@http://localhost:8888/nbextensions/jupyter-js-widgets/extension.js?v=20160504124932:32297:25
```

How do I go about fixing this? Is there a way to check the backend and frontend versions of ipywidgets?

PR #304, opened 2016-01-08
Default values for widget model attributes

SylvainCorlay:
@jdfreder @parente here is the approach I am proposing for specifying default values for widget attributes on the javascript side.

I started with the base widget classes and the `_Bool` widgets.

TODO:
- [x] figure out what to do with the instance of layout widget.
- [x] check commit regarding buffered state diff.
- [ ] some widgets have the same models and view types, but *different default values*. Is it worth inheriting and specifying a custom model for these cases (range sliders vs sliders)

This is an API change from the perspective of custom widget authors who need custom models.

### DIFF
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -249,7 +249,7 @@ define(["./utils",
             // Handle when a widget is updated via the python side.
             this.state_lock = state;
             try {
-                WidgetModel.__super__.set.call(this, state);
+                this.set(state);
                 if (this._first_state) {
                     this.trigger('ready', this);
                     this._first_state = false;
@@ -314,10 +314,16 @@ define(["./utils",
              */
             var return_value = WidgetModel.__super__.set.apply(this, arguments);
 
-            // Backbone only remembers the diff of the most recent set()
-            // operation.  Calling set multiple times in a row results in a
-            // loss of diff information.  Here we keep our own running diff.
-            this._buffered_state_diff = _.extend(this._buffered_state_diff, this.changedAttributes() || {});
+            if (!this._first_state) {
+                // Backbone only remembers the diff of the most recent set()
+                // operation.  Calling set multiple times in a row results in a
+                // loss of diff information.  Here we keep our own running diff.
+                //
+                // However, we don't buffer the initial state comming from the
+                // backend or the default values specified in `defaults`.
+                // 
+                this._buffered_state_diff = _.extend(this._buffered_state_diff, this.changedAttributes() || {});
+            }
             return return_value;
         },
 
--- a/ipywidgets/static/widgets/js/widget.js
+++ b/ipywidgets/static/widgets/js/widget.js
@@ -39,6 +39,15 @@ define(["./utils",
     };
 
     var WidgetModel = Backbone.Model.extend({
+
+        defaults: {
+            _model_module: null,
+            _model_name: "WidgetModel",
+            _view_module: "",
+            _view_name: null,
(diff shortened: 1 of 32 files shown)


jdfreder:
From the sample you have here, I think this is great.

SylvainCorlay:
Test failures seem unrelated.

ipywidgets/static/widgets/js/widget_bool.js
```diff
@@ -10,8 +10,24 @@ define([
...
+        defaults: _.extend(widget.DOMWidgetModel.prototype.defaults, {
+            value: false,
+            description: "",
+            disabled: false,
+            _model_name: "BoolModel"
+        }),
+    });
+
+    var CheckboxModel = BoolModel.extend({
+        defaults: _.extend(BoolModel.prototype.defaults, {
```
jdfreder: This is the line that is causing the tests to fail.  You need to import underscore.

jdfreder:
> Test failures seem unrelated.

They are related, missing import, see above

ipywidgets/static/widgets/js/widget_bool.js
SylvainCorlay: fixed thanks!

SylvainCorlay:
Thanks. I will iterate on this PR to cover the other models.

ipywidgets/static/widgets/js/widget_bool.js
jdfreder: Thanks the new karma tests ;)

SylvainCorlay:
- The color picker widget has a recently added boolean attribute called `short` specifying whether it should take a full widget width and have a text input area or just have a picker button and have a short widget width. We could rename this to "concise".

jdfreder:
> We could rename this to "concise".

Sure

parente:
Ping @lbustelo

SylvainCorlay:
>> We could rename this to "concise".

> Sure

Done

SylvainCorlay:
I have to figure out how to define a default for the instance of layout widget from the js, the default being to instantiate a new one...

jdfreder:
Great work so far!  Looking forward to this one

SylvainCorlay:
Would love to have your opinion on the `buffered_state_diff` commit.

SylvainCorlay:
The issue was that the state from `defaults` was being added to `buffered_state_diff` at the time when the base constructor was called. Then the default values were sent back to the backend.

I changed the code in the following fashion: only buffer state diff after the initial state has been set.

ipywidgets/static/widgets/js/widget.js
```diff
@@ -548,7 +563,11 @@ define(["./utils",
         }
     };
 
-    var DOMWidgetModel = WidgetModel.extend({}, {
+    var DOMWidgetModel = WidgetModel.extend({
+        defaults: _.extend({}, WidgetModel.prototype.defaults, {
+            layout: undefined,
```
jdfreder: Isn't this supposed to be `null`?

jdfreder:
@SylvainCorlay one small comment

ipywidgets/static/widgets/js/widget.js
SylvainCorlay: Well I don't know. We don't allow `None` for the layout widget. 
In fact, even when building it from JavaScript, we should instantiate a layout widget. Not sure what is the best way to do it. That is why I used `undefined`.

jdfreder: That's okay for now, note to self: test edge cases of this.

PR #305, opened 2016-01-08
Box predefined styling was not updating correctly

SylvainCorlay:
It is now using the same method as for `Button`. Besides, there was no reason for `box_style` or `button_style` to allow `None`. There is also a couple of docstring re-formatting.


SylvainCorlay:
This is ready for review.

SylvainCorlay:
As soon as it is merged, I will rebase the default value stuff on it. Actually, I find a ton of minor issues when going over all the default values of all the widgets.

PR #309, opened 2016-01-09
Add border to layout model and use it in Flexbox example notebook

SylvainCorlay:
As discussed with @jdfreder and @ellisonbg.


jdfreder:
:+1:

PR #312, opened 2016-01-11
Use webpack instead of browserify

SylvainCorlay:
@jdfreder @blink1073


SylvainCorlay:
It seems to be working-ish. Two problems
- webpack complaines about `base/js/namespace`, `notebook/js/outputarea` and `services/kernels/comm` which it does not now where to fine.
- the tests rely on our definition of `define` via amdefine.

jdfreder:
I'm not sure I follow, why would the tests rely on `define` any more than the web examples?

SylvainCorlay:
I removed the amdefi e definition of define because webpack understands amd.

blink1073:
@SylvainCorlay, I think you need to use externals for those other modules: https://webpack.github.io/docs/library-and-externals.html

examples/development/web2/index.js
```diff
@@ -1,12 +1,13 @@
 var CodeMirror = require("codemirror");
-require("./node_modules/codemirror/lib/codemirror.css");
+require("css!./node_modules/codemirror/lib/codemirror.css");
```
blink1073: You don't need to specify the loader here, you can do so in the config: https://github.com/phosphorjs/phosphor-tabs/blob/master/example/webpack.conf.js#L14

examples/development/web2/index.js
```diff
@@ -1,12 +1,13 @@
...
+require("css!./node_modules/jupyter-js-widgets/static/widgets/css/widgets.min.css");
 require("./node_modules/codemirror/mode/python/python");
 
 var WidgetManager = require("./manager").WidgetManager;
 
 document.addEventListener("DOMContentLoaded", function(event) {
 
-    var code = require("./widget_code.json").join("\n");
+    var code = require("json!./widget_code.json").join("\n");
```
blink1073: Same here (use the json-loader)

jdfreder:
> I removed the amdefine definition of define because webpack understands amd.

Ah so this breaks usage in node?

blink1073:
Yes, but why would you use widgets in node?

jdfreder:
> Yes, but why would you use widgets in node?

I don't think there is a need to use widgets in node.  Right now I'm using it for unit testing only.  Originally this is why I avoided webpack, in anticipation of unit testing the widgets in node.  

@SylvainCorlay ,
A) Can you make this work for the karma tests?
B) How does this improve the current situation, other than removing `requirejs-shim` (huzzah!) ?

SylvainCorlay:
> A) Can you make this work for the karma tests?

I need to look into it. I am working on the default values thing at the moment.

> B) How does this improve the current situation, other than removing requirejs-shim (huzzah!) ?

The requirejs-shim thing is a nightmare when dealing with custom widget libraries like bqplot.

jdfreder:
@rgbkrk would this work for you in composition?

rgbkrk:
Well even in Electron apps, we'll have access to the DOM. That's not a big deal. Also, webpack is totally fine. I'm a big fan of the hot reload for React components.

examples/development/web2/index.js
rgbkrk: Thanks, good comments. I definitely prefer that the webpack config shows how to load rather than changing the `require` syntax.

SylvainCorlay: Yes, this was my first experiment with webpack. I am updating this afternoon!
One thing that I will have to figure out though is how to remove the `require-css` syntax used with requirejs.

rgbkrk:
One odd thing is that normally projects relying on a package like this would bundle it themselves for webpack.

jdfreder:
> A) Can you make this work for the karma tests?
> I need to look into it. I am working on the default values thing at the moment.

If you can get the tests to work somehow, I'd be really happy to use this approach and throw away that shim! :tada:

jdfreder:
Looks like #303 made this need a rebase

SylvainCorlay:
Done, remains to fix the tests, the rest works fine.

examples/development/web3/webpack.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  module: {
+    loaders: [
+      { test: /\.css$/, loader: "style-loader!css-loader" },
+      { test: /\.json$/, loader: "json-loader" },
+      { test: /\.(jpg|png|gif)$/, loader: "file" },
```
SylvainCorlay: This one is for the loading of jquery-ui which requires some images.

examples/development/web3/webpack.config.js
```diff
@@ -0,0 +1,19 @@
...
+      { test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/octet-stream" },
+      { test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, loader: "file" },
+      { test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=image/svg+xml" }
```
SylvainCorlay: These are for the loading of font-awesome.

examples/development/web3/webpack.config.js
rgbkrk: It's ok to put comments in the webpack config too, it's a regular JS file.

SylvainCorlay: Good point. Adding some comments inline

SylvainCorlay:
Thanks to @blink1073 the tests are now passing. The last item we need to decide on is how to load the main css in the case of the notebook.

I replaced the call to `require-css` with a manual loading of the css. The reason is that it does not seem that  requirejs allows to specify loaders like webpack.

blink1073:
Passing, but you can't dynamically load widgets because of `loadClass`, which will have to be replaced by `requirejs` or `System.import`, since Webpack has no way to resolve the import.

SylvainCorlay:
I moved some things from dependencies to dev dependencies (in a rebase).

examples/development/web/package.json
```diff
@@ -4,17 +4,19 @@
   "description": "Project that tests the ability to npm install jupyter-js-widgets within an npm project.",
   "main": "index.js",
   "scripts": {
-    "prepublish": "browserify -g node-lessify index.js --outfile index.built.js --debug",
+    "prepublish": "webpack index.js index.built.js",
```
jdfreder: nice, simple

rgbkrk: Well, it's the webpack.config.js that does all the heavy lifting. :wink:

jdfreder: o i c.  Not so simple :/

rgbkrk: Webpack is awesome, don't get me wrong.

examples/development/web3/webpack.config.js
```diff
@@ -0,0 +1,21 @@
...
+      { test: /\.css$/, loader: "style-loader!css-loader" },
+      { test: /\.json$/, loader: "json-loader" },
+      // jquery-ui loads some images
+      { test: /\.(jpg|png|gif)$/, loader: "file" },
+      // required to load font-awesome
+      { test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/octet-stream" },
+      { test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, loader: "file" },
+      { test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=image/svg+xml" }
```
jdfreder: Yikes

examples/development/web2/webpack.config.js
```diff
@@ -0,0 +1,21 @@
...
+      { test: /\.css$/, loader: "style-loader!css-loader" },
+      { test: /\.json$/, loader: "json-loader" },
+      // jquery-ui loads some images
+      { test: /\.(jpg|png|gif)$/, loader: "file" },
+      // required to load font-awesome
+      { test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/font-woff" },
+      { test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=application/octet-stream" },
+      { test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, loader: "file" },
+      { test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, loader: "url?limit=10000&minetype=image/svg+xml" }
```
jdfreder: Yikes

jdfreder:
Thanks !  Hopefully this makes things a little less painful for people who want to embed prior to ipywidgets 6.0

PR #314, opened 2016-01-12
Persist widgets as screenshots to the notebook json.

jdfreder:
Finally!  This PR introduces a work around for the inability to render widgets on GitHub, nbviewer, through nbconvert, etc...  A save hook is used to persist canvas renderings of the widget DOM elements to the notebook JSON prior to save.

Ping @SylvainCorlay , @ellisonbg, @captainsafia, @willingc (we can use this for docs), @blink1073 (similar to your export button in the matplotlib widget renderer)

### DIFF
--- a/ipywidgets/static/notebook/js/manager.js
+++ b/ipywidgets/static/notebook/js/manager.js
@@ -6,8 +6,9 @@ define([
     "backbone",
     "services/kernels/comm",
     "jupyter-js-widgets",
-    "../../components/html2canvas/dist/html2canvas"
-], function (_, Backbone, comm, widgets, html2canvas) {
+    "../../components/html2canvas/dist/html2canvas",
+    "./progress-modal"
+], function (_, Backbone, comm, widgets, html2canvas, progressModal) {
     "use strict";
     
     // Work around for a logging bug, reported in https://github.com/niklasvh/html2canvas/issues/543
@@ -128,6 +129,9 @@ define([
         this.deleteSnapshots();
         
         this._init_actions();
+        
+        // Initialize the widget screenshot rendering dialog.
+        this.progressModal = new progressModal.ProgressModal();
     };
     WidgetManager.prototype = Object.create(widgets.ManagerBase.prototype);
 
@@ -223,7 +227,6 @@ define([
         var notifier = Jupyter.notification_area.new_notification_widget('widgets');
         var buildSnapshotsAction = {
             handler: (function() {
-                console.log('clicked!');
                 this.updateSnapshots().then((function() {
                     notifier.set_message('Widgets rendered', 3000);
                 }).bind(this));
@@ -339,33 +342,68 @@ define([
      * @return {Promise<void>} success
      */
     WidgetManager.prototype.updateSnapshots = function() {
-        document.querySelector('#site').style.overflow = 'visible';
-        var cells = Jupyter.notebook.get_cells();
-        return Promise.all(cells.map((function(cell) {
-            var widgetSubarea = cell.element[0].querySelector(".widget-subarea");
-            if (widgetSubarea && widgetSubarea.children.length > 0) {
-                return new Promise((function(resolve) {
-                    html2canvas(widgetSubarea, {
-                        onrendered: function(canvas) {
-                            // Save the screenshot of the canvas to the widget-subarea
-                            var mimetype = "image/png";
-                            widgetSubarea.widgetSnapshot = {
-                                mimetype: mimetype,
-                                data: canvas.toDataURL(mimetype)
-                            };
+        var that = this;
+        
+        // Wait for the progress modal to show before continuing
+        return this.progressModal.show().then(function() {
+            
+            // Disable overflow to prevent the document from having elements
+            // that are scrolled out of visibility.
+            document.querySelector('#site').style.overflow = 'visible';
+            
+            // Render the widgets of each cell.
+            var progress = 0;
+            var renderPromise = Promise.resolve();
+            var cells = Jupyter.notebook.get_cells();
+            cells.forEach(function(cell, index) {
+                renderPromise = renderPromise.then(function() {
+                    var widgetSubarea = cell.element[0].querySelector(".widget-subarea");
+                    if (widgetSubarea && widgetSubarea.children.length > 0) {
+                        
+                        return that.progressModal.setText(
+                            'Rendering widget ' + String(index + 1) + '/' + String(cells.length) + ' ...'
+                        ).then(function() {
+                            return new Promise(function(resolve) {
+                                html2canvas(widgetSubarea, {
+                                    onrendered: function(canvas) {
+                                        
+                                        // Save the screenshot of the canvas to the widget-subarea
+                                        var mimetype = "image/png";
+                                        widgetSubarea.widgetSnapshot = {
+                                            mimetype: mimetype,
+                                            data: canvas.toDataURL(mimetype)
+                                        };
+                                        
+                                        resolve();
+                                    }
+                                });
+                            });
+                        }).then(function() {
+                            return that.progressModal.setValue(++progress/cells.length);
+                        });
+                    } else {
+                        if (widgetSubarea && widgetSubarea.widgetSnapshot) {
+                            delete widgetSubarea.widgetSnapshot;
                             
-                            resolve();
+                            return that.progressModal.setValue(++progress/cells.length);
                         }
-                    });
-                }).bind(this));
-            } else {
-                if (widgetSubarea && widgetSubarea.widgetSnapshot) {
-                    delete widgetSubarea.widgetSnapshot;
-                }
-            }
-        }).bind(this))).then((function() {
-            document.querySelector('#site').style.overflow = '';
-        }).bind(this));
+                    }
+                });
+            });
+            
+            // When all of the rendering is complete, re-enable scrolling in the
+            // notebook.
+            return renderPromise.then(function() {
+                document.querySelector('#site').style.overflow = '';
+            });    
+        
+        // When the entire process has completed, hide the progress modal.
+        }).then(function() {
+            return that.progressModal.hide();
+        }).then(function() {
+            that.progressModal.setText('Rendering widgets...');
+            return that.progressModal.setValue(0);
+        });
     };
     
     /**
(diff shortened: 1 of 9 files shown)


SylvainCorlay:
This is a great idea. I did not know about html2canvas! Does it handle svg?

jdfreder:
Yes, it should!

SylvainCorlay:
:tada:

jdfreder:
I should mention, this is WIP.

ipywidgets/static/widgets/js/manager.js
```diff
@@ -268,6 +288,96 @@ define([
...
+            var widgetSubarea = cell.element[0].querySelector(".widget-subarea");
+            if (widgetSubarea.children.length > 0 && widgetSubarea.widgetSnapshot) {
+                
+                // Get the last screenshot of the widget sub-area
+                var mimetype = widgetSubarea.widgetSnapshot.mimetype;
+                var screenshot = widgetSubarea.widgetSnapshot.data;
+                
+                // Create a mime bundle for the screenshot. Remove
+                // URL information, so only the b64 encoded data 
+                // exists, because that's what the notebook likes.
```
rgbkrk: and we should continue to question this assumption in future versions of both the message spec and the notebook

rgbkrk:
This is great. The on-save is a really neat idea.

ellisonbg:
Wow, does `html2canvas` work in the browser - IOW the live notebook - or just in node?

rgbkrk:
@ellisonbg Browserify (via wzrd.in at least) is certainly barfing on it, haven't tried anything else. If so :unamused:

willingc:
@jdfreder If you are at Cal Poly later this week, I would love to see it live. :surfer:

rgbkrk:
@willingc See https://gist.github.com/jdfreder/4909058126124ab46279

jdfreder:
> Wow, does html2canvas work in the browser - IOW the live notebook - or just in node

It works in the browser.



> @ellisonbg Browserify (via wzrd.in at least) is certainly barfing on it, haven't tried webpack. If so,

Not sure how you are importing it, but be aware it isn't an AMD module, it just pollutes the window with a `html2canvas` class.

rgbkrk:
Ok, cool. I assume they're just not providing commonjs then. I went to tinker and provide a link to a working sample to Brian, gave up at stack trace. Sorry for the alarm!

SylvainCorlay:
This solves the issue of RTD and GitHub rendering of notebooks which is pretty awesome.

Regarding nbviewer, I think that it should actually really support interactive widgets. 
After all, if it is ok to add a screenshot of the rendered widget in the notebook, it should also be ok to add information about the widget state which hopefully should be much smaller, so that widgets can be rendered dynamically on the page.

jdfreder:
Talked to @ellisonbg , and he suggested adding a render button for doing the HTML -> PNG conversion, since it is asynchronous.  I didn't know what icon to use, so I used a truck:

screen shot 2016-01-13 at 3 19 11 pm

Here it is. in action:

demo

Here's the output, when saved:

https://gist.github.com/jdfreder/067476b62ed0f0042f89

Open to suggestions for the icon, otherwise, this is good to go!

SylvainCorlay:
I already use an auto for `auto-run` :)

SylvainCorlay:
@jdfreder the problem you have with save event handler that have to be fully synchronous is also going to arise if you want to save the widget state when the notebook is saved. 

It was not the case before, but since we use (asynchronously loaded) custom serializers, we will have the same issue.

Hence, it is probably something we should tackle at the notebook level.

jdfreder:
> Hence, it is probably something we should tackle in the notebook project.

Yeah I agree, but for now, I think this should be a button.  Then, when that is fixed in the notebook project, I can open an accompanying PR here to do it on save.  However, one thing I discovered is that DOM -> PNG can be really slow with a lot of widgets, which may be annoying on save.

ipywidgets/static/widgets/js/manager.js
```diff
@@ -141,7 +170,7 @@ define([
          * ----------
          * load_callback: function()
          *      function that is called when the widget manager state should be
-         *      loaded.  This function should return a promise for the widget
+         *      loaded.  This function should return a prooutput_metadatamise for the widget
```
SylvainCorlay: ?

jdfreder: Thanks Atom!

minrk:
I don't think this should be in the toolbar, just the menubar (menubar should always be a superset of toolbar actions, anyway). Menubar space seems too precious for something like this.

jdfreder:
> don't think this should be in the toolbar, just the Menu (Menu should always be a superset of toolbar actions, anyway). Toolbar space seems too precious for something like this.
> 
> EDIT: Fixed super-confusing mixture of menubar/toolbar.

A menu is a really good idea, because in the not-so-distant-future we'll need a button for computing possible widget states.  Maybe we'll have more things too in the menu, related to managing widget state.

jdfreder:
Menu , done!

demo

ellisonbg:
Looks like some bugs with the rendering of some widgets...

screen shot 2016-01-15 at 2 06 59 pm

jdfreder:
@ellisonbg thanks for testing this.  I think there may be problems with the library I've selected to do the renderings of the widgets.  I'll look into alternatives

jdfreder:
I investigated some more, part of the problem was the overflow: scroll attribute of the #site div in the notebook.  Temporarily removing that during the rendering seems to work well. Radio buttons, shadows, and select controls styled as a list box do not render.

jdfreder:
Okay, some good news.  I update the html2canvas component to the latest version and a lot of the bugs in the renderings are gone! Radio buttons and checkboxes now render correctly, see: https://gist.github.com/jdfreder/a2773c690bc84f211e15

List boxes still have a bug where only the selected element is rendered, but I've determined this is a bug in html2canvas and will open an issue there.

ellisonbg:
Sounds like progress!

On Wed, Jan 20, 2016 at 1:49 PM, Jonathan Frederic <notifications@github.com
> wrote:

> Okay, some good news. I update the html2canvas component to the latest
> version and a lot of the bugs in the renderings are gone! Radio buttons and
> checkboxes now render correctly, see:
> https://gist.github.com/jdfreder/a2773c690bc84f211e15
>
> List boxes still have a bug where only the selected element is rendered,
> but I've determined this is a bug in html2canvas and will open an issue
> there.
>
> —
> <https://github.com/ipython/ipywidgets/pull/314#issuecomment-173371674>.
>



-- 
Brian E. Granger

jdfreder:
> Sounds like progress!

It is, good progress.  The rendering fixes added in the latest version seemed to have caused a slowdown in the rendering speed, significantly.  It totally makes sense to have this in a menu or triggered by a button now.  I'm going to add a progress bar that allows you to know how long the render will take.

jdfreder:
Here's what this looks like with a nice loading screen.  The loading screen is full screen because scrolling in the notebook needs to be temporarily disabled as the widgets are rendered.

loading

Thinking about this more, it probably makes sense for the menu option to be "Save with widgets".  The function will then automatically invoke a save after the widgets have been rasterized.  Saving using the normal save mechanism wouldn't fave with widget images.  

That way, the user would never accidentally save the notebook with old widget images, nor would the user have to perform two actions to save with widgets.  Opinions?

jasongrout:
just checking - to get the real widgets back, I would have to just re-execute everything again, and the images go away, and real widgets are there again?

jdfreder:
> just checking - to get the real widgets back, I would have to just re-execute everything again, and the images go away, and real widgets are there again?

No, actually the way it works now is that a custom metadata tag `metadata: {isWidgetSnapshot: true}` is added to the widget image outputs.  The widget Javascript, when loaded on the page, makes sure that those outputs are not visible.  Generating the snapshots leaves the live widgets as-is.

jdfreder:
I'm totally mystified as to why the Python tests fail in 0b2ff80 , seeing that I don't modify any Python in this PR.

@SylvainCorlay I rebased, added the download state action to the menu, and now save immediately after snapshot capture.  This is ready for review.  If you decide to test it, don't forget to rebuild your CSS!

jdfreder:
Demo:  

loading

ipywidgets/static/notebook/js/manager.js
```diff
@@ -5,10 +5,16 @@ define([
     "underscore",
     "backbone",
     "services/kernels/comm",
-    "jupyter-js-widgets"
-], function (_, Backbone, comm, widgets) {
+    "jupyter-js-widgets",
+    "../../components/html2canvas/dist/html2canvas",
```
SylvainCorlay: as soon as we separate jupyter-js-widgets, we should make those npm deps.

ipywidgets/static/notebook/js/manager.js
```diff
@@ -108,6 +121,20 @@ define([
         this.notebook.events.on('kernel_connected.Kernel', function(event, data) {
             validate();
         });
+        
+        // Delete the snapshots when the notebook has saved, failed to save, is
+        // loaded from disk, and when the widget manager is constructed.
+        this.notebook.events.on('notebook_saved.Notebook', this.deleteSnapshots.bind(this));
+        this.notebook.events.on('notebook_save_failed.Notebook', this.deleteSnapshots.bind(this));
+        this.notebook.events.on('notebook_loaaded.Notebook', this.deleteSnapshots.bind(this));
```
jdfreder: loaaaaaaaded

ipywidgets/static/notebook/js/manager.js
jdfreder: I added a comment so we don't forget

ipywidgets/static/notebook/js/manager.js
jdfreder: :heavy_check_mark:

SylvainCorlay:
A minor problem. You should probably disable scrolling while rendering the snapshots.

jdfreder:
> A minor problem. You should probably disable scrolling of the notebook while rendering the snapshots.

Last commit hides the scrollbars and fixes the scroll position.

SylvainCorlay:
:+1:

PR #315, opened 2016-01-13
handle non-inspectable functions in interact

### LINKED ISSUE #313: Allow `interact` to work when `inspect.signature` fails
Consider this example, that works:

```python

p = lambda x: print(x)
    
interact(p, x=10);
```

Yet the semantically equivalent:

```python
    
interact(print, x=10);
```

Fails with:

```
ValueError                                Traceback (most recent call last)
<ipython-input-5-24ed27fb4a18> in <module>()
----> 1 interact(print, x=10);

/Users/fperez/dev/ipython/ipywidgets/ipywidgets/widgets/interaction.py in interact(__interact_f, **kwargs)
    321         #        ...
    322         f = __interact_f
--> 323         w = interactive(f, **kwargs)
    324         try:
    325             f.widget = w

/Users/fperez/dev/ipython/ipywidgets/ipywidgets/widgets/interaction.py in interactive(__interact_f, **kwargs)
    193     kwargs = kwargs.copy()
    194 
--> 195     new_kwargs = _find_abbreviations(f, kwargs)
    196     # Before we proceed, let's make sure that the user has passed a set of args+kwargs
    197     # that will lead to a valid call of the function. This protects against unspecified

/Users/fperez/dev/ipython/ipywidgets/ipywidgets/widgets/interaction.py in _find_abbreviations(f, kwargs)
    147     """Find the abbreviations for a function and kwargs passed to interact."""
    148     new_kwargs = []
--> 149     for param in signature(f).parameters.values():
    150         for name, value, default in _yield_abbreviations_for_parameter(param, kwargs):
    151             if value is empty:

/Users/fperez/usr/conda/lib/python3.5/inspect.py in signature(obj, follow_wrapped)
   2987 def signature(obj, *, follow_wrapped=True):
   2988     """Get a signature object for the passed callable."""
-> 2989     return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
   2990 
   2991 

/Users/fperez/usr/conda/lib/python3.5/inspect.py in from_callable(cls, obj, follow_wrapped)
   2737         """Constructs Signature for the given callable object."""
   2738         return _signature_from_callable(obj, sigcls=cls,
-> 2739                                         follow_wrapper_chains=follow_wrapped)
   2740 
   2741     @property

/Users/fperez/usr/conda/lib/python3.5/inspect.py in _signature_from_callable(obj, follow_wrapper_chains, skip_bound_arg, sigcls)
   2228     if _signature_is_builtin(obj):
   2229         return _signature_from_builtin(sigcls, obj,
-> 2230                                        skip_bound_arg=skip_bound_arg)
   2231 
   2232     if isinstance(obj, functools.partial):

/Users/fperez/usr/conda/lib/python3.5/inspect.py in _signature_from_builtin(cls, func, skip_bound_arg)
   2058     s = getattr(func, "__text_signature__", None)
   2059     if not s:
-> 2060         raise ValueError("no signature found for builtin {!r}".format(func))
   2061 
   2062     return _signature_fromstr(cls, func, s, skip_bound_arg)

ValueError: no signature found for builtin <built-in function print>
```

From the user's perspective, this is pretty baffling.

minrk:
skips loading abbreviations from signature and argument validation when signature cannot be retrieved

closes #313

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -10,7 +10,12 @@ try:  # Python >= 3.3
     from inspect import signature, Parameter
 except ImportError:
     from IPython.utils.signatures import signature, Parameter
-from inspect import getcallargs, getargspec
+from inspect import getcallargs
+
+try:
+    from inspect import getfullargspec as check_argspec
+except ImportError:
+    from inspect import getargspec as check_argspec # py2
 
 from IPython.core.getipython import get_ipython
 from . import (Widget, Text,
@@ -203,7 +208,7 @@ def interactive(__interact_f, **kwargs):
     # that will lead to a valid call of the function. This protects against unspecified
     # and doubly-specified arguments.
     try:
-        getargspec(f)
+        check_argspec(f)
     except TypeError:
         # if we can't inspect, we can't validate
         pass
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -10,7 +10,7 @@ try:  # Python >= 3.3
     from inspect import signature, Parameter
 except ImportError:
     from IPython.utils.signatures import signature, Parameter
-from inspect import getcallargs
+from inspect import getcallargs, getargspec
 
 from IPython.core.getipython import get_ipython
 from . import (Widget, Text,
@@ -146,7 +146,13 @@ def _yield_abbreviations_for_parameter(param, kwargs):
 def _find_abbreviations(f, kwargs):
     """Find the abbreviations for a function and kwargs passed to interact."""
     new_kwargs = []
-    for param in signature(f).parameters.values():
+    try:
+        sig = signature(f)
+    except (ValueError, TypeError):
+        # can't inspect, no info from function; only use kwargs
+        return [ (key, value, value) for key, value in kwargs.items() ]
+
+    for param in sig.parameters.values():
         for name, value, default in _yield_abbreviations_for_parameter(param, kwargs):
             if value is empty:
                 raise ValueError('cannot find widget or abbreviation for argument: {!r}'.format(name))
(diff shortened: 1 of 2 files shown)


minrk:
cc @fperez who reported #313 

There is still an error in the print example because `print(x=5)` isn't a valid call, but at least it's the right error:

screen shot 2016-01-13 at 14 21 40

SylvainCorlay:
The test failure is legitimate. `getfullargspec` is new in Python 3.

minrk:
Thanks. Really looking forward to dropping support for legacy python...

minrk:
Arg, `getfullargspec` only works on Python 3, and `getargspec` *doesn't* work on Python 3. Need a switch to handle reasonable Python and Python 2, I guess.

minrk:
It would have been really nice if `getcallargs` raised a different error for bad arguments vs failure to inspect.

minrk:
should be working now.

jdfreder:
Eggs-cellent...

jasongrout:
A switch is apparently what astropy uses: https://github.com/astropy/astropy/blob/c7099d60a1d38b6074801d528ad069f4d06a8202/astropy/utils/decorators.py#L691

fperez:
On Fri, Jan 15, 2016 at 3:07 AM, Min RK <notifications@github.com> wrote:

> Thanks. Really looking forward to dropping support for legacy python...
>

I agree, and I think this year we should start doing so fairly aggressively.

However, the kernel + widgets repo are probably two where we should hold
off for a bit longer providing py2 support, miserable as that may be.
There will continue to be a lot of py2 usage for quite a while, and
offering people the 'full ipython experience', widgets included, would be a
good thing.

Eventually we may hit a point where we decide the cost of py2 support is
just too high for us even in those repos. But I don't think we're quite
there.

-- 
Fernando Perez (@fperez_org; http://fperez.org)
fperez.net-at-gmail: mailing lists only (I ignore this when swamped!)
fernando.perez-at-berkeley: contact me here for any direct mail

PR #319, opened 2016-01-14
Fix bug in comm based server side validation,

jdfreder:
which was causing "Widget Javascript not detected" to be
printed to the console despite the Javascript being
installed correctly

cc @SylvainCorlay

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -8,6 +8,7 @@ in the IPython notebook front-end.
 
 from contextlib import contextmanager
 import collections
+import sys
 
 from IPython.core.getipython import get_ipython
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
:+1: 
We could define this loud error outside of the function (and make it a loud warning :) ). For deprecations like #288, users should not have to enable warning manually.

jdfreder:
> For deprecations like #288, users should not have to enable warning manually.

I think the difference between this and that is that this also uses self.log.warn, which prints to the server's terminal window.  I do this because I want to make sure that the message is seen somewhere.

minrk:
@SylvainCorlay IPython should be showing deprecation warnings by default if they are caused directly by user code.

PR #321, opened 2016-01-15
Update ipykernel, traitlets and notebook

SylvainCorlay:
There is another PR coming with the traitlet API changes. This is a quick cleanup PR
- update notebook and traitlets to latest release. Remove the check for presence of `set_trait`.
- update ipykernel to fix the issue with json serialization of `np.int` type.
- remove the jstests.py file which is related to the old tests.


ipywidgets/jstest.py
```diff
@@ -1,58 +0,0 @@
...
-    return [ WidgetTestController(g, extra_args=options.extra_args) for g in groups ], []
-
-
-def main():
-    with patch.object(jstest, 'prepare_controllers', prepare_controllers):
-        jstest.main()
-
-
-if __name__ == '__main__':
-    main()
```
jdfreder: Fantastic, sorry I forgot to remove this earlier!

PR #323, opened 2016-01-17
Update max-height on dropdown-menu

### LINKED ISSUE #94: Dropdown Widget cut off at the end of the notebook

captainsafia:
Closes #94.

A better solution would be to increase the `min-height` on the `end_space` div at the end of the document to accommodate for longer dropdowns but I'm not sure if we just want to limit changes for this issue to ipywidgets.

dropdown-end-of-notebook

cc: @jdfreder

### DIFF
--- a/ipywidgets/static/widgets/less/widgets.less
+++ b/ipywidgets/static/widgets/less/widgets.less
@@ -298,7 +298,7 @@
     }
 
     .dropdown-menu {
-        max-height: 200px;
+        max-height: 120px;
         overflow: hidden;
         overflow-y: auto;
     }


jdfreder:
I see this works because of the end space div in the notebook.  I was actually thinking of using some Javascript to see if the dropdown would fit prior to displaying it, and if not, add a drop-up css class to the menu that causes it to display above the box instead of below it.  I'm not sure how complex the implementation would need to be to work with bootstrap, but it would also mean the use wouldn't have to scroll the notebook to see the contents of the dropdown.

jdfreder:
From the dev meeting, @captainsafia said she is investigating a JS powered solution. :)

captainsafia:
Yes! Sorry I did not post a note here.

SylvainCorlay:
Should we close this?

captainsafia:
The issue is still pending I'm working on a JavaScript solution.

SylvainCorlay:
:+1:

captainsafia:
Closing because #373 is a better alternative.

Thanks @jdfreder for implementing. :fireworks:

PR #326, opened 2016-01-19
Base css class for Jupyter widgets should not have 'ipy'

SylvainCorlay:
Removing references to IPython in jupyter-js-widgets Javascript.


jdfreder:
Thanks for doing this

PR #327, opened 2016-01-19
Traitlets 4.1 API

SylvainCorlay:
This first PR is doing two things
- moving the `sync=True` and `**widget_serialization` into the `.tag()` call
- using the observe API for interact.
- update the example notebooks where custom widgets are made to use the new traitlets API and the unique "jupyter-js-widgets" entry point for the javascript.

There are many other changes coming for the new traitlets API but I don't want to do everything in the same PR.
One reason is that I noticed that some things which should be done at traitlets cross-validation are done at the observe stage, which *will* cause race conditions in the case where the cross-validation yields a trait error.


minrk:
Don't forget to update requirements in setup.py

SylvainCorlay:
Yeah, I did it in a previous PR.

minrk:
Oops, didn't catch that. Thanks.

SylvainCorlay:
This is ready to be looked at.

examples/notebooks/Custom Widget - Hello World.ipynb
```diff
@@ -225,7 +226,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You first need to import the `widget` and `manager` modules.  You will use the manager later to register your view by name (the same name you used in the back end).  To import the modules, use the `require` method of [require.js](http://requirejs.org/) (as seen below).\n"
+    "You first need to import the `jupyter-js-widgets` module. To modules, use the `require` method of [require.js](http://requirejs.org/) (as seen below).\n"
```
jdfreder: Accidentally deleted the "To **import the **modules,"

SylvainCorlay: ok, replacing with `to import modules.`

examples/notebooks/Date Picker Widget.ipynb
```diff
@@ -817,21 +807,21 @@
    ]
   ],
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 2",
    "language": "python",
-   "name": "python3"
+   "name": "python2"
```
jdfreder: The example notebooks are all supposed to be Python 3

jdfreder: Luckily, I don't think you have to reopen and re-save every one.  Instead, you can probably get away with altering this by hand.

examples/notebooks/Custom Widget - Hello World.ipynb
jdfreder: Ah yes, that's better.

examples/notebooks/Date Picker Widget.ipynb
SylvainCorlay: Fixed. There were a couple of notebooks that were already set to Python 2, which I reverted to python. All squashed in earlier commits.

SylvainCorlay:
This is all ready.

jdfreder:
Sweet!  Thanks

PR #329, opened 2016-01-20
WIP: Update first_state switch logic

SylvainCorlay:
The problem with the old logic is that since were setting `_first_state` to true *after* the first call to set, any other call to set that was synchronously triggered by the first call to `_first_state` would still not be buffered to be sent to the backend. We now change the value of `_first_state` prior to calling `set`.

*I still have to test this in a couple of other scenarios, please do not merge.*


SylvainCorlay:
Will get back to this one tonight.

jdfreder:
> Will get back to this one tonight.

No rush, thanks!

SylvainCorlay:
Closing. This is now part of #383 .

PR #331, opened 2016-01-21
No need for request_state or _resolve_received_state mechanism anymore

SylvainCorlay:
`WidgetModel.request_state` was not called anymore. Removing the function and the promise resolving mechanism in _handle_msg.


jdfreder:
Hurrah for cleanup!

PR #332, opened 2016-01-21
SelectionSlider

rmenegaux:
`SelectionSlider` is a selection widget (inherits from `_Selection`) that works just like a Dropdown / Toggle / Radio, but with a slider interface. 

Right now this duplicates a lot of the slider code (the `render`, `updateDescription` and `update_attr` functions) from `widget_int.js` into `widget_selection.js`, which is pretty bad. I was thinking moving the base slider code into a new file `widget_slider.js` and then having the `IntSlider` and `SelectionSlider` inherit from that. 
What do you guys think?
@SylvainCorlay @jdfreder 

I also added an example in the `Widget List` notebook.

### DIFF
--- a/examples/notebooks/Widget List.ipynb
+++ b/examples/notebooks/Widget List.ipynb
@@ -496,6 +496,27 @@
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### SelectionSlider"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "widgets.SelectionSlider(\n",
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
:+1:

jdfreder:
Hi Roman, nice to see you here.  I think making a base class is a good idea.  Is that something you'd prefer to do here, in this PR, or later?

rmenegaux:
I'd rather re-organize that later in another PR, if that's ok with you.

jdfreder:
Yeah, let me give this a try, and I think we may be good to go.

jdfreder:
I see you don't have support implemented for the readout - that's okay, we can do that later.  Otherwise this looks good and works as expected.  I can tell from the source code that you weren't joking when you said this draws a lot from the existing slider... jquery soup!

jdfreder:
Thanks!

PR #333, opened 2016-01-21
Fixes

SylvainCorlay:
We merged #331 a bit quickly. No idea how I missed this. 

There were remaining calls to `request_state`. One in `new_widget`, which had to be removed anyway now that we now provide js default values, and another in `manager-base`.


jdfreder:
:+1:

PR #338, opened 2016-01-27
Use new traitlets API in domwidget.py and widget_selection.py

SylvainCorlay:
This replaces the calls to `on_trait_change` for widget styling and deprecation.


jdfreder:
:+1:

PR #344, opened 2016-02-01
Fix bug in button click cb sig for interact

### LINKED ISSUE #342: Callback for interact button is wrong
Related to the new traitlet API.  Running 

```python
interact(slow_function,i=FloatSlider(min=1e5, max=1e7, step=1e5),__manual=True)
```

yields 

```
<function __main__.slow_function>
TypeError                                 Traceback (most recent call last)
/home/jon/github.com/ipython/ipywidgets/ipywidgets/widgets/interaction.py in call_f(change)
    249     # Build the callback
    250     def call_f(change):
--> 251         name, old, new = change['name'], change['old'], change['new']
    252         container.kwargs = {}
    253         for widget in kwargs_widgets:

TypeError: 'Button' object is not subscriptable
```

jdfreder:
closes #342 

@SylvainCorlay I think this is related to the new traitlet API changes

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -247,8 +247,7 @@ def interactive(__interact_f, **kwargs):
     container.children = c
 
     # Build the callback
-    def call_f(change):
-        name, old, new = change['name'], change['old'], change['new']
+    def call_f(sender):
         container.kwargs = {}
         for widget in kwargs_widgets:
             value = widget.value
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
The problem is that we use `call_f` in the same way for both `button.on_click` (which oddly passes a button instance to it) and `observe` which calls it with the change dict.

jdfreder:
Ah okay, well the unpacked arguments weren't being used anyways.  I change this to *args to be clear that the arguments are unknown.

PR #352, opened 2016-02-05
Fix controller axis and button model names

SylvainCorlay:
Since we removed the state request and use js-side default values, we need to instantiate the right model from the start.


jdfreder:
:+1:

PR #353, opened 2016-02-07
Fix iopub callback invocation

parente:
Currently, every iopub handler is invoked for every message type in the shim. Here's a fix that also checks the message type.

Things I'm not sure about:

* Is iopub.output supposed to be invoked for display_data only? Or is it supposed to handle other messages as well?
* I didn't touch onReply since there's only one callback in there at the moment.

### DIFF
--- a/jupyter-js-widgets/src/services-shim.js
+++ b/jupyter-js-widgets/src/services-shim.js
@@ -203,8 +203,13 @@ Comm.prototype._hookupCallbacks = function(future, callbacks) {
                     callbacks.iopub.status(msg);
                 } else if (callbacks.iopub.clear_output && msg.msg_type === 'clear_output') {
                     callbacks.iopub.clear_output(msg);
-                } else if (callbacks.iopub.output && msg.msg_type === 'display_data') {
-                    callbacks.iopub.output(msg);
+                } else if (callbacks.iopub.output) {
+                    switch(msg.msg_type) {
+                        case 'display_data':
+                        case 'execute_result':
+                            callbacks.iopub.output(msg);
+                            break;
+                    };
                 }
             }
         };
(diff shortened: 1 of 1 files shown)


minrk:
I expect output should be called for both display_data and execute_result. Otherwise :+1:.

jdfreder:
@parente thanks for this!  I think this is good.  FWIW the shim was modeled after this, https://github.com/jupyter/notebook/blob/master/notebook/static/services/kernels/kernel.js#L891

parente:
Added handling of `execute_result` too.

> FWIW the shim was modeled after this

Does it make more sense then to have just one handler function per channel name instead of breaking it down by channel name and select message types?  Or are there clients already relying on the channel+type callbacks?

In our use, we wind up setting every iopub callback to the same function anyway, one that switches on message type.

jdfreder:
> Does it make more sense then to have just one handler function per channel name instead of breaking it down by channel name and select message types? Or are there clients already relying on the channel+type callbacks?

The manager inside the jupyter/notebook widget manager Javascript relies on the `callbacks` object behaving how it is described in kernel.js.  It probably makes more sense for this logic to live downstream, just because the shim is supposed to make jupyter-js-services behave exactly like that API.

I think you're the first consumer of this API, so I'm happy to iterate on things upstream (here).  I've merged this, but it'll make sense to revert it if we agree the logic should live downstream.

jdfreder:
> Currently, every iopub handler is invoked for every message type in the shim. Here's a fix that also checks the message type.

Actually, is it possible that this is a problem with jupyter-js-services ?

jdfreder:
@parente gah sorry for the noise, it's totally something I just missed when I was implementing it: https://github.com/jupyter/notebook/blob/master/notebook/static/services/kernels/kernel.js#L806

This is definitely something that should live here, upstream! thanks!

PR #359, opened 2016-02-09
Test the jupyter-js-widgets examples

jdfreder:
This should help prevent the web examples from going stale, and thus bring some stability for components relying on this tech @SylvainCorlay @parente @lbustelo


parente:
Nice. This will help.

PR #360, opened 2016-02-09
remove unneeded cd ../.. from npm test

minrk:
since these are run in subprocesses, cwd is not preserved


jdfreder:
Thanks!

minrk:
Oh, whoops. That's me working on too many PRs at once. Thanks!

PR #361, opened 2016-02-09
keep deprecated ipywidgets.find_static_assets

minrk:
to avoid breaking notebook 4.x startup


jdfreder:
Thanks!   I also need to make ipywidgets depend on widgetsnbextension

PR #363, opened 2016-02-09
Update test scripts so they distinguish between ff/ch for example tests

jdfreder:
Also update the script for Travis because it's calling the test script by the wrong name.


jdfreder:
Trigger rebuild...

jdfreder:
The is good to go

PR #373, opened 2016-02-16
Drop the dropdown upwards if it's too long.

### LINKED ISSUE #94: Dropdown Widget cut off at the end of the notebook

jdfreder:
closes #94 
alternative to #323

cc @captainsafia

### DIFF
--- a/jupyter-js-widgets/src/widget_selection.js
+++ b/jupyter-js-widgets/src/widget_selection.js
@@ -47,6 +47,7 @@ var DropdownView = widget.DOMWidgetView.extend({
             .addClass('dropdown-toggle')
             .addClass('widget-combo-carrot-btn')
             .attr('data-toggle', 'dropdown')
+            .click(this._showDropdown.bind(this))
             .append($('<span />').addClass("caret"))
             .appendTo(this.$buttongroup);
         this.$droplist = $('<ul />')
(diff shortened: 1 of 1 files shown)


captainsafia:
This looks great! Thanks for taking care of this. Apologies for slacking on it.

jdfreder:
Np!  It actually was a lot more straight forward than I anticipated

jupyter-js-widgets/src/widget_selection.js
```diff
@@ -60,6 +61,37 @@ var DropdownView = widget.DOMWidgetView.extend({
...
+    _showDropdown: function(e) {
+        // Don't allow bootstrap to show the dropdown!
+        e.stopImmediatePropagation();
+        e.preventDefault();
+
+        // Get the bottom of the dropdown label, and the bottom of the nb site.
+        // The difference is the maximum height of the dropmenu when displayed
+        // below the button.
+        var droplabelRect = this.$droplabel[0].getBoundingClientRect();
+        var siteRect = document.querySelector('#site').getBoundingClientRect();
```
jhpedemonte: `#site` is a Notebook-ism. We don't have that in the Dashboard Server. Is there a way to make this more generic?

jdfreder: Yes, I think so.  I believe, because of the nature of the notebook we can get away with subbing in document.body.

jdfreder: Thanks for catching this BTW!  It's easy for me to slip on things like this.  We should probably have some better non-notebook unit tests to prevent this....

jdfreder: #389

jhpedemonte: Thanks for the quick response. Yeah, non-notebook tests would be nice, but I understand that everyone is busy. Is there anyone else using the widgets in a non-notebook env? (Thebe is built on the notebook JS and copies the page structure, I believe)

blink1073: We could use the same class name across implementations (jupyter-js-notebook uses jp-Notebook)

SylvainCorlay: I have been tinkering with a gallery for bqplot using static widgets, but I need ipywidgets 5.0 to be released before moving forward with this.

A really good place to have a couple of static widgets would be ipywidgets' GitHub pages and the jupyter website itself.

PR #383, opened 2016-02-21
Allow initial values for model attributes to be specified on creation of the models

SylvainCorlay:
- The base widget model constructor now takes the standard backbonejs `attributes` argument so that we can create widget models with initial values.
- The private helper function `_deserialize_state` is now a class method. We need that so that we can deserialize state before instantiating the model, and pass the full deserialized state to the model constructors.
- A consequence is that `_deserialize_state` cannot take the model instance as a second argument anymore. *And therefore, the same is also true for the custom serializers*. Custom deserializers were actually not using this second argument, except the model deserialization (`IPY_MODEL_***` to model instance) which in fact only needed a reference to the widget manager. Hence, the second argument passed to `_deserialize_state` and the custom serializers is now the widget manager itself.
- The `_first_state` attribute has been removed from the base model, simplifying our overload of backbone.Model.set.
- The widget controller does not use the `ready` event that was removed since it is not needed anymore. This should solve the last issues with Controller widget persistence.

This is a negative contribution in the number of lines, so a good sign!

### DIFF
--- a/jupyter-js-widgets/src/widget.js
+++ b/jupyter-js-widgets/src/widget.js
@@ -9,7 +9,7 @@ var Backbone = require("backbone");
 var $ = require("./jquery");
 
 
-var unpack_models = function unpack_models(value, model) {
+var unpack_models = function unpack_models(value, manager) {
     /**
      * Replace model ids with models recursively.
      */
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
@jdfreder this seems to complete the blocking issue on persistence for widgets created from the front-end.

minrk:
I like the resulting centralization of deserialization here. :+1:

jdfreder:
Nice, this is great cleanup @SylvainCorlay !

blink1073:
:tada:

jupyter-js-widgets/src/widget.js
```diff
@@ -194,7 +172,7 @@ var WidgetModel = Backbone.Model.extend({
                         for (var i=0; i<buffer_keys.length; i++) {
                             state[buffer_keys[i]] = buffers[i];
                         }
-                        return that._deserialize_state(state);
+                        return that.constructor._deserialize_state(state, that.manager);
```
jhpedemonte: Shouldn't this be `that.widget_manager`?

jdfreder: You're right, see #388

SylvainCorlay: Thanks! :+1:

PR #392, opened 2016-02-24
Use MD instead of RST for the docs

jdfreder:
because it makes writing docs more fun for me and it's consistent with the daily GitHub workflow.

I have no intention to break RTD the docs support, but soon I want a github.io page for the widgets which will be the "prefered" rendering location of the docs.  This will allow @SylvainCorlay and I to have control over rendering widget demos in the docs.

@SylvainCorlay in a later PR I think I'm going to try http://bruth.github.io/jekyll-docs-template/ to create a simple GitHub.io page, what do you think?


SylvainCorlay:
@jdfreder this is awesome! Thanks for tackling the github pages!

I like the jekyll docs examples. Which part of the pages would you like to use it for?

minrk:
You can use markdown on RTD, and I think it makes sense to consolidate the docs in the same place as much as possible.

I *really* don't like including docs builds in the same repo as the code, since it adds a huge amount of unused bloat to clones. So if we decide gh-pages is preferable, it should be a dedicated repo.

jdfreder:
> I really don't like including docs builds in the same repo as the code, since it adds a huge amount of unused size to clones. So if we decide gh-pages is preferable, it should be a dedicated repo.

Agreed, I'm happy to go ahead and remove it in this PR.  I've been playing with http://ipython.org/ipywidgets/ (the gh-pages branch) and it seems like this will work well.  I'm a little worried about diverging from RTD because IPython lives there...  Maybe I'll try to figure out a way to redirect to http://ipython.org/ipywidgets/ nicely.

Thoughts?

jdfreder:
Shoot, with jekyll + github.io, I'm not quite sure how to handle multiple versions of the docs.  I think we at the very least need master + stable.

SylvainCorlay:
We just had a quick chat about this with @jdfreder .

Regarding the choice of Jekyll, another popular static site generator is Pelican (which is Python-based instead of Ruby-based). We might be more comfortable troubleshooting it since it is part of our ecosystem.

I saw that @rossant was using Pelican for his personal blog and has some blog entries on how to use it with Github pages.

@rossant, do you think that Pelican is a good fit for narrative documentation of a project? We will also embed some JavaScript in the page to create live interactive widgets in the examples...

minrk:
And what would be the benefits of those over using markdown in Sphinx? Which gets us automatic builds for all versions no hosting, no repos to deal with, common build tools, etc.?

If the goal is just to use markdown as the source format, I'm all for it, but there are a lot of benefits of sphinx+rtd that we don't have to lose if that's the only motivator.

minrk:
For reference, the [JupyterHub docs](https://github.com/jupyter/jupyterhub/tree/master/docs/source) are markdown for the most part, but since it's sphinx I can still use a couple of rst files for the autodoc API bits that don't need much manual writing.

jdfreder:
cc @willingc

rossant:
Do you know http://www.mkdocs.org/ ?
Pelican is good but might be overkill for documentation. I would favor sphinx, mkdocs or another documentation library over pelican if possible.

willingc:
@jdfreder @SylvainCorlay @minrk Let's begin discussing this tomorrow. I'm fine with Markdown as a source. I'm not ok, yet, with moving all ipywidgets docs off of RTD. While I think we should explore ideas for the best way to document the widgets, I'm -1 on having to maintain different documentation workflows for each project.

For now, let's begin determining what is needed for documenting the content, and worry less about whether to host on github.io/jekyll. Pelican, while I like it, would add significant overhead and maintenance of the docs.

While I see @minrk's point on docs adding to the github repo size, I think decoupling the doc content source from the repo would be a premature action and likely result in the docs falling out of date given the number of repos that we currently have.

Ideally, I would like to see us be able to support rst, md, ipynb, some of the cool js doc tools on a global Jupyter scope.

willingc:
I would fully support helping with development of @rossant's @ipymd to support the current notebook version since it would give us better support of notebooks within the sphinx docs.

jdfreder:
@minrk in response to your question

Benifits:

- Live unconnected widget support
- Control over page layout

Cons:

- Multiple versions are hard
- Automating doc releases isn't clear
- Not consistent with other RTD docs

Whether we continue to deploy on RTD or in a github.io page, we still have the problem that this repo really hosts 3 projects, two of which need serious narrative doc content and will be targeted to mostly entirely different audiences.

jdfreder:
@rossant I didn't know about mkdocs, but it looks great.  I particularly like this theme: https://github.com/chrissimpkins/cinder

willingc:
@jdfreder mkdocs is cool. @minrk and I looked at it when we were thinking about jupyterhub docs. Yet, I'm not yet convinced that it provides more than Sphinx/RTD. I've been playing around with nicer themes than the RTD theme for our RTD docs.

rossant:
I think you can use Markdown in RTD via mkdocs:
http://read-the-docs.readthedocs.org/en/latest/getting_started.html#in-markdown
http://read-the-docs.readthedocs.org/en/latest/builds.html?highlight=mkdocs#mkdocs

jdfreder:
Thanks @rossant !  If we roll our own theme, even if it's based off of an existing theme, it looks like we may be able to inject some Javascript on RTD.  The ability to put Javascript on the page is the big motivator of this discussion 
  - because it allows us to put widgets on the page  
    - which will allow us to demo things  
    - and showcase cool examples, like bqplot  
    - and show dynamic layout traits of the widgets  
  - and it allows the user to play with the APIs on the page  

I totally agree that moving off of RTD has a cost, both in reader experience because of consistency and project doc contributor workflow.

willingc:
@jdfreder Sounds like we are in sync. Havingmyou and @SylvainCorlay working through this with me will likely improve not just the ipywidget docs but all of the Jupyter docs. I have some ideas which I think with your knowledge will help us get there sooner :sunrise: 

@rossant Thanks for sharing the links re: mkdocs. You can also use markdown in RTD if you use recommonmark with sphinx. https://github.com/jupyter/jupyterhub/blob/master/docs/source/conf.py

jdfreder:
Okay, I've been able to get CDN content and embeded Javascript working.  Relative imports I don't have working yet.

willingc:
@jdfreder There's a new version of Sphinx released today. You might want to see if that helps.

SylvainCorlay:
In any case, the version of sphinx that comes in anaconda is a bit outdated at the moment

jdfreder:
So I've made a lot of progress here.  It's not quite as pretty as the jekyll-docs-template, but I was able to get local Javascript on the page and automagically replace .MD links with .HTML so that the docs will render on Github and on RTD as fancy HTML.  I'm working on cleaning things up aesthetically, then I'm going to work towards getting the widget.js in particular on the page.

jdfreder:
Here's a work in progress preview, which will update as I go: http://jons-widgets-fork.readthedocs.org/en/nicedocs/index.html

willingc:
@jdfreder Nice!

Here's a sample (very rough) of another doc theme (sphinx_bootstrap_theme) http://test-notebook.readthedocs.org/en/sphinx-alpha/

jdfreder:
Okay, I'm going to stop fiddling with the style a bit.  I'll start working on getting the jupyter-js-widgets javascript on the page tomorrow...

willingc:
@jdfreder You might find the theme preview on this helpful: https://pythonhosted.org/sphinxjp.themes.basicstrap/index.html Here's the repo: https://github.com/tell-k/sphinxjp.themes.basicstrap

jdfreder:
ooh I like that one @willingc !  I ended up pulling bootstrap into the one I have applied now.  

I'm currently trying to figure out the mechanics of sphinx with Markdown.  Specifically TOC and content structuring which Markdown lacks.  Also, still trying to figure out the best way to pull node content into the mix automagically.   I think a small Python script called by conf.py may do the trick...

willingc:
@jdfreder You can do TOC easily with RST. The rest of the content structuring we should be fine with markdown. Perhaps an addition to conf.py will work. Let's talk tomorrow/later. See the jupyterhub docs for the combination of RST for TOC and md for the rest.

jdfreder:
@minrk , @willingc filled me in on the JupyterHub docs.  Tomorrow we are going to work together to incorporate some of the stuff from there in addition to getting, building, and hosting the widget static JS and an example of a live widget.

We are also trying to figure out how/if notebook files play a role in this story.  Maybe docs -> notebooks, maybe notebooks -> docs, maybe notebooks !== docs, it's not clear.

We are also keeping our eyes open for clean, clear, crisp, easy on the eyes themes - but that's lower priority because we can change that fairly easily later down the line.

minrk:
Sounds great!

docs/source/_static/index.js
```diff
@@ -0,0 +1,82 @@
...
+
+// Load styling
+require('jquery-ui/themes/smoothness/jquery-ui.min.css');
+require("jupyter-js-widgets/css/widgets.min.css");
+
+const widgets = require('jupyter-js-widgets');
+console.info('jupyter-js-widgets loaded successfully');
+
+class WidgetManager extends widgets.ManagerBase {
+  displayWidgetState(models, el) {
```
jdfreder: renderWidgetState

willingc:
@jdfreder We're going to want to pin the sphinx dependency to 1.3.6+ or the "Edit on GitHub" link in the upper right will 404. Not sure that the feature is still in our custom template. Yet, the link failure was a Sphinx bug fixed in the 1.4alpha and backported to 1.3.6.

jdfreder:
Thanks @willingc !  I think that is actually not a part of our template.  IIRC it would need to be custom anyways because we are using MD and ipynb source files.

docs/source/dev_install.md
```diff
@@ -0,0 +1,73 @@
...
+=================
+
+<script>
+w({
+  "e43a868c5c0b4db39f0416d0fe9cc476": {
+    "model_name": "IntSliderModel",
+    "model_module": null,
+    "state": {
+      "_model_name": "IntSliderModel",
+      "_view_name": "FloatSliderView",
```
SylvainCorlay: You should not need to specify _model_name or _view_name here

docs/source/dev_install.md
```diff
@@ -0,0 +1,73 @@
+Developer Install
+=================
+
+<script>
+w({
+  "e43a868c5c0b4db39f0416d0fe9cc476": {
+    "model_name": "IntSliderModel",
+    "model_module": null,
```
SylvainCorlay: If you rebase this on master, you will probably need to replace the model-module with jupyter-js-widgets.

docs/source/dev_install.md
```diff
@@ -0,0 +1,73 @@
...
+    "state": {
+      "_model_name": "IntSliderModel",
+      "_view_name": "FloatSliderView",
+      "description": "testtesttesttesttesttesttesttesttesttest",
+       "min": 0,
+      "max": 4,
+      "value": 0,
```
SylvainCorlay: Same for `value` and `min`.

docs/source/_static/index.js
```diff
@@ -0,0 +1,82 @@
...
+      if (this._models[id]) {
+        displays.push(this._models[id].then(model => {
+          return this.display_model({}, model, { el })
+        }));
+      } else {
+        displays.push(this.new_model({
+          model_id: id,
+          model_name: models[id].model_name,
+          model_module: models[id].model_module,
+        }, models[id].state).then(model => this.display_model({}, model, { el })));
```
SylvainCorlay: I am not sure if we should necessarily display all the models in the state with `el` as the parent, because of the case of child widgets.

The widget_state object should have a notion of what are the models that are displayed at the top-level, and what are those that are only child widgets of the former.

SylvainCorlay: here: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/manager-base.js#L407

SylvainCorlay: Hence whatever widget model in our script tag is displayed as a top-level thing in the hierarchy should  have an attribute "views" like [here](https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/examples/web2/widget_state.json).

jdfreder:
@SylvainCorlay here I moved the embed manager into jupyter-js-widgets and updated the docs to use that.  I also added an embed helper to generate the `text/html` data for given widget state.

SylvainCorlay:
@jdfreder I think that the moved embed widget manager has the same issue as before, which is that you display all widgets top-level. Instead, we should use the `views` list which contains the root-level list of views, like in the web2 and web4 examples.

SylvainCorlay:
We could move the web4 widget manager in `jupyter-js-widgets` and make jupyter-js-widgets export the `w` function.

jdfreder:
@SylvainCorlay It doesn't make sense to export the w function without the manager because the manager has to be on the page for the w function to work.

SylvainCorlay:
> @SylvainCorlay It doesn't make sense to export the w function without the manager because the manager has to be on the page for the w function to work.

Yes, I meant *also* export the `w` function.

SylvainCorlay:
@jdfreder do you have a minute to appear.in?

widgetsnbextension/src/manager.js
```diff
@@ -421,12 +430,81 @@ WidgetManager.prototype._updateCellSnapshots = function(cell, index) {
...
+    } else if (state instanceof Object) {
+        var states = [];
+        _.each(state, function(value, key) {
+            states.push(that._traverseWidgetState(value));
+        });
+        return _.flatten(states);
+    } else {
+        return [];
+    }
+};
```
SylvainCorlay: At the moment, we only reference peer widget models that are listened to as as child widgets, which would ensure that this works, but it is not clear that it would always be the case.

README.md
```diff
@@ -1,23 +1,31 @@
...
+Learning becomes an immersive, plus fun, experience. Researchers can easily see how changing inputs to a model impact the results. We hope you will add ipywidgets to your notebooks, and we're here to help you get started.
+
+## Features
+A [demonstration notebook](https://github.com/ipython/ipywidgets/blob/master/examples/Index.ipynb) provides an overview of the interactive widgets, including:
 
+- sliders
+- progress bars
+- text boxes
+- toggle buttons and checkboxes
+- display areas
```
SylvainCorlay: gamepads!

README.md
```diff
@@ -29,82 +37,30 @@ or
...
+See the [examples](docs/source/examples.md) section of the documentation. The widgets are being used in a variety of ways; some uses can be seen in these notebooks:
 
-## Resources
-### ipywidgets
 - [Demo notebook of interactive widgets](https://github.com/ipython/ipywidgets/blob/master/examples/Index.ipynb)
-- [Documentation for ipywidgets](http://ipywidgets.readthedocs.org/en/latest/) [[PDF](https://media.readthedocs.org/pdf/ipywidgets/latest/ipywidgets.pdf)]
-- [Issues](https://github.com/ipython/ipywidgets/issues)
-- [Technical support - Jupyter Google Group](https://groups.google.com/forum/#!forum/jupyter)
+- Example 1
+- Example 2
```
SylvainCorlay: ?

jdfreder:
Okay @willingc and @SylvainCorlay , this is good to go now.  I'd like to follow it up with another PR where I add new content.  In its current state this PR uses the new embed widget manager, to allow the embedding of widgets into the docs, it also uses the new jupyter-sphinx-theme for both an improved style and the ability to integrate Javascript.

SylvainCorlay:
Arg, this comes with all the commits from `6.x`.

jdfreder:
Arg!  Good catch, I must have rebased against the 6.x branch instead of current master.

SylvainCorlay:
Note: I noticed that you are using ES6's `const` keyword in the embed manager.

SylvainCorlay:
Did you get a chance to do a rebase?

jdfreder:
I totally did, I'm having some *really* weird GH problems.  The 6.x commits shown here don't show up in my local git log!!

SylvainCorlay:
> I totally did, I'm having some really weird GH problems. The 6.x commits shown here don't show up in my local git log!!

Did you reset your local master to upstream/master?

jdfreder:
I did, I'm trying another machine now...

jdfreder:
Okay, it took some magic but I was able to repair the branch.  Before we merge, I'm going to squash some of these commits, there are too many!!!

SylvainCorlay:
So we are not using embedded widget yet right?

jdfreder:
@SylvainCorlay okay, everything is squashed and rebased.  To answer your question, we are including it, but not using it yet in this PR.   I removed the code in the page that was using it because it shouldn't have belonged there.

SylvainCorlay:
btw, I pushed to master the use of addEventListener instead of onload.

jdfreder:
> btw, I pushed to master the use of addEventListener instead of onload.

You should not push to master for things like that, only bumping versions for releases.  It helps me when you open PRs, even if super small, because then I can be notified of your changes and keep track of them for my own info.

SylvainCorlay:
>> btw, I pushed to master the use of addEventListener instead of onload.

> You should not push to master for things like that, only for bumping versions for releases. It helps me when you open PRs, even if super small, because then I can be notified of your changes and keep track of them for my own info.

Sorry, it was outstanding from #450 as per comments.

jupyter-js-widgets/src/manager-base.js
```diff
@@ -426,11 +432,15 @@ ManagerBase.prototype.set_state = function(state, displayOptions) {
...
             // Display the views of the model.
             if (state[model.id] !== undefined) {
-                return Promise.all(_.map(state[model.id].views, function(options) {
-                    // Display the model using the display options merged with the
-                    // options.
-                    return that.display_model(undefined, model, Object.assign({}, options, displayOptions));
-                }));
+                // Display the model using the display options merged with the
+                // options.
+                if (displayOptions.displayOnce && state[model.id].views) {
```
SylvainCorlay: What is the role of displayOnce? This seems to be a new option.

jdfreder: Changes the default set_state behavior from displaying a widget view for every "view" in the state to only once.

SylvainCorlay: :+1:

SylvainCorlay:
I think that it is worth merging now. It is going to be a pain to rebase again...

SylvainCorlay:
:shipit: :+1: :tada: :racehorse:

willingc:
Yay!!!!! :cake: for @jdfreder, @SylvainCorlay, and the team

widgetsnbextension/src/manager.js
```diff
@@ -533,6 +533,141 @@ WidgetManager.prototype._rasterizeEl = function(el) {
 };
 
 /**
+ * Update the widget snapshots for a single cell
+ * @param  {object} cell
+ * @param  {number} index - of the cell
+ * @return {Promise}
+ */
+WidgetManager.prototype._updateCellSnapshots = function(cell, index) {
```
SylvainCorlay: @jdfreder this seems to be a rebase issue. All these functions already exist in the file elsewhere.

jdfreder: I can take a look at this today

PR #393, opened 2016-02-25
IE11 svg does not support classList

### LINKED ISSUE #376: IE11 svg and classList
With IE11, svg elements don't support classList (we use it in `addClass` / `removeClass`). Besides, `className` is read-only.

SylvainCorlay:
Fixes #376.


jasongrout:
Thanks!

PR #395, opened 2016-02-25
Move callbacks() to notebook manager specialization

SylvainCorlay:


jupyter-js-widgets/src/manager-base.js
```diff
@@ -67,6 +67,14 @@ ManagerBase.prototype.loadClass = function(class_name, module_name, registry) {
     return utils.loadClass(class_name, module_name, registry);
 };
 
+ManagerBase.prototype.setViewOptions = function(options) {
```
jdfreder: This is a great idea!

PR #399, opened 2016-03-01
Remove comm:dead and comm:live events

SylvainCorlay:
This is an attempt to remove the `comm:[dead/live]` event bubbling mechanism.

### DIFF
--- a/widgetsnbextension/src/extension.js
+++ b/widgetsnbextension/src/extension.js
@@ -43,13 +43,6 @@ var handle_cell = function(cell) {
     if (cell.cell_type==='code') {
         var area = new widgetarea.WidgetArea(cell);
         cell.widgetarea = area;
-        var died = function () {
-            area._widget_dead();
-        }
-        cell.notebook.events.on('kernel_disconnected.Kernel', died);
-        cell.notebook.events.on('kernel_killed.Kernel', died);
-        cell.notebook.events.on('kernel_restarting.Kernel', died);
-        cell.notebook.events.on('kernel_dead.Kernel', died);
     }
 };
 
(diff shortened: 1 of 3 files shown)


widgetsnbextension/src/extension.js
```diff
@@ -43,6 +43,13 @@ var handle_cell = function(cell) {
     if (cell.cell_type==='code') {
         var area = new widgetarea.WidgetArea(cell);
         cell.widgetarea = area;
+        var died = function () {
+            area._widget_dead();
+        }
+        cell.notebook.events.on('kernel_disconnected.Kernel', died);
+        cell.notebook.events.on('kernel_killed.Kernel', died);
+        cell.notebook.events.on('kernel_restarting.Kernel', died);
+        cell.notebook.events.on('kernel_dead.Kernel', died);
```
SylvainCorlay: I should probably call `off` on disposal of the widget area.

jdfreder: +1, I don't think we were doing that before, but it's probably good to add if it's not too much effort.

jdfreder: Oh, I see, we were doing it before

jdfreder: Can you add it in a cb for the widget dispose event?

SylvainCorlay: You mean the disposal of the area?

widgetsnbextension/src/widgetarea.js
```diff
@@ -96,10 +92,6 @@ WidgetArea.prototype._create_elements = function() {
                 for (var i = 0; i < that.widget_views.length; i++) {
                     var view = that.widget_views[i];
                     view.remove();
-
-                    // Remove widget live events.
-                    view.off('comm:live', that._widget_live);
-                    view.off('comm:dead', that._widget_dead);
```
jdfreder: @SylvainCorlay I read quickly and mistaked these lines as the old unregister point.

widgetsnbextension/src/extension.js
jdfreder: Yes, sorry

SylvainCorlay:
This is ready for review.

jdfreder:
Thanks!

PR #400, opened 2016-03-01
Partially Remove jquery

dwillmer:
This is open for comments at Sylvain's request.

- [X] Remove ES6-style lambdas.
- [X] Callbacks need fixing.
- [x] Decide what to do with dropdowns  - jquery/other.
- [X] Remove example ipynb's from commits.
- [ ] Squash the commits down.

We have removed jquery usage from most widgets, as a precursor to removing completely. We have left jquery-ui present in order to use the slider, whilst we work out our approach to maintaining a consistent slider view across browsers/platforms.


jupyter-js-widgets/src/manager-base.js
```diff
@@ -191,6 +191,8 @@ ManagerBase.prototype.new_widget = function(options, serialized_state) {
             options_clone.model_id = utils.uuid();
         }
         return that.new_model(options_clone, serialized_state);
+    }).catch((error) => {
```
SylvainCorlay: ES6

jdfreder:
:eyes: 

Thanks for opening this!  Just a comment for later, and you may have already been aware of this, but make sure not to commit the changes to the example notebooks unless you plan on changing the contents of the notebooks.  If you do want to change the contents, you'll want to make sure you do so on a Py3 kernel and that you save using the widgets menu instead of the normal notebook save - this will allow you to save with widget snapshots, so that they can be exported in our docs.

dwillmer:
@jdfreder yep, thanks, is already in the WIP list above.

SylvainCorlay:
I was thinking that the move of setElement to initialize could be in a separate PR that we could merge before 5.0. Is it too dependent on the rest of the changes that you guys are making here?

SylvainCorlay:
Besides, would it be possible to rebase this (at least on itself to remove the merge commits and all), and revert the notebook changes?

dwillmer:
Yep, of course, both of those are in the todo list above.

jdfreder:
I think @dwillmer is going to rebase -i and squash

dwillmer:
:+1:

dwillmer:
Just a reminder - this is a big change, is work in progress, and is only PR'd at Sylvain's request.

dwillmer:
No, this was a typo from the jquery removal.

jdfreder:
Looks like some of the recently merged ipywidgets 5.0 release work conflicts with this.  If I were you, I'd wait till the very end to do the rebase on top of master (along with what ever commit squashing you want to do).  I say that because there will probably be a few more bug/behavioral fix type PRs prior to 5.0 being released that may cause more conflicts.  I promise that I won't use any jquery in these bug fixes! :)

dwillmer:
Pinging @SylvainCorlay @jdfreder -

This is ready for review.

I haven't squashed the commits for 2 reasons - firstly there's other people with copies of this branch, and secondly i don't want to hide the large contributions from @afshin.

Happy to do whatever you want :)

SylvainCorlay:
@dwillmer a rebase on master would remove the merge commits though

README.md
```diff
@@ -100,7 +100,7 @@ This will run the test suite using `karma` with 'debug' level logging.
 ## Resources
 ### ipywidgets
 - [Demo notebook of interactive widgets](https://github.com/ipython/ipywidgets/blob/master/examples/Index.ipynb)
-- [Documentation for ipywidgets](http://ipywidgets.readthedocs.org/en/latest/) [[PDF](https://media.readthedocs.org/pdf/ipywidgets/latest/ipywidgets.pdf)]
+- [Documentation for ipywidgets](http://ipywidgets.readthedocs.org/en/latest/) [[PDF](https://media.readthedocs.org/pdf/ipywidgets/latest/ipywidgets.pdf)
```
SylvainCorlay: missing ]?

jupyter-js-widgets/examples/web/test/index.js
```diff
@@ -11,23 +11,21 @@ const widgetsRendered = new Promise((resolve, reject) => {
...
         });
     });
-    describe('html', () => {
-        it('exists', () => {
+    describe('html', function() {
+        it('exists', function() {
             expect(document.querySelector('div.widget-html')).to.be.ok;
         });
-        it('correct value', () => {
+        it('correct value', function() {
```
SylvainCorlay: I think that this file was ES6 already, so we probably don't need to change that.

jupyter-js-widgets/examples/web3/src/index.js
```diff
@@ -29,7 +29,7 @@ document.addEventListener("DOMContentLoaded", function(event) {
         var inputarea = document.getElementsByClassName("inputarea")[0];
         var editor = CodeMirror(inputarea, {
             value: code,
-            mode: "python",
+            mode: 'python',
```
SylvainCorlay: Same here, the web3 example uses babel, hence we can probably revert the changes on index.js

jupyter-js-widgets/test/widget_test.js
```diff
@@ -212,18 +212,19 @@ describe("Widget", function() {
...
             expect(changeCallback.callCount).to.equal(2);
 
-            this.widget.on_some_change(['a', 'b'], someChangeCallback, this.widget);
-            this.widget.set_state({ a: false, b: false });
-            return this.widget.state_change;
-        }).then(() => {
+            that.widget.on_some_change(['a', 'b'], someChangeCallback, that.widget);
+            that.widget.set_state({ a: false, b: false });
+            return that.widget.state_change;
+        }).then(function() {
```
SylvainCorlay: same here

afshin:
@SylvainCorlay The ES6 => ES5 changes were probably me being too zealous. It seems, though, that it would be nice if the whole codebase follows the same conventions. What do you think?

SylvainCorlay:
> @SylvainCorlay The ES6 => ES5 changes were probably me being too zealous. It seems, though, that it would be nice if the whole codebase follows the same conventions. What do you think?

It is also to reduce the amount of noise on git. 

The large number of commits + this type of changes everywhere and changes in the code formatting make the relevant changes difficult to isolate...

jupyter-js-widgets/src/widget_bool.js
```diff
@@ -60,71 +66,70 @@ var CheckboxView = widget.DOMWidgetView.extend({
...
-        this.setElement($('<button />')
-            .addClass('jupyter-widgets widget-toggle-button btn btn-default')
-            .attr('type', 'button')
-            .on('click', function (e) {
-                e.preventDefault();
-                that.handle_click();
-            }));
-        this.$el.attr("data-toggle", "tooltip");
-        this.listenTo(this.model, "change:button_style", this.update_button_style, this);
+        this.el.className = 'jupyter-widgets widget-toggle-button';
```
jasongrout: Using `className` instead of `classList.add` means that any attributes the container assigns are obliterated, so this isn't the same as jquery's `addClass` that it is replacing. I see some commits in the middle of the PR where there was a `classList.add`. Why the change to use `className`?

afshin:
That's true, but mix and match code conventions:

- some places used single-quote strings, some used double-quotes
- some places used jquery to add even listeners, others used the backbone `events` hash
- some places had `function_name : function (param_name)` and others had `function_name: function(param_name)`

... also can add to confusion when looking at the code.

jupyter-js-widgets/src/widget_bool.js
afshin: @jasongrout you're absolutely right, and I believe we only use `className` in cases like this where we "own" the element because `setElement` was used, thereby already obliterating what was there and in cases where the element is private.

dwillmer:
Happy to revert the ES5 changes in the test file, but i also agree with Darian - it's better to be consistent rather than to have the source ES5 and the tests ES6.

SylvainCorlay:
@jdfreder decided to use babel in the test files. I will let him decide. Regardless, I think that is not really the crux on this PR.

SylvainCorlay:
One good way to squash reversion is to do the following: make the all the reversions in one single commit and rebase -i by interleaving every commit with the that commit (and the squash statement).

```
pick first commit
squash reversion
pick second commit
squash reversion
pick third commit
squash reversion
```

jdfreder:
The tests and examples are intentionally ES6.  Soon the codebase will be ES6 too.  Please leave them as ES6 so we don't have to re-translate back in the future, sorry for the PITA.

jupyter-js-widgets/src/widget_bool.js
jasongrout: A widget view object can be created (i.e., that is where setElement should be called), and then the container can assign classes to the element, and then render the widget. This may be theoretical at this point (especially since setElement was in the render up to this point), but using className would break things in this case. For example, it would probably be cleaner for the generic jupyter-widgets class to be applied by the widget manager.

afshin: Fair enough. The only *guaranteed* safe place to use `.className` is in the same function that calls `.setElement`, so it's probably a good idea to not use it at all unless they are grouped together. I'll update this.

SylvainCorlay:
So as per discussion on gitter, in addition to the merge commits that should not be in the history, it seems that a lot of commits were repeated in this PR (at least the 13 first commit). One of you rebased his branch on master and then merged the other version of the same commits into his branch.

jupyter-js-widgets/package.json
```diff
@@ -54,8 +54,7 @@
     "webpack": "^1.12.11"
   },
   "dependencies": {
-    "backbone": "1.2.0",
-    "bootstrap": "^3.3.5",
+    "backbone": "^1.2.3",
```
SylvainCorlay: As per discussion in person, we must pin backbone to `1.2.0`. We have noticed a breaking change between 1.2.0 and 1.2.2. (This specific regression seemed to have been fixed in 1.2.3 though) but the crux is that backbone folks admittedly don't follow semver and we can't use these types of loose version requirement with it.

jdfreder: We saw the same kind of breaking change from Backbone 1.1.0 to 1.2.0 IIRC

SylvainCorlay: (From 1.2.0 to 1.2.2)

jdfreder: No, 1.1.0 (see https://github.com/ipython/ipython-components/blob/master/bower.json#L5)  Immediately post big split, pre 4.0, we upgraded which caused major headaches because of backwards incompat. changes in backbone.

SylvainCorlay: Yes! See https://github.com/ipython/ipywidgets/pull/246 :)

jdfreder: Yes, we've had this breakage more than once. :(  It's unfortunate that everyone can't agree to semver.  I wonder what Backbone's version system is.  Time related?  Arbitrary?

jdfreder:
@afshin said to go ahead and close this in favor of #427

PR #401, opened 2016-03-01
Allow regular string representation of commless widgets

SylvainCorlay:
If one does
```python
s1 = IntSlider()
s1
```
and then 
```python
s1.close()
```
Then any further attempt to "display" the s1 python object is going to yield an exception.

What is done here instead is that comm-less widgets fall back to the regular `__repr__`.


jdfreder:
Thanks!

PR #402, opened 2016-03-02
javascript link act on initial data.

SylvainCorlay:
Now that we are initializing widget models with data, we must do more than registering callbacks in initialize, we must also act on the initial data. This is a case where it was missing (javascript link). This fixes the web2 example.


jdfreder:
Thanks!

PR #403, opened 2016-03-03
Use new trailets API for Range widgets

SylvainCorlay:
- The range widgets were doing validation on trait change, which was allowing intermediate inconsistent states for the widgets to exist and be notified to the front-end.

  Now we use custom cross validation to validate that `lower <= upper` for `_FloatRange` and `min <= lower <= upper <= max` for `_BoundedFloatRange`.

- Cross validation occurs before any notification is sent, which solves the problem of intermediate inconsistent states, but it comes with constraints:

  With the `hold_trait_notifications` context manager, we allow delaying cross-validation and trait notifications to the time when the context is exited. At that time, all the cross-validation are called. If any of them fails, all the changes done with the context manager are reverted.

  However, we don't control the order of the cross validation calls which are set by the order of the iteration in a dictionary. Therefore, we must enforce that the operations done in the custom validator **commute**. One simple way to enforce this is to only allow the cross-validation to reject a change or change the proposed attribute value, but **not** change other trait attributes. (See traitlets documentation [here]( https://github.com/ipython/traitlets/blob/d13f22d93a826dedd2199c08cd89791cf393f16f/traitlets/traitlets.py#L798).)

    For example, if you have a `min` and `max`, you should not have max push min down if set to low and min push max up if set too high. Otherwise, the following code has unpredictable outcomes:
 
   ```python
   # min, max is 0, 1 
   with w.hold_trait_notifications():
       w.min = 2
       w.max = -2

   print(w.min, w.max)  # could give (-2, -2) or (2, 2)
   ```
   *So the new behavior for the range widget is that we reject if the min is set too high or the max is set too low*, which is a behavior change.

- Range widgets are "value" widgets in that like many other core widgets, they represent a "value", which here is a tuple for the  lower and upper bounds of the range.

  Originally, two other trait attributes (`lower` and `upper`) were added to facilitate direct access to the lower and upper bounds. Since the link was done on trait change, again, intermediate inconsistent state was possible. Using cross validation to do the link was not a solution for the same reason explained above.

   Instead, I use a *property* to bind `upper` and `lower` to the `value` tuple attribute. It is a behavior change in that you cannot register to change of upper or lower, and you must register to changes of `value` instead. Doing that also solves the issue met when the constructors has initial values for both `upper`, `lower` on the one hand and `value` on the other hand.

### DIFF
--- a/ipywidgets/widgets/tests/test_interaction.py
+++ b/ipywidgets/widgets/tests/test_interaction.py
@@ -14,6 +14,8 @@ import nose.tools as nt
 
 from ipykernel.comm import Comm
 import ipywidgets as widgets
+
+from traitlets import TraitError
 from ipywidgets import interact, interactive, Widget, interaction
 from ipython_genutils.py3compat import annotate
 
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
cc @chronitis and @jdfreder

jdfreder:
:+1:

SylvainCorlay:
This is ready

jdfreder:
Thanks!

PR #405, opened 2016-03-04
Refactor Selection widgets

SylvainCorlay:
Same type of refactoring using the new traitlets API to simplify the selection and multi-selection widgets.

One change here is that the `selection_label` attribute was removed and it is directly `value` that is synced with the front-end, but with a custom serializer that actually sends the label through the wire.

It allows to remove a complicated logic of locks that was put in place to cross-update `label` and `value`.

Besides, like for the range widgets, this fixes the potential inconsistent state issues that was due to validating things on change of attributes rather than before.

### DIFF
--- a/ipywidgets/widgets/tests/test_interaction.py
+++ b/ipywidgets/widgets/tests/test_interaction.py
@@ -615,21 +615,21 @@ def test_multiple_selection():
 
     # degenerate multiple select
     w = smw()
-    check_widget(w, value=tuple(), options=None, selected_labels=tuple())
+    check_widget(w, value=tuple(), options=None)
 
     # don't accept random other value when no options
-    with nt.assert_raises(KeyError):
+    with nt.assert_raises(TraitError):
         w.value = (2,)
-    check_widget(w, value=tuple(), selected_labels=tuple())
+    check_widget(w, value=tuple())
 
     # basic multiple select
     w = smw(options=[(1, 1)], value=[1])
     check_widget(w, cls=smw, value=(1,), options=[(1, 1)])
 
     # don't accept random other value
-    with nt.assert_raises(KeyError):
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Besides, like for the range widgets, this fixes the potential inconsistent state issues that was due to validating things on change of attributes rather than before.

SylvainCorlay:
Finally, this removes much more code than it adds, which is a really good thing.

jdfreder:
Looks like the multi select Python test is broken

SylvainCorlay:
This is ready.

SylvainCorlay:
Not anymore.

SylvainCorlay:
This can be merged as-is. I have more simplification to come but I cannot do what I want because of a couple of bugs I found in traitlets regarding union traits which I will fix.

I will get back to this later.

ipywidgets/widgets/tests/test_interaction.py
```diff
@@ -637,31 +637,14 @@ def test_multiple_selection():
...
-    check_widget(w, value=(1,))
-
-    # don't accept selected_label (from superclass)
-    with nt.assert_raises(AttributeError):
-        w.selected_label = 3
-
-    # don't return selected_label (from superclass)
-    with nt.assert_raises(AttributeError):
-        print(w.selected_label)
+    check_widget(w, value=(1, 2))
```
jdfreder: It's not clear to me why you removed these tests.  Shouldn't these errors still be thrown?

jdfreder: Ah because you remove selected label

jdfreder:
I think later we'll want to revisit the subject of user API convenience regarding this Widget.

PR #417, opened 2016-03-11
Add set ex to abort on non-zero return and print as exec

jdfreder:


jdfreder:
cc @blink1073

blink1073:
:brohugs:

PR #418, opened 2016-03-12
Filter default values from widget state

SylvainCorlay:
@jdfreder as per discussion.


jdfreder:
Woot!

SylvainCorlay:
We should probably also be able to get rid of the `views` list, and the serialization of the cell.

PR #419, opened 2016-03-12
Specify model module in new_widget for Controller

SylvainCorlay:
This is now necessary since we removed the registry.


jdfreder:
:+1:

PR #420, opened 2016-03-13
requirejs -> window.require

SylvainCorlay:
@jdfreder


blink1073:
This tricks webpack into not bundling it?

SylvainCorlay:
`requirejs` is not a function...

blink1073:
It is added when you use require.js

SylvainCorlay:
Yes, but it's an object...

jdfreder:
Requirejs works as a function too

SylvainCorlay:
Not when I load requirejs like this: https://github.com/SylvainCorlay/embed-bqplot/blob/master/index.html#L3.

blink1073:
What I'm interested in is using window.System so we can swtich to SystemJS for webpack 1 and 2.0.

PR #423, opened 2016-03-15
Add web4 example with inline widget state

SylvainCorlay:
@jdfreder this is essentially the web2 example manager and the script tag trick that you used for the doc. 

This `index.built.js` is probably what we would want cdn-ed for this to work anywhere with the future embed button.


SylvainCorlay:
I would like to move that manager to jupyter-jd-widgets and call it  like Embedded manager.

jdfreder:
Thanks!

PR #433, opened 2016-03-18
Add DropdownView keyboard bindings.

afshin:


jupyter-js-widgets/src/widget_selection.js
```diff
@@ -187,10 +284,21 @@ var DropdownView = widget.DOMWidgetView.extend({
...
+            }
+            // Close the drop list.
             this.droplist.classList.remove('mod-active');
+            // Remove global event listener.
+            document.removeEventListener('keydown', this.keydown);
             return;
         }
 
+        // Add a global keydown listener for drop list events.
+        document.addEventListener('keydown', this.keydown);
```
jdfreder: Why did you have to use a global keydown listener?  Does an element specific event listener not work?

afshin: Because the drop list has to be attached to `body` and it floats, and because there is no element that you'll necessarily be on that I can guarantee will catch the keystrokes. It's the way controls like this always have to work, unfortunately.

jdfreder: That's unfortunate

jdfreder:
Thanks! :cake:

jupyter-js-widgets/src/widget_selection.js
afshin: FYI, here's Bootstrap attached a `document` listener that it then filters with its `toggle` selector:

https://github.com/twbs/bootstrap/blob/master/js/dropdown.js#L162

jdfreder: Thanks

PR #434, opened 2016-03-21
Add initial accordion implementation

jdfreder:


jdfreder:
I must have been half asleep when I opened the PR for this, cc @afshin

PR #437, opened 2016-03-25
Bugfix: IntSlider() not loading without jQuery UI being loaded.

afshin:
I couldn't get `IntSlider` to load without having jQuery UI `require`d in this way.

cc: @jdfreder @SylvainCorlay


jdfreder:
IIRC this change breaks the menu in jupyter/notebook because it overrides the jqui there.  Will have to test before merging.

PR #438, opened 2016-03-25
Add new TabView implementation

afshin:
This is a new implementation of `TabView` that does not use Bootstrap or jQuery. It uses the phosphor `TabBar`.

cc: @jdfreder @SylvainCorlay @jasongrout @dwillmer


afshin:
@jdfreder @SylvainCorlay @jasongrout @dwillmer 

By the way, I am not clear why `_titles` needs to be a dictionary in the back-end, but that comes with front-end consequences because the `children` array and the `_titles` dictionary can easily fall out of sync with each other. It seems to me that the natural data structure for `_titles` ought to be an array just like `children`.

FYI, here is the bugfix: https://github.com/ipython/ipywidgets/pull/438/commits/a1f5259203642550472353778806a45f6bcbf36c

afshin:
Actually if you look at log output in Travis: https://travis-ci.org/ipython/ipywidgets/jobs/119256390 on line 1008, I'm pretty sure this is the error with `IntSlider`, and may require https://github.com/ipython/ipywidgets/pull/437 first.

PR #447, opened 2016-03-31
Use unicode string (fixes Exploring Graphs example)

SylvainCorlay:
@jdfreder


SylvainCorlay:
This should be forward ported to 6.0 too.

PR #450, opened 2016-04-02
Embed no function

SylvainCorlay:
@jdfreder another take on the how to embed widgets, which is to use a `text/json` mime type.

The problem I see is that I must use windows.onload.

### DIFF
--- a/jupyter-js-widgets/src/embed-webpack.js
+++ b/jupyter-js-widgets/src/embed-webpack.js
@@ -7,13 +7,12 @@
 // Load jquery and jquery-ui
 var $ = require('jquery');
 require('jquery-ui');
-$.getQueryParameters = window.$.getQueryParameters;
-$.urldecode = window.$.urldecode;
 window.$ = window.jQuery = $;
+require('jquery-ui');
 
 // Load styling
 require('jquery-ui/themes/smoothness/jquery-ui.min.css');
-require("../css/widgets.min.css");
+require('../css/widgets.min.css');
 
 const widgets = require('./index');
 console.info('jupyter-js-widgets loaded successfully');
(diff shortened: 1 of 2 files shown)


jupyter-js-widgets/src/embed-webpack.js
```diff
@@ -7,13 +7,12 @@
 // Load jquery and jquery-ui
 var $ = require('jquery');
 require('jquery-ui');
-$.getQueryParameters = window.$.getQueryParameters;
```
SylvainCorlay: I don't  understand why you were refering to `window.$`at this point.

jdfreder: These lines are left over from the docs, without these lines the docs will break.  However, I don't think these lines should exist here, instead the fix belongs in the docs.

SylvainCorlay: but how is `window.$` defined at this point?

jdfreder: By the jquery in the theme

jdfreder: Ithink the real problem here is that the jquery version of the doc theme !== the version specified in the package.json

SylvainCorlay: ok, so indeed, it should be moved to the docs thing. You are right.

jupyter-js-widgets/src/embed-webpack.js
```diff
@@ -21,37 +20,25 @@ console.info('jupyter-js-widgets loaded successfully');
...
 // Magic global widget rendering function:
-function renderWidgetState(widgetStateObject) {
-
-  // Create the container div and insert it into the DOM immediately above the
-  // current script tag.
-  console.info('Inserting widget(s)...');
-  var container = document.createElement('div');
-  container.innerHTML = `
-    <div class="widget-loading">
-      <span class="glyphicon glyphicon-hourglass glyphicon-spin" aria-hidden="true"></span>
```
jdfreder: Here you remove the loading animation, you should also remove the `glyphicon-spin` class from the css which I had to add manually.

jdfreder: We can address this in another PR

PR #459, opened 2016-04-06
Use npmcdn

SylvainCorlay:
`https://npmcdn.com/jupyter-js-widgets@0.0.17` already serves the base index.js file.

- [x] get the version number dynamically

### DIFF
--- a/widgetsnbextension/src/embed_widgets.js
+++ b/widgetsnbextension/src/embed_widgets.js
@@ -11,7 +11,7 @@ var embed_widgets = function() {
                 'drop_defaults': true
             }).then(function(state) {
                 var data = JSON.stringify(state, null, '    ');
-                var value = '<script src="http://perso.crans.org/corlay/built/index.built.js"></script><script>w(' + data + ');</script>';
+                var value = '<script src="https://npmcdn.com/jupyter-js-widgets@0.0.17/dist/embed.js"></script><script type="application/vdn.jupyter-embedded-widgets">' + data + '</script>';
                 var content = $('<textarea/>')
                     .attr('readonly', true)
                     .css({'width': '100%', 'min-height': '250px'})
--- a/widgetsnbextension/src/embed_widgets.js
+++ b/widgetsnbextension/src/embed_widgets.js
@@ -5,13 +5,13 @@
 
 var embed_widgets = function() {
     return new Promise(function(resolve) {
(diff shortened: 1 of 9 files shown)


widgetsnbextension/src/embed_widgets.js
```diff
@@ -11,7 +11,7 @@ var embed_widgets = function() {
                 'drop_defaults': true
             }).then(function(state) {
                 var data = JSON.stringify(state, null, '    ');
-                var value = '<script src="http://perso.crans.org/corlay/built/index.built.js"></script><script>w(' + data + ');</script>';
+                var value = '<script src="https://npmcdn.com/jupyter-js-widgets@0.0.17/dist/embed.js"></script><script>w(' + data + ');</script>';
```
SylvainCorlay: If we can get the version of the package, that is even better. Otherwise this will need to be added to the release instructions.

SylvainCorlay:
@rgbkrk

rgbkrk:
Nice!

rgbkrk:
Super cool how npmcdn works by integrating into the `npm run build` script.

widgetsnbextension/src/embed_widgets.js
```diff
@@ -11,7 +11,7 @@ var embed_widgets = function() {
                 'drop_defaults': true
             }).then(function(state) {
                 var data = JSON.stringify(state, null, '    ');
-                var value = '<script src="http://perso.crans.org/corlay/built/index.built.js"></script><script>w(' + data + ');</script>';
+                var value = '<script src="https://npmcdn.com/jupyter-js-widgets@0.0.17/dist/embed.js"></script><script type="application/vdn.jupyter-embedded-widgets">' + data + '</script>';
```
SylvainCorlay: Do you know how I can get the package version number of npm dynamically?

SylvainCorlay:
```javascript
version = require('./package.json').version;
```

SylvainCorlay: Actually, we would need to expose that as a variable in `jupyter-js-widgets` since this file is in widgetsnbextension.

rgbkrk: `var version = require('./package.json').version;` is normally quite correct. Do you have to dig it out of `node_modules` here?

PR #461, opened 2016-04-07
Fix new_comm call

SylvainCorlay:
@jdfreder this fixes the issue reported by @jasongrout regarding the signature of `manager.new_comm` ie the difference between [services/kernels/comm#L33]( https://github.com/jupyter/notebook/blob/41d6da235cbf3bcf6d7f818e11a066e0fd12ff8b/notebook/static/services/kernels/comm.js#L33) and [manager#L588](https://github.com/ipython/ipywidgets/blob/6.x/widgetsnbextension/src/manager.js#L588).

The widget list notebook change is just replacing Latex with Label since Latex is deprecated.

### DIFF
--- a/examples/Widget List.ipynb
+++ b/examples/Widget List.ipynb
@@ -42,7 +42,7 @@
     {
      "data": {
       "text/plain": [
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
This is obviously not the only way to do this, we could rather pass the `metadata` to the `data` attribute in the `comm_open` message.

jdfreder:
Thanks @SylvainCorlay ! Fish cake :fish: :fish_cake:

PR #469, opened 2016-04-07
Fix replaceWith, a jqueryism, with DOM operations.

jasongrout:
The box widget uses a jquery function replaceWith at https://github.com/ipython/ipywidgets/blob/4936b86a87804053eb3d719ff67b527e7f5beb4b/jupyter-js-widgets/src/widget_box.js#L195 (see http://api.jquery.com/replacewith/ for documentation). Since the dummy element is now a div, `replaceWith` is undefined and box widgets won't display their children.

CC @afshin @dwillmer, @jdfreder, @SylvainCorlay


afshin:
@jasongrout whoa, good spot!

PR #472, opened 2016-04-08
Warn when user is running a version of the notebook pre jupyter/notebook/pull/1279

jdfreder:
https://github.com/jupyter/notebook/pull/1279

This is much better than the catastrophic find_static_assets function
does not exist failure which causes the notebook server to not launch.

cc @minrk


Carreau:
Will that need backport ?

jdfreder:
No, it's only a problem that occurs on 5.0dev

Carreau:
Thanks.

PR #475, opened 2016-04-08
nbextension_name

SylvainCorlay:
@jdfreder this does only the renaming and some simplification in index.js.

If you had a dev install of widgetsnbextension, I recommend you disable and uninstall the nbextension before pulling.


SylvainCorlay:
@jdfreder @jasongrout @parente @jhpedemonte FYI

jdfreder:
Thanks!

PR #476, opened 2016-04-08
[WIP] Fixes a bug where JS links were not getting persisted to notebook widg state

jdfreder:


SylvainCorlay:
Why would you not want to persist links?

PR #479, opened 2016-04-10
Require font-awesome as part of the styling for embedder

SylvainCorlay:


SylvainCorlay:
I need to figure out how to set up the `publicPath`attribute in the webpack config so that it looks for the font-awesome assets (png / ttf / woff[2] / eot) in the same directory as the embed.js file.

At the moment, I can set the absolute path with respect to the URL root or another URL. If we decide to bless npmcdn, I can specify that.

SylvainCorlay:
Done, the static assets will be in the same directory as embed.js on npmcdn. To fully test, we must merge and make a release (0.0.20).

SylvainCorlay:
Works perfectly with the cdn.

PR #481, opened 2016-04-10
State should be deserialized

SylvainCorlay:
In the new differential `manager.set_state` we were not using custom deserializers.


jdfreder:
I'm surprised the behavior wasn't already doing this, good catch!

SylvainCorlay:
Jon, we have a false positive issue again with the tests apparently. Chrome and Firefox complain that Promise is not defined.

PR #484, opened 2016-04-11
Fix custom widget example

SylvainCorlay:
This makes the custom widget example notebooks work despite the removal of the registry. It was actually rather simple, using `define` and `requirejs.undef`.


SylvainCorlay:
PS: the output is cleared on these notebook since I don't think that it adds value in this case.

jdfreder:
> PS: the output is cleared on these notebook since I don't think that it adds value in this case.

That's okay, we can add the static widget representations later at little / no effort.  Thanks for the updates!

PR #487, opened 2016-04-11
Explicitly handle missing requirejs failure.

jdfreder:
cc @jasongrout @SylvainCorlay

### DIFF
--- a/jupyter-js-widgets/src/utils.js
+++ b/jupyter-js-widgets/src/utils.js
@@ -66,14 +66,23 @@ function loadClass(class_name, module_name, registry) {
 
             // If the module is jupyter-js-widgets, we can just self import.
             var modulePromise;
-            if (module_name === 'jupyter-js-widgets') {
+            var requirejsDefined = typeof window !== 'undefined' && window.requirejs;
+            if (requirejsDefined) {
+                if (module_name !== 'jupyter-js-widgets' || window.requirejs.defined('jupyter-js-widgets')) {
+                    modulePromise = new Promise(function(innerResolve, innerReject) {
+                        window.requirejs([module_name], function(module) {
+                            innerResolve(module);
+                        }, innerReject);
+                    });
+                } else if (module_name === 'jupyter-js-widgets') {
+                    modulePromise = Promise.resolve(require('../'));
+                }
+            } else if (module_name === 'jupyter-js-widgets') {
                 modulePromise = Promise.resolve(require('../'));
             } else {
-                modulePromise = new Promise(function(innerResolve, innerReject) {
-                    window.require([module_name], function(module) {
-                        innerResolve(module);
-                    }, innerReject);
-                });
+                // FUTURE: Investigate dynamic loading methods other than require.js.
+                throw new Error(['In order to use third party widgets, you ',
+                    'must have require.js loaded on the page.'].join(''));
             }
 
             modulePromise.then(function(module) {


jupyter-js-widgets/src/utils.js
```diff
@@ -66,11 +66,18 @@ function loadClass(class_name, module_name, registry) {
 
             // If the module is jupyter-js-widgets, we can just self import.
             var modulePromise;
-            if (module_name === 'jupyter-js-widgets') {
-                modulePromise = Promise.resolve(require('../'));
+            var requireJSDefined = Boolean(typeof window !== 'undefined' && window.requirejs);
```
SylvainCorlay: In strict mode, you can always undefined was not defined right? Then you should be able to compare with undefined?

jupyter-js-widgets/src/utils.js
```diff
@@ -66,11 +66,18 @@ function loadClass(class_name, module_name, registry) {
...
+            if (!requireJSDefined || module_name === 'jupyter-js-widgets') {
+                if (module_name === 'jupyter-js-widgets') {
+                    modulePromise = Promise.resolve(require('../'));
+                } else {
+                    // FUTURE: Investigate dynamic loading methods other than require.js.
+                    throw new Error(['In order to use third party widgets, you ',
+                        'must have require.js loaded on the page.'].join(''));
+                }
```
SylvainCorlay: why nesting? 
```
if (module_name === 'jupyter-js-widgets') {
    modulePromise = Promise.resolve(require('../'));
} else if (requireJSDefined) {
    modulePromise = ...
} else {
    throw new Error(
```

SylvainCorlay:
Thanks! I have a couple of inline remarks but besides that it is 👍

jupyter-js-widgets/src/utils.js
jdfreder: The typeof keyword returns a string with the type name.  `undefined` is the type name of `undefined` (I feel silly typing that).  You can't compare to `undefined` because you'll get a runtime error that the `window` variable hasn't been defined yet.

jupyter-js-widgets/src/utils.js
jdfreder: Much better!  I feel silly for not doing it this way, thanks!

jdfreder:
Comments addressed, I also changed the logic so users can override jupyer-js-widgets definitions if they have requirejs installed.

jupyter-js-widgets/src/utils.js
SylvainCorlay: Yes, that is right,

jupyter-js-widgets/src/utils.js
```diff
@@ -66,14 +66,18 @@ function loadClass(class_name, module_name, registry) {
...
+                    requirejs([module_name], function(module) {
                         innerResolve(module);
                     }, innerReject);
                 });
+            } else if (module_name === 'jupyter-js-widgets') {
+                modulePromise = Promise.resolve(require('../'));
+            } else {
+                // FUTURE: Investigate dynamic loading methods other than require.js.
+                throw new Error(['In order to use third party widgets, you ',
+                    'must have require.js loaded on the page.'].join(''));
```
SylvainCorlay: lol, that won't work, you need to check for `module_name === 'jupyter-js-widgets'`first.

jupyter-js-widgets/src/utils.js
```diff
@@ -66,14 +66,18 @@ function loadClass(class_name, module_name, registry) {
 
             // If the module is jupyter-js-widgets, we can just self import.
             var modulePromise;
-            if (module_name === 'jupyter-js-widgets') {
-                modulePromise = Promise.resolve(require('../'));
-            } else {
+            if (typeof requirejs !== 'undefined') {
                 modulePromise = new Promise(function(innerResolve, innerReject) {
-                    window.require([module_name], function(module) {
+                    requirejs([module_name], function(module) {
```
SylvainCorlay: window.require is probably better because it does not get picked up by webpack *and* you don't need to define requirejs.

jupyter-js-widgets/src/utils.js
SylvainCorlay: Even though `extension` and `embed` both define the amd module, bare `jupyter-js-widgets` does not, and you want to be able to use it on a page that has requirejs loaded without having to declare the amd module.

jdfreder: Crud, I was hoping this would work so users could override widgets...  :/

jupyter-js-widgets/src/utils.js
SylvainCorlay: (I have seen cases where loading require would not allow requirejs to be used as a function.)

jupyter-js-widgets/src/utils.js
SylvainCorlay: Only @jdfreder can define jupyter-js-widgets (jdfreders's facts).

jupyter-js-widgets/src/utils.js
jdfreder: I forgot webpack picks it up, arg!  I was trying to solve the use case where `window` is undefined - pure node env.

SylvainCorlay: I am trying to reproduce that issue I had, but window.require works for me.

jupyter-js-widgets/src/utils.js
jdfreder: :clap: lol

jupyter-js-widgets/src/utils.js
jasongrout: @jdfreder, you don't *ever* have to feel silly typing seemingly obvious statements about javascript typing, since often they are decidedly non-obvious! :)

jdfreder:
Okay @SylvainCorlay , check out the last commit.  It should handle all the cases we've talked about in this thread so far.  

- requirejs exists, use it  
- requirejs exists and module name = 'jupyter-js-widgets', load from requirejs **if defined**, otherwise load locally  
- no requirejs and module name = 'jupyter-js-widgets', load locally  
- error

SylvainCorlay:
Ps, what is the point of over-riding jupyter-js-widgets in require? If it is overridden, this is the that in which we would be already right (in the case of notebook, where jupyter-js-widget is an amd module)?

jdfreder:
I can see many different reasons for overriding, I think the two most important are

- The ability to redefine core widget views in contexts where the ones we have now wont play nice or as well as native implementations.  i.e. Atom+Hydrogen or nteract/nteract
- The ability to customize these things, beyond simple themes!  I could see users making cool packages to do this :)

SylvainCorlay:
ok - do you mind squashing the 4 commits?

jdfreder:
Not at all, squashed!

jdfreder:
Thanks!

PR #488, opened 2016-04-12
Persist widget view cell information to nb metadata

jdfreder:
@SylvainCorlay this is what we talked about.

cc @minrk 

Right now I'm persisting this to doc level metadata, so GUIDs are in one place, but really it it's cell level information.  Should I persist it per cell instead?

### DIFF
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -75,9 +75,7 @@ var WidgetManager = function (comm_manager, notebook) {
             // Load the initial state of the widget manager if a load callback was
             // registered.
             if (WidgetManager._load_callback) {
-                that.clear_state(true).then(function () {
-                    return WidgetManager._load_callback.call(that);
-                }).then(function(state) {
+                WidgetManager._load_callback.call(that).then(function(state) {
                     that.set_state(state);
                 }).catch(widgets.reject('Error loading widget manager state', true));
             }
@@ -190,12 +188,21 @@ WidgetManager.set_state_callbacks = function (load_callback, save_callback, opti
 var url = [window.location.protocol, '//', window.location.host, window.location.pathname].join('');
 var key = 'widgets:' + url;
 WidgetManager.set_state_callbacks(function() {
-    if (localStorage[key]) {
-        return JSON.parse(localStorage[key]);
+    if (Jupyter.notebook.metadata.widgets) {
+        return return Promise.resolve(Jupyter.notebook.metadata.widgets.state);
     }
-    return {};
+    return Promise.resolve({});
 }, function(state) {
-    localStorage[key] = JSON.stringify(state);
+    Jupyter.notebook.metadata.widgets = {
+
+        // Persisted widget state version
+        version: '1.0.0',
+
+        // Only persist the views of the widget
+        state: _.mapObject(state, function(widgetState) {
+            return _.pick(widgetState, 'views');
+        }),
+    };
 });
 
 WidgetManager.prototype._handle_display_view = function (view) {


widgetsnbextension/src/manager.js
```diff
@@ -190,12 +188,18 @@ WidgetManager.set_state_callbacks = function (load_callback, save_callback, opti
...
-        return JSON.parse(localStorage[key]);
-    }
-    return {};
+    return Promise.resolve(Jupyter.notebook.metadata.ipywidgets.state);
 }, function(state) {
-    localStorage[key] = JSON.stringify(state);
+    Jupyter.notebook.metadata.ipywidgets = {
+
+        // Persisted widget state version
+        version: '1.0.0',
```
SylvainCorlay: Should this rather be the version of jupyter-js-widgets (or rather of the notebook widget manager)?

I am not sure about this because I think that eventually the extension version info will be in the `_model_module`  and `_view_module` variables.

SylvainCorlay:
Awesome thanks!

widgetsnbextension/src/manager.js
```diff
@@ -190,12 +188,18 @@ WidgetManager.set_state_callbacks = function (load_callback, save_callback, opti
 var url = [window.location.protocol, '//', window.location.host, window.location.pathname].join('');
 var key = 'widgets:' + url;
 WidgetManager.set_state_callbacks(function() {
-    if (localStorage[key]) {
-        return JSON.parse(localStorage[key]);
-    }
-    return {};
+    return Promise.resolve(Jupyter.notebook.metadata.ipywidgets.state);
```
SylvainCorlay: you need to handle the case where `ipywidgets` is not defined. (Should it rather be called widgets?)

jdfreder: :white_check_mark:

widgetsnbextension/src/manager.js
```diff
@@ -190,12 +188,21 @@ WidgetManager.set_state_callbacks = function (load_callback, save_callback, opti
...
+        return return Promise.resolve(Jupyter.notebook.metadata.widgets.state);
     }
-    return {};
+    return Promise.resolve({});
 }, function(state) {
-    localStorage[key] = JSON.stringify(state);
+    Jupyter.notebook.metadata.widgets = {
+
+        // Persisted widget state version
+        version: '1.0.0',
```
SylvainCorlay: Still not sure about that part.

widgetsnbextension/src/manager.js
jdfreder: It could be the version of widgetsnbextension, because it's specific to the notebook.  Really it's a separate version number though, like nbformat it may not change along with major releases.  The purpose of the version number at the top level here is different than a version number per model.  It's used to identify the storage format of the state.

widgetsnbextension/src/manager.js
jdfreder: It could be the version of widgetsnbextension, because it's specific to the notebook. Really it's a separate version number though, like nbformat it may not change along with major releases. The purpose of the version number at the top level here is different than a version number per model. It's used to identify the storage format of the state.

SylvainCorlay: I am hesitant to add a version number that we are not using from the start. The risk is that it would be stale - but I see where you come from.

jdfreder: If it grows stale, we remove it.  If we need to discern between state versions later and we don't start versioning now, that makes things more difficult.

SylvainCorlay: How about we make it the version of the widgetnbextension node package.

SylvainCorlay: When we release `ipywidgets`, we will probably want to make it a 1.0 for `widgetsnbextension`.

jdfreder: Yeah I think that's okay for now, we may realize later we want the two decoupled, like nformat and notebook versioning is now

jdfreder:
merged :)

SylvainCorlay:
Given the `return return` thing that was in the code, I think that it was one of those false positives in the tests where Promise is not defined.

PR #492, opened 2016-04-12
Fix testing

SylvainCorlay:


jdfreder:
Oh no, did we break the tests at some point?

SylvainCorlay:
The web example complain about Promise not being defined.

PR #493, opened 2016-04-12
Fallback to npmcdn in the embed manager

### LINKED ISSUE #489: npmcdn + semver
npmcdn supports semver. 

You can request  
 - `https://npmcdn.com/requirejs@~2.1.0` 

and you get 

- `https://npmcdn.com/requirejs@2.1.22`

Same with the other ways to define semver ranges. Should we use that `~` operator in our embedable code snippet so that we get the latest backward compatible code?

The user would have a manager compatible with his state plus bug fixes.

SylvainCorlay:
Closes #489 .

### DIFF
--- a/jupyter-js-widgets/src/embed-manager.js
+++ b/jupyter-js-widgets/src/embed-manager.js
@@ -25,3 +25,18 @@ EmbedManager.prototype.display_view = function(msg, view, options) {
 EmbedManager.prototype._get_comm_info = function() {
     return Promise.resolve({});
 };
+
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
After that, we have a working static embedder entirely relying on cdns. It requires the custom models to follow a certain scheme in how they are cdn-ed.

We can discuss the scheme itself before we merge this.

SylvainCorlay:
I think I got a nice solution to handle widget model versions with the cdn. Next PR.

PR #496, opened 2016-04-13
Use functional programming where it's more expressive

jdfreder:
Follow up of #495

@SylvainCorlay


jupyter-js-widgets/src/manager-base.js
```diff
@@ -367,16 +367,11 @@ ManagerBase.prototype.get_state = function(options) {
...
-                                        v.push(view.options);
-                                    }
-                                }
-                            }
-                            return v;
+                            return _.values(views).filter(function(view) {
+                                    return view.options !== undefined && view.options.root;
+                                }).map(function(view) {
+                                    return view.options;
+                                });
```
SylvainCorlay: :+1:

PR #497, opened 2016-04-13
Add es6 Promise polyfill for IE11 support

SylvainCorlay:


jdfreder:
:+1:

jupyter-js-widgets/webpack.config.js
```diff
@@ -1,4 +1,5 @@
 var version = require('./package.json').version;
+require('es6-promise').polyfill();
```
jdfreder: Is this the right place to put this?  Seems like you meant to put it in embed-webpack.js

PR #499, opened 2016-04-13
Object.assign is not supported by IE11.

SylvainCorlay:
Some IE love.


jdfreder:
Thanks, this is definitely something I did
I remember writing the code :P

PR #504, opened 2016-04-14
Add jupyter client as a build req for the notebook files

jdfreder:
@willingc I'm seeing a `jupyter_client` missing dependency error on RTD.  I'm not quite sure what of nbsphinx is triggering the error.  Let me know if you see it too.  If you do, we'll add it to the sphinx theme.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
+jupyter_client
 sphinx==1.3.6
 jupyter_sphinx_theme==0.0.6


jdfreder:
Ah here's some good information to know:

https://github.com/spatialaudio/nbsphinx/blob/0.2.6/nbsphinx.py#L389-L397

nbsphinx attempts to execute a notebook if no outputs exist and yet code cells exist.

willingc:
Thanks Jon.

willingc:
@jdfreder We should probably add a readthedocs.yml and environment.yml to build using conda on RTD

jdfreder:
I don't use conda myself, but I do not object to switching

willingc:
Which repo do you want me to do a pr against.

jdfreder:
A PR against ipython/ipywidgets directly would be nice!

jdfreder:
Thanks :grin:

willingc:
okie dokie

PR #506, opened 2016-04-14
Include one output in custom widget notebook as a workaround

jdfreder:
to prevent nbsphinx from building.

See https://github.com/spatialaudio/nbsphinx/pull/39


jdfreder:
cc @SylvainCorlay

Issue #41, opened 2015-06-02
How to use the widgets from other language kernels?


minad:
Is this also part of jupyter? It should be very well possible to share the js part and only keep a small model and logic part on the kernel side.

jdfreder:
It is.  At one point we were trying to decide whether or not to put the repository under the Jupyter org.  We decided to leave it here because of its Python content.  Other kernels have implemented the widget spec, IHaskel and IJulia to name two.  It's not well documented yet, but totally doable.

minrk:
@minad the kernel-side and frontend-side will be split like the rest of ipython:jupyter when the interfaces between them (loading, widget messages, etc.) stabilize. Since those interfaces are *highly* unstable at this point, the Python-side and js-side are really a single active research project for now, from a practical standpoint. Other kernels can definitely use the JS provided with this package, but we are pretty far from any API stability promises.

minad:
Ok, good. It would be great if the js/notebook side could be stabilized, documented and shared between the kernels.

But no hurry. It is already very good what you have already achieved with jupyter in terms of multi lang support and I really appreciate your work. I stalled further development of iruby up to that point where I felt that jupyter is ready enough and I think I guess we will just do the same concerning the widgets. It would still be great if you let me know if the widget thing starts to converge...

minrk:
Yup, that's definitely the plan. Just not there yet.

minad:
great :clap:

lbustelo:
Where is the discussion of how IPy Widgets move forward in the language agnostic world of Jupyter? I would like to understand if there are plans or experimentation taking place.

jdfreder:
@lbustelo there's not much planning to be done, it's just a matter of moving the files around.  However, we'd like to wait for the API to stabilize before splitting the repo.  Only time will tell when the API has stabilized.

jasongrout:
For reference, we have now split the backend (ipywidgets) and frontend (jupyter-js-widgets). They still live in the same repo, but are completely separate packages. I think it's right to close this as fixed at this point.

PR #507, opened 2016-04-14
Add all widget outputs to custom widget example

jdfreder:
@SylvainCorlay unfortunately I have to have all of the outputs for it to not run.

### DIFF
--- a/docs/source/examples/Custom Widget - Hello World.ipynb
+++ b/docs/source/examples/Custom Widget - Hello World.ipynb
@@ -103,7 +103,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {
     "collapsed": false
    },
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
What do you mean by "for it to not run"?

jdfreder:
I mean to prevent nbsphinx from trying to run the notebook which requires extra dependencies on RTD that we don't need.  I opened a PR against nbsphinx adding a flag to disable this, but it hasn't been merged yet.  https://github.com/spatialaudio/nbsphinx/pull/39

mgeier:
@jdfreder Just to clarify: you don't need to have all outputs, one output is enough for `nbsphinx` not to execute the notebook. In #506, however, you added only an `execution_count` and not an output.

Probably I should check for those, too?

I'll take care of your PR momentarily.

jdfreder:
@mgeier I definitely executed the cell, it may be that the cell didn't result in any displayed output.  You probably want to correct your auto execute logic to account for that.  Jupyter widgets are also not output, so executing a cell with widgets only will probably cause failure in the auto execute logic too.

mgeier:
OK, I'll extend the auto execute logic, see #41.

> Jupyter widgets are also not output

What are they then?
How do they look like in the representation provided by `nbformat`?

mgeier:
@jdfreder FYI, I've extended the auto-execute logic in `nbsphinx` master.

PR #510, opened 2016-04-14
Add support for conda builds on RTD

willingc:
@jdfreder As discussed. Rendered on my [test version](https://test-widgets.readthedocs.org/en/conda-rtd/index.html)

It looks like the move of examples broke some links to images. Let me know if you need help with that.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,0 +1,14 @@
+name: ipywidget_docs
+dependencies:
+- python=3.5
+- sphinx_rtd_theme
+- jinja2
+- tornado
+- nbformat
+- jupyter_client
+- ipykernel
+- ipywidgets
+- pip:
+  - sphinx==1.3.6
+  - nbsphinx
+  - jupyter_sphinx_theme==0.0.6
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,0 +1,4 @@
+conda:
+    file: docs/environment.yml
+python:
+    version: 3


jdfreder:
Thanks @willingc !  And thanks for the heads up about the images- I thought everything was self contained in the examples folder.  Which images are missing/broken?

jdfreder:
Oh wow, looks like they are all broken.  This has something to do with the b64 images encoded in the text/html mimetype I think, cc @SylvainCorlay

willingc:
Also when I built locally just now there were like 500+ warnings.

jdfreder:
500 !  That's a big number

jdfreder:
Looks like they are all related to the image extraction process... hmmm

willingc:
I've been able to get it to 16. Though not too gracefully.

jdfreder:
Feel free to open a PR - I was going to re-execute the example notebooks with the latest and greatest in widget rasterization, which should fix the problem by using the html/text mimetype.

willingc:
I'm going to wait until you are done. I excluded some directories which helped with the warnings but with unfortunate side effect of creating other warnings.

jdfreder:
Okay, unfortunately I've hit a bug in one of the notebooks which is prohibiting me from running it.  So it will take "a few" more minutes than expected.  (a few == undefined number)

willingc:
I think we may need to add `IPython.sphinxext.ipython_console_highlighting` to extensions in the theme so we have IPython3 lexer support.

jdfreder:
The bug I found turned out to be a symptom of an out of date ipython/traitlets.

willingc:
Are you getting a build with no warnings?

jdfreder:
No, still see the warnings :/  But it is building!  Did you want to open a PR with the work you said you had done to get it to work partially?

willingc:
The PR will be split between here and the jupyter_sphinx_theme

willingc:
I'm down to 14 warnings if that helps.

willingc:
Actually the one here is one line in conf.py:

``exclude_patterns = ['examples', 'generated']``

willingc:
I pushed the changes to jupyter_sphinx_theme to the ``fix-widgets`` branch, @jdfreder

jdfreder:
Thanks!

PR #513, opened 2016-04-14
Update widget metadata

jdfreder:


jdfreder:
@willingc here are the updates.  Looks like there's still a ton of warnings... :frowning:

PR #521, opened 2016-04-19
Use wheel `data_files`.

SylvainCorlay:
The `widgetsnbextension` package is a package that does not have the ipywidgets' python. All it does is installing the `jupyter-js-widgets` nbextension on the notebook server.

The problem with the new notebook nbextension mechanism is that it does not handle dependencies. For example, if you do `pip install [bqplot/pythreejs/ipyleaflet]`, then the `widgetsnbextension` package gets installed as a dependency, but the nbextension package is not installed. 

One must then do

```
jupyter nbextension install --py widgetsnbextension
jupyter nbextension enable --py widgetsnbextension
jupyter nbextension install --py [bqplot/pythreejs/ipyleaflet]
jupyter nbextension enable --py [bqplot/pythreejs/ipyleaflet]
```

wouhou, a **5 steps** installation!

What this does is make the `jupyter-js-widgets` nbextension a `data_file` attribute of the wheel. So that the installation of the nbextension is just a byproduct of the one of the widgetsnbextension python package, and pip's dependency installation takes care of it.

Now the only thing that we need to do in **notebook** side is enable a list of dependency to also *enable* when you enable an extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,6 @@ or
 
 Then run
 
-    jupyter nbextension install --py widgetsnbextension
     jupyter nbextension enable --py widgetsnbextension
 
 See the [Installation](docs/source/user_install.md) section of the documentation for additional details.
--- a/docs/source/dev_install.md
+++ b/docs/source/dev_install.md
@@ -28,8 +28,8 @@ available in the PATH used with sudo.
         npm run update:widgets
         pip install -v -e .
         cd ..
-        jupyter nbextension install --py --symlink widgetsnbextension
-        jupyter nbextension enable --py widgetsnbextension
+        jupyter nbextension install --py --symlink --user widgetsnbextension
+        jupyter nbextension enable --py --user widgetsnbextension
 
 3. Dev-install of the package (run from ipywidgets repo directory):
 
--- a/widgetsnbextension/setup.py
+++ b/widgetsnbextension/setup.py
@@ -202,6 +202,12 @@ setup_args = dict(
         'sdist': js_prerelease(sdist, strict=True),
         'jsdeps': NPM,
     },
+    data_files      = [(
+            'share/jupyter/nbextensions/jupyter-js-widgets', [
+                'widgetsnbextension/static/extension.js',
+                'widgetsnbextension/static/extension.js.map'
+        ]),
+    ],
     include_package_data = True,
 )
 


SylvainCorlay:
@jasongrout @jdfreder @Carreau

ellisonbg:
I like this idea!

SylvainCorlay:
OK, I updated the instructions on development installs so that dev version of the javascript gets precedence over the sys-prefix install.

jdfreder:
What exactly is happening here?  How are you bypassing the nbextension install step?  Does nbextension check Python module data?

SylvainCorlay:
> How are you bypassing the nbextension install step?

✨ magics ✨

SylvainCorlay:
Do you want to hop in appear-in

jdfreder:
Sure, https://appear.in/ipywidgets

blink1073:
magic

SylvainCorlay:
@blink1073 I love that meme.

minrk:
This is great for virtualenvs, conda envs. Some people may notice that it won't work for `--user` installs in some contexts, and a couple of other possible cases where data_files don't end up where we look (and Python doesn't record where it puts data_files, so you can't look it up).

jasongrout:
@jdfreder - this uses the new `--sys-prefix` option for installing nbextensions. Basically, now nbextensions are looked for in the old `--user` and global locations, and *also* in a directory in python's `sys.prefix` path. With this PR, we'll copy the nbextension into the `sys.prefix` path when you install the nbextension.

jasongrout:
@minrk - okay, we still have the option of installing manually, of course.

I'm curious - when will python put data_files somewhere where we can't find them?

SylvainCorlay:
In any case, user install is evil :)
   - user directories are checked before all others. So if you have
multiple python installs, and you installed a package as a user for one, it
will be used by all.
   - user directories are not located in consistent places depending on
your os, python distribution.

People who use user installs deserve whatever happen to them :)

Sylvain

jasongrout:
@SylvainCorlay - it sounds like you want to modify the install instructions again :).

SylvainCorlay:
well yes, but no :)

minrk:
data_files should work in most common situations (`--user` installs on non-Linux are the most likely exception, I think). As long as there's a command that you can run to install after-the-fact, or in a different place, it should be okay.

SylvainCorlay:
@bollwyvl @damianavila FYI

PR #523, opened 2016-04-20
Adding note about installing the nbextension assets

tmtabor:


SylvainCorlay:
We actually don't need the install step anymore, only the enabling.

tmtabor:
Fixed.

SylvainCorlay:
Thanks!

Issue #466, opened 2016-04-07
6.x dropdown widget doesn't scroll with page


jasongrout:
Notice that scrolling the page when the dropdown is open shows the dropdown does not scroll with the page. Presumably this is from absolutely positioning the dropdown, rather than inserting it into the dom and relatively positioning it.

![dropdown-positioning](https://cloud.githubusercontent.com/assets/192614/14364520/49e75d7e-fcd7-11e5-8c28-87b89c6f3f12.gif)

CC @afshin

SylvainCorlay:
Fixed in #524

PR #524, opened 2016-04-20
Scrolls and mousedowns dismiss dropdown.

afshin:
This PR resolves https://github.com/ipython/ipywidgets/issues/466

- [x] `window` `scroll` events dismiss a dropdown (which is similar to native behavior).
- [x] `document` `mousedown` events dismiss a dropdown (which is similar to native behavior).
- [x] Bugfix: drop list `max-height` and `top` were calculated incorrectly on some occasions.

cc: @SylvainCorlay @jdfreder @jasongrout

### DIFF
--- a/jupyter-js-widgets/less/widgets.less
+++ b/jupyter-js-widgets/less/widgets.less
@@ -430,7 +430,7 @@ ul.widget-dropdown-droplist {
     list-style-type: none;
     position: absolute;
     margin: 0;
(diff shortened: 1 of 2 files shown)


sccolbert:
Should this be `mousedown` instead of `click`?

afshin:
@sccolbert `mousedown` is better, I've updated.

PR #525, opened 2016-04-20
Update Image Processing example to work with current skimage.

fperez:
The `io.Image` class was removed from skimage. Added equivalent local utility,
and also added `@interact` example to browse other images in the `skimage.data`
gallery.

Thanks to @stefanv for the original fix.


blink1073:
Nice!  :+1:

fperez:
Thanks @blink1073! I made some small fixes, all done now.

fperez:
Thanks!

PR #528, opened 2016-04-21
Filter view information in notebook metadata

SylvainCorlay:
@jdfreder this is removing some of the churn that we were having in the notebook metadata with the view info.


SylvainCorlay:
backported to 5.x

PR #530, opened 2016-04-22
Revert to put widget_class in the data portion of comm message

### LINKED ISSUE #522: widget_class information on comm_open should be placed in data
The current code merged in #461 places the information about `widget_class` in a the `meta-data` section of the `comm_open` message. This is a huge deal, because nowhere in the spec around Comm support (and I'm basing the statement on [this](http://jupyter-client.readthedocs.org/en/latest/messaging.html#opening-a-comm)) does it say that handlers to comm message (open, msg, close, info) need access to the `meta-data` area. That being the case, both the Comm support in IRKernel and Toree/SparkKernel, do not make this area of the message available to handlers. In my opinion, the goal of Comm is to simplify communication and not to expose all the pieces of the Kernel Protocol.

I propose that this is changed so that the information about `widget_class` is passed within the `data` area, a much easier fix for declarativewidgets and do not involve changes to the kernels.

/cc @SylvainCorlay @jdfreder @jasongrout @parente

lbustelo:
Resolves #522

### DIFF
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -593,12 +593,12 @@ WidgetManager.prototype.deleteSnapshots = function() {
     }).bind(this));
 };
 
-WidgetManager.prototype._create_comm = function(comm_target_name, model_id, metadata) {
+WidgetManager.prototype._create_comm = function(comm_target_name, model_id, data) {
     var that = this;
     return this._get_connected_kernel().then(function(kernel) {
-        if (metadata) {
-            return kernel.comm_manager.new_comm(comm_target_name, {},
-                                                that.callbacks(), metadata, model_id);
+        if (data) {
+            return kernel.comm_manager.new_comm(comm_target_name, data,
+                                                that.callbacks(), {}, model_id);
         } else {
             return new Promise(function(resolve) {
                 requirejs(["services/kernels/comm"], function(comm) {
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -139,7 +139,7 @@ class Widget(LoggingConfigurable):
     @staticmethod
     def handle_comm_opened(comm, msg):
         """Static method, called when a widget is constructed."""
-        class_name = str(msg['metadata']['widget_class'])
+        class_name = str(msg['content']['data']['widget_class'])
         if class_name in Widget.widget_types:
             widget_class = Widget.widget_types[class_name]
         else:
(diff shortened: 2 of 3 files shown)


lbustelo:
@SylvainCorlay and @jasongrout I think this is what you guys meant with keeping it symmetric. Both sides are now using the `data` portion of the comm message to share `widget-class` during the `comm_open`.

widgetsnbextension/src/manager.js
```diff
@@ -593,12 +593,12 @@ WidgetManager.prototype.deleteSnapshots = function() {
...
-WidgetManager.prototype._create_comm = function(comm_target_name, model_id, metadata) {
+WidgetManager.prototype._create_comm = function(comm_target_name, model_id, data) {
     var that = this;
     return this._get_connected_kernel().then(function(kernel) {
-        if (metadata) {
-            return kernel.comm_manager.new_comm(comm_target_name, {},
-                                                that.callbacks(), metadata, model_id);
+        if (data) {
+            return kernel.comm_manager.new_comm(comm_target_name, data,
+                                                that.callbacks(), {}, model_id);
```
SylvainCorlay: From a jupyter-js-widgets perspective, this is a backward incompatible change.

Could we do 
```
return kernel.comm_manager.new_comm(comm_target_name, data,
                                               that.callbacks(), data, model_id)
```
with a command saying that only the first one will be kept?

lbustelo: @SylvainCorlay Hmm... not sure I understand the backwards incompatibility wrt jupyter-js-widgets. That code only implements a stub. I do see a problem if the widgetsnbextension and ipywidgets get out of sync.

I'm fine with making the change. How about a target version for removal? Or does it even matter?

SylvainCorlay: hmm, indeed, this change is in the nbextension manager.

SylvainCorlay:
@lbustelo I am 👍 on this change generally, and I would like to push out a 5.0.1 or 5.1.0 pretty quickly with that change so that you guys can iterate.

@jdfreder @jasongrout it is time to say something !

jasongrout:
@SylvainCorlay: I have this to say. Can we *please* merge my 6.x branch to master? :)

This looks okay to me, but I haven't thought a lot about ramifications, nor have I tested it.

jdfreder:
I think this is okay, as long as the information is available on comm open.

lbustelo:
I tested the changes with R and Scala and with this I'm able to continue adding the declarativewidgets support in top of the widgetnbextension by just supporting the new comm target_name.

PR #535, opened 2016-04-26
Lowering the message for the version check to warn

### LINKED ISSUE #531: jupyter.widget.version open call should only take place when the kernel is Python
The version check does not seem to cause any issues, but it does print an error message on the js console. Maybe this check should look at the kernel.info and be performed only for Python kernels.

Something along the lines of this https://github.com/jupyter-incubator/declarativewidgets/blob/7659d9def805b4f7066abeedb3225c09ac153522/elements/urth-viz-ipywidget/urth-viz-ipywidget.html#L104

lbustelo:
resolves #531


SylvainCorlay:
Backported.

PR #537, opened 2016-04-27
Lowering the message for the version check to warn

SylvainCorlay:
@lbustelo my original pull from master failed when I released 5.1.0 and #535 is the only missing thing in the release. I am reopening the PR on master here.


lbustelo:
That is fine... this is not a big deal at all.

PR #545, opened 2016-04-29
Min max width height

### LINKED ISSUE #533: Layout widget should also expose [min/max]_[width/height] properties
I don' t remember why we did not include this in the Layout attributes, but it is really missing in the case of widget for that 
 - should stretch in a flexbox where align-items is `stretch`
 - have a minimal size, in case align-items is unset.

SylvainCorlay:
Fixes #533 .


SylvainCorlay:
Backported.

PR #553, opened 2016-05-03
Using collapse() the 'parent' attribute remains undefined.

nastasi-oq:
Using collapse() the 'parent' attribute remains 'undefined' and the accordion doesn't work correctly.


jdfreder:
Thanks @nastasi-oq !

Issue #544, opened 2016-04-28
Review of 6.x changes


SylvainCorlay:
We spent some time today with @jasongrout reviewing the changes in the 6.x branch (PR #538).

 - In `utils.js`
   - `escape.html` should use `textContent` instead of `innerHTML`
   - `typeset` should use `Array.prototype.map` instead of a for loop.

- In lots of places, the double quotes were replaced with single quotes and `'` had to be escaped, which highlighted that we should replace `couldn't` with `could not` in most error messages.

- In `widget_bool.js` why is the `fa` class removed from `<i />`? 

- In `widget_box.js`, missing quotes around class `vbox`.

- In the color picker, `addEventListener` is called twice on the color picker, while in the original widget, it was called once on the picker and once on the text input.

- In `widget-int.js` we are still using the bootstrap class `form-control`. #529 seems to be fixing it.

SylvainCorlay:
cc @afshin

afshin:
@SylvainCorlay @jasongrout,

> escape_html should use textContent instead of innerHTML

Just to be clear, which of these behaviors do we want? (Please see attached screenshot.)

<img width="344" alt="screen shot 2016-04-29 at 12 49 19 pm" src="https://cloud.githubusercontent.com/assets/159529/14915204/facfa50a-0e08-11e6-8c5f-340b09a49ce5.png">

It seems clear that we do not want the first or the last. Of the middle two, one *escapes* the HTML and one strips it. Given that the function is called `escape_html`, I imagine we want the one with all the HTML character entities in it, but please confirm.

> typeset should use `Array.prototype.map` instead of a `for` loop.

Could you explain the rationale for this? `typeset` is called in a lot of places:

1. A `for` loop is faster than `map`.
2. Perhaps even the `for` loop is unnecessary: are there actually cases where `typeset` is returning an array? The reason it currently does is because we were trying to keep a 1:1 fidelity with the `jQuery` version, but if it's actually the case that individual elements are passed in instead of collections, perhaps we can get rid of the loop altogether.

> In lots of places, the double quotes were replaced with single quotes and `'` had to be escaped, which highlighted that we should replace `'couldn\'t'` with `'could not'` in most error messages.

Good call.

> In `widget_bool.js` why is the fa class removed from `<i />`?

Because it is not necessary. An `<i>` tag can be given a `FontAwesome` icon without needing that class. For example, the caret in the `DropDownView` widget has the following style:

```
i.widget-caret {
[...]

PR #555, opened 2016-05-03
Issue 544 bugfixes

afshin:
These are bugfixes for https://github.com/ipython/ipywidgets/issues/544

- [x] Color picker widget respects widget width
- [x] Dropdown widget respects widget width
- [x] Progress bar widget renders `'vertical'` mode correctly
- [x] Fix Dropdown widget `keydown` handler inside notebook

### DIFF
--- a/jupyter-js-widgets/less/widgets.less
+++ b/jupyter-js-widgets/less/widgets.less
@@ -229,6 +229,10 @@
     white-space: nowrap;
     width : @widget-width-short;
 
+    i.fa {
+        padding-right: 3px;
+    }
+
     .outside-shadow-2dp();
 
     &:hover {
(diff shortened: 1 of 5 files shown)


afshin:
screen shot 2016-05-03 at 7 24 19 pm

afshin:
FYI, I added the `.fa` class back in for button icons.

screen shot 2016-05-03 at 9 03 21 pm

jasongrout:
Darian, those changes look good; thanks. One more thing about the dropdown - if the size of the dropdown has changed, the actual dropdown list no longer matches the width of the control (see the image below - the grey box is a different width than the control). Can we get the list to match the width of the control?

screen shot 2016-05-03 at 5 05 31 pm

afshin:
@jasongrout Good call, the drop list width needs to be calculated at run-tim in JS (it used to just be a static 300px).

jasongrout:
Nice, thanks! I also investigated a bit more deeply what the `<select>` behavior is on different platforms over in #562. No conclusions there, but there is precedence for having the dropdown list box width be more than the dropdown control size.

PR #557, opened 2016-05-03
Simplify typeset function.

afshin:
The `typeset` function can be dramatically simplified because we do not pass element collections to it. It was implemented to support collections because it was ported over from a `jQuery` implementation and `jQuery` objects are collections by default. But it is not necessary in the current codebase.

cf. https://github.com/ipython/ipywidgets/issues/544

### DIFF
--- a/jupyter-js-widgets/src/utils.js
+++ b/jupyter-js-widgets/src/utils.js
@@ -156,29 +156,10 @@ function reject(message, log) {
  * text: option string
  */
 function typeset(element, text) {
-    if (arguments.length > 1) {
-        if (element.length) {
-            for (var i = 0; i < element.length; ++i) {
-                var el = element[i];
-                el.textContent = text;
-            }
-        } else {
-            element.textContent = text;
-        }
+    element.textContent = text;
+    if (window.MathJax) {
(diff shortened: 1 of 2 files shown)


sccolbert:
The docs for this function should be update to explain the new call signature.

jupyter-js-widgets/src/utils.js
```diff
@@ -143,42 +143,22 @@ function reject(message, log) {
...
-    if (arguments.length > 1) {
-        if (element.length) {
-            for (var i = 0; i < element.length; ++i) {
-                var el = element[i];
-                el.textContent = text;
-            }
-        } else {
-            element.textContent = text;
-        }
+    if (text) {
```
jasongrout: Let's unconditionally set the text. Suppose I have a label widget which sets the content based on a text string coming from python. If the user for whatever reason sets the content to `''`, this function will now keep the original text since `''` is false in Javascript.

jasongrout: Or even better, we can keep it optional, but explicitly check to see if `text` is undefined.

afshin:
@jasongrout 👍

jasongrout:
Thanks!

SylvainCorlay:
:+1:

PR #560, opened 2016-05-04
A widget must be attached to the DOM before it is typeset.

### LINKED ISSUE #550: Label widget doesn't work second time
In the Widget List example notebook, there is an example of the Label widget:

```python
widgets.Label(
    value="$$\\frac{n!}{k!(n-k)!} = \\binom{n}{k}$$",
)
```
The first time I run the cell, it looks fine. If I run it a second time, though, I just get "Math processing error". Right-clicking on the error and selecting "Error message" says "Error: Cannot read property 'firstChild' of null"

CC @afshin.

jasongrout:
MathJax needs this so it can measure things related to the font size, etc.

Fixes #550

### DIFF
--- a/jupyter-js-widgets/src/widget.js
+++ b/jupyter-js-widgets/src/widget.js
@@ -759,7 +759,7 @@ var DOMWidgetViewMixin = {
     },
 
     typeset: function(element, text){
-        utils.typeset(element, text);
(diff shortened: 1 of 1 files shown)


jasongrout:
CC @SylvainCorlay

SylvainCorlay:
👍

Issue #562, opened 2016-05-04
Dropdown behavior in ipywidgets 6.x


jasongrout:
@afshin wrote a custom dropdown widget so we can have nice styling. I looked into how various dropdown widgets work in different platforms for corner cases. In the following cases, the 'list box' is the list box that pops up when clicking on the control that displays the options to select. 

At this stage, I don't have any suggestions for improved dropdown behavior, but just wanted to record observations of various behavior.

On Chrome on OS X 10.10, a `<select>` widget has the following behavior (see https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/ControlswithMenus.html)

![dropdown](https://cloud.githubusercontent.com/assets/192614/15023003/61b91124-11fc-11e6-9067-049841d2529c.gif)

1. the list box is essentially a popup menu which is modal and covers the dropdown control so that the selected item is over the control. The height is the minimum of the screen height (i.e., larger than the browser window) and the natural height of the list box, and the ends have arrows indicating whether there is more content above or below the displayed content.
2. The widget takes over scrolling when the list box is displayed, and scrolling changes the selection while keeping the list box open *if* the list box is bigger than the screen. Scrolling *doesn't* affect the page. Scrolling also doesn't do anything if the list box is open and is shorter than the screen height.
3. For normal cases, the width of the list box seems to be the minimum of the dropdown control width and the width needed to fit the content of the list box, up to a maximum of about 172 characters. After a list entry is about 6309 characters, though, the entry seems to be ignored in the width calculation (e.g., it fits the second longest entry). 
4. By default, the list box left edge lines up with the dropdown box left edge. However, the list box is adjusted left or right so that the entire list box displays on the screen, even if it means that the list box left edge does not align with the dropdown control left edge.

For Chrome on Windows 7, I see:

1. The list box is a dropdown below the control with a fixed height of 20 items.
2. You can scroll inside the dropdown list. Moving your mouse outside the list box and scrolling the page causes the list box to disappear with no change in selection.
3. The width is the minimum of the dropdown control width and the content width.
4. The list box left edge is always the drop down box left edge, even if the content width goes off the edge of the physical screen - the list box also just goes off the edge of the screen and you don't see the scrollbars at the left edge)
4. Putting thousands of characters in an `<option>` *really* slows down the control so that selects lags the mouse movement *really* bad and the list box also takes maybe a second to display.
`

jasongrout:
[...]

PR #566, opened 2016-05-05
Do not cycle through options when edges are hit via keyboard navigation.

afshin:
cf. https://github.com/ipython/ipywidgets/issues/562

cc: @jasongrout @blink1073 @SylvainCorlay


jasongrout:
Thanks!

Issue #567, opened 2016-05-05
Dropdown selection and key behavior


jasongrout:
I'm experimenting more with the dropdown in master. Here are a couple of points:

1. When an option is selected, then the dropdown is opened, the current value should be selected. This also makes sure that keyboard up/down events start from the currently selected item instead of the top/bottom.
4. Any time the selected value in the list box is set (either on open or on keyup/keydown), the new selected value should be visible in the list box. We see this problem on keyup/keydown (when using the keyboard to select an option, when the selected option goes out of the visible area, the list box doesn't scroll to make the selected option visible.)

CC @afshin - can you look at these? Thanks!

afshin:
Sure thing.

PR #570, opened 2016-05-06
Bugfixes for issue 567

afshin:
Bugfixes for https://github.com/ipython/ipywidgets/issues/567

- [x] When drop list opens, make current selection active.
- [x] Scroll to active item in a drop list if it is not visible.


jasongrout:
Awesome, thanks!

PR #572, opened 2016-05-08
bump less to 2.7

### LINKED ISSUE #547: jupyter-js-widgets will not build with node v6.0.0
There is an issue with the Less compiler: https://github.com/less/less.js/issues/2881 that causes it to fail silently on node v6.0.0, thus breaking `npm install` in `jupyter-js-widgets`. Before it is resolved by the Less developers, we should at least be aware that it may cause issues for users who have upgraded to the latest version of `node`.

cc: @jdfreder @SylvainCorlay @jasongrout

minrk:
works on node 6

closes #547


jdfreder:
Thanks min!

PR #585, opened 2016-05-13
Simplify installation

jdfreder:


widgetsnbextension/package.json
```diff
@@ -33,5 +33,8 @@
     "rimraf": "^2.5.1",
     "style-loader": "^0.13.0",
     "webpack": "^1.12.12"
+  },
+  "optionalDependencies": {
+    "jupyter-js-widgets": "2.0.0-dev"
```
SylvainCorlay: How about we let it be a dependency and use `file: ` all the time except on releases?

widgetsnbextension/package.json
```diff
@@ -7,8 +7,9 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "webpack",
-    "update:widgets": "rimraf node_modules/jupyter-js-widgets; npm install file:../jupyter-js-widgets; npm run build",
-    "prepublish": "npm run build"
+    "update": "rimraf node_modules/jupyter-js-widgets; npm install file:../jupyter-js-widgets; npm run build",
+    "prepublish": "npm run build",
+    "nbextension": "pip install -v -e . && jupyter nbextension install --py --symlink --user widgetsnbextension && jupyter nbextension enable --py --user widgetsnbextension"
```
SylvainCorlay: Since we probably want to separate this for good sometimes, we cannot assume that widgetsnbextension is present...

minrk: I agree that maybe we should avoid calling setup.py from npm. We just finished breaking those cycles in notebook. In particular, this picks a very specific user-dev-install that not everyone should do.

jdfreder:
Comments addressed.  I made the bash script accept and arg so the user can alter whether or not `--user` is used in the nbext stuff.

dev-install.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+nbExtFlags=$1
```
minrk: probably want `set -e` to exit on error

dev-install.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "'npm -v' failed therefor npm is not installed.  In
+    order to perform a developer install of ipywidgets you must
+    have both npm and pip installed on your machine!"
+    exit 1
+fi
+pip --version
+if [ $? -eq 0 ]; then
+    echo pip is installed
+else
+    echo "'pip --versopm' failed therefor pip is not installed.
```
blink1073: Typo

widgetsnbextension/package.json
```diff
@@ -24,8 +24,8 @@
     "backbone": "^1.2.3",
     "html2canvas": "^0.5.0-beta4",
     "jquery": "^2.2.0",
-    "jupyter-js-widgets": "2.0.0-dev",
-    "underscore": "^1.8.3"
+    "underscore": "^1.8.3",
+    "jupyter-js-widgets": "file:../jupyter-js-widgets"
```
SylvainCorlay: 👍

SylvainCorlay:
LGTM

SylvainCorlay:
I fixed the typo in jon's branch.

blink1073:
There's another typo in that line.

blink1073:
And also Min's comment about the flag.

SylvainCorlay:
> There's another typo in that line.

I fixed the `therefor`. is that it?

dev-install.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+nbExtFlags=$1
+
+npm -v
+if [ $? -eq 0 ]; then
+    echo npm is installed
+else
+    echo "'npm -v' failed therefor npm is not installed.  In
```
blink1073: Ah, this is the other typo.

jdfreder:
Okay, I added the set -e to the bash

dev-install.sh
jdfreder: Thanks, not having it has definitely caused me trouble before.

docs/source/dev_install.md
```diff
@@ -15,33 +15,18 @@ available in the PATH used with sudo.
...
-        cd ..
-        jupyter nbextension install --py --symlink --user widgetsnbextension
-        jupyter nbextension enable --py --user widgetsnbextension
-
-3. Dev-install of the package (run from ipywidgets repo directory):
+1. Navigate into the cloned repo and install
 
         cd ipywidgets
-        pip install -v -e .
+        bash dev-install.sh --user
```
SylvainCorlay: Do we want to recommend a user install?

jdfreder: I think that's what we were already doing.  I don't really have an opinion on the mater, other than --user has always worked well for me.

PR #590, opened 2016-05-17
Align-items with stretch by default in HBox and VBox

SylvainCorlay:
As discussed with @jasongrout this is probably the behavior expected when people use the `Box` widget with the `HBox` and `VBox` functions, especially when nesting them...


SylvainCorlay:
Jason actually is the one who pressed the Green Button.

SylvainCorlay:
backported.

PR #591, opened 2016-05-18
only filter stored snapshots

### LINKED ISSUE #554: All output_area elements are getting rerender on a notebook save
We noticed this recently and seems to be due to https://github.com/ipython/ipywidgets/blob/3a06dc11311e4d290227d9270729b15c3824b2ca/widgetsnbextension/src/manager.js#L592. This function executes on each notebook `save` and does not seem to filter which `output_area` to redisplay.

It is causing [this](https://github.com/jupyter-incubator/declarativewidgets/issues/339) issue in declarativewidgets.

/cc @SylvainCorlay @parente

minrk:
don't re-render outputs on the page

This should fix the problem of js outputs being re-executed on save.

I was a bit surprised that I didn't have to remove the snapshots from the page, but I don't seem to, because they don't seem to be there when this is called.

cc @SylvainCorlay and @jdfreder for review, in case there's something else needed here.

closes #554

### DIFF
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -595,11 +595,18 @@ WidgetManager.prototype.deleteSnapshots = function() {
 
         // Remove the outputs with isWidgetSnapshot: true.
         if (cell.output_area) {
-            // filter out outputs that are snapshots
-            // only touch the data, not the page.
+            // Filter out outputs that are snapshots
+            // Only touch the data, not the page
             cell.output_area.outputs = cell.output_area.outputs.filter(function (output) {
                 return !(output.metadata && output.metadata.isWidgetSnapShot);
             });
+
+            // Remove the corresponding elements from the page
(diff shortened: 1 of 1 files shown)


jdfreder:
I'm a little worried that this is a race condition, in that if ipywidgets gets loaded later the outputs will be on the page.  I'll take a look though.  Thanks for the PR!

jdfreder:
@minrk this does not remove the widget snapshots for me, because the widget JS is loaded after the notebook contents have already rendered on the page.

Here's a gif illustrating the order of events:
noclear

I'll take a look at this now, expect a PR to your PR soon...

jdfreder:
https://github.com/minrk/ipywidgets/pull/3 :smile:

minrk:
Thanks for the extra fix, @jdfreder!

SylvainCorlay:
LGTM.

SylvainCorlay:
backported

PR #611, opened 2016-05-25
Allow for multiple embed managers

SylvainCorlay:
I started working on a bootstrap page to make a gallery like http://www.htmlwidgets.org/showcase_leaflet.html and found a big in the inline embedder that occurs when there are multiple widget managers on the page.

This fixes the issue.


SylvainCorlay:
backported

PR #613, opened 2016-05-26
Improve installation instructions

SylvainCorlay:


docs/source/user_install.md
```diff
@@ -3,21 +3,19 @@ Installation
...
 
+- With pip:
+
 ``` sourceCode
 pip install ipywidgets
+jupyter nbextension enable --py widgetsnbextension
 ```
 
-or
+When using virtualenv, the <code>--sys-prefix</code> option may be required for
```
jasongrout: I think sys-prefix could be:
```
`--sys-prefix`
```

docs/source/user_install.md
```diff
@@ -3,21 +3,19 @@ Installation
...
+the extension enabling command in order to keep the environment isolated.
```
jasongrout: extension-enabling, I think.

jasongrout:
A belated +1 to updating these! I put a few comments inline too.

docs/source/user_install.md
SylvainCorlay: This is unmodified from the original one...

PR #617, opened 2016-05-29
separate fileupload methods into a python module

denfromufa:
Please follow this discussion:
https://groups.google.com/forum/#!topic/jupyter/s8MUMgbZTzY

### DIFF
--- a/docs/source/examples/File Upload Widget.ipynb
+++ b/docs/source/examples/File Upload Widget.ipynb
@@ -6,29 +6,23 @@
    "metadata": {
     "collapsed": false
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Overwriting upload.py\n"
+     ]
+    }
+   ],
    "source": [
+    "%%writefile upload.py\n",
     "from __future__ import print_function # py 2.7 compat.\n",
     "import base64\n",
     "import ipywidgets as widgets # Widget definitions.\n",
-    "from traitlets import Unicode # Traitlet needed to add synced attributes to the widget."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "This is a custom widget that allows the user to upload file data to the notebook server.  The file data is sent via a statefull `value` attribute of the widget.  The widget has an upload failed event that fires in the front-end and is echoed to the back-end using a custom msg."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {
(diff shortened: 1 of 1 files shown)


minrk:
I don't think adding `%%writefile` to the example notebook isn't as nice as the code being either:

- in the notebook, or
- in a separate file, without `%%writefile`

Can I ask why you would prefer to do this? I see from the linked discussion that you want to, but not why.

denfromufa:
So that I can re-use this upload widget between notebooks as a python module
.

denfromufa:
@minrk what is wrong with %%writefile?

minrk:
There's nothing wrong with `%%writefile`, but I think the example is clearer if it is not used, by either leaving the code in the notebook, or taking it out and importing it. Using `%%writefile` means that the code isn't in a separate file until after the notebook has been run, so the code isn't available to be imported unless the notebooks are executed in the correct order.

I'm confused about how adding `%%writefile` to this example notebook makes code available to your other notebooks. What notebooks are these? How are they getting a widget from a module that doesn't exist until an example notebook from this repo has been executed? Are you planning to add more notebook examples to this repo that re-use the upload widget?

denfromufa:
Not planning to add any other notebooks, the problem with making a separate
`upload.py` in repo is that I have not seen this here. Looks like all
notebooks are stand-alone.
 The problem with having this widget defined completely in the notebook is
that it cannot be re-used in other notebooks as a dependency. Copy-pasting
huge cells just to make files upload does not seem a good idea. On the
other hand - facotring out upload.py makes it easy to re-use in other
notebooks.
Am I missing something?

jasongrout:
> Not planning to add any other notebooks, the problem with making a separate
> `upload.py` in repo is that I have not seen this here. Looks like all
> notebooks are stand-alone.
>  The problem with having this widget defined completely in the notebook is
> that it cannot be re-used in other notebooks as a dependency. Copy-pasting
> huge cells just to make files upload does not seem a good idea. On the
> other hand - facotring out upload.py makes it easy to re-use in other
> notebooks.

The purpose of this example was to show how to write a simple widget, but the medium we had to write in was a notebook. I completely agree that if you want to use this widget in your day-to-day work, you should take the code out and put it in a python file (or even better, in a javascript file and load it as an extension). We encourage you to do that in your project. However, for a documentation example, we should probably leave it in the notebook.

Would you like to add a remark to the notebook, suggesting that a user copy the file into a python file they can import if they want to actually use the example with different notebooks?

denfromufa:
Just close this pull request :) I have the reference if I need to come back.

jasongrout:
Okay, sounds good. Thanks for thinking about how to use the code better!

PR #619, opened 2016-05-31
WIP Media widget, providing access to the camera

jasongrout:
This is still very much a work-in-progress:
- [ ] Expose the full MediaConstraints api
- [ ] figure out why it's slow
- [ ] provide a trigger to take a screenshot
- [ ] is there a way to record video/audio?

@SylvainCorlay :)

### DIFF
--- a/jupyter-js-widgets/src/widget_media.js
+++ b/jupyter-js-widgets/src/widget_media.js
@@ -1,0 +1,78 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+'use strict';
+
+var widget = require('./widget');
+var utils= require('./utils');
+var _ = require('underscore');
+
+
+/**
+ * From https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
+ */
+var promisifiedOldGUM = function(constraints) {
+
+  // First get ahold of getUserMedia, if present
+  var getUserMedia = (navigator.getUserMedia ||
+      navigator.webkitGetUserMedia ||
+      navigator.mozGetUserMedia);
+
+  // Some browsers just don't implement it - return a rejected promise with an error
+  // to keep a consistent interface
+  if(!getUserMedia) {
+    return Promise.reject(new Error('getUserMedia is not implemented in this browser'));
+  }
+
+  // Otherwise, wrap the call to the old navigator.getUserMedia with a Promise
+  return new Promise(function(resolve, reject) {
+    getUserMedia.call(navigator, constraints, resolve, reject);
+  });
+		
+}
+
+// Older browsers might not implement mediaDevices at all, so we set an empty object first. From https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
+if(navigator.mediaDevices === undefined) {
+  navigator.mediaDevices = {};
+}
+
+// Some browsers partially implement mediaDevices. We can't just assign an object
+// with getUserMedia as it would overwrite existing properties.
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
👍

jasongrout:
By the way, Sylvain and I discovered it is "slow" in that it only seems like the video preview is working when the browser repaints (e.g., scrolling, or changing window size).

dmadeka:
@jasongrout Ideally, we should also have the ability to transfer the stream at some time interval over to the python side? (if the record is done, then analysis can be done directly with Python for NLP type stuff right?)

jupyter-js-widgets/src/widget_media.js
```diff
@@ -0,0 +1,78 @@
...
+    initialize: function() {
+        // Get the camera permissions
+        this.stream = navigator.mediaDevices.getUserMedia({audio: false, video: true});
+    }
+});
+
+var MediaView = widget.DOMWidgetView.extend({
+
+    initialize: function() {
+        this.setElement(document.createElement('video'));
```
dmadeka: `this.el.autoplay = true`

to stop the 'slow' rendering

jasongrout:
@dmadeka - do you want to take these commits and finish it up to whatever looks good to you and submit it as a separate PR?

gnestor:
@jasongrout Let me know how I can help with this one...

dmadeka:
@gnestor I got a little far with this actually. So, the question is what do you send back to the python side? I was serializing the model using the `Byte` traitlet. But I think this is totally inefficient to continuously sync the stream. It seems to require something intelligent to know when to send the stream back to the python side, or might need something eventful to send just the diff..

if we are able to send the stream back though, we can use the voice-recognition/nlp tools of Python to enable voice-activated applications in the notebook!

dmadeka:
@gnestor At least that was my initial goal, not sure what @jasongrout or @SylvainCorlay envision for the widget

gnestor:
@dmadeka Would it be possible or efficient to just write the stream to the filesystem and relay the file path and stream status back to IPython?

dmadeka:
@gnestor I haven't tried that actually, should the user specify the path in Python and then the JS side sends a custom message on some condition for the stream? 

By the way, I know the stream is written to a DOM File, maybe we can just access that? Im not too sure about how that works

gnestor:
Ya I'm not sure either. I think that file path can be an optional argument for the Media widget and if none is given, it can default to a temp directory. If the stream is already being written somewhere and we can access that file path, even better!

jasongrout:
We could take this a lot of directions, like both of you mention. How about for now, we just have an event that will take a snapshot and save that image as a png data url? I think that's a decent minimally viable widget.

Unfortunately, now the code needs to be updated to typescript, but the current widget code should provide a good template.

As for recording to the filesystem, that gets tricky. Very tricky. The widget doesn't have access to the filesystem itself, and doesn't really have access to the notebook server. Instead, it only has access to the kernel through a comm channel. 

I suppose we could record to a buffer and then periodically sync that buffer to the kernel. We'd probably need a custom serializer that would only transmit new data.

gnestor:
@jasongrout Ok so it sounds like there is no precedent for a widget like this? In that case, I agree that it would behoove us to go with a minimum viable widget first (take a still screenshot) . I can update to Typescript and try to implement this, unless you would like to see this through.

jasongrout:
Go for it. I'm happy to let you do this.

I've noticed you've been doing a lot on widgets lately. Thanks!

gnestor:
Ok, cool! This will be a good learning experience! I'm happy to help!

SylvainCorlay:
@dmadeka do you want to complete this for 6.0?

maartenbreddels:
I think so, we can let it evolve in ipywebrtc faster than here.

PR #623, opened 2016-06-01
Bump pre-release version to -dev.0 so we can easily have multiple dev releases

jasongrout:


jasongrout:
I pushed this release to npm with the `dev` tag.

SylvainCorlay:
I don't understand. You pushed with the `.0` suffix?

jasongrout:
Yes. We can have dev.0, dev.1, dev.2, etc. I needed some published version I could start working with in jupyterlab.

Issue #14, opened 2015-05-05
Typescript


jdfreder:
Should we convert the widgets to Typescript?
Or instead to ES6?

My feeling is that as long as we can maintain easy (unchanged) access to our Javascript users, we should give Typescript a shot.

Adding to next week's dev meeting.

jasongrout:
+1 to typescript

jdfreder:
More thoughts, I've done this type of migration before.  I think we should initially move to ES6 and then consider moving to Typescript (breaking it into two steps is easier).  Leaving open as 5.0.

bollwyvl:
I was just doing a bit of refactoring of some older widgets into ES6, and hit the [Backbone and ES6 Classes issue](https://github.com/jashkenas/backbone/issues/3560). [Follow-up here](http://benmccormick.org/2015/07/06/backbone-and-es6-classes-revisited/) with workarounds. Basically, static properties are not really supported, and backbone uses them all over, `events`, etc. and we've added a few `serializers`. There are a few uses, such as `events` in ButtonView, but the more troubling `serializers` in `WidgetModel` (which has given me some trouble) might be a showstopper, as almost anybody that does custom work is going to have to monkey with this. One solution would be make sure we can accept functions for any custom properties (right now, `serializers` must be an POJO), which would clean up the inheritance story.

[TypeScript supports static properties](http://www.typescriptlang.org/Handbook#classes-static-properties), but they end up as `Class.someProperty`, and not `this.someProperty`, which probably won't work.

Since there will be a relatively narrow time when one WON'T need a transpiler, going straight to ES7, which supports decorators, might be attractive. In the follow-up, that example really does read nicely in comparison to the others.

In the near term, I'm going to make do with one of the workarounds (when I pick one) but this may be a thing to watch out for when picking a direction!

jdfreder:
Yeah I read about that problem regarding backbone recently.  I'm glad that you're heading the investigation on this @bollwyvl , and am curious to see what method you choose.  Please ping me. Thanks! :)

bollwyvl:
I'm leaning towards the es7 decorator approach. Used appropriately, i think
they offer a really nice visual way to inject just enough code, a la flask.
Used badly, well:
http://blog.jooq.org/2015/04/15/how-jpa-2-1-has-become-the-new-ejb-2-0/

However, in the case of serializers, i think we should accept/expect a
static method, as they nearly always need inheritance, albeit by _.extend
[...]

PR #627, opened 2016-06-10
Migrating to typescript

jasongrout:
This PR migrates many of the widgets to use typescript. We also start using a phosphor widget under the hood to provide the dom element. This will allow us in the future to use the phosphor event system to get good resize events, etc.

I also found and fixed several bugs from the typing (like the listen functions don't take a context).

Still todo: 
- [ ] testing framework
- [ ] check to make sure it works on the current notebook still
- [ ] make sure bqplot, pythreejs, ipyleaflet still work, as sanity checks for custom widgets.


SylvainCorlay:
Would you be ok walking me through the changes over a call. Since the ts + internal changes in the files makes it difficult to review, I would prefer not having too much changes accumulating on the top of this one...

jasongrout:
Definitely. We can do this today, or I can keep working on changes...

jasongrout:
and I'm also happy to rebase anything to make it easier to review. I realize it's quite a patch dump.

blink1073:
Will be nice when we don't have thousands of lines of `.d.ts` files checked in...

lbustelo:
declarativewidgets builds on top of the WidgetModel, so I'm assuming that this should not affect us. Thoughts?

jasongrout:
It *should* be backwards compatible, so hopefully it won't affect you, @lbustelo. We'll watch this closely, though.

jasongrout:
Another bug bites the dust! https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_int.js#L83 - the w3 spec says the things that you can set this to are *strings* `"true"`, `"false"`, or `"inherit"` (https://www.w3.org/TR/2012/WD-html5-20121025/editing.html#attr-contenteditable), otherwise a syntax error should be thrown. Typescript typing caught this.

jasongrout:
Another one bites the dust! https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_int.js#L566 should also be a string (see https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSPrimitiveValue).

jasongrout:
Here's a serious bug that typing complained about: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_int.js#L448

jasongrout:
Another one bites the dust! https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_string.js#L105 (calls the method with undefined, but the method will throw an error because of this)

jasongrout:
Typing makes another bug bite the dust! https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_string.js#L98 - setAttribute's value should be a string (see https://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-F68F082)

jasongrout:
Typing catches another bug! https://github.com/ipython/ipywidgets/blob/d3dc17843435539e8030d08197627c858d853223/jupyter-js-widgets/src/widget_selectioncontainer.js#L158 - textContent is a function on a jquery element. For an html element, it is an attribute that is set.

SylvainCorlay:
It would make it easier to review without the renaming to typescript. Is the `.ts` extension required?

jasongrout:
> It would make it easier to review without the renaming to typescript. Is the .ts extension required?

That's part of the reason I renamed the files first, and then made the changes to actually convert them to typescript.

But git is also smart about this: `git difftool -t meld --find-renames=10% master` is your friend!

jasongrout:
Another bug! In https://github.com/ipython/ipywidgets/blob/d3dc17843435539e8030d08197627c858d853223/jupyter-js-widgets/src/widget_selection.js#L751, the options parameter is shadowed by the variable created a few lines into the function.

jasongrout:
Still todo: 
- [ ] testing framework
- [ ] check to make sure it works on the current notebook still
- make sure bqplot, pythreejs, ipyleaflet still work, as sanity checks for custom widgets.

SylvainCorlay:
This PR fixes #14.

jasongrout:
Several more bugs! (several styling bugs, but these weren't caught by typing, but instead by going through all of the examples)

jasongrout:
@SylvainCorlay - tests pass for me, but it appears that travis has errors. Do tests pass for you?

SylvainCorlay:
Merging after in-person review and discussion with Jason, so that we can iterate. We are aware of the testing issue.

PR #639, opened 2016-06-21
Don't write empty widget state to metadata

### LINKED ISSUE #636: shouldn't write empty widget state to notebook metadata
Opening and saving a notebook adds an empty:

```python
{ 'widgets':
  'state': {},
  'version': '1.2.3',
}
```

to all notebooks. If there's no widget state to be saved, it should write no metadata to avoid this clutter and churn.

minrk:
This is two related changes, the first of which I'm pretty sure is the right thing to do, but I'm less sure about the second:

1. Don't save empty widget state dicts in metadata if there's no state to save.
2. Exclude widgets with no views from the metadata, since no information is persisted for these widgets when there are no views attached to them.

closes #636

### DIFF
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -200,16 +200,21 @@ WidgetManager.set_state_callbacks(function() {
     }
     return Promise.resolve({});
 }, function(state) {
-    // Only persist the views of the widget
-    var stateToSave = _.mapObject(state, function(widgetState) {
-        return _.pick(widgetState, 'views');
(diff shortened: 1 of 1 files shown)


Carreau:
+1

SylvainCorlay:
👍 Quick remark: we already don't save the state of widgets in the metadata, only the list of views.

I am +1 on not even including the list of views when it is empty.

SylvainCorlay:
LGTM. Merging.

SylvainCorlay:
Backported

PR #640, opened 2016-06-21
try default msg_throttle = 1

minrk:
It seems to offer a smoother experience than 3, which causes jerks as a slider moves across values


SylvainCorlay:
👍 I also checked with selections and bqplot and this seems to improve things.

PR #653, opened 2016-06-29
widgetsnbextension is not zip_safe

minrk:
Prevents installation of zipped-up eggs, for which the nbextension install will fail

cc @nthiery who encountered this.


SylvainCorlay:
Thanks! I really thought I fixed this one already. Must have been missed in a commit in a rebase.

SylvainCorlay:
Backported

PR #656, opened 2016-07-02
Convert testing to typescript and expect.js.

jasongrout:
It turns out that phosphor requires a newer version of Chrome than Chrome 37, which is the chromium available on Travis. Upgrading Chrome to the newest stable apparently requires a newer Ubuntu than available on Travis. So I'm giving up for now and disabling the Chrome on Travis.


jasongrout:
I think this is ready to merge.

blink1073:
testallthethings

blink1073:
(LGTM).

PR #658, opened 2016-07-04
Bugfixes

afshin:
- [x] Fix broken `web` example.
- [x] Fix broken `web2` example.
- [x] Fix broken `web4` example.
- [x] Use `getComputedStyle` to calculate border width instead of hard-coding it in `widget_selection`.
- [x] Fix missing `$` (`window.jQuery`) error in `widget_box`.
- [x] Fix occasionally broken destructuring initialization in `widget_link`.
- [x] Formatting - remove trailing whitespace in `widget`.
- [x] 	Only use `floor` or `ceil` when writing to the DOM in `widget_selection`.

cc @jasongrout @SylvainCorlay

### DIFF
--- a/jupyter-js-widgets/examples/web/manager.js
+++ b/jupyter-js-widgets/examples/web/manager.js
@@ -29,6 +29,7 @@ WidgetManager.prototype._create_comm = function() {
     return Promise.resolve({
         on_close: () => {},
         on_msg: () => {},
-        close: () => {}
+        close: () => {},
+        send: () => {}
     });
 }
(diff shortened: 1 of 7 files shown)


SylvainCorlay:
Thanks a lot for this @afshin I think that the typescript transition is going to really improve the jupyter-js-widgets codebase.

afshin:
Cheers! Yeah, I think you're right about TS :)

jasongrout:
Thanks @afshin!

PR #665, opened 2016-07-12
Add a JupyterLab extension npm package

jasongrout:


jasongrout:
@SylvainCorlay, the green button is below :).

jasongrout:
@SylvainCorlay - this should be ready for review now.

jasongrout:
Note to self: It's possible the commit deleting the phosphor css breaks widgets in the current notebook, since the current notebook probably doesn't load the phosphor css. Likely we should update the widgetsnbextension to include `phosphor/lib/base.css`

PR #685, opened 2016-07-21
Bug fix: hard code jquery-ui version that includes smooth theme.

afshin:
The `jquery-ui` version in the current `package.json` was `^1.10.5`, which is a version that does not exist. This causes `npm` to instead download `jquery-ui@1.12.0`:
```
$ npm install --save jquery-ui@^1.10.5
temp@ /Users/af/Desktop/temp
└── jquery-ui@1.12.0
```

That version does not come bundled with themes.

The last version of `jquery-ui` that includes the `smoothness` theme which the webpack bundle needs for building is `jquery-ui@1.10.4`, so this PR hard codes that version.

### DIFF
--- a/jupyter-js-widgets/package.json
+++ b/jupyter-js-widgets/package.json
@@ -72,7 +72,7 @@
     "bootstrap": "^3.3.5",
     "d3-format": "^0.5.1",
     "jquery": "^2.1.4",
-    "jquery-ui": "^1.10.5",
+    "jquery-ui": "1.10.4",
     "jupyter-js-services": "^0.15.1",
     "lolex": "^1.4.0",
     "phosphor": "^0.5.0",


SylvainCorlay:
Thanks!

jasongrout:
Yes, thanks! I'll also take a look at moving to no-ui slider...

SylvainCorlay:
@jasongrout support for native slider is pretty broad now. We could probably use no-ui only for the range.

Btw, in which time zone are you?

afshin:
Good question.

jasongrout:
Hey, even IE 11 supports `<input type="range">`! I guess the only thing is if we want to style the slider, it will be much easier with noui slider, I think. Especially if we want the range slider to styled the same as the normal slider.

jasongrout:
Time zone? The 24-hour time zone, man. :)

SylvainCorlay:
Styling the range input:
https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/

jasongrout:
Yes, it's possible. We'll need to coordinate styles, though, and we lose on consistency between range sliders and normal sliders as far as capabilities.

SylvainCorlay:
I just came across https://github.com/Wildhoney/ngRangeSlider.

jasongrout:
Very interesting!

afshin:
But this is an Angular directive ... it's only lightweight if you're in an Angular app, otherwise it's huge. Clever idea, though.

SylvainCorlay:
yeah, I just meant the idea of using native range input.

SylvainCorlay:
http://stackoverflow.com/questions/4753946/html5-form-slider-with-two-inputs-possible

afshin:
Cool, re: the no-dependency modification of the wildhoney slider.

PR #691, opened 2016-07-25
Document overflow-x/y properties

SylvainCorlay:


hainm:
the `Widget Styling.ipynb` looks very nice. It's great if more notebook examples would be added.

PR #701, opened 2016-08-07
Updating broken links

PedroGFonseca:
Assumed the correct links were to [this page](https://github.com/ipython/ipywidgets/tree/master/docs/source/examples)

### DIFF
--- a/docs/source/examples/Index.ipynb
+++ b/docs/source/examples/Index.ipynb
@@ -32,12 +32,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "- [Using Interact](Using Interact.ipynb)\n",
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for the PR!

These should in fact all be relative URLs so that they resolve correctly in different contexts (nbviewer, github, live notebook most importantly). Where do you see that these links are broken?

jdfreder:
Ditto the thanks, but as @minrk says, the links should be relative.  If you view the same file you changed [on GitHub in master](https://github.com/ipython/ipywidgets/blob/master/docs/source/examples/Index.ipynb), you'll notice the links work correctly.  The links not working may be a symptom of how you have your server setup.

PR #705, opened 2016-08-08
CUnicode -> Unicode

SylvainCorlay:
It is a common mistake for users to try to set `width` or `height` to an integer value which is not a valid value for the property. This makes it clearer that a string is expected.


minrk:
Yup, in general `C` traits should be avoided unless there's a really good reason, especially CUnicode.

PR #712, opened 2016-08-14
interactive: Add argument for suppressing result output.

josteinbf:
The flag can typically be used when the output is large, e.g. a numpy
array.

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -199,6 +199,7 @@ def interactive(__interact_f, **kwargs):
     f = __interact_f
     co = kwargs.pop('clear_output', True)
     manual = kwargs.pop('__manual', False)
+    output_result = kwargs.pop('__output_result', True)
     kwargs_widgets = []
     container = Box(_dom_classes=['widget-interact'])
     container.result = None
@@ -247,7 +248,7 @@ def interactive(__interact_f, **kwargs):
                 if co:
                     clear_output(wait=True)
                 container.result = f(**container.kwargs)
-                if container.result is not None:
+                if output_result and container.result is not None:
                     display(container.result)
         except Exception as e:
             ip = get_ipython()


josteinbf:
How should parameters like these to `interactive` be documented? I see that `clear_output` is not documented.

jdfreder:
Clear output should be documented in the interactive example notebook.   What's your use case for this flag?

josteinbf:
I did not add the clear output flag, and I do not know what the use case for that flag is. If you are asking about the use case for the output_result flag, see the first comment above.

jdfreder:
I think the use of interact without output rendering is a little strange because it's not what interact was intended for.  Interact was intended to be used to explore the input/output relationship of functions.  General UIs are usually constructed "manually" by instantiating the widgets directly.  That said, I don't see why this would hurt (other than API bloat / scope creep).  I'll differ to @ellisonbg or @SylvainCorlay 's judgement for a :+1: or :-1: .

As an aside, you should be able to accomplish a similar effect by wrapping your function in a lambda that returns nothing.

SylvainCorlay:
I am not a good judge at all since not a big fan of the `interact` way of doing things altogether. Using widgets directly is already very simple. 

For a python beginner, `interact` is too magic: decorator syntax, introspection, widget abbreviation, nontrivial behavior wrt output.

My only objective with interact at the moment is to make it compatible with the API of sage so that sage's migration to ipywidgets is as easy as possible.

raphaeltimbo:
I was looking for something like this... I hope it will be merged.
For example, Lorenz Differential Equations.ipynb has a huge array as output and sometimes I only want to see the plot. Delete the line return 'array' in this case is a possibility (all the code in inside the notebook), but sometimes the function that I am using is in a .py file and things become more complicated.

jdemeyer:
I'm not commenting on whether this is a good idea or not, but if you do implement, it should be an option (see #771).

jasongrout:
Here's a situation where output isn't wanted:

```python

def f(x,y):
    print(x)
    print(y)
    return 'Huge printed output: %s'%(x+y)

interact(f, x=10, y=(1,20))
```

and here's a solution that uses the functools.wraps method:
```python
from functools import wraps
@interact(x=10, y=(1,20))
@wraps(f)
def myfun(*args, **kwargs):
    f(*args, **kwargs)
```
I'm on the fence whether we should encourage users to use something like the above workaround, or provide an option. Thoughts from @josteinbf?

jasongrout:
I'm going to close this since the wrapping solution above has a smaller API surface and hews more closely to the ideas of the interact decorator.

Issue #711, opened 2016-08-13
Multiple kernels causes duplicate widget javascript error


jeffyjefflabs:
I have a Jupyter extension that creates a second kernel on the fly while viewing a notebook.  That results in a ["widget with that name already exists"](https://github.com/jupyter/notebook/blob/41d6da235cbf3bcf6d7f818e11a066e0fd12ff8b/notebook/static/base/js/notificationarea.js#L68) error.  I was able to work around it by applying this patch after running the `jupyter nbextension enable` command:

```
--- /usr/share/jupyter/nbextensions/jupyter-js-widgets/extension.js.orig
+++ /usr/share/jupyter/nbextensions/jupyter-js-widgets/extension.js
@@ -394,7 +394,7 @@
         * Registers manager level actions with the notebook actions list
         */
        WidgetManager.prototype._init_actions = function() {
-           var notifier = Jupyter.notification_area.new_notification_widget('widgets');
+           var notifier = Jupyter.notification_area.widget('widgets');
            this.buildSnapshotsAction = {
                handler: (function() {
                    this.updateSnapshots().then((function() {
```

Calling [`widget`](https://github.com/jupyter/notebook/blob/41d6da235cbf3bcf6d7f818e11a066e0fd12ff8b/notebook/static/base/js/notificationarea.js#L39) instead of [`new_notification_widget`](https://github.com/jupyter/notebook/blob/41d6da235cbf3bcf6d7f818e11a066e0fd12ff8b/notebook/static/base/js/notificationarea.js#L66) will use the existing widget if it already exists or create a new one if it does not.  It looks like [this](https://github.com/ipython/ipywidgets/blob/b434cbcdce8771b70e0f57ea10a66091c8fb7042/widgetsnbextension/src/manager.js#L248) is the original location of that code -- could the same change be made there?

jdfreder:
That should be okay, can you open a PR with your patch?  Multiple kernels is something we haven't tried yet, I'm surprised this is the only bug you found.

jeffyjefflabs:
Fixed in #713 - thanks!

PR #713, opened 2016-08-15
Resolve duplicate widget error

jeffyjefflabs:
This function will use the existing widget if it exists or create a new one if not.  See issue #711.


jdfreder:
Thanks!

PR #722, opened 2016-08-18
Use isinstance(x, Mapping) to check for dict-like types

jdemeyer:
Using the collections ABC is better than listing possible types.


SylvainCorlay:
:+1: this looks good to me. Waiting for Travis to get green.

PR #727, opened 2016-08-20
Update for JupyterLab 0.3.0

jasongrout:
Updates that work with the third party extension mechanism in JupyterLab 0.3.0


jasongrout:
Passes (minimal) testing, and only affects really the lab extension, so merging.

PR #731, opened 2016-08-22
Hide overflow of widget descriptions

SylvainCorlay:
- overflow hidden for widget description and button caption.
 - slightly smaller min-width for slider readout
 - larger widget description width.


willingc:
LGTM. Thanks @SylvainCorlay.

Issue #738, opened 2016-08-25
Resetting an attribute to default doesn't sync


justinmc:
Setting an attribute in a model back to its original default is not synced to the server as it should be.

```js

    // This will be successfully synced to Python
    model.set('myparam', 'notthedefault');
    model.save();

    ...

    // If model's default is empty string, this will not be
    // synced, so Python will still see myparam as "notthedefault"
    model.set('myparam': '');
    model.save();
```

I believe this is caused by [this line](https://github.com/ipython/ipywidgets/blob/eb4dcd9956d88ae842bff7822a9c32fc038906dd/jupyter-js-widgets/src/widget.ts#L312) in WidgetModel's `sync` method, which is passing what seems to be an unrelated `options` object into [get_state](https://github.com/ipython/ipywidgets/blob/eb4dcd9956d88ae842bff7822a9c32fc038906dd/jupyter-js-widgets/src/widget.ts#L207), which expects a boolean.

I will attempt a pull request really quick, but I'm not totally sure what the original intent was in passing the options object here.

SylvainCorlay:
Closed with #739

PR #739, opened 2016-08-25
Fix model change to default not syncing

justinmc:
See my issue here: https://github.com/ipython/ipywidgets/issues/738

Looking at another call to `model.get_state` as an example [here](https://github.com/ipython/ipywidgets/blob/eb4dcd9956d88ae842bff7822a9c32fc038906dd/jupyter-js-widgets/src/manager-base.ts#L350), I believe this is just a simple typo where `options` was passed instead of `options.drop_defaults` [here](https://github.com/ipython/ipywidgets/blob/eb4dcd9956d88ae842bff7822a9c32fc038906dd/jupyter-js-widgets/src/widget.ts#L312).

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -309,7 +309,7 @@ class WidgetModel extends Backbone.Model {
             return false;
         }
 
-        var attrs = (method === 'patch') ? options.attrs : model.get_state(options);
+        var attrs = (method === 'patch') ? options.attrs : model.get_state(options.drop_defaults);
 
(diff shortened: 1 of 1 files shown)


justinmc:
@jdfreder @SylvainCorlay Can someone take a look at this when you get a chance?  Easy one liner that hopefully fixes a big bug.

SylvainCorlay:
Thanks!
This looks good to me. I will do more testing tomorrow on what may be the other implications.

justinmc:
@SylvainCorlay Thank you!

SylvainCorlay:
Thanks again. This will be backported to the next 5.x.

PR #744, opened 2016-08-31
Support more general container types as abbreviations

jdemeyer:
This pull request adds or improves support for various container types used as abbreviations for `@interact`:

1. Properly support a list of 2-tuples: the first element of each tuple should be converted to a string.
2. Support any kind of `Mapping` instead of just `dict` (and `OrderedDict` which subclasses `dict`).
3. Support any kind of `Iterable` instead of only `list`.

Tests are added for all new cases.


minrk:
Looks good to me, thanks for the tests! @SylvainCorlay anything I'm forgetting?

SylvainCorlay:
Can we wait until tomorrow? I am afk tonight.

minrk:
Yup, no problem.

jdemeyer:
ping...

PR #745, opened 2016-08-31
Make interactive a class to allow future customizations

jdemeyer:
This changes `interactive` from a function to a class to allow future customizations using subclassing. There are no user-visible changes at all.

For example, it should be possible to customize the way how abbreviations are converted to widgets.


SylvainCorlay:
👍 after in-person review of the PR.

PR #748, opened 2016-08-31
Avoid conversion using type(value)

jdemeyer:
We should not assume that the following works to convert `1` to the same "type" as `value`:
```
t = type(value)
x = t(1)
```
In particular, this would break for many SageMath types.

As alternative, doing arithmetic with the given value does work for SageMath types and all standard Python types.

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -52,8 +52,8 @@ def _get_min_max_value(min, max, value=None, step=None):
         if not isinstance(value, Real):
             raise TypeError('expected a real number, got: %r' % value)
         if not value:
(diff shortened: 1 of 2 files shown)


ipywidgets/widgets/interaction.py
```diff
@@ -52,8 +52,8 @@ def _get_min_max_value(min, max, value=None, step=None):
         if not isinstance(value, Real):
             raise TypeError('expected a real number, got: %r' % value)
         if not value:
```
minrk: Can you make this `if value == 0`? It makes it a little clearer to me that this is definitely a number and not the None condition above.

minrk:
Thanks!

PR #751, opened 2016-09-04
Add CommManager.comms shim

philippjfr:
The current jupyter-js-services shim for the CommManager, which gets used by project's like the dashboard_server does not mirror the API of the real [CommManager](https://github.com/jupyter/notebook/blob/master/notebook/static/services/kernels/comm.js) completely, which is generally no problem but I was hoping to at least expose a means to look up existing comms. bokeh 0.12.2 will use this part of the comm_manager API to register an on_msg callback on an existing Comm. In future bokeh will probably correctly set up a notebook extension and/or display plots as a widget so this should no longer be needed but for the time being it would be great if we could expose the ``comms`` on the shim. Tested with bokeh 0.12.2rc2 and the current dashboard_server master.

### DIFF
--- a/jupyter-js-widgets/src/services-shim.ts
+++ b/jupyter-js-widgets/src/services-shim.ts
@@ -47,6 +47,7 @@ namespace shims {
              */
             new_comm(target_name: string, data: any, callbacks: any, metadata: any, comm_id: string): Comm {
                 var comm = new Comm(this.jsServicesKernel.connectToComm(target_name, comm_id));
+                this.register_comm(comm);
                 comm.open(data, callbacks, metadata);
                 return comm;
             };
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
ping @jasongrout

jupyter-js-widgets/src/services-shim.ts
```diff
@@ -58,9 +59,10 @@ namespace shims {
              *                         comm is made.  Signature of f(comm, msg).
              */
             register_target (target_name, f) {
-                var handle = this.jsServicesKernel.registerCommTarget(target_name, function(jsServicesComm, msg) {
+                var handle = this.jsServicesKernel.registerCommTarget(target_name, _.bind(function(jsServicesComm, msg) {
```
jasongrout: You can use an arrow function here to automatically bind this:

```
(jsServicesComm, msg) => {
```

philippjfr: Thanks, will do.

philippjfr:
@jasongrout I made the change you suggested. Do you have any other objections? Otherwise this is ready to merge I think.

PR #759, opened 2016-09-08
Allow arbitrary options to be passed to model.save by client

abelnation:
Small change to allow additional options to be passed to backbone's `model.save` method.  This would allow, e.g. passing `success`/`failure` callbacks so that code can be executed after save successfully completes.

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -411,9 +411,11 @@ class WidgetModel extends Backbone.Model {
      *
      * This invokes a Backbone.Sync.
      */
-    save_changes(callbacks?) {
+    save_changes(callbacks?, options={}) {
         if (this.comm_live) {
-            let options: any = {patch: true}
+            _.defaults(options, {
+                patch: true
+            });
             if (callbacks) {
                 options.callbacks = callbacks;
             }


jupyter-js-widgets/src/widget.ts
```diff
@@ -411,9 +411,11 @@ class WidgetModel extends Backbone.Model {
      *
      * This invokes a Backbone.Sync.
      */
-    save_changes(callbacks?) {
+    save_changes(callbacks?, options={}) {
```
SylvainCorlay: `options?: any`?

jasongrout:
@abelnation - this won't do anything - the problem is that we override the actual sync method, and we never call the success/fail callbacks because we are using comms, which don't give a response to requests. See, for example, the comment at https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget.ts#L362.

We use some of the backbone infrastructure, but we've really customized backbone a *lot*, so much that sometimes it's not really recognizable as backbone anymore. I think eventually we'll just finish the job and eliminate our dependency on backbone.

jasongrout:
(I almost did this exact thing today when trying to help fix #833, but then I realized it wouldn't do anything anyway for the above reasons.)

Thanks for the suggestion. I'm going to close this for now, since it would involve a lot more work to have these success/fail callbacks work.

PR #762, opened 2016-09-12
Change most functions in interaction.py to methods of interactive

jdemeyer:
In order to allow users to override certain behaviours of `interactive`, change most functions from `interaction.py` to become methods of the class `interactive`. In the process, some functions are renamed and the two functions `_widget_from_abbrev` and `_widget_abbrev` have been merged to one method `widget_from_abbrev`.

Depends on #725.

Needed for https://github.com/OpenDreamKit/OpenDreamKit/issues/94

### DIFF
--- a/ipywidgets/widgets/__init__.py
+++ b/ipywidgets/widgets/__init__.py
@@ -1,5 +1,6 @@
 from .widget import Widget, CallbackDispatcher, register, widget_serialization, handle_version_comm_opened
 from .domwidget import DOMWidget
+from .valuewidget import ValueWidget
 
 from .trait_types import Color, EventfulDict, EventfulList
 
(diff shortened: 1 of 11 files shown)


SylvainCorlay:
LGTM. Are you still adding to this?

jdemeyer:
No, I consider it done.

SylvainCorlay:
ok, so I guess that adding the `get_interact_value` / `get_value` default implementation that returns self.value comes in another PR?

jdemeyer:
> ok, so I guess that adding the get_interact_value / get_value default implementation that returns self.value comes in another PR?

See #725

SylvainCorlay:
woohoo!

PR #765, opened 2016-09-12
VBox and HBox should be classes

jdemeyer:
Make `HBox` and `VBox` classes instead of functions to allow for subclassing.


SylvainCorlay:
👍

SylvainCorlay:
Instead of doing that, we could use a default value generator for layout

```python
@default('layout')
def _default_layout(self):
    return Layout(align_items='stretch', display='flex')
````

SylvainCorlay:
This would prevent sending the original state and then updates to the layout...

PR #767, opened 2016-09-13
Use default value generator for HBox and VBox

SylvainCorlay:
This should allow user to override the layout of vbox and hbox.


SylvainCorlay:
ping @jdemeyer FYI

PR #771, opened 2016-09-14
Interact options

jdemeyer:
Implement #430:
```
@interact.option(manual=True)
def function(...):
      pass
```

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -155,21 +155,30 @@ class interactive(Box):
         # Otherwise, it is triggered for every trait change received
         # On-demand running also suppresses running the function with the initial parameters
         if self.manual:
-            self.manual_button.on_click(self.call_f)
+            self.manual_button.on_click(self.update)
 
             # Also register input handlers on text areas, so the user can hit return to
             # invoke execution.
             for w in self.kwargs_widgets:
                 if isinstance(w, Text):
-                    w.on_submit(self.call_f)
+                    w.on_submit(self.update)
         else:
             for widget in self.kwargs_widgets:
-                widget.observe(self.call_f, names='value')
+                widget.observe(self.update, names='value')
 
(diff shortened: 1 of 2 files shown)


jdemeyer:
No idea why Travis CI is failing here. It seems to be complaining about Javascript-related issues while I didn't make any changes to that.

SylvainCorlay:
kicked the build

SylvainCorlay:
Leaving this open to give @carreau the opportunity to comment.

jdemeyer:
Let me just remind that only the last 2 commits need to be reviewed (the first commits are from #762).

SylvainCorlay:
@jdemeyer if you rebase, we should only see those commits.

jdemeyer:
> if you rebase, we should only see those commits.

`<rant>`
I consider that a GitHub bug/misfeature since those first commits are literally (with the same hash) in `master` now, so it shouldn't show them in the diff.

This is IMHO yet another thing that Trac gets right but GitHub does not.
`</rant>`

Anyway, should I rebase or merge in `master` or just leave it?

SylvainCorlay:
I completely agree. I am out for lunch and my desktop computer has 2 ongoing 3-pane merge windows open. I am constantly in the middle of a rebase.

Meld is my ide :)

But I prefer rebase over merge and here the rebase would not require any manual action...

jdemeyer:
Done.

Carreau:
I'm happy with that, I'll trust Sylvain on it, haven't touched widgets in a while.

minrk:
Nice!

jasongrout:
Noting for future people: one huge issue with this implementation is that it can't be used as a decorator anymore:
```python
@interact.options(manual=True)(x=(0,10))
def f(x):
    print(x)
```
gives a syntax error. That's because decorators are syntactically limited to a single function call. A way around it is to store the partial interact as another variable:

```python
myinteract = interact.options(manual=True)
@myinteract(x=(0,10))
def f(x):
    print(x)
```
(for convenience, an ipywidgets.interact_manual is defined with this option.)

PR #774, opened 2016-09-15
Update to jupyterlab-extension-builder 0.4.0.

jasongrout:


widgetslabextension/scripts/buildExtension.js
```diff
@@ -7,11 +7,6 @@ var buildExtension = require('jupyterlab-extension-builder').buildExtension;
...
   name: 'widgetslabextension',
-  entryPath: './lib/plugin.js',
-  extractCSS: true,
-  config: {
-      output: {
-        path: path.join(process.cwd(), 'widgetslabextension', 'static'),
-    }
-  }
+  entry: './lib/plugin',
+  outputDir : 'widgetslabextension/static'
```
blink1073: This works without a leading `./`?

jasongrout: yes. It uses npm's `path.resolve`. Though it's clearer with a leading `./`, so I'll put it in.

jasongrout:
Thanks!

PR #775, opened 2016-09-16
Reset to the correct default

### LINKED ISSUE #632: Layout css trait default values unused
It appears in https://github.com/ipython/ipywidgets/blob/076f69845373aeecb55a3cfc6314d37823c79878/jupyter-js-widgets/src/widget_layout.js#L10 that the default values would be used when unlayouting an element, but no, it's assumed that the default value is always `''` in https://github.com/ipython/ipywidgets/blob/076f69845373aeecb55a3cfc6314d37823c79878/jupyter-js-widgets/src/widget_layout.js#L123.

SylvainCorlay:
@jasongrout  Fixes #632


jupyter-js-widgets/src/widget_layout.ts
```diff
@@ -112,7 +112,7 @@ class LayoutView extends DOMWidgetView {
         this._traitNames.forEach((trait) => {
             this.displayed.then((parent) => {
                 if (parent) {
-                    parent.el.style[this.css_name(trait)] = '';
+                    parent.el.style[this.css_name(trait)] = css_properties[trait];
```
SylvainCorlay: I was wondering if we should have a private `set_css(el, name, value)` which in fact does `el.style.removeProperty(name)` is `value` is an empty string.

SylvainCorlay: d

SylvainCorlay:
Rebased. 
This is another take on it. I basically adopt the bqplot approach which is to consider that `None` stands for an unset css property. (and not a css property of value 'unset' :) )

PR #778, opened 2016-09-16
WIP More jquery removal

SylvainCorlay:


SylvainCorlay:
@afshin this and #776 were leftover from the jquery removal. I am not sure what the `collapse` was supposed to do.

afshin:
Collapsing is just for accordions, I think. Good eye catching those jQuery methods.

SylvainCorlay:
I just wonder how this even works, unless bootstrap does voodoo with the css selectors that we are using. `page.collapse` is only used when updating from the backend or another view.

afshin:
Yeah, you're right, I can't tell where that class name comes from, either.

jasongrout:
This is superseded by #837.

jasongrout:
And it was bootstrap - took me a while to realize it wasn't actually magic.

PR #783, opened 2016-09-21
Move widgetslabextension -> jupyterlab_widgets/jupyterlab-widgets

jasongrout:
For discussion - on pypi, the name is `jupyterlab_widgets`, on npm it's `jupyterlab-widgets`. Should these be consistent? It's a bit confusing, but each makes sense in their own ecosystem. Thoughts, @SylvainCorlay, @ellisonbg?


jasongrout:
(CC @blink1073, @afshin, @sccolbert too)

We're going to have this pattern a lot, where a python package provides an npm package to the server. Do we strive to have the same names, or convert underscores to dashes, or some other convention to match up the python and npm packages?

jasongrout:
In discussion with @blink1073, @afshin, and @blink1073, we decided that having consistency between the package names was more important than respecting npm convention, so an npm package of jupyterlab_widgets it is!

minrk:
PyPI normalizes underscores to hyphen as well, so `pip install jupyterlab-widgets` will work. Conda doesn't do this, which causes confusion. Personally, I would be inclined to use hyphen in distribution names everywhere, as that seems to be the more common convention, but we didn't do that for the first round of Jupyter packages.

PR #801, opened 2016-09-29
Migrate to typescript 2

SylvainCorlay:
Besides I am requiring `jupyter-js-widgets` by file path like in widgetsnbextension in the labextension.

@jasongrout


SylvainCorlay:
PS: this is required for the update to jupyterlab 0.4.x

SylvainCorlay:
Self marging since this is required to work with the latest jupyterlab. re-ping @jasongrout .

PR #807, opened 2016-10-02
Improve build on RTD

willingc:
- Update `environment.yml` and `requirements.txt` to align package versions
- Replace `bash build-local.sh` in `build-rtd.sh` as RTD would not execute the bash command but will execute the individual commands contained in the script.
- Updated the suffix for `dev_doc` as it is a `.rst` file not `.md`. Also updated the environment name.
- Simplified and documented `conf.py` to be more consistent with Sphinx and other Jupyter projects 
- In `conf.py` enabled support for rst, md, and ipynb since all are needed

What to iterate on
- There are a number of nbsphinx image warnings though docs build
- There are still the JavaScript errors in `Widget Styling.ipynb`
- There are some places where data type is not recognized. Perhaps not recognizing `application/vnd.jupyter.widget` as a recent change?

Rendered Docs: http://test-widgets.readthedocs.io/en/fix-build/

### DIFF
--- a/docs/build-rtd.sh
+++ b/docs/build-rtd.sh
@@ -4,4 +4,7 @@ nvm install 5
 nvm use 5
 nvm alias default 5
 npm install -g npm
-bash build-local.sh
+cd ../../jupyter-js-widgets
+npm install
+cd ../docs
+npm install
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,6 +1,7 @@
 name: ipywidget_docs
 channels:
 - conda-forge
+- defaults
 dependencies:
 - python=3.5
 - sphinx_rtd_theme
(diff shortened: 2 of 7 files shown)


SylvainCorlay:
Thanks a lot!

SylvainCorlay:
Hum, the build on rtd still fails.

SylvainCorlay:
I wonder if this is because of the alpha requirement to the rtd theme. Do we need to require the rtd theme at all?

willingc:
@SylvainCorlay More likely it's that RTD's build often times out or doesn't clear old cruft. I will try to kick it a few times. Sometimes wiping the build in the versions menu helps. Other times it takes disabling and reenabling the particular branch build.

willingc:
@SylvainCorlay Victory! Cajoling RTD got the build passing there.

SylvainCorlay:
Awesome thanks!

PR #810, opened 2016-10-05
Update Widget List.ipynb

gnestor:
Update Widget List.ipynb to include all widgets and properties/keys for each widget.

https://github.com/gnestor/ipywidgets/blob/widget-list/docs/source/examples/Widget%20List.ipynb


gnestor:
@SylvainCorlay Am I missing widgets? On a side note: Is there a way to view this notebook live without checking out this branch, etc? I tried nbviewer and Binder and neither seem to support ipywidgets. If I created a Docker image I could use tmpnb to host it, but that's a lot of work...

SylvainCorlay:
Thanks, awesome!

PR #819, opened 2016-10-06
2.0

jasongrout:
Update jupyter-js-widgets to 2.0.0, but publish it with the next tag:

```
npm publish --tag=next
```

We do this so that semver for `^2.0.0` works correctly, so we don't keep breaking things when we make a patch release of this dev version.

### DIFF
--- a/jupyter-js-widgets/package.json
+++ b/jupyter-js-widgets/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jupyter-js-widgets",
-  "version": "2.0.0-dev.23",
+  "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
:+1:

SylvainCorlay:
extensions depend on jupyterlab_widgets. should we have a `jupyterlab_widgets 0.1.0 tag=next` ?

jasongrout:
Extensions should just depend on the semver they need, I think. The only reason we did this is because of the weird semantics around dev pre-releases. We don't have that problem with jupyterlab_widgets

PR #824, opened 2016-10-11
PR #725 broke the fixed() functionality of @interact. This fixes it.

jasongrout:
See #725.

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -507,3 +507,9 @@ class fixed(HasTraits):
     description = Unicode('', help="Any Python object")
     def __init__(self, value, **kwargs):
         super(fixed, self).__init__(value=value, **kwargs)
+    def get_interact_value(self):
+        """Return the value for this widget which should be passed to
+        interactive functions. Custom widgets can change this method
+        to process the raw value ``self.value``.
+        """
+        return self.value


jasongrout:
CC @jdemeyer, @SylvainCorlay.

jdemeyer:
If something is undocumented and untested, it's not surprising that it breaks. I would suggest that you add a testcase.

jdemeyer:
Speaking of `fixed`, I don't really understand why it's not just a widget. This `fixed` class looks like an ugly hack that somebody once added.

jasongrout:
> If something is undocumented and untested, it's not surprising that it breaks. I would suggest that you add a testcase.

Point taken. I noticed it because the Image Processing example notebook broke (which is sort of a manual test case :).

jasongrout:
> Speaking of fixed, I don't really understand why it's not just a widget. This fixed class looks like an ugly hack that somebody once added.

It's sort of the python half of a widget, which is all that is needed here since it is just maintaining the value in the kernel.

I'm certainly open to introducing currying or partial functions in a different way. Perhaps we could just rewrite the examples to use `functools.partial`.

jasongrout:
> This fixed class looks like an ugly hack that somebody once added.

I just talked to @ellisonbg, who I think implemented it originally, and it appears it was a concession to simplify things for beginners.

SylvainCorlay:
:+1:

minrk:
I'm in favor of using and encouraging and documenting `functools.partial` over `fixed`. I think we can keep `fixed` for those who aren't comfortable with currying without any real downside. A test for exercise would definitely be good, though.

PR #828, opened 2016-10-13
fixing setup.py for widgetsnbextension

ssunkara1:
This is needed for installing in windows

### DIFF
--- a/widgetsnbextension/setup.py
+++ b/widgetsnbextension/setup.py
@@ -128,7 +128,9 @@ class NPM(Command):
         
     def has_npm(self):
         try:
-            check_call(['npm', '--version'])
+            ## shell=True needs to be passed for windows to look at non .exe files.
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
There is another instance of `check_call` in setup.py. Don't we need to make the same change there too?

ssunkara1:
@SylvainCorlay I changed that too. I removed a function which seemed to be not doing much. Have a look.

SylvainCorlay:
Thanks

SylvainCorlay:
This looks good to me. I will wait a little bit to let @jasongrout comment.

@ssunkara1 when this gets in, would you be OK to contribute the same change in the cookiecutter ?

ssunkara1:
@SylvainCorlay sure. I will do that.

jasongrout:
This looks good to me too.

Issue #829, opened 2016-10-13
Order of functions called in WidgetModel


ssunkara1:
Hey all,

This is more of a question than an issue. So. in `WidgetModel` previously, the function `constructor` is called before the function `set`. This is probably what I would have expected.

But in the current master, the function `initialize`, which looks like a replacement for the `constructor`, is called after the function `set`. I wouldn't have expected that. 

I wanted to find out if this change is intentional and if yes, any reason why this was changed?

Thanks.

Version details:
By previously, I meant `ipywidgets` 5.2.2. And now, I am running off `master`.

ssunkara1:
There seems to be a difference between the `constructor` and `initialize`. 

https://github.com/jashkenas/backbone/issues/720

But I guess we want to use `constructor` instead of `initialize`?

ssunkara1:
@jasongrout

gnestor:
@jasongrout It looks like you switched from `constructor` to `initialize` in https://github.com/ipython/ipywidgets/commit/386bcd61f65ba8c80309aaa4080748ae1a6c9088. @ssunkara1 Can you elaborate about your use case (e.g. what you would like to do before initializing the model)?

jasongrout:
Yes, it was intentional. See https://github.com/ipython/ipywidgets/issues/677#issuecomment-253834228 for more details.

jasongrout:
(Happy to discuss further if needed. @ssunkara1 and I discussed this in-person as well.)

PR #832, opened 2016-10-14
Initializing buffered_state_diff correctly

ssunkara1:
In widgets, we changed from over-riding `constructor` to over-riding `initialize`. But `initialize` is called after set and I think we want `_buffered_state_diff` initialized before, because in `set` we try to update the `_buffered_state_diff`. 

After a discussion with @jasongrout, we decided that this is the minimally intrusive change that we can make to have `_buffered_state_diff` initialized before `set` is called.

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -88,7 +88,6 @@ class WidgetModel extends Backbone.Model {
         let comm = options.comm;
 
         this.state_change = Promise.resolve();
-        this._buffered_state_diff = {};
         this.pending_msgs = 0;
         this.msg_buffer = null;
         this.state_lock = null;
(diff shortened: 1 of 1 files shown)


ssunkara1:
Not sure if anything else needs to be fixed. As I found this: https://github.com/ipython/ipywidgets/pull/266

@SylvainCorlay Can you just check how this will affect creating widgets from the javascript side itself? I think because of the previous change, that will be affected.

ssunkara1:
This addresses the question raised in #829

jasongrout:
Is this superseded by #833?

ssunkara1:
@jasongrout #833 is on top of this. This is just fixing the bug with the `constructor` and `initialize` call. The `_buffered_state_diff` change is a separate fix.

jasongrout:
This looks good to me.

PR #833, opened 2016-10-14
Buffered state diff bug fix

ssunkara1:
This is on top of my PR #832. 

I think, the root issue is that in WidgetModel the attribute _buffered_state_diff is not set correctly on initialization. So when the first sync from javascript happens, the change handlers of all the attributes are notified. Also, if any of the other attributes changed, they are being overwritten by the old values which came back in the comm message.

### DIFF
--- a/jupyter-js-widgets/src/manager-base.ts
+++ b/jupyter-js-widgets/src/manager-base.ts
@@ -216,9 +216,9 @@ abstract class ManagerBase<T> {
             options_clone.comm = comm;
             let widget_model = this.new_model(options_clone, serialized_state);
             widget_model.then(model => {
-+                model.sync('create', model);
-+                return model;
-+            });
+                model.sync('create', model);
+                return model;
+            });
         }, () => {
             // Comm Promise Rejected.
             if (!options_clone.model_id) {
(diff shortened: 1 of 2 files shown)


ssunkara1:
@SylvainCorlay I think this change might cause an issue with loading widgets just from the `javascript` side. I do not know the exact code, but I believe that it relies on `_buffered_state_diff` being set on `initialization`. 

If you can point me to how we can recognize if a widget is initialized from `javascript` side how I can identify, I'll make changes to ensure that the `_buffered_state_diff` is set correctly.

SylvainCorlay:
@ssunkara1 I will dig into this logic when I am back from Milan. In the meantime, you can test with widgets created from the js side:

- the panzoom widget in bqplot is created by the toolbar on the js side.
- controller axes and buttons are created from the js side.

jasongrout:
Looking at this...

jasongrout:
Submitted a PR to your PR.

jupyter-js-widgets/src/manager-base.ts
```diff
@@ -214,7 +214,11 @@ abstract class ManagerBase<T> {
         return commPromise.then((comm) => {
             // Comm Promise Resolved.
             options_clone.comm = comm;
-            return this.new_model(options_clone, serialized_state);
+            let widget_model = this.new_model(options_clone, serialized_state);
+            widget_model.then(model => {
```
jasongrout: I think you'll need to return this promise: `return widget_model.then...`

ssunkara1:
Thanks @jasongrout Tested this. Seems to be good. Have a look.

jasongrout:
This looks good to me. Thanks for working on this!

PR #837, opened 2016-10-17
Make the Accordion widget not depend on Bootstrap.

jasongrout:
I think this is the last of the widgets that depend on Bootstrap. It certainly needs some CSS love, and maybe we can use CSS transitions instead of jQuery? However, this is a start that we can iterate from.


jasongrout:
CC @afshin, @SylvainCorlay for comments and/or merging.

jasongrout:
@ssunkara1, @dmadeka  - if you have some time, do you mind looking at this accordion widget implementation and giving feedback? It's *very* basic, but is a starting point for iterating (especially iterating on styling...)

SylvainCorlay:
Should this allow us to remove bootstrap from the package.json of jupyter-js-widgets and the web examples?

jasongrout:
Oh yeah! Hopefully! I'll try...

jasongrout:
@SylvainCorlay, try this out. I cleaned up the examples, found and fixed a bug in IntSlider, and did a bunch of other related clean-up tasks.

jasongrout:
Since the example cleanup was a big project in its own right, I made it #843 to make this easier to review.

SylvainCorlay:
We could allow None for the selected index?

gnestor:
I agree with ya regarding UX. Your implementation seems Iike an improvement
to me. I'm trying to merge but I'm using a VPN to access Github because of
this Dyn DDOS issue and its super slow.

PR #839, opened 2016-10-18
Stop key event propagation.

jasongrout:
In JupyterLab, key events were bubbling up from inputs to be considered as shortcut keys in the application. This ensures that key presses in inputs stay in inputs.

As an example, this would create a new cell in a notebook in JupyterLab:

```
import ipywidgets as w
w.Text()
```
and then pressing `a`.


jasongrout:
CC @SylvainCorlay for review.

PR #840, opened 2016-10-19
interactive: rename constructor -> factory

jdemeyer:
Bikeshedding my own code... "factory" is a better and more standard name for the thing which I called "constructor" before.


jasongrout:
Looks good to me. Thanks!

PR #841, opened 2016-10-19
Remove the widgetsnbextension jQuery dependency.

jasongrout:
This was causing problems since the jQuery versions were mismatched between jupyter-js-widgets and widgetsnbextension.


jasongrout:
Tested using the Widgets | Embed menu item.

PR #843, opened 2016-10-19
Clean up examples

jasongrout:
This includes #837.

* remove bootstrap
* fix an error with button icons
* convert babel to typescript
* clean out other unused dependencies
* add explanatory text to first example
* get rid of global jQuery in the embed-webpack.ts module


gnestor:
I checked this out:

* The accordion widget in Widget List.ipynb looks good
* I tested the examples in jupyter-js-widgets and all looks good except I'm running into a CORS exception in web3:

```
XMLHttpRequest cannot load http://localhost:8888/api/kernelspecs?1477062144700. 
No 'Access-Control-Allow-Origin' header is present on the requested resource. 
Origin 'null' is therefore not allowed access. 
The response had HTTP status code 404.
```

Separate issue: I noticed this on master, but does the tab widget look like this for you?

image

jasongrout:
Tabs are a separate issue - we need to include the phosphor css, I think.

For web3, did you follow the instructions in the README in that directory? There's a flag there to get rid of the warning.

Let's merge the Accordion PR first, before this one. I can then rebase this one, if it's cleaner to do so.

gnestor:
@jasongrout Merged https://github.com/ipython/ipywidgets/pull/848... You think this is ready, too?

jasongrout:
Yep, I think it's ready.

PR #847, opened 2016-10-21
Add no wrapping and overflow indication to descriptions and buttons

jasongrout:
This enforces the design constraint that widget descriptions and buttons have a single line of text.


ssunkara1:
+1 Looks good @jasongrout

PR #848, opened 2016-10-21
Get tabs working in the classic notebook.

jasongrout:
CC @gnestor.


gnestor:
Much better! Merging...

image

PR #849, opened 2016-10-21
In JLab, handle the case where a widget model is not found.

jasongrout:
This comes up when opening a notebook where widgets were already displayed and saved. Really, we should make the rendering system smarter, so that we can give up, and the rendering system will continue to try to find a new renderer.


jasongrout:
See https://github.com/jupyterlab/jupyterlab/issues/1139 for the rendermime issue.

PR #860, opened 2016-10-29
Use the new css in the case of the legacy notebook

SylvainCorlay:
- So that we have only one flavor of the css to maintain.
- Variables from lab are defined in a separate css file only loaded in the case of the legacy notebook.


SylvainCorlay:
- The legacy notebook loads some jquery-ui theme which has some effect on the styling of the widgets, which still look ok.
- The tabs look a bit different in legacy notebook. Looking into it.
- Need to use myth.

jasongrout:
> Need to use myth.

I just pushed a few commits to your branch that use postcss in webpack. See what you think.

jasongrout:
I pushed some more changes. Here's what I'm thinking: Two files are copied to `/lib`, and eventually will become the two entry points to the css. `widgets.css` will be a compiled version offering a default styling that does not require any fancy CSS features, while `widgets-base.css` will assume that various variables are defined and allows for customization. The vast majority of people will probably use `widgets.css`. Lab will use `widgets-base.css`, of course.

jasongrout:
One thing I noticed in the widget list: the vertical progress bar is not vertical and not the same color as the horizontal one is:
screen shot 2016-10-30 at 1 07 01 am

jasongrout:
Oh, never mind about the bar. I didn't see the `bar_style='info'` part.

jupyter-js-widgets/css/widgets-base.css
```diff
@@ -350,11 +361,6 @@
 /* Widget Slider */
 
 .widget-slider {
-    /* Local Variables for the slider */
-    --jp-widgets-track-thickness: 6px;
-    --jp-widgets-handle-thickness: 12px;
-    --jp-widgets-handle-size: var(--jp-widgets-inline-height);
-    --jp-widgets-slider-border-width: var(--jp-widgets-border-width);
```
SylvainCorlay: Why moving those to root?

jasongrout:
See the commit message

jupyter-js-widgets/css/widgets-base.css
jasongrout: See the merge commit. Basically, the preprocessors need them in root because the preprocessors don't know about the DOM structure, so they can't correctly scope them.

SylvainCorlay: gotcha

SylvainCorlay:
I am on appear.in

jupyter-js-widgets/examples/web3/src/widgets.css
```diff
@@ -0,0 +1,10 @@
+/* Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
+/* We import all of these together in a single css file because the Webpack
+loader sees only one file at a time. This allows postcss to see the variable
+definitions when they are used. */
+
+@import "../node_modules/jupyter-js-widgets/css/labvariables.css";
```
SylvainCorlay: This should work with `jupyter-js-widgets/css/labvariables.css`

jupyter-js-widgets/examples/web/index.js
```diff
@@ -2,9 +2,7 @@
 // Create a widget manager instance.
 var WidgetManager = require('./manager').WidgetManager;
 var uuid = require("jupyter-js-widgets").uuid;
-require('jquery-ui/themes/base/all.css');
-require('jupyter-js-widgets/css/widgets.min.css');
-
+require('jupyter-js-widgets/lib/widgets.css');
```
SylvainCorlay: is `lib` part of what is distributed in the `jupyter-js-widgets` npm package.

SylvainCorlay: If not, should widgets.css be also in dist?

jasongrout: That's another thing - I think right now a lot of stuff is distributed with the npm package, and we can trim that down to the compiled files. I think we should do that in a separate PR.

jupyter-js-widgets/examples/web3/src/widgets.css
jasongrout: Done.

PR #862, opened 2016-10-31
Slim down the npm package.

jasongrout:
Also, put all of the css in the /css directory.


jasongrout:
I'm open to a better name than `widgets.built.css`...

minrk:
makes sense as a name to me

PR #864, opened 2016-10-31
Move embed files to their own directory.

jasongrout:
Continuing #862, this moves the embed files to their own directory. Since the embedding is more properly implemented as a separate package (like the widgetsnbextension or jupyterlab_widgets), we'd rather not have the files in with the source files. However, it's a small amount of code and makes the jupyter-js-widgets library itself useful on webpages, so I think it's okay to keep it in the package.


SylvainCorlay:
lgtm.

PR #865, opened 2016-10-31
Fix css

jasongrout:
Fixes some remaining issues with moving the CSS files.


jasongrout:
Ping @SylvainCorlay for review...

PR #870, opened 2016-11-02
Add datafiles for jupyterlab_widgets wheel

SylvainCorlay:


jasongrout:
This means that it's no longer necessary to do the labextension install step, at least if you want to install to sys-prefix, right?

SylvainCorlay:
That is right. Although it remains necessary for dev install.

minrk:
Nice!

minrk:
Nice! This can put things in the wrong place for `--user` installs, but the consequences aren't dire.

SylvainCorlay:
`--user`  is   👿

PR #878, opened 2016-11-04
Make version warning more explicit.

jasongrout:


minrk:
Can this be done on the Python side as well? That's the warning users typically see.

jasongrout:
Better?

minrk:
Thanks!

PR #879, opened 2016-11-04
Audit of Selection widgets

### LINKED ISSUE #876: Select broken in 6.0.0b2
Not sure if you guys are already aware of bugginess in Select in the beta releases, but I tried upgrading and noticed Selects weren't working so thought I'd submit this.  Setting `value` on a Select seems to be broken in 6.0.0b1 and b2.  

    import ipywidgets
    select = ipywidgets.Select(
         options={'ok': 'ok', 'yes': 'yes', 'maybe': 'maybe', 'whatevs': 'whatevs'},
         width='100px',
         height='20px'
    )
    select.value = 'ok'
    display(select)

The select shows up as blank (`None` selected) regardless of what I do programmatically.  Also, observing `value` does not call the given function when manually changing the selected option.

jasongrout:
* Fixed a few places where the .equals attribute was not used
* Clarified documentation.

Fixes #876.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -6,6 +6,10 @@ Represents an enumeration using a widget.
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 from collections import Mapping
+try:
+    from itertools import izip
+except ImportError:  #python3.x
+    izip = zip
 
 from .domwidget import DOMWidget
 from .valuewidget import ValueWidget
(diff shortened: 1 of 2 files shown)


jasongrout:
CC @SylvainCorlay and @justinmc for review.

SylvainCorlay:
There was some subtlety about non-serializable values for selection widgets which was the reason why the model value synced with the JS was the key...

I will be back in 30min and review.

SylvainCorlay:
OK, I reviewed and this looks good to me. The crux of the serialization thing is that the custom [serializer / deserializer] ar `_value_to_label`  and `_label_to_value`.

PR #888, opened 2016-11-11
Factor out labels on widgets

jasongrout:
This continues the work to factor out labeling handling from the various widgets to make sure that it is done consistently.


jasongrout:
@SylvainCorlay - I think this is ready for review.

PR #891, opened 2016-11-11
[WIP] Refactor @interact widget creation to make default values consistent.

### LINKED ISSUE #226: @interact + select does not respect default value from function signature.
Not sure if it's a bug, but seem users expect the following ...

``` python
from ipywidgets.widgets import Select

@interact(v=Select(options=['a','b']))
def function(v='b'):
    print(v)
```

To print `b`, it prints `a`:

screen shot 2015-11-04 at 15 36 42

Setting value in the options explicitely works. 

Real example is of course more complex...

jasongrout:
Fixes #226.

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -249,38 +249,30 @@ class interactive(Box):
 
     def widget_from_abbrev(self, abbrev, default=empty):
         """Build a ValueWidget instance given an abbreviation or Widget."""
+        widget = None
         if isinstance(abbrev, ValueWidget) or isinstance(abbrev, fixed):
-            return abbrev
+            widget = abbrev
 
-        if isinstance(abbrev, tuple):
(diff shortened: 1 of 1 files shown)


jasongrout:
A design question brought up by a test failure: what should the default be in this case, where an int slider is created?

```python
@interact(x=5)
def f(x=1):
    print(x)
```

Before, `x=5` was the default, but after this PR, `x=1` is the default. Which should it be?

CC @Carreau, @SylvainCorlay

jasongrout:
Same design question for this code, which also caused a test failure:

```python
@interact(x="hi")
def f(x="default"):
    print(x)
```
What should the default be?

jasongrout:
Hold off on merging this - I posted an example to #226 that makes me think we should reject the PR.

jasongrout:
Closing since we decided not to do #226.

Issue #895, opened 2016-11-12
Spec


SylvainCorlay:
Two important areas where widgets would benefit from a formal specification

1. A json schema for the static widget serialized manager.

   This should be the simplest part since we effectively use this to power http://jupyter.org/widgets.html

   - It is likely to be complemented with a semver requirement for model modules.
   - At the moment, only the front-end can generate that json, and we should also have the backend be able to do it.

2. The message spec for display / update / custom messages and asynchronous state requests.

   This will also require a detailed description of the custom serialization mechanism.

SylvainCorlay:
Ping @rgbkrk

SylvainCorlay:
I would love to have feedback on the form the spec should have

1. First item

 - documentation + jsonschema for the format.

2. Second item

 - is there a standard or a good model I should use for inspiration for the spec?

rgbkrk:
Excellent to see this!

> is there a standard or a good model I should use for inspiration for the spec?

On the slightly more informal route there's [the jupyter message spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html#general-message-format). I personally think that the widgets should carve out a space on that spec that is more tightly defined than just the comms. It can use the comms -- the missing piece is what those messages are and how they get invoked (who does the `comm_open`, what it means, etc.)

If more formal, there are RFCs and W3C docs. It doesn't have to be that though. This can be refined and updated in the same way that the jupyter message protocol and notebook format get updated. They do both have JSON Schema as well.

The original spec for the widgets was in [IPEP 23](https://github.com/ipython/ipython/wiki/IPEP-23:-Backbone.js-Widgets). The current implementation of the widgets certainly surpasses that and should wholly replace it. The primary goal is to have something that can be versioned and worked with over time.

SylvainCorlay:
Notes for self (on the plane to Frankfurt)

## Widget Manager State Schema

used e.g. in https://github.com/jupyter/jupyter.github.io/blob/master/assets/pythreejs-example.json

```js
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "description": "Jupyter Interactive Widget State JSON schema.",
    "type": "object",
    "additionalProperties": true,
    "additionalProperties" : {
        "type": "object",
        "properties": {
            "model_name": { "type": "string" },
            "model_module": { "type": "string" },  // AMD loadable module for model implementation
            "state": {
                "type": "object",
                 "additional_properties": true
            },
            "views": {
                "type": "array"
            }
[...]

PR #896, opened 2016-11-13
get_manager_state static method

SylvainCorlay:
This adds a static method to `Widget` to serialize the widget manager from the python side instead of the JavaScript side.

This will be required by the sphinx extension.

This is also related to #895


ipywidgets/widgets/widget.py
```diff
@@ -286,13 +295,14 @@ def get_state(self, key=None):
         else:
             raise ValueError("key must be a string, an iterable of keys, or None")
         state = {}
-        traits = self.traits() if not PY3 else {} # no need to construct traits on PY3
```
SylvainCorlay: here

PR #899, opened 2016-11-14
Add semver string to model state

SylvainCorlay:
OK, this seems to be doing the job.

Tested with jupyterlab, legacy notebook and embedded widgets woot!

### DIFF
--- a/jupyterlab_widgets/src/index.ts
+++ b/jupyterlab_widgets/src/index.ts
@@ -185,10 +185,10 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   /**
    * Load a class and return a promise to the loaded object.
    */
-  protected loadClass(className: string, moduleName: string, error: any): any {
-    let mod: any = this._registry.get(moduleName, '*');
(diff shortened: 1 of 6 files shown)


jupyterlab_widgets/src/index.ts
```diff
@@ -185,8 +185,8 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   /**
    * Load a class and return a promise to the loaded object.
    */
-  protected loadClass(className: string, moduleName: string, error: any): any {
-    let mod: any = this._registry.get(moduleName, '*');
+  protected loadClass(className: string, moduleName: string, moduleVersion: string, error: any): any {
+    let mod: any = this._registry.get(moduleName, moduleVersion);
     if (!mod) {
       return Promise.reject(`Module ${moduleName} not registered as a widget module`);
```
jasongrout: Can you change the error message to something like:

```
`Module ${moduleName}, semver range ${moduleVersion} is not registered as a widget module`
```

SylvainCorlay: Done.

Issue #904, opened 2016-11-15
Fix alignment of text in widgets


jasongrout:
Some of the widgets have slightly vertically offset text. I checked this using:

```python
from ipywidgets import *

d = 'description: '
v = 'value'
o = [v, 'value2']
VBox([
HBox([
  BoundedFloatText(description='BoundedFloatText: '),
  BoundedIntText(description='BoundedIntText: '),
  Button(description='Button'),
  Checkbox(description='Checkbox: '),
  ColorPicker(description='ColorPicker: '),
]), HBox([
  Dropdown(options=o, description='Dropdown: '),
  FloatProgress(description='FloatProgress: '),
  FloatRangeSlider(description='FloatRangeSlider: '),
  FloatSlider(description='FloatSlider: '),
]), HBox([
  FloatText(description='FloatText: '),
  HTML(value='HTML'),
  IntProgress(description='IntProgress: '),
  IntRangeSlider(description='IntRangeSlider: '),
  IntSlider(description='IntSlider: '),
]), HBox([
  IntText(description='IntText'),
  Label(value='Label'),
  RadioButtons(description='RadioButtons: ', options=o),
  Select(description='Select: ', options=o),
  SelectMultiple(description='SelectMultiple: ', options=o),
]), HBox([
  SelectionSlider(description='SelectionSlider: ', options=o),
  Text(description='Text: ', value=v),
  Textarea(description='Textarea: ', value='\n'.join(o)),
  ToggleButton(description='ToggleButton'),
]), HBox([
  ToggleButtons(description='TogButtons: ', options=o),
  Valid(description='Vaild: '),
])])
```
and then using a horizontal guideline to check vertical alignment of text across all of the widgets.

Here it is on master with Jupyter notebook:

<img width="1409" alt="screen shot 2016-11-16 at 11 36 31 am" src="https://cloud.githubusercontent.com/assets/192614/20356262/ffc3cf1e-abf1-11e6-9124-74a1f1ac9712.png">

and on JupyterLab:

<img width="1423" alt="screen shot 2016-11-16 at 11 47 31 am" src="https://cloud.githubusercontent.com/assets/192614/20356453/8a735f80-abf2-11e6-8d05-9761c61d9a51.png">

jasongrout:
Here is what it looks like after a few tweaks in JLab. I added a horizontal guide one pixel below the baseline for the label.

[...]

PR #908, opened 2016-11-16
Fix alignment for some widgets and labels by judicious additions of line-heights

jasongrout:
See #904 for before. Here it is after on Jupyter Notebook:

screen shot 2016-11-16 at 11 16 24 am

and on JupyterLab:

screen shot 2016-11-16 at 11 11 35 am


jasongrout:
CC @cameronoelsen, @ellisonbg

jasongrout:
In particular, the HTML and Label controls are better vertically aligned (though HTML still seems to be a pixel off), and the labels for SelectMultiple and Textarea are aligned to the other control labels.

PR #919, opened 2016-11-25
Make HBox and VBox use classes instead of explicitly setting the styles.

jasongrout:
We run into problems when we set the element style of phosphor widgets - phosphor hides widgets by adding a css class, so if we set the display style of an element explicitly, it overrides phosphor hiding the element. This means that the element behaves badly in tabs, etc.

### DIFF
--- a/ipywidgets/widgets/widget_box.py
+++ b/ipywidgets/widgets/widget_box.py
@@ -53,19 +53,15 @@ class Box(DOMWidget):
 @register('Jupyter.VBox')
 class VBox(Box):
     """Displays multiple widgets vertically using the flexible box model."""
-
-    @default('layout')
-    def _default_layout(self):
-        return Layout(display='flex', flex_flow='column')
+    _model_name = Unicode('VBoxModel').tag(sync=True)
+    _view_name = Unicode('VBoxView').tag(sync=True)
 
 
 @register('Jupyter.HBox')
 class HBox(Box):
     """Displays multiple widgets horizontally using the flexible box model."""
-
-    @default('layout')
-    def _default_layout(self):
-        return Layout(display='flex')
+    _model_name = Unicode('HBoxModel').tag(sync=True)
+    _view_name = Unicode('HBoxView').tag(sync=True)
 
 
 @register('Jupyter.Proxy')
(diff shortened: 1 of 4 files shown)


jasongrout:
Another more far-reaching change is to disallow setting display in the Layout widget. Thoughts, @SylvainCorlay? It seems that people mainly use the Layout.display attribute to hide or show a widget - perhaps we should make a new widget that does that (with css), and deprecate the Layout.display attribute? You'd use this new widget like:

```python
h = Hideable(my_widget)
h.visible = True
h.visible = False
```

jasongrout:
Or we could make Layout special-case the display attribute and use phosphor to hide the widget. That way it would work for every widget without having to wrap it.

SylvainCorlay:
:+1: on all the changes of the PR as of now.

Unsure about the layout attribute question since layout systematically exposes all css properties that impact layout...

jasongrout:
I would say that it's not a usual usecase to need the full power of the display attribute - usually you just want to set it to none or the default, where the default should be determined by the view on the javascript side and/or css. Accordingly, I think it would be fine to treat it special - display is true or false. 

In fact, I don't think it would be terrible to combine both display and visible to a tri-state attribute for display, hide, or collapse - that's often a point of confusion, the difference between display and visible.

jasongrout:
Essentially, the problem with the layout is that it takes precedence over css styling, and that can break things like phosphor hiding a widget.

SylvainCorlay:
In the examples such as flexbox based things, we also use `display: flex;`

SylvainCorlay:
Happy to merge this one and carry on the discussion on `display`.

SylvainCorlay:
We could make the `display` property an enum disallowing `hidden`. However, when the value differs from the default, it would prevent phosphor from hiding it...

jasongrout:
> We could make the display property an enum disallowing hidden. However, when the value differs from the default, it would prevent phosphor from hiding it...

Yep, there's no way to win with exposing display, I think.

jasongrout:
> In the examples such as flexbox based things, we also use display: flex;

The box widget now is basically a blank box with display: flex...

ssunkara1:
@SylvainCorlay @jasongrout What was decided on this? So what is the right way to make a widget hidden?

jasongrout:
Set `mywidget.layout.display = 'none'`, and to unhide `mywidget.layout.display=None` should work.

jasongrout:
Or `mywidget.layout.display=''` should work too, and maybe be less confusing than `'none'` vs `None`.

PR #922, opened 2016-11-30
Use relative paths, otherwise the examples/* stuff gives a runtime error

infinity0:


jasongrout:
Thanks @infinity0, and sorry that this seems to have languished.

ping @SylvainCorlay - can you take a look at this?

SylvainCorlay:
Hum, I don't see a runtime error on either ReadTheDocs or locally.

infinity0:
I can't remember the exact context, but it was needed for [Debian packaging](https://packages.debian.org/sid/all/python-ipywidgets-doc/filelist). Might only be for 5.2.2. examples/ is stored under the same directory as _static so a "_static" relative paths means examples/_static which is wrong.

infinity0:
Also this patch makes it more robust to path changes you make in the git tree.

SylvainCorlay:
This is not needed for 6.0 anymore since we don't have static assets in the documentation. Closing.

Issue #910, opened 2016-11-16
Provide a way to easily create and layout widgets for something like interact


jasongrout:
From @gvwr at https://github.com/ipython/ipywidgets/issues/715#issuecomment-260986825:

Further to the previous comments, I've tried and failed to layout widgets to my satisfaction.
I'd really appreciate improved documentation on the usage of the layout widgets - in particular the usage of `Box`, `HBox` & `VBox` with `interact`, `@interact` & `interactive`.

The issues I've encountered are:

1. Layout of multiple widgets in a two column layout - I've only managed to achieve this with a kludgy attempt using `display` with the children of `interactive`, and I can't believe this is the best/correct way.
2. `interact` printing the function name between my widgets and plots - not pretty
3. My kludgy implementation of `display` with `interactive` doesn't display my plot when the cell is first run, only when a widget is used.

I can supply my test notebook if required.

jasongrout:
@gwvr - it would be great to see your test notebook, if you can link it here.

As it stands now, Box, HBox, and VBox aren't meant to be used with interact or interactive. If you want the system to lay out the widgets very simply and automatically, use interact or interactive. If you want control over how the widgets are layed out, use the HBox and VBox, creating your own widgets. Unfortunately, that leaves a pretty wide gap in the middle between writing one line to automatically having a layout, and dozens of lines of code to construct every widget you need and hook them all up to a function. It sounds like we need to have some middle ground - perhaps make it easy to construct a layout of widgets (explicitly constructing the widgets, or perhaps exposing the function that automatically creates widgets), and hook all of the widgets up to a single function.

gwvr:
A middle ground would be great. It's hooking multiple widgets up to the function(s) manually that scares me.

My example notebook with a plotting function that I've been using to get to grips with ipywidgets.

My kludgey method for laying out the widgets looks like this:
```python3
w = interactive(makeplot,
                title=widgets.Text(value='Hello World', placeholder='Type something', description='Title:', disabled=False),
                amplitude=widgets.FloatSlider(value=1, min=-5, max=5, description='Amplitude:'),
                step_size=widgets.FloatSlider(value=0.1, min=0.01, max=0.1, step=0.01, description='Step size:'),
                periods=widgets.FloatSlider(value=5, min=1, max=20, description='Periods:'),
                noise_scale=widgets.FloatSlider(value=0.1, min=0.01, max=2, description='Noise:'),
[...]

PR #930, opened 2016-12-02
Add interact_hookup function

gwvr:
Allows separation of widget definition and UI, as per https://github.com/ipython/ipywidgets/issues/910

Feedback on the following would be good:
a) function name
b) docstring
c) have I put the function in the right place?

Thanks


jasongrout:
Thanks! How about `interactive_output` for a name? I still don't quite like that name, but it's better than `hookup`. Also, you're right that we need a docstring. And I think the function is in a great place. Thanks again!

gwvr:
Superseded by https://github.com/ipython/ipywidgets/pull/977

PR #936, opened 2016-12-02
jupyter-widgets-schema

SylvainCorlay:
as per discussion with @jasongrout


jupyter-widgets-schema/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "jupyter-widget-schema",
+  "name": "jupyter-widgets-schema",
   "version": "0.0.1",
```
jasongrout: Please use 0.1.0 - we've been burned with 0.0.x versions before.

jupyter-js-widgets/package.json
```diff
@@ -82,7 +82,7 @@
   },
   "dependencies": {
     "@jupyterlab/services": "^0.34.2",
-    "jupyter-widget-schema": "file:../jupyter-widget-schema/",
+    "jupyter-widgets-schema": "file:../jupyter-widgets-schema/",
```
jasongrout: I'll change this to the released version when I release the beta.

PR #948, opened 2016-12-06
Serialize widget state to metadata on save of the notebook

SylvainCorlay:
This will be required to

 - render widgets in sphinx documentation with nbsphinx.
 - render widgets in nbviewer.


dmadeka:
So currently widgets don't render in nbviewer? Is there a timeline on that?

SylvainCorlay:
> So currently widgets don't render in nbviewer? Is there a timeline on that?

6.0!

PR #794, opened 2016-09-28
WIP Resize of change of layout attribute

SylvainCorlay:
@jasongrout @sccolbert comments?

Now I understand what is going on: 
- The box widget should not add children to the DOM manually but call attach.
- since onAfterAttach triggers jupyter-js-widgets'  `display` event, we should make it not trigger the displayed event.

PR #949, opened 2016-12-07
Send resize message when the layout changes

jasongrout:
Also, change LayoutView to use the already-available parent option.

Fixes https://github.com/bloomberg/bqplot/pull/371#issuecomment-265224920

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -791,6 +791,7 @@ class DOMWidgetView extends WidgetView {
                             // effect first.
                             postMessage(this.pWidget, ResizeMessage.UnknownSize);
                         });
+                        postMessage(this.pWidget, ResizeMessage.UnknownSize);
                         return view;
                     });
                 }).catch(utils.reject('Could not add LayoutView to DOMWidgetView', true));
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -16,8 +16,13 @@ import {
 } from 'phosphor/lib/ui/widget';
 
 import {
-    Message
+    postMessage, Message
 } from 'phosphor/lib/core/messaging';
+
+import {
+    ResizeMessage
+} from 'phosphor/lib/ui/widget';
+
 
 /**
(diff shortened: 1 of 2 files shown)


jasongrout:
@ssunkara1 - can you verify that this fixes the bqplot issue?

ssunkara1:
@jasongrout Sure will check this. Thanks

jasongrout:
Also fixes #945

SylvainCorlay:
Could we also remove the triggering of `displayed` for child widgets by the box widgets?

The phosphor machinery should propagate the attach events and DOMWidgets now triggers its own `displayed` event when attached.

jasongrout:
@SylvainCorlay, The box widget already uses phosphor machinery: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_box.ts#L299

The proxy widget needs to be fixed to use phosphor machinery, but let's make that a different PR.

jasongrout:
> The proxy widget needs to be fixed to use phosphor machinery, but let's make that a different PR.

See #978 and #979.

ssunkara1:
@jasongrout The change looks good for bqplot. The `resize` handlers are triggered once the `Layout` is applied. Thanks!

SylvainCorlay:
I am inclined to merge. 

Although in the original #794, we were listening to the layout changes from the widget model (instead of the layout view), which allowed to stop listening to the old layout when a new value was assigned.

I am unsure that this implementation prevents a zombie layout view to trigger resize events when its model changes.

SylvainCorlay:
^ @jasongrout

jasongrout:
> Although in the original #794, we were listening to the layout changes from the widget model (instead of the layout view), which allowed to stop listening to the old layout when a new value was assigned.

That's much better, and means we don't have this weird logic to pass the parent into the child. I'll change this.

jasongrout:
CC @SylvainCorlay, @ssunkara1

jasongrout:
There are a lot of extra commits in here. Let me clean them out.

jasongrout:
Okay, I think it's cleaned up now, and ready to test.

PR #950, opened 2016-12-07
Better error message when options is not specified

### LINKED ISSUE #903: Dropdown throws confusing error when options is not specified
This apparently used to work (it's in the alignment example notebook):

```python
from ipywidgets import Dropdown
Dropdown(description='Foobar')
```

It doesn't now, but the error is very confusing. We should make a validator if we want to insist that the options list is nonempty.

ivanov:
closes #903

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -27,9 +27,11 @@ def _value_to_label(value, obj):
     # TODO: make a separate validation function so this doesn't have
     # to redo the work of parsing the options object.
     options = obj._make_options(obj.options)
+    if len(options) == 0:
+        raise TraitError("`options` must not be empty.")
     try:
-        # use an iterator approach as a shortcut
-        return next(k for (k, v) in options if obj.equals(v, value))
+        # return the first label whose value is equal to the desired value
+        return next(l for (l, v) in options if obj.equals(v, value))
     except StopIteration:
         raise KeyError(value)
 


ipywidgets/widgets/widget_selection.py
```diff
@@ -15,21 +15,23 @@
 from .valuewidget import ValueWidget
 from .widget import register
 from traitlets import (Unicode, Bool, Any, Dict, TraitError, CaselessStrEnum,
-                       Tuple, List, Union, observe, validate)
+                       Tuple, List, Union, observe, default, validate)
```
jasongrout: We don't need this import, right?

ivanov: good catch - fat fingered my `git add -p`. Fixed now

ipywidgets/widgets/widget_selection.py
```diff
@@ -27,9 +27,11 @@ def _value_to_label(value, obj):
     # TODO: make a separate validation function so this doesn't have
     # to redo the work of parsing the options object.
     options = obj._make_options(obj.options)
+    if len(options) == 0:
+        raise TraitError("`options` must not be empty.")
     try:
-        # use an iterator approach as a shortcut
-        return next(k for (k, v) in options if obj.equals(v, value))
+        # return the first label whose value is equal the desired value
```
jasongrout: Sorry, one more thing: "equal *to* the desired value"

ivanov: Not at all, thank you! Fixed the typo, now.

jasongrout:
Thanks!

PR #956, opened 2016-12-08
Make widget state saving an opt-in

### LINKED ISSUE #638: reconsider saving widget state to notebook metadata by default
It seems quite rare that restoring widget state makes sense as often as opening a notebook does. In general, it only makes sense if the kernel has not been restarted (i.e. across page refresh, but nothing else). As such, I find it quite frustrating to work with widgets now that they save a bunch of data to my notebook documents that don't make sense for anyone but me (or even me a day later). It's doubly troubling that the state apparently cannot be cleared (#637).

I believe saving widget state to notebook metadata makes sense sometimes, but that seems to be the exception rather than the rule, so I don't think the current default behavior of saving widget state in notebook metadata is desirable.

SylvainCorlay:
Close #638.


jbednar:
Excellent!  Thanks.

PR #957, opened 2016-12-08
jupyterlab@^0.11.0

gnestor:
labextension breaks when running jupyterlab@^0.11

### DIFF
--- a/jupyterlab_widgets/package.json
+++ b/jupyterlab_widgets/package.json
@@ -6,7 +6,7 @@
   "typings": "lib/index.d.ts",
   "dependencies": {
     "jupyter-js-widgets": "^2.0.12",
-    "jupyterlab": "^0.10.0",
+    "jupyterlab": "^0.11.0",
     "phosphor": "^0.7.0"
   },
   "devDependencies": {


jasongrout:
Should we change it to `~0.10.0` then?

gnestor:
From my undestanding, `~0.10.0 := >=0.10.0 <0.11.0` whereas `~0 := >=0.0.0 <1.0.0`

https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1

jasongrout:
Aren't we targeting `>=0.10.0 <0.11.0`?

gnestor:
The purpose of this commit is to make interop between ipywidgets and jupyterlab more flexible by allowing higher minor versions, because currently I'm running jupyterlab 0.11.2 and ipywidgets depends on ^0.10.0.

blink1073:
The two are not rectifiable, because 0.10 -> 0.11 is considered a major version release.  The solution is for ipywidgets to bump to 0.11, especially because 0.10 has known issues with state restoration.

gnestor:
Ok, I see. I'll update the commit...

jasongrout:
> The two are not rectifiable, because 0.10 -> 0.11 is considered a major version release. The solution is for ipywidgets to bump to 0.11, especially because 0.10 has known issues with state restoration.

Agreed. Grant, can you make sure it works in 0.11.0? I haven't had time to check things yet.

gnestor:
I tested and Widget List.ipynb is rendering with no issues.

ivanov:
I'm going to go for and merge this. After all, "laugh in the face of danger" is the official jupyter motto, is it not? 

Happy hacking! :bowtie:

PR #979, opened 2016-12-12
Remove the Proxy and PlaceProxy widgets

### LINKED ISSUE #978: Update proxy widget to use Phosphor to attach children widgets
The proxy widget currently doesn't use the phosphor mechanism, so phosphor messages stop at the proxy widget boundary.

jasongrout:
Fixes #978 (so the fix is in the history), and then removes the widgets. We don't feel there is a compelling usecase for Proxy (using a Box with a single child has much the same effect), and the PlaceProxy widget gets very tricky, especially when there are multiple copies of a notebook on a single page, as in JupyterLab.


jasongrout:
@SylvainCorlay - do we actually use the proxy widget at all anymore? We don't use it internally except for the PlaceProxy widget, which we both think should be deleted. I'm inclined to delete the Proxy widget as well.

SylvainCorlay:
Besides the place proxy, we used `Proxy` a lot to apply styles to the contained widget.

jasongrout:
Yes, I think we should encourage people to use the Box with a single child in the cases where they need a placeholder with a specific layout.

jasongrout:
(@SylvainCorlay and I discussed this, and agreed to merge this PR when I fixed the import issue that was causing tests to fail)

SylvainCorlay:
Fyi. The proxy widget approach to styling allowed two views of the same widget to be styled differently when displayed in different context.

(One in a proxy, one e.g. in an output area).

This was useful for the figure builder which is a complex gui but not much otherwise.

hainm:
@jasongrout I used to use PlaceProxy to put ipywidgets' GUI to `nglview` GUI in full screen mode.

The idea is to mimics this: http://arose.github.io/ngl/ but using Python (via ipywidgets) to write the GUI and add that to the existing Javascript code.

hainm:
picture2

maartenbreddels:
I'm also using my custom fullscreen code, maybe there needs to be support for fullscreen somewhere.

SylvainCorlay:
> I'm also using my custom fullscreen code, maybe there needs to be support for fullscreen somewhere.

That would be context dependent (jupyterlab vs jupyter) so there might not be a clean way to have this in jupyter-js-widgets.

maartenbreddels:
Why not, I mean, fullscreen is fullscreen, it shouldn't care about the rest.

SylvainCorlay:
Right, I misread and thought this was full browser window.

jasongrout:
Very nice, @hainm! We have a much more flexible system for rendering widgets now that widgets are normal displayed outputs - let's see if there is a way to use that to get what you want.

Is that full screen UI running in the the notebook, or in a custom webpage?

hainm:
> Is that full screen UI running in the the notebook, or in a custom webpage?

it's running in the notebook. 

```python
import nglview
nglview.demo()
```

--> Then you can double-click to the widget to go to fullscreen mode.

My idea is that to build GUI controls by `ipywidgets`, then insert it to fullscreen mode.

PR #980, opened 2016-12-12
Use phosphor attach method instead of native DOM's appendChild

SylvainCorlay:
So that the phosphor-based-widgets properly render in embedded contexts.

### DIFF
--- a/jupyter-js-widgets/examples/web3/src/manager.ts
+++ b/jupyter-js-widgets/examples/web3/src/manager.ts
@@ -1,5 +1,8 @@
 import * as widgets from 'jupyter-js-widgets';
 import './widgets.css';
+import 'phosphor/styles/base.css';
+import * as PWidget from 'phosphor/lib/ui/widget';
+
 
 export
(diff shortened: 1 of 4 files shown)


jupyter-js-widgets/examples/web3/src/manager.ts
```diff
@@ -1,5 +1,8 @@
 import * as widgets from 'jupyter-js-widgets';
 import './widgets.css';
+import 'phosphor/styles/base.css';
+import * as PWidget from 'phosphor/lib/ui/widget';
```
jasongrout: Personally, I like `import { Widget } from 'phosphor/lib/ui/widget';` more.

SylvainCorlay: ok although `Widget` could be a jupyter-js-widget thing...

SylvainCorlay: Feel free to change if you also want to remove mathjax.

PR #983, opened 2016-12-12
Phosphor selection containers

jasongrout:
I think we should write phosphor widgets to do the heavy lifting of the javascript interaction, and then simply use them from Jupyter widgets. This is a WIP for tabs and accordion widgets.

The tab widget is almost straight from phosphor, tweaked to not use absolute positioning.


jasongrout:
@SylvainCorlay - can you take a look at this? I think it works okay for an initial take, and I'd like to get some wider feedback.

jasongrout:
CC @rmenegaux

PR #984, opened 2016-12-13
changing package.json for windows

ssunkara1:
The `;` is not valid in Windows. I also see that for other commands in here and in `jupyter-js-widgets` we `&&` where we are only executing the subsequent commands if the previous one succeeds.


SylvainCorlay:
This looks good to me, but this probably started off an old version and needs a rebase.

ssunkara1:
@SylvainCorlay Rebased this. Thanks!

PR #987, opened 2016-12-14
widget-subarea do not shrink correctly

martinRenou:
I'm developing a nbExtension, and I found out that when my elements are big enough, and window's width is thin, widget-subarea do not shrink at all, so my elements overflow from widget area. Setting min width of sub area to 0px fix this issue.

### DIFF
--- a/widgetsnbextension/css/outputarea.css
+++ b/widgetsnbextension/css/outputarea.css
@@ -29,6 +29,7 @@
 }
 
 .widget-area .widget-subarea {
+    min-width: 0px;
     padding: 0.4em 0 0.4em 0;
     display: flex;
     flex: 2;


jasongrout:
Can you post a screenshot showing the problem?

martinRenou:
You can find what I try to explain on the first point (Minimum content sizing of flex items not honored) of this page: https://github.com/philipwalton/flexbugs (content being the widget-subarea, and container the widget-area).
I'll post a screenshot with my example tomorrow morning. Thanks for your answer.

martinRenou:
Sorry... I misread the article, that's not same issue! However, I'll post a screenshot of my issue tomorrow morning.

jasongrout:
Thanks. I'm not sure how to reproduce the issue from your description, but I think a screenshot (before/after) will help crystalize exactly what the issue is.

Thanks!

martinRenou:
Well, to reproduce this issue, you need to create a nbExtension. Create a view/model object, and in the render method do something like:
```javascript
let myEl = document.createElement('div');
$(myEl).css({ display: 'flex', flex: '1 1 0', height: '100px', 'background-color': 'blue' });
this.el.appenChild(myEl);
```
Then when rendering the view, the container (widget-subarea) will fit it's content size, so far no problem. But if you reduce the width of the window, the widget-subarea will not correctly shrink. And setting min-width of widget-subarea to 0px fix this issue.
I already found similar issue in other project, and this fix did the job. Maybe I misunderstand something ?

martinRenou:
So, on my example, I've got my view that is first correctly placed on window:



Then I grow the window size:



From then, no problem. Widget-subarea correctly grows and its content adapt itself. But if I shrink the window size:



Widget-subarea doesn't shrink... And it seems like setting min-width to 0px fiw this issue.

martinRenou:
That's weird, I just tried the example with the blue background, and it works properly... But with the second example and the same parameters, that's not the same result... I don't really understand what's going on

SylvainCorlay:
> It sounds like my canvas blocks the shrinkage.

That could be it. This could also be a browser bug. Do you see the same thing with e.g. Firefox (assuming you use chrome).

martinRenou:
Hi Sylvain,

>  Do you see the same thing with e.g. Firefox

yes, it's the same with Firefox. As you can see on the screenshot, it's widget-subarea that doesn't shrink. It maybe respect its content size ? And as my canvas wait for widget-subarea to change its size (by listening to window resize, and using container.getBoundingClientRect), it doesn't wont shrink until its container shrink ?

SylvainCorlay:
In any case, I don't have an issue with setting 0px for the minimum width of the widget subarea in 6.0.

However, in 6.0, you will also have all the benefits from the new resize events that we get out of phosphorjs (the JavaScript library powering jupyterlab).

martinRenou:
Interesting! That was so bad that only the window had a resize event. Thanks for your answers.

If you are interested in that, we are developing a javascript library for 3D mesh data visualization, you can try that here: [SciviJS](https://demo.logilab.fr/SciviJS/). And we are currently integrating it to Jupyter Notebook. :smiley:

SylvainCorlay:
That looks really really cool. I would love to hear more about it if you have some time for a demo at some point if you could give us a demo.

There is a lot of interest with 3D with ipywidgets in the notebook

 - pythreejs and the massive refactoring by @abelnation
 - the autodesk molecular design toolkit
 - nglview
 - and now SciviJS

SylvainCorlay:
It could be interesting to get people who build upon ipywidgets to talk after we release 6.0, and identify pieces that could be built as common tools. (e.g. I am building sphinx extensions for galleries of visualization libraries based on widgets, eventful containers, live collaboration in jupyterlab, our plans with dashboarding etc).

I am thinking of setting up a bi-weekly video meeting (every two weeks) with core ipywidgets devs and custom widget authors...

martinRenou:
I didn't know about autodesk molecular design toolkit and nglview, that's cool!

I do not know how much it can be said that ipyscivijs will be a custom widget, of course it is, because it will extend DOMWidget. But it's based on a pure Javascript library (there's no python in SciviJS) and we don't want to synchronize everything with the python kernel, it's useless. Our idea is to synchronize a dictionary defining the tree of Blocks and its parameters. I don't know how they did with nglview and molecular design toolkit, but we choose to not use pythreejs because of its "massive" synchronization with kernel...

BTW, we didn't use ipywidgets like sliders for SciviJS, for the same reasons. Maybe you could be interested in make it possible to create ipywidgets that are not synchronized with kernel ?

jasongrout:
> Maybe you could be interested in make it possible to create ipywidgets that are not synchronized with kernel ?

That's kind of the whole point of ipywidgets - automatic synchronization with the kernel. If you are not doing synchronization, and the data flow is one-time and one-way (i.e., load data once from the kernel), it may be much easier to do this as a custom output renderer.

That said, I'm in the middle of pulling out some of the UI control logic from ipywidgets to create standalone components for tabs, accordion, etc. Those UI controls could be reused outside of ipywidgets for whatever.

jasongrout:
> it may be much easier to do this as a custom output renderer.

One nice thing about doing this as a custom mimetype renderer is that it is easily turned into a file viewer in JupyterLab, as well as working with static output like notebooks converted to web pages, etc.

jasongrout:
I don't see any problem with making the min-width 0px as well, so merging. Thanks!

jasongrout:
(but I do agree that this is more of an issue with the widget area fitting the content, and the way you are checking the size).

martinRenou:
@jasongrout So it sounds like it's not ipywidgets that we need.. However that was just like a first test of integration in Jupyter. Thanks for your answers, it really helped.
Thanks for the merge, it could help someone else!

PR #990, opened 2016-12-15
Update to jupyterlab 0.11.3

SylvainCorlay:
ping @jasongrout


hainm:
@SylvainCorlay any update for jupyterlab `0.12.1` soon? thanks.

```bash
X is not version compatible with installed JupyterLab version 0.12.1
    Expects JupyterLab version ^0.11.3 from packaged module @jupyterlab/nbwidgets@0.6.4/lib/plugin.js
    Expects JupyterLab version ^0.11.3 from packaged module @jupyterlab/nbwidgets@0.6.4/lib/output.js
```

PR #994, opened 2016-12-21
Buffered_state_diff bug fix backport

ssunkara1:
Backporting the fix from #833 tp 5.x. I think this is valuable because the issue results in a irritating bug where the traits of the `widget` are being reset back to their previous values.


ssunkara1:
@jasongrout @SylvainCorlay

ssunkara1:
@jasongrout @SylvainCorlay What do you guys think about this? I feel we should have this fix backported to `5.x` as well. Because the issue is big when the traits of a widget are being reset to previous values.

SylvainCorlay:
I am ok with this.

PR #995, opened 2016-12-22
Add web5 example with tmpnb backend

SylvainCorlay:
This example of static widgets uses the `tmpnb` backend.


nscozzaro:
I just spent time trying to get the web5 example to work, but it appears it no longer works because tmpnb.org is retired, is that correct?
If so, maybe the web5 example should be updated/replaced?

PR #1013, opened 2017-01-03
Explain the JLab package is experimental

### LINKED ISSUE #1004: Status of the lab extension in ipywidgets 6.0.0?
Since jupyterlab will probably still have a few 0.X releases after ipywidgets 6.0 is released, should we declare that the jupyterlab extension is experimental when we release 6.0?

It would be nice to coordinate with the jlab team if there is a 0.x version that they expect to stay around for some time to be released soon.

jasongrout:
Fixes #1004.

### DIFF
--- a/jupyterlab_widgets/README.md
+++ b/jupyterlab_widgets/README.md
@@ -1,7 +1,9 @@
 Jupyter Widgets JupyterLab Extension
 ====================================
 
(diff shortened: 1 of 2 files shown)


jasongrout:
CC @SylvainCorlay

Issue #946, opened 2016-12-06
Progress bar style not updating if value starts at 0


jasongrout:
With notebook ce575f7a5ab625ea044709659cf6aebe264098e0:

```python
from ipywidgets import IntProgress
from IPython.display import display
p = IntProgress(min=0, max=100, value=0)
p.bar_style = 'info'
p.value = 50
display(p)
```
doesn't display the right color in the progress bar, but this does:
```python
from ipywidgets import IntProgress
from IPython.display import display
p = IntProgress(min=0, max=100, value=50)
p.bar_style = 'info'
display(p)
```
<img width="492" alt="screen shot 2016-12-06 at 1 29 56 pm" src="https://cloud.githubusercontent.com/assets/192614/20944431/48f690c4-bbb8-11e6-9503-353a5283dddd.png">

SylvainCorlay:
Fixed in #1017

PR #1017, opened 2017-01-04
update_mapped_classes fixes

SylvainCorlay:
This is a proposal for a fix to #946 .

1. The reason for that bug is when `ProgressView.render` is run, `model.previous('bar_style')` and `model.get('bar_style')` are both equal to `info`, so that the difference performed by `update_classes` is nil, and nothing gets applied.

   So what I did is create an extra `set_bar_style` which will make the difference with an empty list instead.

   For the non-mapped classes that can be applied, this is what the base `DOMWidgetView` was doing.

2. I made `class_map` a static variable in all the widget views that use one.

3. In the case of `ButtonView`, mapped classes and top-level element classes properties both edit the same class list, which causes conflicts. One way to solve this would be to move the button element down in the DOM. @jasongrout what do you think of this?

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -866,6 +866,12 @@ class DOMWidgetView extends WidgetView {
         this.update_classes(old_classes, new_classes, el || this.el);
     }
 
+    set_mapped_classes(class_map, trait_name, el?) {
+        var key = this.model.get(trait_name);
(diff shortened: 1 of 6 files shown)


jasongrout:
> One way to solve this would be to move the button element down in the DOM. @jasongrout what do you think of this?

Sure, why not?

jasongrout:
Other than the conflict and the (minor) comment, this looks good to me.

PR #1023, opened 2017-01-05
Allow dev install without JupyterLab; clean up script

jdemeyer:
Improve the `dev-install.sh` script:

1. Make it executable

2. First, check all versions, then do `set -e` (otherwise the version checking breaks).

3. Make the installation of `jupyterlab_widgets` optional if JupyterLab is not installed.

4. Make the console output nicer.


jasongrout:
Thanks!

PR #1024, opened 2017-01-05
interactive.widget_from_abbrev should be a classmethod

jdemeyer:
All the `widget_from_...` methods can be called from the `interactive` class (without an instance) because they are static methods. The only exception is `widget_from_abbrev` (which calls the other methods). Fix this by making `widget_from_abbrev` a class method. Note that it cannot be a static method, since we need to call other methods.


SylvainCorlay:
Thanks, this makes perfect sense.

PR #1027, opened 2017-01-06
Add core widget base class providing module semver ranges

### LINKED ISSUE #1002: Provide other value than '*' for core widgets' [model/view]_module_version
This probably requires creating two base widgets for all core widgets... above `DOMWidget` and `Widget` respectively.

SylvainCorlay:
Closes #1002.

On the Python side, I use multiple inheritance to specify the version range for all core widgets without impacting the base `Widget` and `DOMWidget` classes that are used in third-party libraries.

On the JS side, I still need to make the change.

Proposed solution: use `'^' + require('../package.json').version`

@jasongrout thoughts?

### DIFF
--- a/ipywidgets/widgets/widget_core.py
+++ b/ipywidgets/widgets/widget_core.py
@@ -1,0 +1,15 @@
+"""Base widget class for widgets provided in Core"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from .widget import Widget
+from .._version import __frontend_version__
+
+from traitlets import Unicode
+
+class CoreWidget(Widget):
+
+    _model_module_version = Unicode(__frontend_version__).tag(sync=True)
+    _view_module_version = Unicode(__frontend_version__).tag(sync=True)
+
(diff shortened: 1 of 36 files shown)


ipywidgets/widgets/widget_core.py
```diff
@@ -0,0 +1,18 @@
+"""Base widget class for widgets provided in Core
+
+Represents a container that can be used to group other widgets.
```
jasongrout: Delete this line?

jasongrout:
> Proposed solution: use '^' + require('../package.json'),version

That sounds consistent to me.

jasongrout:
Maybe this?
```
`^${require('package.json').version}`
```

SylvainCorlay:
> Maybe this?
> `^${require('../package.json').version}`

This is cool :)

SylvainCorlay:
So on the js side, I have `CoreWidgetModel` and `CoreDomWidgetModel` since `default()` will not work well with multiple inheritance.

SylvainCorlay:
So if the semver req specified by the backend for a model is `^4.3.0`, and the installed/available js version is `4.4.3`, what happens when the backend opens a widget, is that the value for `model_module_version` stored in the model is `^4.3.0`.

When we save the widget state, this value is different from the js-side default (^4.4.3) and therefore is the one used in the serialization.

An idea would be to have a custom serializer for these attributes to restrict the semver range.

SylvainCorlay:
@ellisonbg @jasongrout

SylvainCorlay:
After in-person discussion with @jasongrout on this, we decided that this was still the right approach.

PR #1031, opened 2017-01-06
Add __repr__ to Widget for testing/debugging

jdemeyer:
This is useful in particular when using the Python [doctest](https://docs.python.org/2/library/doctest.html) framework to test widgets.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -123,6 +123,34 @@ class Widget(LoggingConfigurable):
     widgets = {}
     widget_types = {}
 
+    def __repr__(self):
+        """
+        Textual representation of this widget. This is meant for
+        testing and debugging.
+        """
+        # List of attributes to show with default values. The attribute
+        # is printed if it exists and its value is not equal to this
+        # default value.
+        attributes = [("value", None),
+                      ("min", None),
+                      ("max", None),
+                      ("step", None),
+                      ("options", None),
+                      ("description", ""),
+                      ("children", [])]
+        r = ""
+        for (attr, default) in attributes:
+            try:
+                v = getattr(self, attr)
+            except AttributeError:
+                continue
+            if default == v:
+                continue
+            if r:
+                r += ", "
+            r += attr + "=" + repr(v)
+        return "%s(%s)" % (type(self).__name__, r)
+
     @staticmethod
     def on_widget_constructed(callback):
         """Registers a callback to be called when a widget is constructed.


SylvainCorlay:
This is work in progress?

jdemeyer:
> This is work in progress?

No, it works. What are you missing?

SylvainCorlay:
Because you are hardcoding a bunch of attributes to display by defaults which don't make sense for most widgets.

Looping over the keys and displaying the current value might be more valuable in most cases.

jdemeyer:
> Because you are hardcoding a bunch of attributes to display by defaults which don't make sense for most widgets.

Sure, that's true. I agree that it's ugly, but it works for the use case it's meant for (debugging/testing).

> Looping over the keys and displaying the current value might be more valuable in most cases.

Problem is: that would become unreadable soon. There are a lot of keys and I chose to display only the most important ones.

jasongrout:
An idea behind repr is that it gives you code that reproduces the object (https://docs.python.org/3/library/functions.html#repr). Since you are deliberately ignoring attributes, I think it better to put it in angle brackets. Also, perhaps you can have a depth argument which controls the recursion depth, perhaps with a default value of 2 or something.

jdemeyer:
> Since you are deliberately ignoring attributes, I think it better to put it in angle brackets.

No problem. I could easily add those if you want.

> Also, perhaps you can have a depth argument which controls the recursion depth, perhaps with a default value of 2 or something.

I have no idea how you would implement that...

jasongrout:
>> Also, perhaps you can have a depth argument which controls the recursion depth, perhaps with a default value of 2 or something.
> I have no idea how you would implement that...

I was thinking a default `depth=2` attribute in the signature for `__repr__`, and then explicitly calling children's `__repr__` with `depth=depth-1`, with an immediate return if depth is 0. If it's not good to change the signature for `__repr__`, then perhaps a helper function that does the recursive bit. I'm just trying to think of a good way to compromise between readability and comprehensiveness. If you wanted the full stack, then you'd have to explicitly call `__repr__` or the helper function with a sufficient depth.

SylvainCorlay:
```python
def print_widget(widget):
    for key in widget.keys:
        if getattr(widget, key) != getattr(widget.__class__, key).default_value:
            print(key + '=' + str(getattr(widget, key)))
```

jdemeyer:
@SylvainCorlay: one thing I don't like is that the `options` keyword from `Dropdown` becomes `_options_labels`. This does not look so nice:
```
Dropdown(_options_labels=(u'1', u'2', u'3', u'4', u'5'))
```

SylvainCorlay:
Indeed, although `_options_labels` is what is being actually synchronized in case the values are not json-serializable.

infinity0:
What's the progress on this? I have to figure out how to integrate this into Debian's ipywidgets package, or else disable 68 doctests in Debian's sagemath package.

In terms of depth recursion for `repr()` I think this is not Pythonic, Python generally expects `x == eval(repr(x))` unless x contains cyclic references or non-representable things like filehandles. Otherwise, artifical depth cutoffs are better featured in `str()` or some other pretty-printer function.

infinity0:
Are the other behavioural changes being discussed compatible with the *behaviour* of the existing commit? If so, I could just backport it to Debian's ipywidgets now without having to wait for this discussion to conclude, on the assumption that an equivalent behaviour would be in ipywidgets 7.x/8.x anyways.

jasongrout:
> one thing I don't like is that the options keyword from Dropdown becomes _options_labels. This does not look so nice:

FYI, we've now revamped how options work in selection widgets.

jasongrout:
> I think this is not Pythonic, Python generally expects x == eval(repr(x)) unless x contains cyclic references or non-representable things like filehandles.

As far as Sage goes, Sage crossed that bridge a *long* time ago - a huge number of things in Sage have reprs that would not produce an equivalent object if they were evaluated.

jasongrout:
> Are the other behavioural changes being discussed compatible with the behaviour of the existing commit?

It's been a while since I thought about this. Looking at it again, I really agree with https://github.com/jupyter-widgets/ipywidgets/pull/1031#issuecomment-271276948 above - I don't think the base Widget class is the right place to have a list of default attributes. Those attributes are pretty specific to our bundled widgets, while Widget is used to build many other types of widgets in third party packages (e.g., bqplot, pythreejs, ipyvolume, etc.).

In an interactive system, printing out the entire state of the widget also is onerous.

So to move forward, we could (a) make the __repr__ depend on the widget type (i.e., the slider widget would have a nice repr with some of the values in the widget, which would be different than the values in the text widget, or the tab widget, for example).

OR for the purposes of doctesting, we could just print the result of the `get_state()` method. It wouldn't recurse into children widgets, but would be a start.

jasongrout:
Note that the `get_state` method takes several arguments that may help here. You can give it a key or list of keys to filter the state by:

```python
from ipywidgets import *
x = IntSlider()
x.get_state(key=['value', 'min', 'max', 'step'])
```
and you can also have it drop any key that is equal to the default:
```python
from ipywidgets import *
x = IntSlider(value=10)
x.get_state(drop_defaults=True)
```

vidartf:
What about such a repr string:
```python
def __repr__(self):
    class_name = self.__class__.__name__
    traits = self.traits()
    signature_entires = []
    for key in self.keys:
        value = getattr(self, key)
        if key[0] == '_' or self._compare(getattr(self, key), traits[key].default_value):
            continue
        signature_entires.append('%s=%s' % (key, value))
    signature = ', '.join(signature_entires)
    return '%s(%s)' % (class_name, signature)
```

vidartf:
Example:
```python
import ipywidgets
w1 = ipywidgets.IntSlider(min=0, max=100)
w2 = ipywidgets.IntText()
box = ipywidgets.HBox(children=[w1, w2])
repr(box)
```
gives:
```python
'HBox(children=(IntSlider(layout=Layout(), style=SliderStyle()), IntText(layout=Layout(), style=DescriptionStyle())), layout=Layout())'
```

vidartf:
Having decent repr strings is key for reproducability, which is of course a big thing among Jupyter users!

jasongrout:
> `signature_entires.append('%s=%s' % (key, repr(value)))`

More succinctly: `signature_entires.append('%s=%r' % (key, value))`

I think that's an okay approach to experiment with.

jasongrout:
> `signature_entires.append('%s=%s' % (key, repr(value)))`

More succinctly: `signature_entires.append('%s=%r' % (key, value))`

I think that's an okay approach to experiment with.

jdemeyer:
Superseded by #1398

PR #1048, opened 2017-01-18
Delete margin for HBox and VBox

jasongrout:
The margin causes a lot of extra padding as these boxes are nested deeply.

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -69,17 +69,14 @@
 .widget-box {
     box-sizing: border-box;
     display: flex;
+    margin: 0;
 }
 
 .widget-hbox {
-    box-sizing: border-box;
-    display: flex;
     flex-direction: row;
 }
 
 .widget-vbox {
-    box-sizing: border-box;
-    display: flex;
     flex-direction: column;
 }
 


SylvainCorlay:
In 5.x, these values were set so that alignment of naturally sized widgets would be preserved. Is it the case here?

jasongrout:
> In 5.x, these values were set so that alignment of naturally sized widgets would be preserved. Is it the case here?

Do you have a code sample to test this?

The problem I noticed is that, for example, if you have an HBox that has both a VBox containing a widget and another widget, the widget inside the vbox is not aligned with the widget outside the vbox, since there is an extra margin. I remember us discussing having padding on boxes and deciding to eliminate it precisely because boxes are often deeply nested for building complicated UIs, and every level of nesting adds some padding.

SylvainCorlay:
> The problem I noticed is that, for example, if you have an HBox that has both a VBox containing a widget and another widget, the widget inside the vbox is not aligned with the widget outside the vbox, since there is an extra margin. I remember us discussing having padding on boxes and deciding to eliminate it precisely because boxes are often deeply nested for building complicated UIs, and every level of nesting adds some padding.

It is not the case in 5.x. Maybe it broke since.

PR #1058, opened 2017-01-20
WIP Send box children resize notifications

### LINKED ISSUE #1054: Box should call resize on children?
The following code on the current bqplot beta shows that we have an issue with resize events. In this case, the bqplot code needs to measure its size to redraw when its container changes size. When in an HBox, the first figure is created, measures its size and draws itself. Then the second figure is inserted, which shares the space with the first widget, and so by default resizes the space for the first widget. However the first widget never gets notified that it has been resized.

Should the HBox call resize on all its widgets when there is a change to the children? Probably...

```
from ipywidgets import HBox, Layout
from bqplot import pyplot as plt
plt.figure()
plt.plot([1,5,2,3])
fig = plt.current_figure()
fig.layout = Layout(flex='1', min_width='500px', min_height='600px')
HBox([fig, fig])
```

jasongrout:
When box children are updated, after they are all inserted their sizes may have changed. This sends resize events to the children for whatever children that care.

For example, bqplot depends on resize events to measure and draw the svg graph the appropriate size.

Fixes #1054

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -557,17 +557,17 @@ class LabeledDOMWidgetModel extends DOMWidgetModel {
  *   will be called in that context.
  */
 export
-class ViewList {
-    constructor(create_view, remove_view, context) {
+class ViewList<T> {
+    constructor(create_view: (model: any, index: any) => T | Promise<T>, remove_view: (view: T) => void, context) {
         this.initialize(create_view, remove_view, context);
     }
 
-    initialize(create_view, remove_view, context) {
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Even though there might be another issue with bqplot, I still think that this is required. Surprised we weren't doing this before.

jasongrout:
note to self: open a new PR with the updated code I haven't pushed yet :).

jasongrout:
> note to self: open a new PR with the updated code I haven't pushed yet :).

For example, code making sure the resize event is sent the first time the list is populated...

SylvainCorlay:
Ah https://github.com/bloomberg/bqplot/issues/414 !

PR #1066, opened 2017-01-31
Selection container title dict should have string keys to be consistent with JSON

### LINKED ISSUE #1065: Tab widget's internal _titles dictionary has its keys changed from int type to unicode
After a state change (ie: a different tab is selected), the tab widget's internal `_titles` dictionary's keys get changed from type int to type unicode. 

This happens because `ipywidgets/widgets/widget.py:Widget.set_state` uses a de-serialized json update that has unicode keys instead of integers.

This probably applies to all the `_SelectionContainer` classes.

Any 'easy' fix would be to `index = unicode(int(index))` in the set_title and get_title methods, but I find this too hacky.

Attached is a notebook which exhibits the behavior: Out [2] shows the `_titles` dict before any tabs have been clicked, and the last output is after a tab has been clicked:
[tab_widget.zip](https://github.com/ipython/ipywidgets/files/742233/tab_widget.zip)

jasongrout:
Thanks to @mgeplf for suggesting the fix.

Fixes #1065

### DIFF
--- a/ipywidgets/widgets/widget_selectioncontainer.py
+++ b/ipywidgets/widgets/widget_selectioncontainer.py
@@ -9,6 +9,7 @@ pages.
 
 from .widget_box import Box, register
 from traitlets import Unicode, Dict, CInt
+from ipython_genutils.py3compat import unicode_type
 
 
 class _SelectionContainer(Box):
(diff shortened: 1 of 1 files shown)


mgeplf:
Looks good to me.  Thanks for the quick fix!

PR #1069, opened 2017-02-01
Create placeholder HTML widget for missing models

### LINKED ISSUE #1042: Embedder: provide a way to render all known widgets and skip unknown
Otherwise, adding a widget type is a backward incompatible change in `jupyter-js-widgets` or for a widget library.

SylvainCorlay:
Fixes #1042.

## In the notebook:

screen shot 2017-02-06 at 7 20 51 pm

## In a web context:

screen shot 2017-02-06 at 7 23 03 pm

## On Styling:

The table should probably have the new table styling of @ellisonbg but that is out of the scope of this PR IMO.

### DIFF
--- a/jupyter-js-widgets/src/manager-base.ts
+++ b/jupyter-js-widgets/src/manager-base.ts
@@ -131,10 +131,8 @@ abstract class ManagerBase<T> {
      * Takes a requirejs success handler and returns a requirejs error handler.
      * The default implementation just throws the original error.
      */
-    require_error (success_callback, version: string) {
-        return function(err) : any {
-            throw err;
-        };
+    require_error (success_callback, failure_callback, version: string) {
+        return failure_callback;
     };
 
     /**
--- a/jupyter-js-widgets/src/manager-base.ts
+++ b/jupyter-js-widgets/src/manager-base.ts
@@ -12,13 +12,17 @@ import {
 } from './widget';
 
 import {
+    HTMLModel
+} from './widget_string';
+
+import {
     shims
 } from './services-shim';
 
 
 /**
  * The options for a model.
- * 
+ *
  * #### Notes
  * Either a comm or a model_id must be provided.
  */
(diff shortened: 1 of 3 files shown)


jupyter-js-widgets/src/manager-base.ts
```diff
@@ -310,9 +314,22 @@ abstract class ManagerBase<T> {
...
+                model_id: model_id,
+                comm: options.comm,
+            }
+            var widget_model = new HTMLModel({}, modelOptions);
+            widget_model.once('comm:close', function () {
+                delete that._models[model_id];
+            });
+            let placeholder =
+                `<div>
+                  <div>ould not create model:</div>
```
blink1073: Could*

SylvainCorlay: c'mon  I pushed this 15sec ago 😄

jupyter-js-widgets/src/manager-base.ts
```diff
@@ -310,9 +314,22 @@ abstract class ManagerBase<T> {
...
+                comm: options.comm,
+            }
+            var widget_model = new HTMLModel({}, modelOptions);
+            widget_model.once('comm:close', function () {
+                delete that._models[model_id];
+            });
+            let placeholder =
+                `<div>
+                  <div>Could not create model:</div>
+                  <div> + ${error.message} +  </div>
```
jasongrout: We don't need the + signs - can we delete them?

jasongrout: Also, can we add model_name, model_module, and model_module_version in the error message to help in debugging?

jasongrout:
I haven't tested this yet, but a read-through of the code looks good to me, modulo some comments.

jupyter-js-widgets/src/manager-base.ts
SylvainCorlay: OK. pluses are leftovers.

SylvainCorlay: Will be using an html table with all the information.

SylvainCorlay:
I replaced the simple div with a more informative table.

SylvainCorlay:
The table should probably have the new table styling of @ellisonbg but that is out of the scope of this PR IMO.

Issue #812, opened 2016-10-05
Hands off approach to styling


gnestor:
ipywidgets only provides an API for adding/removing CSS classes and the Style API. The Style API provides the ability to set layout CSS traits in a mutable fashion. All other styling APIs are deprecated.

gnestor:
@SylvainCorlay Has this already been implemented?

SylvainCorlay:
Nope, not yet. We are probably going to enable something similar to the `layout` class.

PR #1070, opened 2017-02-01
Widget styling

### LINKED ISSUE #992: Add a .style attribute for control-specific styles
We don't have a way of changing the button color anymore, for example, and various other attributes of a control. This is frustrating some people trying to move to current master.

@SylvainCorlay and I suggest exploring having a `.style` attribute that contains control-specific style information. For example, a Button's style attribute would be a `ButtonStyle` traitlet that had things like color, font color, font weight, etc. We'll have to pick the attributes carefully, and we don't want to expose the DOM structure.

CC @ssunkara1, @dmadeka, @ccherukuri, @ChakriCherukuri, @rmenegaux

SylvainCorlay:
Fixes #992.

I have not yet exposed many properties but the main logic is already there.

Widgets now have a style attribute behaving in a similar way to `layout`. However, these are not directly mapped to css properties of the main `el` element of the view. Instead, in the front-end logic, we specify the css selector for the DOM subtree, css property name, and default value for the attribute.

This allows to define things like  `Slider.style.handle_color` controlling the color of the slider handle without exposing the DOM structure of the widget.

screen shot 2017-02-02 at 4 04 44 pm

### DIFF
--- a/ipywidgets/widgets/__init__.py
+++ b/ipywidgets/widgets/__init__.py
@@ -6,7 +6,7 @@ from .trait_types import Color, Datetime
 
 from .widget_core import CoreWidget
 from .widget_bool import Checkbox, ToggleButton, Valid
-from .widget_button import Button
+from .widget_button import Button, ButtonStyle
 from .widget_box import Box, Proxy, PlaceProxy, HBox, VBox
 from .widget_float import FloatText, BoundedFloatText, FloatSlider, FloatProgress, FloatRangeSlider
 from .widget_image import Image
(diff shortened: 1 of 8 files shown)


SylvainCorlay:
@jasongrout feedback welcome. I had to fight a bit with typescript so I guess there are more elegant ways to do this. Once the logic is set, I will add the attributes to expose in each widget.

gnestor:
I think this PR also closes https://github.com/ipython/ipywidgets/issues/812

jasongrout:
I think this looks good for now. Another PR can add more values. I'm not really in favor of the break between the javascript and python DOMWidgetModels (one has a style, one doesn't), but I think this is a good point to iterate from.

PR #1071, opened 2017-02-01
[WIP] Eventful Traits

rmorshea:
Migrating from [traitlets/#327](https://github.com/ipython/traitlets/pull/327) following [traitlets/#352](https://github.com/ipython/traitlets/issues/352).

cc @minrk @SylvainCorlay @jdfreder @jasongrout @ellisonbg 

# Tasks

+ [ ] Technical Review:
    + [ ] `spectate.py`
    + [ ] `eventful.py`
    + [ ] `traits.py`
+ [ ] Testing:
    + [x] Draft:
        + [x] `spectate.py`
        + [x] `eventful.py`
        + [x] `traits.py`
    + [ ] Final:
        + [ ] `spectate.py`
        + [ ] `eventful.py`
        + [ ] `traits.py`
+ [ ] Features:
    + [ ] Documentation:
        + [x] `spectate.py`
        + [x] `eventful.py`
        + [ ] `traits.py`
    + [ ] Eventful TraitTypes:
        + [x] Eventful Dictionary
        + [x] Eventful Lists
        + [ ] Eventful Sets
    + [x] Eventful Base Clase:
        + [x] Callback registration
        + [x] Event notifications by type
        + [x] Event supersets (e.g. `extend` is a superset of `append`)
    + [x] Watchable Types:
        + [x] `Spectator` for instances
        + [x] `MethodSpectator` descriptor to wrap methods
        + [x] Intuitive interface for exposing methods to spectators.
    + [ ] Implement `traitlets.EventReport`:
        + [ ] Add rollback heuristics (depends on [ipython/traitlets/#251](https://github.com/ipython/traitlets/pull/251)).

# Summary

An alternative to [traitlets/#278](https://github.com/ipython/traitlets/pull/278)

While [traitlets/#278](https://github.com/ipython/traitlets/pull/278) enables custom events to be built into `TraitType` subclasses, via custom classes with event hooks. This pull request exposes a generic `Eventful` base class that will automatically create these event hooks. Eventful subclasses need only defined these three class attributes:

+ `event_map` used to distribute notifications to `@observer` handlers:
	+ `{'setitem': '__setitem__'}` maps `obj.a[k] = v` to `@observer('a', type='setitem')`
        + However, the event data that is passed to an observer must be defined by trait callbacks.
+ `klass` (optional) the base class from which eventful classes are created:
	+ `klass` can also by inferred from the type of a default value.
	+ To create an eventful dictionary `klass` should be `dict`
+ `type_name` the name of the eventful type:
	+ `type(obj.a).__name__ == type_name`

# Trait Callbacks and Event Reporting

In order to report an event that has occurred on a trait value, callbacks are required. To make callbacks that trigger before, and after an event (hereafter referred to as beforebacks and afterbacks respectively) define methods `_before_<event-type>` or `_after_<event-type>` on `Eventful` subclasses (e.g. `_before_setitem` and `_after_setitem`). Beforebacks and Afterbacks give you the tools to create your notifications.

### Beforebacks
+ Signature: `(value, call)`
	1. `value`: the value held by the trait.
	2. `call`: a bunch with the keys:
		+ `name`: name of the method called
		+ `args`: the arguments that method was called with
		+ `kwargs`: the keyword arguments the method was called with
+ Return: `None`, value, or an afterback
	1. If `None` no notifications are distributed to `@observer` handlers.
	2. If a value is returned, it is distributed to `@observer` handlers under `change['event']`.
	3. The afterback is a callable with a signature `(returned)`
		+ `returned` is the output of the called method.
		+ The afterback returns `None` to prevent notifications, or a value as in [ii].

### Afterbacks
+ Signature: `(value, answer)`
	1. `value`: the value held by the trait.
	2. `answer`: a bunch with the keys:
		+ `name`: name of the method called
		+ `value`: the value returned by that call
		+ `before`: if a beforeback was defined, this is the value it returned.
+ Return: `None` or a value
	1. `None` prevents notifications
	2. Any other value gets distributed to `@observer` handlers under `change['event']`

# Examples

We can easily create an `Eventful` subclass which notifies when new items are assigned:

```python
class EItems(Eventful):
    
    # we will infer the class from a default value
    event_map = {'setitem': ('__setitem__')}
    type_name = 'eitems'
    
    @staticmethod
    def _before_setitem(inst, call):
        """Expect call.args[0] = key"""
        key, = call.args[0]
        try:
            old = inst[key]
         except:
            old = Undefined
        return key, old
    
    @staticmethod
    def _after_setitem(inst, answer):
        key, old = answer.before
        if inst[key] != old:
            return Bunch(key=key, old=old, new=new)
```

We also include basic implementations for eventful lists and eventful dictionaries.

See: [ipywidgets/widgets/trait_types/traits.py](https://github.com/rmorshea/ipywidgets/blob/eventful/ipywidgets/widgets/trait_types/traits.py)

# Abstracted Events

It's often the case that an object's method acts like it wraps several other method calls (e.g. `list.extend` acts like it wraps calls to `list.append`). To address this the method `Eventful.abstracted` provides a context manager yielding an "abstracted" event that spoofs method calls. The events which those calls would have produced are then then captured and repackaged into a single list of events that get sent to event observers. Consider this example:

```python
class WatchedLengthList(Eventful, List):
    
    event_map = {"append": "append", "extend": "extend"}
    
    def _before_append(self, value, call):
        return Bunch(index=len(value), value=call.args[0])
    
    def _before_extend(self, value, call):
        extension = call.args[0]
        with self.abstracted(value, "append") as append:
            for e in extension:
                append(e)
        return append


class HasEventfuls(HasTraits):
    elist = WatchedLengthList()
    
    @observe("elist", type="extend")
    def _extension_observe(self, report):
        print(report)

obj = HasEventfuls()
obj.elist = []
obj.elist.extend([1, 2, 3])
```

Printed Report:

```
{'owner': <__main__.HasEventfuls object at 0x10af102b0>, 'name': 'elist', 'type': 'extend', 'event': [{'value': 1, 'index': 0}, {'value': 2, 'index': 0}, {'value': 3, 'index': 0}]}
```


jasongrout:
@SylvainCorlay - it sounds like you'd be the most natural one to review and discuss this PR. I'm setting it to the Future milestone for now.

jasongrout:
> I'm setting it to the Future milestone for now.

(of course, feel free to reassign if you think we should get this in for version 7.

rmorshea:
Is there a scheduled release date for version 7? Given the size of this PR I should probably give it a pretty thorough rundown myself since it's been several months since I last worked on it.

jasongrout:
> Is there a scheduled release date for version 7?

There isn't a scheduled release date, but we're hoping to release in the next few weeks. I finished most of the major things I had planned to work on for this release, and now I'm focusing on testing and docs.

rmorshea:
@jasongrout I've made some updates, but there are still two things that are holding this back:

1. [ipython/traitlets/#251](https://github.com/ipython/traitlets/pull/251) needs to be merged to add rollback capabilities.
2. I need to write tests for new ``EList`` and ``EDict`` trait types

rmorshea:
In short, the limiting factor here will likely be a [traitlets 5.0 pre-release](https://github.com/ipython/traitlets/issues/401). @jasongrout, do you think that we can sync up the ipywidgets and traitlets releases? If not, then eventful traits will have to wait.

jasongrout:
> @jasongrout, do you think that we can sync up the ipywidgets and traitlets releases?

What is the timeframe for a traitlets release? I don't want to hold up an ipywidgets 7.0 release (which hopefully will happen by the end of May) for a traitlets release, but this could also go in a 7.1.

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+    }
+    type_name = 'edict'
+
+    def _before_update(self, inst, call):
+        new = call.args[0]
+        new.update(call.kwargs)
+        call_list = []
+        return self.redirect(
+            None, 'setitem', inst,
+            args=new.items())
```
jdfreder: Why is the above broken into three lines? Why not just:

```python
return self.redirect(None, 'setitem', inst, args=new.items())
```

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _before_clear(self, inst, call):
+        return self.redirect(
+            None, 'delitem', inst,
+            args=inst.keys())
+
+    @staticmethod
+    def before_setitem(inst, call):
+        """Expect call.args[0] = key"""
```
jdfreder: Can you expand on this description? Also, is `before_setitem` supposed to be private? 

It's strange that it's called `before_setitem` and it returns a callback (which is supposed to be used after the item is set?).

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+class EDict(eventful.Eventful, traitlets.Dict):
+
+    event_map = {
+        'setitem': ('__setitem__', 'setdefault'),
+        'delitem': ('__delitem__', 'pop'),
+        'update': 'update',
+        'clear': 'clear',
+    }
+    type_name = 'edict'
```
jdfreder: Same with the type name

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+            new = inst.get(key, traitlets.Undefined)
+            if not equivalent(old, new):
+                return traitlets.Bunch(key=key, old=old, new=new)
+        return after_setitem
+
+    _before_setitem = before_setitem
+    _before_delitem = before_setitem
+
+
+class EList(eventful.Eventful, traitlets.List):
```
jdfreder: EventfulList ?

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+    event_map = {
+        'append': 'append',
+        'extend': 'extend',
+        'setitem': '__setitem__',
+        'reverse': 'reverse',
+        'sort': 'sort',
+    }
+    type_name = 'elist'
```
jdfreder: Here too

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def before_setitem(self, inst, index):
```
jdfreder: Can you add a doc string for this? And why is this a class member when the other, for the EDict, is static?

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+            except:
+                new = traitlets.Undefined
+            if not equivalent(old, new):
+                return traitlets.Bunch(index=index, old=old, new=new)
+        return after_setitem
+
+    def _before_setitem(self, inst, call):
+        return self.before_setitem(inst, call.args[0])
+
+    def rearrangement(self, origin, inst):
```
jdfreder: Doc string here, and ideally on every function, even private ones.

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+    def _before_reverse(self, inst, call):
+        return self.rearrangement(inst, 'reverse')
+
+    def _before_sort(self, inst, call):
+        return self.rearrangement(inst, 'sort')
+
+    def _before_extend(self, inst, call):
+        size = len(call.args[0])
+        return self.redirect('append', 'setitem', inst,
+            args=[(i+len(inst),) for i in range(size)])
```
jdfreder: Can you add a comment for this line? It's not immediately obvious with this comprehension.

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+        return event
+
+
+class Redirect(object):
+
+    def __init__(self, origin, target, inst, args, kwargs):
+        self.origin = origin
+        registry = inst._instance_spectator._callback_registry
+        # Eventful._setup_events registers first
+        b, a = registry.get(target, [(None, None)])[0]
```
jdfreder: b and a force me to go read the docs for `registry.get`. Can you use names here?

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+            self.owner.notify_change(Bunch(
+                type=origin, event=event,
+                name=self.trait.name,
+                owner=self.owner))
+        return event
+
+
+class Redirect(object):
+
+    def __init__(self, origin, target, inst, args, kwargs):
```
jdfreder: s/inst/instance/

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+        # trigger afterbacks
+        return self.trigger(value)
+
+    @staticmethod
+    def trigger(value):
+        pass
+
+
+class Eventful(TraitType):
+
```
jdfreder: Doc strings for classes too.

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+        self._active_events = []
+        self.setup_events()
+
+    def event(self, type, on, before=None, after=None):
+        for method in (on if isinstance(on, (list, tuple)) else (on,)):
+            self._active_events.append((type, method, before, after))
+        return self
+
+    def setup_events(self):
+        for name, on in self.event_map.items():
```
jdfreder: s/on/event/ - right? Assuming an event map is a map of names to events.

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(Eventful, self).__init__(*args, **kwargs)
+        default_value = self.default()
+        if self.klass is None and default_value is not Undefined:
+            self.klass = type(default_value)
+        self._active_events = []
+        self.setup_events()
+
+    def event(self, type, on, before=None, after=None):
```
jdfreder: Is `on` here the same as in setup_events below? If so, I suggest renaming it to `event`.

ipywidgets/widgets/trait_types/eventful.py
```diff
@@ -0,0 +1,180 @@
...
+        for method in (on if isinstance(on, (list, tuple)) else (on,)):
+            self._active_events.append((type, method, before, after))
+        return self
+
+    def setup_events(self):
+        for name, on in self.event_map.items():
+            for method in (on if isinstance(on, (tuple, list)) else (on,)):
+                before = getattr(self, "_before_"+name, None)
```
jdfreder: nit: spaces between `+` here and below

rmorshea:
@jdfreder thanks for the review. I will add docstrings to the TODO list in the description - they are desperately needed in this PR! No need for further review until then, as docs should make doing so much easier.

ipywidgets/widgets/trait_types/__init__.py
rmorshea: 1. The description was more of a personal note. I will be writing a complete set of docstrings.
2. You're right, it'd be better to just merge `_before_setitem` and `before_setitem`.

## Why allow the return of "afterbacks"?

This is what enables redirects.

A `Redirect` is essentially a before/afterback (b/a) pair wrapped in one object:

1. `Redirect.__init__` : Gathers the `target` b/a, calls the beforeback, and creates a closure that holds the beforeback's result and will execute the afterback when called.
2. `Redirect.__call__`: triggers the aforementioned closure, which runs the afterback.

Because callable objects returned by beforebacks act like a afterbacks. A "redirection" of a notification can be performed by returning a `Redirect` from a beforeback.

Let me know if this makes sense.

ipywidgets/widgets/trait_types/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+            return super(InstanceDict, self).validate(obj, self.klass(**value))
+        else:
+            return super(InstanceDict, self).validate(obj, value)
+
+    def make_dynamic_default(self):
+        return self.klass(*(self.default_args or ()),
+                          **(self.default_kwargs or {}))
+
+
+class EDict(eventful.Eventful, traitlets.Dict):
```
rmorshea: This was just the naming convention used by @SylvainCorlay in https://github.com/ipython/traitlets/pull/278. I'm okay changing it though since the `Eventful`  prefix is more descriptive than `E`.

ipywidgets/widgets/trait_types/__init__.py
rmorshea: I'll change this - for some reason I always want to make every line approximately the same length in Python.

ipywidgets/widgets/trait_types/__init__.py
rmorshea: This should be static, and similarly to `EList` I think I'll just merge this into `_before_setitem`

ipywidgets/widgets/trait_types/__init__.py
rmorshea: I'm going to have a `redirect` decorators rather than the method:

```python
@cause("setitem")
def _before_extend(self, value, call, setitem):
    for i in range(len(call.args[0])):
        setitem(i + len(value))
```

Hopefully this is more intuitive than what's currently implemented?

ipywidgets/widgets/trait_types/eventful.py
rmorshea: I think I'm going to use value. Even when I was writing this I kept confusing myself over whether it was the "instance" was the trait or the value being set.

ipywidgets/widgets/trait_types/eventful.py
rmorshea: I think I'll replace `on` with `methods` since it's supposed to refer to the methods that the "event" will be registered to.

ipywidgets/widgets/trait_types/eventful.py
rmorshea: Maps arbitrary names (used for the `_before_<name>` and `_after_<name>` convention) to methods.

rmorshea:
@jdfreder, any chance I could do a voice chat without about this pr. Have some design questions I'd like to hash out.

jdfreder:
Yeah, I could GVC tomorrow afternoon, PST - does that work for you?

rmorshea:
@jdfreder, still a lot to go, but there's also a lot of added docstrings and tests now.

rmorshea:
I've added a ton of docstrings and a relatively thorough test suite for this pull request.

Let me know if anything doesn't make sense (docs, code, or otherwise) - there's a lot here.

cc: @jdfreder @SylvainCorlay

rmorshea:
Updated the pull request checklist and accurate summary of some user facing features.

Madhu94:
@rmorshea Will this be in the next release of widgets ?

rmorshea:
@Madhu94 there is actually a workshop coming up in January where I may get a chance to work on this. If I make enough progress there, it's possible it could make it into 7.2, but I can't make any promises on that.

Madhu94:
@rmorshea Thanks. Let me know if there is anything I can do to help move this PR along (it is hard to see  exactly what's pending)

rmorshea:
Closed in favor of: https://github.com/ipython/traitlets/pull/466

rmorshea:
This is quite old, however I'll just mention that I don't think this will get merged into Traitlets.

For anyone interested in this kind of behavior with Traitlets I would suggest looking at [Spectate](https://github.com/rmorshea/spectate). This PR will never be able to deliver the kinds of granular change notifications Spectate provides, so if you absolutely need to track changes to mutable data types from within Traitlets, I would suggest implementing a `TraitType` that integrates Spectate and Traitlets. 

To jump start ideas, consider spectate's `Oracle` object which allows you to track nested changes:

```python
from spectate.mvc import Oracle, Dict, view

o = Oracle(
    a=Dict(
        x=Dict(
            y=1
        )
    )
)

@view(o)
def changed(events):
    for e in events:
        print(e)

o.a["x"].update(y=2, z=3)
```

```
{'key': 'y', 'old': 1, 'new': 2, 'location': ('a', 'x')}
{'key': 'z', 'old': Undefined, 'new': 3, 'location': ('a', 'x')}
```

TK-21st:
@rmorshea I have a need for an eventful `Dict` in my new ipywidget since the JS library. The use-case is that I have a dictionary like
```
{
  object1:
    {
      data: [...] (a list or tuple),
      properties:
        {
          color: (tuple),
          state: string,
          ... (other properties potentially dicts)
        }
    }
}
```
The JS side uses a library we developed in the past that relies on a custom-made eventful object object and hence allows for event propagation through a nested dict like the one above.

To simplify python side, I want to keep the same data-structure and am wondering if the ipywidget JS side would work with `spectate`, i.e. would the following work
```
//python
class myClass(DOMWidget):
    def __init__():
       ...
        this. o = Oracle(
            a=Dict(
                x=Dict(
                    y=1
                )
            )
        )
      ...
//javascript
class myWidgetView extends DOMWidgetView{
    initialize(){
        ...
        this.model.on('change:o', do_something, this);
    }
    ...
} 

```

rmorshea:
@TK-21st unfortunately I removed the `Oracle` class from a later release of Spectate because it wasn't working as intended and I didn't have time to fix it. I'd be happy to help you reimplement an `Oracle` class for Spectate and integrate it into a custom `TraitType`. Doing so would revive the dream of "eventful" traits 🙌

rmorshea:
The documentation for Spectate has been greatly improved. The section on defining you own model types is probably what you'd want to look at:

https://python-spectate.readthedocs.io/en/latest/usage/create-models.html

PR #1075, opened 2017-02-03
Make the base, label, and readout font colors configurable.

jasongrout:
This continues making our theming more flexible.


jasongrout:
Fixes #1075.

Issue #1073, opened 2017-02-02
Make it possible to change the min-width on textbox, select, etc.


jasongrout:
Right now, there are min widths set on inner elements (like inputs or the actual select control) that prevent the `.layout.min_width` attribute from controlling the minimum width of a control. Audit these width settings and make sure that it's possible to change the widths using the `.layout` attributes.

PR #1078, opened 2017-02-03
Widths

jasongrout:
Many widgets (such as text, select, slider, etc.) now let you set the width with `widget.layout.width`.

This does not fix the multiple select, which needs to be overhauled.

This is a partial fix to #1073.

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -370,7 +370,7 @@
     font-size: var(--jp-widgets-font-size);
     padding: var(--jp-widgets-input-padding);
     flex-grow: 1;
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
lgtm

SylvainCorlay:
@jasongrout can hop on gitter?

jasongrout:
This may change the widths of widgets in a layout, but tries to make it even more consistent by making the width just depend on the outermost container's width.

jasongrout:
It looks like since the default value for flex-shrink is 1, if the widgets would overflow the output area, they do shrink until they fit. However, if they do fit, they are the 'correct' widths and lay out nicely.

jasongrout:
#1081 continues this.

PR #1082, opened 2017-02-03
Add HTMLMath widget.

### LINKED ISSUE #1072: Widget to render html + math?
Since we don't have a Latex widget anymore, and the Label widget now has strict size limits, the HTML widget seems like the appropriate widget to render math in. However, it doesn't, it just puts the raw html on the page without typesetting it.

Should we (a) enable math rendering (b) make an option on the html widget to render math, or (c) make a new HTMLMath widget?

It's a one-line change to render math, but I think it makes sense to keep the stock HTML widget as plain as we can. I think having a new HTMLMath widget makes sense.

CC @SylvainCorlay.

jasongrout:
Also fix HTML and HTMLMath widget so that the description label works.

Fixes #1072.

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -37,51 +37,54 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "collapsed": false
+   },
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "{'Jupyter.Accordion': ipywidgets.widgets.widget_selectioncontainer.Accordion,\n",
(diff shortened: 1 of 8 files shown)


SylvainCorlay:
I am a bit confused by what is the scope of the HTML math widget.

 - is it an html widget
 - is it a multi-line MathJax-rendered widget?

jasongrout:
It is merely an HTML widget with mathematics parsing and typesetting. I don't think we should make the default HTML widget do math typesetting.

SylvainCorlay:
> I don't think we should make the default HTML widget do math typesetting.

I agree with that. Similarly, I wonder if this is the place for general HTML. Would it make sense to limit this to typesetting?

jasongrout:
I think the common use case will be HTML plus typesetting.

SylvainCorlay:
Do you have a minute to hop on gitter?

PR #1087, opened 2017-02-07
Allow a selection widget to have no options

### LINKED ISSUE #1086: 'Select' widget does not support empty list in options any more
Previously *Select* widget supported empty list in the options. Now it throws a TraitError saying "'options' must not be empty"!

Please enable Select to get initialized with an empty list (and have value attribute initialized to None). We have use cases where the Select list has no options initially and later gets populated based on some event.

jasongrout:
Fixes #1086.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -28,17 +28,21 @@ def _value_to_label(value, obj):
     # TODO: make a separate validation function so this doesn't have
     # to redo the work of parsing the options object.
     options = obj._make_options(obj.options)
-    if len(options) == 0:
-        raise TraitError("`options` must not be empty.")
-    try:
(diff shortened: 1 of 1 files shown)


jasongrout:
Test code:

```python
from ipywidgets import *
Select(options=[])
```

SylvainCorlay:
The code looks good to me. I am doubtful this is a good idea to support that behavior.

jasongrout:
Two situations came up:

1. initializing a select box and later adding the options. Without this, you have to put in a default option, then remember to remove it when having another option.

2. Making it possible for a user to remove options - without this, you have to be careful about special-casing when the list becomes an empty list.

Basically, this makes handling the empty-option case easy for the user, harder for us. I think that's better than making it harder for the user, but more straightforward for us.

PR #1089, opened 2017-02-08
Make various layout properties enums.

### LINKED ISSUE #740: CaselessStrEnum for layout attributes?
Hi,

I have hard time about silent error.

For example, I try

``` python
from ipywidgets import Text

t = Text()
t

# set visibility
t.layout.visibility = True # and 'true'
```

nothing happens, no error message.
turns out that `t.layout.visibility = 'hidden' # or visible`

It's great help if error message is clearer. Since [Layout](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget_layout.py) only support a subset of CSS, what's about using CaselessStrEnum rather the Unicode?

jasongrout:
Fixes #740.

### DIFF
--- a/ipywidgets/widgets/widget_layout.py
+++ b/ipywidgets/widgets/widget_layout.py
@@ -3,9 +3,10 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import Unicode, Instance
+from traitlets import Unicode, Instance, CaselessStrEnum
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
:+1: on this.

jasongrout:
(You could just merge it :)

SylvainCorlay:
I am leaving soon. I did not want to take upon myself to merge without testing.

PR #1092, opened 2017-02-09
fix ES2015 feature of getter

ivanov:
We get a javascript error with this syntax on older browsers.

cc @jasongrout


jasongrout:
Thanks!

PR #1093, opened 2017-02-10
Move LabeledDOMWidgetModel and LabeledDOMWidgetView to exports so others can subclass

jasongrout:
This provides a standardized way for 3rd party widgets to handle the description labels.


SylvainCorlay:
OK, it makes sense. Thanks.

PR #1094, opened 2017-02-10
remove more of ES2015/ES6 features

ivanov:
follow up to #1092 - older browser won't work without these changes. 

With these fixes in place, the next stumbling block I hit is an error of "WeakMap is not defined". I'll open a separate issue for that.

### DIFF
--- a/widgetsnbextension/src/widget_output.js
+++ b/widgetsnbextension/src/widget_output.js
@@ -66,7 +66,7 @@ var OutputView = widgets.DOMWidgetView.extend({
     },
 
     render: function(){
-        let renderOutput = (outputArea) => {
+        var renderOutput = function(outputArea) {
             this.output_area = new outputArea.OutputArea({
                 selector: this.el,
                 config: this.options.cell.config,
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks. I guess I was used to having Typescript compile these down to es5.

As for the WeakMap, that may be where we need to draw the line - that's a phosphor requirement, IIRC. What browsers are we trying to support here?

jasongrout:
One of the major things arrow functions give us is capturing this. If we switch away from them, then we'll need to be careful how we use `this` inside the function (i.e., don't use it). I think there's at least one `this` in one of the functions you are converting.

widgetsnbextension/src/widget_output.js
```diff
@@ -66,7 +66,7 @@ var OutputView = widgets.DOMWidgetView.extend({
     },
 
     render: function(){
-        let renderOutput = (outputArea) => {
+        var renderOutput = function(outputArea) {
             this.output_area = new outputArea.OutputArea({
```
jasongrout: We'll need to capture this, like `var that = this`, and then use `that` inside the function instead of `this`.

ivanov: Oh, right. Thanks. I tried with the next commit, but I'm already pretty far out of my comfort zone so please tell me if I need to make other changes

jasongrout: Thanks, your changes look great! I'll try to test and merge tomorrow. Thanks for finding and correcting this oversight!

jasongrout:
This seems to work great. Thanks again!

Issue #817, opened 2016-10-06
Labels should allow HTML code?


jdemeyer:
Currently, the label (i.e. the `description` attribute) of a widget is displayed as plain text (with LaTeX support). It would be useful to allow arbitrary HTML.

---
## Summary
Should the description accept html? Right now it doesn't by design, but we could change that.

SylvainCorlay:
I think that it currently allow latex equations.

jasongrout:
In JupyterLab, `IntSlider(description='<u>hi</u> $x^2$')` does *not* render the latex *until* mathjax is loaded to render some other equation. The html is not rendered since we use textContent: https://github.com/ipython/ipywidgets/blob/91b9ae5de02f9618642d95f72bd0088b9f64b1a9/jupyter-js-widgets/src/utils.ts#L155

gnestor:
In notebook, LaTeX in labels works but not HTML:

![image](https://cloud.githubusercontent.com/assets/512354/19604440/209c1698-976a-11e6-906c-b4003f46eedb.png)

jasongrout:
#888 seems to fix the issue with the latex not rendering immediately in my example above.

Should the description accept html? Right now it doesn't by design, but we could change that.

jasongrout:
ping @SylvainCorlay - what do you think? Allowing html code would be pretty natural, but may not be compatible with the label's width/height restrictions. I think I'm +0 on it.

SylvainCorlay:
Let's move this forward. I also have no opinion on this. +0 :)

jasongrout:
Oh, and the other problem is that having a label with `<` would no longer be as easy.

jasongrout:
but I think the added flexibility of html is probably worth it.

jasongrout:
> It would be useful to allow arbitrary HTML.

@jdemeyer - is this important for Sage compatibility?

jdemeyer:
> @jdemeyer - is this important for Sage compatibility?

Depends what you mean "important". It's not very high priority.

But HTML in labels is certainly something that people did use when making interacts in the Sage Notebook. So it's not a theoretical thing.

PR #1096, opened 2017-02-10
Handle font-weight

SylvainCorlay:


jasongrout:
I think we won't need this if https://github.com/ipython/ipywidgets/issues/817 goes in, and we just allow the button label to be html.

SylvainCorlay:
That is for the button desciption.

PR #1098, opened 2017-02-10
Clear widget areas when output is cleared

### LINKED ISSUE #637: Clear widgets when a cell's output is cleared in classic notebook
Since clear output doesn't clear widgets, we need a clear widgets action in the Widget menu.

jasongrout:
Fixes #637.

### DIFF
--- a/widgetsnbextension/src/extension.js
+++ b/widgetsnbextension/src/extension.js
@@ -62,18 +62,32 @@ function register_events(Jupyter, events) {
         handle_cell(cells[i]);
     }
 
-    // Listen to cell creation and deletion events.  When a
-    // cell is created, create a widget area for that cell.
     events.on('create.Cell', function(event, data) {
         handle_cell(data.cell);
     });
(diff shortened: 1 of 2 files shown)


jasongrout:
This needs jupyter/notebook#2173 in order to properly clear output areas.

jasongrout:
Rebased for the new event name (see #2173)

marscher:
Just a stupid question, why is the widget not removed when we call close on the backend side? I seems like there are some left overs after closing (namely an empty div element in the dom).

jasongrout:
Great question. We discussed this also in https://github.com/jupyter-widgets/ipywidgets/issues/1845 -- basically it's a limitation of the notebook. However, I think https://github.com/jupyter-widgets/ipywidgets/issues/1845#issuecomment-350482904 points out a way we can get visually what we want without messing up the notebook.

Let's take the discussion to that issue so it's all in one place.

marscher:
Thanks for the references and quick response!

PR #1100, opened 2017-02-10
Fix restoration of widgets on refresh.

jasongrout:
Before, this whole clause would error out when it got to an output message that was not a display_data message.


jasongrout:
Do you ever sleep?

PR #1101, opened 2017-02-10
Allow html in labels

### LINKED ISSUE #817: Labels should allow HTML code?

jasongrout:
Fixes #817

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -960,7 +960,8 @@ class LabeledDOMWidgetView extends DOMWidgetView {
         if (description.length === 0) {
             this.label.style.display = 'none';
         } else {
-            this.typeset(this.label, description);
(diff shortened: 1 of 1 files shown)


jasongrout:
This doesn't allow html in buttons - if we want to do that, let's make a different ticket?

SylvainCorlay:
Html in buttons are a good way to give the opportunity to break layout and all. I am not sure this is a good idea.

jasongrout:
> I am not sure this is a good idea.

html in labels, or buttons, or both?

SylvainCorlay:
Buttons would be more dangerous than labels.

Issue #1077, opened 2017-02-03
Problem trying to set the height (and width?) of a Textarea widget


nleclercq:
Hi all,
Forcing the height of a Textarea doesn't work properly (I'm using ipywidgets 6 beta 6).
Attached is a picture illustrating the problem.
Am I doing something wrong ?
N.

![textarea_size_problem](https://cloud.githubusercontent.com/assets/19185982/22597690/f448b870-ea30-11e6-8041-804f3f9d51c4.jpg)

jasongrout:
Related: I'm working on making sure setting the widths works in #1078.

jasongrout:
To make it easier to reproduce, here is the code from the above picture:
```python
from IPython.display import display
from ipywidgets import Button, Layout, Textarea, HBox, VBox
l = Layout(flex='0 1 auto', height='40px', min_height='40px', width='auto')
t1 = Textarea(value='TA: height=40px', layout=l)
t2 = Textarea(value='TA: height=40px', layout=l)
vb = VBox([t1, t2], layout=Layout(flex='1 1 auto', width='auto'))
display(t1,t2)
```

jasongrout:
The problem is that the rows (which we hardcode) overrides the height setting: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_string.ts#L113

We should expose 'rows' (and 'cols'?) to the user as a Textarea attribute.

jasongrout:
(for a user that wants to set a specific height, they'd probably want to set the rows to be 1 and then control the height with `.layout.height`).

jasongrout:
Should we always just set the rows to 1, and set a default height that the user can change? That makes it harder to have a textarea that is an integral multiple of the line height, but makes it easier to tile the widgets in a nice layout. @SylvainCorlay, @cameronoelsen

jasongrout:
Sylvain and I discussed this. I'm going to move forward with setting rows to 1 by default, setting a natural height in css based on the default lineheight (to still give an integral number of rows).

Then if the user wants a text area of a certain size, they can just set the `.layout.height`.

We also might expose the rows attribute, so it is easy for a user to size the widget to an integral number of rows. Then it will be the user's responsbility to also set the height to something that makes sense.

abulka:
Sorry to comment on a closed issue, but I've been struggling for hours to make the textarea auto expand to the height of its parent container.  Buttons do it, bees do it - but not the text area.

```
[...]

PR #1102, opened 2017-02-11
WIP Make text area the same height as enclosing div

jasongrout:
WIP fix for #1077

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -366,6 +366,11 @@
     line-height: var(--jp-widgets-inline-height);
 }
 
+.widget-textarea textarea {
+    height: 100%;
+    line-height: normal;
+}
+
 .widget-text input[type="text"], .widget-textarea textarea {
     box-sizing: border-box;
     border: var(--jp-border-width) solid var(--jp-border-color1);


jasongrout:
This still doesn't quite do it - it stretches the textarea to be however high the enclosing box is, so, for example, we get a longer textbox if we have
```python
from ipywidgets import *

s = Textarea(description='hi')
i = IntSlider(description='another')
HBox([s,VBox([i,i,i,i,i])])
```
However, the textbox looks about 5 rows high in this case:
```python
from ipywidgets import *

s = Textarea(description='hi')
i = IntSlider(description='another')
HBox([VBox([s,i]),VBox([i,i,i,i,i])])
```
though in that case, the slider under the textarea doesn't align with the slider next to it.

Perhaps we give up on having integral-row textboxes, and just set the size to be about 3 widgets high (with inter-spacing). Another problem is that the widget and the textarea seem to be offset by one pixel when the align is baseline and the line-height is normal. Perhaps we need to set a lineheight to know for sure how much padding so that the top of the textbox is aligned with the top of the label and the baselines are also aligned.

@cameronoelson, thoughts about sizing the textarea compared to the other controls?

jasongrout:
I reset the line-height to normal, since it was inheriting some 20px line height from somewhere in the notebook css. I think a line height of 28px as in most of the widgets is way to spread out for lines in a text box.

SylvainCorlay:
Is this still WIP?

jasongrout:
Yes. I'll work on this today.

SylvainCorlay:
This was fixed in #1113

PR #1113, opened 2017-02-15
Fix textarea layout logic

### LINKED ISSUE #1077: Problem trying to set the height (and width?) of a Textarea widget

SylvainCorlay:
Closes #1102 .
Fixes #1077 .

A new `rows` attribute controls the natural height and actually sets the `rows` property of the `textarea` tag. Setting it to the default trait value `None` let the browser use its default value which is typically 2.

Layout properties override this natural height when set.

textarea


SylvainCorlay:
Same unrelated test failure on master. Merging.

PR #1117, opened 2017-02-16
Tab Styling Issue Fix

ssunkara1:
Right now, the `TabPanel` has overflow set to `visible` because of being classed as `jupyter-widgets`. The same styling is set explicitly for the `TabBar`. But the `TabPanel` contents has `overflow` set to `hidden`.

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -880,6 +880,11 @@ ul.widget-dropdown-droplist.mod-active {
     overflow-y: visible;
 }
 
+.jupyter-widgets .p-TabPanel-tabContents {
+    overflow-x: visible;
+    overflow-y: visible;
+}
+
 .jupyter-widgets.widget-tab .p-TabBar.p-mod-horizontal > .p-TabBar-content {
     align-items: flex-end;
     margin-bottom: 0;


ssunkara1:
@jasongrout @SylvainCorlay

jasongrout:
On the other hand, perhaps this calls into question why jupyter-widgets sets overflow to visible. Perhaps it shouldn't override phosphor's `override: hidden`?

SylvainCorlay:
> On the other hand, perhaps this calls into question why jupyter-widgets sets overflow to visible. Perhaps it shouldn't override phosphor's override: hidden?

As far as I remember, several widgets in the widgets list were broken if we did not do so. It was a big difference with the earlier version.

In any case, if we make it "hidden", tooltips in bqplot figures will be come a lot harder to make work in the classic notebook at the same time as in JLab...

jasongrout:
> In any case, if we make it "hidden", tooltips in bqplot figures will be come a lot harder to make work in the classic notebook at the same time as in JLab...

IIRC, the tooltips in bqplot are done 'right', in that they are absolute positioned divs at the top layer, not relatively positioned divs next to the plots, so they shouldn't be affected at all.

SylvainCorlay:
> IIRC, the tooltips in bqplot are done 'right', in that they are absolute positioned divs at the top layer, not relatively positioned divs next to the plots, so they shouldn't be affected at all.

Not anymore, they are in the figure's div at the moment.

ssunkara1:
@jasongrout @SylvainCorlay The tooltips right now are placed in the same `div` as the `svg`. So changing the `overflow` to `hidden` will affect the tooltips, i.e., they will be clipped.

ssunkara1:
@SylvainCorlay @jasongrout so what is the final consensus on this?

jasongrout:
1. We're currently factoring out the tooltip logic used in jupyterlab so it can be used in other projects.

2. I think we should keep it the way it is, and maybe even set other containers to overflow:hidden...

jasongrout:
Closing as a 'we should fix the real issue with tooltips'

ssunkara1:
Along with tooltips, we should also make `TabPanel` consistent with the rest of the containers. I thought that was the main issue here :)

jasongrout:
Good call. I think we ought to test taking off the overflow:visible.

PR #1122, opened 2017-02-17
WIP Phosphor update

jasongrout:
Phosphor is coming up on a 1.0 release - this migrates our code to the pre-1.0 release.

### DIFF
--- a/ipywidgets/_version.py
+++ b/ipywidgets/_version.py
@@ -3,4 +3,4 @@
 
 version_info = (6, 0, 0)
 __version__ = '.'.join(map(str, version_info))
-__frontend_version__ = '~2.1.4'
+__frontend_version__ = '~2.2.0'
--- a/jupyter-js-widgets/package.json
+++ b/jupyter-js-widgets/package.json
@@ -77,12 +77,16 @@
     "spawn-sync": "^1.0.14",
     "style-loader": "^0.13.1",
     "typedoc": "^0.5.0",
-    "typescript": "~2.0.10",
+    "typescript": "~2.1.0",
     "url-loader": "^0.5.7",
     "webpack": "^1.12.11"
   },
   "dependencies": {
     "@jupyterlab/services": "^0.35.0",
+    "@phosphor/algorithm": "0.1.0",
+    "@phosphor/widgets": "0.1.0",
+    "@phosphor/messaging": "0.1.0",
+    "@phosphor/utilities": "0.1.0",
     "@types/backbone": "^1.3.33",
     "@types/semver": "^5.3.30",
     "ajv": "^4.9.0",
@@ -93,7 +97,6 @@
     "jquery-ui": "^1.12.1",
     "jupyter-widgets-schema": "^0.1.1",
     "lolex": "^1.4.0",
-    "phosphor": "^0.7.0",
     "scriptjs": "^2.5.8",
     "semver": "^5.1.0",
     "underscore": "^1.8.3"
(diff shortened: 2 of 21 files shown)


jasongrout:
This isn't going to work with JLab while JLab is using phosphor 0.7 - we'll have conflicting versions of phosphor, and thus it breaks compiling the widget plugin for phosphor.

So I'm bumping this to 6.1, with the idea that we can merge this after Jupyter moves to Phosphor 1.0.

jasongrout:
Actually, maybe this just *might* work. If I leave jupyterlab_widgets with phosphor 0.7.0, but I do upgrade typescript there to 2.1.0 (so I can compile against jupyter-js-widgets), I get just one error:

```
src/output.ts(106,7): error TS2415: Class 'OutputView' incorrectly extends base class 'DOMWidgetView'.
  Types of property 'pWidget' are incompatible.
    Type 'Panel' is not assignable to type 'Widget'.
      Types of property 'disposed' are incompatible.
        Type 'ISignal<Widget, void>' is not assignable to type 'ISignal<Widget, void>'. Two different types with this name exist, but they are unrelated.
          Types of property 'connect' are incompatible.
            Type '(slot: Slot<Widget, void>, thisArg?: any) => boolean' is not assignable to type '(slot: Slot<Widget, void>, thisArg?: any) => boolean'. Two different types with this name exist, but they are unrelated.
              Types of parameters 'slot' and 'slot' are incompatible.
                Type 'Slot<Widget, void>' is not assignable to type 'Slot<Widget, void>'. Two different types with this name exist, but they are unrelated.
                  Types of parameters 'sender' and 'sender' are incompatible.
                    Type 'Widget' is not assignable to type 'Widget'. Two different types with this name exist, but they are unrelated.
                      Types of property 'title' are incompatible.
                        Type 'Title' is not assignable to type 'Title<Widget>'.
                          Types of property 'changed' are incompatible.
                            Type 'ISignal<Title, void>' is not assignable to type 'ISignal<Title<Widget>, void>'.
                              Types of property 'connect' are incompatible.
                                Type '(slot: Slot<Title, void>, thisArg?: any) => boolean' is not assignable to type '(slot: Slot<Title<Widget>, void>, thisArg?: any) => boolean'.
                                  Types of parameters 'slot' and 'slot' are incompatible.
                                    Type 'Slot<Title<Widget>, void>' is not assignable to type 'Slot<Title, void>'.
                                      Types of parameters 'sender' and 'sender' are incompatible.
                                        Type 'Title' is not assignable to type 'Title<Widget>'.
```

jasongrout:
By avoiding the type-checking right at the interface of the old and new phosphor (by setting the OutputWidget output area to have type any), we can get this to compile. It even sort of works 

@sccolbert, @blink1073 - do you see any huge problems with mixing phosphor pre-1.0 with JLab at this point?

jasongrout:
(The tabs and accordion widgets don't work in either the classic notebook or Lab - I'll look into this more...)

sccolbert:
The only problems you'll be likely to have are widgets which cross the boundaries of the modules. So dragdrop widgets between modules wont work. But you'll probably be "okay" other than that.

jasongrout:
Thanks. @SylvainCorlay - let's definitely tackle this after 6.0. It's subtle enough that I don't think we should try to push it in at this point.

blink1073:
The output area itself is what it used as the drop value, so it *should* work fine.

jasongrout:
@SylvainCorlay - any comments on this before I merge it?

PR #1140, opened 2017-02-21
Make selected range thicker

SylvainCorlay:
This reverts a change that was done in #1130 . 

It makes the selected interval thicker (again) in the range slider:

Thin (before this PR):

thin

Thick (after this PR):

thick


SylvainCorlay:
This is important if the color of the selected range cannot be easily differentiated from the rail color.

jasongrout:
+1 to making it a bit thicker - having it so thin bothered me a bit too for the same reason.

PR #1141, opened 2017-02-21
Don't set height of checkbox, which sometimes causes abnormally large checkboxes

### LINKED ISSUE #1139: Check checkbox size on windows and linux
Apparently the checkbox is abnormally large on linux and windows, but looks fine on OS X. Perhaps instead of `large` at https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/css/widgets-base.css#L329, we should give a specific percentage, like 120%.

jasongrout:
Fixes #1139.

CC @maartenbreddels


SylvainCorlay:
@jasongrout can you hop on gitter?

Issue #1142, opened 2017-02-21
Dropdown button_style attribute doesn't work


jasongrout:
Since we use the select element for dropdowns now, we completely ignore the button_style attribute.

jasongrout:
Perhaps we just delete the dropdown style attribute? We don't provide those style attributes on other input widgets, only on buttons.

Essentially this is just deleting this attribute: https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget_selection.py#L229

jasongrout:
Actually, given our choices to (at least temporarily) change the dropdown to a select, perhaps we literally rename it and deprecate the dropdown? In the future with a custom dropdown, we can just change the select to a dropdown?

SylvainCorlay:
I have no opinion on this.

SylvainCorlay:
Actually I prefer to keep the dropdown. We have a phosphor implementation of it and really nice styling. I guess that it can get back in when we fix the issue.

jasongrout:
It's broken right now, so I think we should either remove the attribute or fix it (i.e., add styles to the select widget to do fulfill this).

jasongrout:
I'm happy to remove the attribute - any problems with that?

SylvainCorlay:
ok I think that it is fine to remove the attribute.

PR #1145, opened 2017-02-21
Remove dropdown style attribute and unused dropdown code

jasongrout:
Since we are currently using the select to provide the dropdown interface, I'm removing the nonfunctional button_style attribute. I'm also removing the unused code. We will likely use logic from this code when writing a true phosphor dropdown widget, and the code will be in the repo history.


jasongrout:
Fixes #1142. CC @SylvainCorlay.

SylvainCorlay:
works for me.

jasongrout:
I'll wait until tests pass before merging...

PR #1148, opened 2017-02-22
Make the date picker consistent with the text box theming

jasongrout:


jasongrout:
@SylvainCorlay, @ellisonbg - we missed this.

SylvainCorlay:
Good catch.

PR #1149, opened 2017-02-23
Make slider css more robust and customizable

jasongrout:
This PR addresses two problems with the slider css:

1. Make the slider css more specific than the jquery-ui css, so that even if another extension includes the jquery-ui css after the widgets css is loaded, the widgets customizations don't get overridden.

2. Make the default slider handle colors abstracted into css variables.


jasongrout:
CC @SylvainCorlay, @ellisonbg.

SylvainCorlay:
LGTM

PR #1156, opened 2017-02-25
Suppress exceptions during output widget redirection.

### LINKED ISSUE #1155: interact: tracebacks should be in the output area
When an `interactive` function raises an exception, the traceback will always remain shown below the normal output area, even if some future run of the interactive function completes successfully. If several runs raise an exception, all tracebacks are shown. This is all very confusing, since it's hard to see whether the interactive function worked or raised an exception.

What should happen instead is that tracebacks while running interactive functions are treated like normal output. If the function raises an exception: show the traceback in the same output area used for normal output.

Context: In SageMath, it is quite common to use `eval()` on a `Text` widget, such that the user can input `sin(x)` and have it understood as the function `x |-> sin(x)`. Of course, something like `sin(x` will raise an exception. This is just to say that exceptions can be quite common.

jasongrout:
Exceptions are already captured and shown, so there’s no need to keep raising them.

Fixes #1155.

### DIFF
--- a/ipywidgets/widgets/widget_output.py
+++ b/ipywidgets/widgets/widget_output.py
@@ -58,6 +58,8 @@ class Output(DOMWidget, CoreWidget):
             ip.showtraceback((etype, evalue, tb), tb_offset=0)
         self._flush()
         self.msg_id = ''
+        # suppress exceptions, since they are shown above
+        return True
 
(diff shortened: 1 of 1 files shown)


jdemeyer:
Together with #1162, this fixes #1155.

PR #1162, opened 2017-02-25
Call get_interact_value inside "with out" block

jdemeyer:
This way, exceptions raised by `get_interact_value()` will be shown in the output area. Related: #1155


SylvainCorlay:
Thanks

PR #1166, opened 2017-02-27
Use built css instead of original css3 for embedder

### LINKED ISSUE #1152: Switch to compiled css for the widget embedder
The current version uses the css variables and won't display on IE.

SylvainCorlay:
Fixes #1152 .

Waiting on @jasongrout to comment:  we could also use the post-css loader in the embedder webpack.

### DIFF
--- a/jupyter-js-widgets/package.json
+++ b/jupyter-js-widgets/package.json
@@ -71,6 +71,7 @@
     "postcss-cli": "^2.6.0",
     "postcss-cssnext": "^2.8.0",
     "postcss-import": "^8.1.2",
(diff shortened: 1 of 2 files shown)


jasongrout:
Either way. I'd probably use the webpack loader to be consistent with widgetsnbextension, but the purpose of the widgets.built.css was to make it simple to use a pre-compiled version.

SylvainCorlay:
OK, now using `postcss-loader`.

SylvainCorlay:
@jasongrout I let you merge this one.

PR #1167, opened 2017-02-27
Range readout using dash

jdemeyer:
For the readout format of a range slider, use "A – B" instead of "A-B" for a range `(A, B)`. This looks much better. The old format looked especially bad with negative numbers: compare "-5 – 5" to "-5-5".


SylvainCorlay:
Lgtm.

PR #1168, opened 2017-02-27
Center button icon when no descripton

### LINKED ISSUE #1160: Button icon has 4px margin, throwing it off center when no text
When there is just an icon, and no text, the icon is off center in the button because it has a 4px right margin: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/css/widgets-base.css#L121

We should figure out a way so that the icon could be centered when it is the only thing on the button.

Putting as 6.0, but could be bumped if the fix isn't easy.

SylvainCorlay:
Fixes #1160.

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -128,6 +128,10 @@
     content: "\200b"; /* zero-width space */
 }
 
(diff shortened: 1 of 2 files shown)


jasongrout:
Thanks!

Issue #1186, opened 2017-03-06
Repeat button for Play


maartenbreddels:
Some animation may benefit from having a repeat option, for instance here:
https://www.astro.rug.nl/~breddels/test/wave/
I'm not sure if a repeat button should always be shown, but I don't think it will hurt.

jasongrout:
A reverse button and maybe a speed of play button would also be interesting.

maartenbreddels:
Would be nice if these could be optional, to not clutter the gui.

SylvainCorlay:
@maartenbreddels PR welcome!

maartenbreddels:
Ok, but dev-install.sh often fails for me, a new clean clone might work again, I may try.

SylvainCorlay:
I don't actually use dev_install.sh!

maartenbreddels:
I always get the same, I start from a clean fork, everythings works. When I run `npm run update` from nbwidgetsextension a 2nd time, it fails running `npm install file:../jupyter-js-widgets`
```
air2 ~/src/ipywidgets/widgetsnbextension (master ✗)$ npm install file:../jupyter-js-widgets                                                                         ipywidgets-dev

> jupyter-js-widgets@2.1.4 prepublish /Users/maartenbreddels/src/ipywidgets/jupyter-js-widgets
> npm run build


> jupyter-js-widgets@2.1.4 build /Users/maartenbreddels/src/ipywidgets/jupyter-js-widgets
> npm run build:src && npm run build:src-embed && npm run build:css && webpack


> jupyter-js-widgets@2.1.4 build:src /Users/maartenbreddels/src/ipywidgets/jupyter-js-widgets
> tsc --project src

src/manager-base.ts(4,20): error TS2307: Cannot find module 'underscore'.
src/services-shim.ts(10,20): error TS2307: Cannot find module 'underscore'.
src/widget.ts(6,20): error TS2307: Cannot find module 'underscore'.
src/widget_bool.ts(16,20): error TS2307: Cannot find module 'underscore'.
src/widget_box.ts(33,20): error TS2307: Cannot find module 'underscore'.
src/widget_button.ts(20,20): error TS2307: Cannot find module 'underscore'.
src/widget_color.ts(12,20): error TS2307: Cannot find module 'underscore'.
src/widget_controller.ts(12,20): error TS2307: Cannot find module 'underscore'.
src/widget_core.ts(11,20): error TS2307: Cannot find module 'underscore'.
[...]

PR #1190, opened 2017-03-07
new: repeat button for Play widget, can be hidden

maartenbreddels:
As discussed in #1186 this adds a repeat button. I've chosed the fa-retweet button, since the fa-repeat does not look like a repeat button.
Futhermore I've included what I think is a fix for Play, it now *includes* max, since that is more consistent with the slider's max behavious.

```python
import ipywidgets
p = ipywidgets.Play(min=0, max=3, interval=500, show_repeat=True)
s = ipywidgets.IntSlider(min=p.min, max=p.max, value=p.value)
ipywidgets.jslink((p, 'value'), (s, 'value'))
ipywidgets.HBox([p, s])
```

hide repeat button, and turn on repeat:
```python
p.show_repeat = False
p._repeat = True
```

### DIFF
--- a/jupyter-js-widgets/src/widget_int.ts
+++ b/jupyter-js-widgets/src/widget_int.ts
@@ -725,6 +725,8 @@ class PlayModel extends BoundedIntModel {
             _playing: false,
             interval: 100,
             step: 1,
+            _repeat: false,
+            show_repeat: true,
         });
     }
     initialize(attributes, options) {
@@ -739,12 +741,16 @@ class PlayModel extends BoundedIntModel {
     loop() {
         if (this.get('_playing')) {
             var next_value = this.get('value') + this.get('step');
-            if (next_value < this.get('max')) {
+            if (next_value <= this.get('max')) {
                 this.set('value', next_value);
                 window.setTimeout(this.loop.bind(this), this.get('interval'));
             } else {
                 this.set('value', this.get('min'));
-                this.set('_playing', false);
+                if(this.get('_repeat')) {
+                    window.setTimeout(this.loop.bind(this), this.get('interval'));
+                } else {
+                    this.set('_playing', false);
+                }
             }
             this.save_changes();
         }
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Awesome!

maartenbreddels:
> Futhermore I've included what I think is a fix for Play, it now includes max, since that is more consistent with the slider's max behavious.
Is it an idea to take this in a seperate PR, since it will be nice to have this consistent for 7.0, and leave the extra button for 7.x or higher.
Since now Play goes from min to max-1, while a slider goes from min to max (inclusive), having this consistent makes sense.

jasongrout:
I think either way makes sense.

maartenbreddels:
So, do you mind if I merge this then, I've tested it myself. Should the change in behaviour of Play go in some changelog?

jupyter-js-widgets/src/widget_int.ts
```diff
@@ -660,12 +662,16 @@ class PlayModel extends BoundedIntModel {
     loop() {
         if (this.get('_playing')) {
             var next_value = this.get('value') + this.get('step');
-            if (next_value < this.get('max')) {
+            if (next_value <= this.get('max')) {
                 this.set('value', next_value);
                 window.setTimeout(this.loop.bind(this), this.get('interval'));
             } else {
                 this.set('value', this.get('min'));
```
jasongrout: Can we only reset it to the min if we are repeating? Reverting to the min after a straight play through has been frustrating to me, instead of just stopping at the end and freezing the state there.

jasongrout:
Yes, there is a changelog at https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/changelog.md

jupyter-js-widgets/src/widget_int.ts
maartenbreddels: I agree!

maartenbreddels:
$#@% messed up rebasing, any idea what I did? :)

jasongrout:
Looks like you figured it out.

maartenbreddels:
Yes, I had to `git push --force ...`
Sorry I had to dismiss your review :), but I didn't see any other option.

PR #1192, opened 2017-03-07
Fix Item ordering in 'Select*', 'ToggleButtons' and 'RadioButtons' widgets

### LINKED ISSUE #1191: Ordering of elements in 'Select*', 'ToggleButtons' and 'RadioButtons' is inconsistent when updating 'options' on 5.x
On the 5.x branch, when updating the `options` of any of the above-mentioned widgets the option ordering in the DOM is inconsistent with the option ordering in the model. I would expect that if
`widget.options = ['d', 'a', 'c']` in the notebook, then the options presented in the DOM should be displayed in the same order.

This also gives rise to a weird effect in the `Select*` widgets where selecting option `a` causes option `c` to be selected.

See this gist [over here](https://gist.github.com/jbweston/c9d751d6a109ef2613da5db955be668e) to see what I mean

This bug is not present on `master`.

jbweston:
Fixes #1191.

I'm not sure if you are still accepting pull requests for bugs on the 5.x branch, but given jupyter-incubator/dashboards_server#317 I still need ipywidgets 5.x, so thought I would make this change available upstream.

I essentially change the logic to be more inline with that on the current `master` branch ([example](https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_selection.ts#L87)),
where instead of updating the DOM with elements and then deleting any which are no longer present, we instead rebuild the list of DOM elements from scratch.

Now elements in the DOM respect the ordering of elements in the `options` of the model.

### DIFF
--- a/jupyter-js-widgets/src/widget_selection.js
+++ b/jupyter-js-widgets/src/widget_selection.js
@@ -331,7 +331,9 @@ var ToggleButtonsView = widget.DOMWidgetView.extend({
          * changed by another view or by a state update from the back-end.
          */
         if (options === undefined || options.updated_view != this) {
-            // Add missing items to the DOM.
+            // Remove all options from the DOM and re-insert, to preserve
+            // ordering with respect to 'model._options_labels'.
+            this.$buttongroup.text('')
             var items = this.model.get('_options_labels');
             var icons = this.model.get('icons');
             var previous_icons = this.model.previous('icons') || [];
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks! I'm not sure if we'll be releasing another 5.x release, but this will be in it if we do.

jasongrout:
Since we now are on ipywidgets version 7, we don't plan any more 5.x releases. Thanks again for your work on this.

PR #1194, opened 2017-03-08
Any descendant of the state can be a binary_type

maartenbreddels:
Before you could only do sth like this when serializing a state:
```python
def array_to_binary(ar, obj=None):
    ar = ar.astype(np.float64)
    return memoryview(ar)
```
However, you'd like to be able to send shape along. This patch makes this possible, e.g.:
```python
def array_to_binary(ar, obj=None):
	ar = ar.astype(np.float64)
	return {'data': memoryview(ar), 'shape': ar.shape}
```
To be able to do this, state is seperated in `state_without_buffers` (state on the js side), and a `state_with_bufffers`, where every dict items that is a buffer is pop'ed, and every list item replaced by None. If `state_with_bufffers` was not treated seperately, the state would first be deserialized completely, and later on the buffers were 'patched in`. By keeping the `state_with_bufffers` seperate, is will first put in all the buffers in place, and deserialize then afterwards.
Maybe it's not the cleanest solution, but my understanding of the code only goes this far.
This does not treat serialization (frontend -> backend), before continuing, I'd like to get 'approval' of this approach.


maartenbreddels:
Oh, and respect for the ones responsible for thinking out the method on doing binary transfers, this is truly a good idea!

jasongrout:
Why not just save the entire array+size+dtype into a binary format? For example, we could implement this: https://docs.scipy.org/doc/numpy/neps/npy-format.html#format-specification-version-1-0

maartenbreddels:
That would be an option, but I also need serializing a dict of numpy arrays in the future, so even if you could solve some options with what you propose, this is more flexible. And I guess that requires a memcpy (although it's probably in the order of megabytes, so it shouldn't be an issue).

SylvainCorlay:
Regardless of the method used for the case of numpy arrays, I think that allowing binary *fields* will be necessary for other use cases.

maartenbreddels:
Yes, also other metadata, units, descriptions etc, you don't want to put it in binary form on each side of the wire, this would make life much easier.

jasongrout:
I hesitate to start doing things that involve picking apart objects and replacing their innards. Are we replacing things arbitrarily deep in the nesting here?

SylvainCorlay:
We can probably address all possible use cases by allowing only top-level binary keys.

maartenbreddels:
Yes, any level deep, as far as lists and dicts go. We could not allow that, but that doesn't change the code at all (except for the recursive part). It is not any different than `_widget_to_json`.

maartenbreddels:
It will actually require hardcoded checks, so as it is now, it's the easiest to maintain I think.

maartenbreddels:
I think something that makes this complex (and it was already the case before this PR), is that first an object get created/deserialized here: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/manager-base.ts#L346 
And then afterwards, the buffer items gets pushed in here: https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget.ts#L175
 and the rest gets deserialized.
Maybe if this can be done differently, like the deserialize part not being called at two places, the state wouldn't required to be split in two parts. Then buffer items could be popped, and list items set to null, and later filled in.

Because now, if say my numpy array 'x' (with the missing buffer value on the js side), get deserialized, and then later the buffer gets filled in, and then the deserialize isn't getting called anymore.

jasongrout:
> We can probably address all possible use cases by allowing only top-level binary keys.

This is essentially what we have now, one level up at the object's state keys. Some keys can be binary, some can be JSON.

What would we need to isolate this to a custom serializer (like the widgets serializer)? For example, what if we allowed multiple buffers to be associated with one key (to have possible zero-copy semantics).

(I'm not totally against this idea, but it certainly makes the serialization more complicated, so I want to explore some simpler ideas, or ways to move the complexity out of the base codebase)

maartenbreddels:
Actually, that is also what @SylvainCorlay thought, but it is more like 0 levels deep.

`state` is a dict with all values, lets say we have a widget with a property 'x', then (igoring all the other meta data), state would be `state = {'x': memoryview(ar)}`, and that is allowed and get properly treated in [_split_state_buffers](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget.py#L261) and on the js side. 
However, we'd like to allow at least `state = {'x': {'data': memoryview(ar), 'shape': ar.shape, ....}}`, so it's an extra level deeper we need to go. And well, 1 level or N levels, it's all the same code.

maartenbreddels:
I agree that it makes it more complex, but I think the complexity start at the two places where the deserialization is happening. If, say at object construction we would already have the buffers available, that could make this simpler, we just pop buffers out of the dicts, and replace lists values with sentinels, and put them back in place directly before the object is constructed. Now the object is first constructed without anything that has a buffer in it missing (before it was only level 0 keys missing that had buffer object), and then an update is adding in the missing buffer objects.
Ah.... now that I am writing this, I understand where the issue is:
[Widget.open](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget.py#L213) first constructs the object without buffers, and then sends the state again. So my guess is, when this issue is fixed, it can at least be a bit simpler (as in not having state and state_with_buffers, but just 1 state)

SylvainCorlay:
@jasongrout the current mechanism allows top-level properties of the *object* to by memory views, not top-level properties of a given trait attribute.

jasongrout:
> @jasongrout the current mechanism allows top-level properties of the object to by memory views, not top-level properties of a given trait attribute.

Right - any particular trait (i.e., a key/value in the object's state) can be binary.

jasongrout:
> However, we'd like to allow at least state = {'x': {'data': memoryview(ar), 'shape': ar.shape, ....}}, so it's an extra level deeper we need to go. And well, 1 level or N levels, it's all the same code.

Yes, I understand what we're trying to do. Another way to think about this proposal is that we are simply extending our base JSON encoder/decoder to handle binary objects anywhere in it. (The popping of binary buffers happens *after* the object's custom serializer has run, right?)

What if we allowed any particular trait to have arbitrary JSON metadata associated with it? We sort of have that now - if a trait is binary, some metadata is associated with it indicating which buffer the data is in. What if the serializer could return both a value and a metadata dict? And the deserializer could take both a data value and a metadata dict.

Also, the difference between what we have now (any trait can be encoded in binary) and a level-1 binary replacer is that the serialization/deserialization happens at the trait level, so with the level-1 binary replacer, a single trait can (very easily) be encoded as both binary data and JSON data, whereas with what we have now, there needs to be some extra magic going on to associate the JSON data with the binary data in two different traits.

maartenbreddels:
I like the idea, but I can think of an example which doesn't really fit into this idea. Say a masked array, it  would have two binary blobs associated with it, and I wouldn't call the array with the mask the meta data. It would make more sense to write it like this.
```python
   ...
   return {'data': memoryview(data), 'mask': memoryview(mask), 'shape' data.shape}
```

jasongrout:
Yes, I'm agreeing with this PR's idea more and more, especially as I think about it as being just a way of augmenting our JSON format to take advantage of the binary buffers we have available.

maartenbreddels:
Ok, good to know. The split `state` and `state_with_buffers` can be avoided if the issue in `Widget.open` is fixed, but it's not a lot of code. I'll comment the code on what can be changed if this issue is fixed. So I guess the next step would be to implement serialize (from js to py)?

jasongrout:
Can you show an example of what a deeper nesting path would look like? a state dictionary of `{'x': {'arr': [ar, ar]}}` would give `[x, 'arr', 0], [x, 'arr', 1]` as paths?

Implementing the serialization/deserialization in js definitely should be done too.

Thanks for tackling this! It's been a frustration we've had for a while (particularly with numpy arrays).

maartenbreddels:
Ok, implemented, demo notebook here: https://gist.github.com/maartenbreddels/ac41cdcd22c8f84c9fc639dea7c90848
it's a model that takes x, and doubles every items, put it to y. As an extra check i also included a 'nested' property:
```python
{'data': mv, 'shape': ar.shape, 'nested': {'list': [mv, 'test']}}
```
that gets passed along from x to y, just to see demonstate it handles nested items ok, in the last cell you see it is property serialize and deserialized.

maartenbreddels:
I've answered your question in the docstring of _split_state_buffers. Futhermore, (in a previous commit) I've renamed buffer_keys by buffer_paths, since it better covers its usage.
Comments on the js/ts are welcome, since I'm not so comforable with it as I am with Python.

I also tried to keep the code simple (like a for loop in _handle_msg, instead of using reduce), to keep the code more readable for those who are not used to functional parts, but I don't know what you guidelines are on this.

maartenbreddels:
FYI: Found some bugs in this PR, will commit new changes later.

SylvainCorlay:
@maartenbreddels are you still making changes to this PR?

SylvainCorlay:
Quick note, the `_split_state_buffers` could may be changed into a free function.

maartenbreddels:
When I used controller, I noticed a few things, first 'circular references' in objects caused endless loops. Second, when object state were modified, it could modify the model's data, so we needed cloning. Then in noticed, that for instance when the controller's model is serialized, is serialized a lot, for instance I noticed the WidgetManager was serialized with all it's promises. Cloning that didn't work, so I only clone when needed (seems better anyway). Then on the Python side, a small issue with deleting a key twice. Seeing if I can add a unittest for that now.
And yes, _split_state_buffers should go outside the class.

ipywidgets/widgets/widget.py
```diff
@@ -302,8 +302,9 @@ def seperate_buffers(substate, path, buffer_paths, buffers):
         # any part of the state that has buffers needs to be treated seperately
         # since of a issue as indicated in .open(..)
         for key in set([k[0] for k in buffer_paths]):
-            state_with_buffers[key] = state[key]
-            del state[key]
+            if key in state: # if we didn't remove it allready
+                state_with_buffers[key] = state[key]
+                del state[key]
```
maartenbreddels: actually, i see it's a set now.. so I have no ideas why the del statement every failed

maartenbreddels:
Ok, this is it. Let me know if the js/ts code is acceptable.

maartenbreddels:
I managed to trigger a bug again, added the unittest and fix. 

To let you know about the performance, on my laptop, I can transfer 1e7 x, y, z float values, that is 1e7*4*4byte=114MB, without any issues (may take a few second at most). At my desktop, I managed to do a factor of 10 more (1e8!). So transfering a gig of data to the browser was possible (in maybe more time, but it was possible).

jasongrout:
Nice! How does this compare with base64-encoding the data?

maartenbreddels:
@jeammimi can you comment on this? I haven't worked with base64 encoding, but compared to JSON i'd say a factor of ~60 (~minute vs ~second). Although #1195 may have caused a factor 2 slowdown in any of the benchmarkts. But at least >10x.

ipywidgets/widgets/widget.py
```diff
@@ -50,6 +50,55 @@ def _json_to_widget(x, obj):
...
+
+    As an example:
+    >>> state = {'plain': [0, 'text'], 'x': {'ar': memoryview(ar1)}, 'y': {'shape': (10,10), 'data': memoryview(ar2)}}
+    >>> widget._split_state_buffers(state)
+    ({'plain': [0, 'text']}, {'x': {}, 'y': {'shape': (10, 10)}}, [['x', 'ar'], ['y', 'data']],
+     [<memory at 0x107ffec48>, <memory at 0x107ffed08>])
+    """
+
+    def seperate_buffers(substate, path, buffer_paths, buffers):
+        # remove binary types from dicts and lists, and keep there key, e.g. {'x': {'ar': ar}, 'y': [ar2, ar3]}
```
vidartf: "keep there key" -> "keep their keys" ?

ipywidgets/widgets/widget.py
```diff
@@ -223,6 +272,10 @@ def open(self):
             if buffers:
                 # FIXME: workaround ipykernel missing binary message support in open-on-init
                 # send state with binary elements as second message
+                # TODO: if this gets fixed, _split_state_buffers does not need to have a seperate
```
jasongrout: Have you tried it? It seems that Comm supports buffers in the init: https://github.com/ipython/ipykernel/blob/master/ipykernel/comm/comm.py#L50

ipywidgets/widgets/widget.py
```diff
@@ -50,6 +50,55 @@ def _json_to_widget(x, obj):
...
+    state_with_buffers is a dict where any of it's decendents is is a binary_type.
+
+    As an example:
+    >>> state = {'plain': [0, 'text'], 'x': {'ar': memoryview(ar1)}, 'y': {'shape': (10,10), 'data': memoryview(ar2)}}
+    >>> widget._split_state_buffers(state)
+    ({'plain': [0, 'text']}, {'x': {}, 'y': {'shape': (10, 10)}}, [['x', 'ar'], ['y', 'data']],
+     [<memory at 0x107ffec48>, <memory at 0x107ffed08>])
+    """
+
+    def seperate_buffers(substate, path, buffer_paths, buffers):
```
jasongrout: sep*a*rate - spelled with an 'a' in the middle.

ipywidgets/widgets/widget.py
maartenbreddels: I did not, but I will try

jasongrout: Looks like it was added about 1.5 years ago: https://github.com/ipython/ipykernel/commit/64c69213be7629bbf8735a7fabdb4e47977775c9

jasongrout: Cool, the GitHub UI shows which tags the commit is part of - shows that the fix is in ipykernel 4.2 and later...

ipywidgets/widgets/tests/test_traits.py
```diff
@@ -19,3 +21,19 @@ class TestColor(TraitTestBase):
     _good_values = ["blue", "#AA0", "#FFFFFF"]
     _bad_values = ["vanilla", "blues"]
 
+
+class TestBuffers(TestCase):
+    def test_state_with_buffers(self):
+        mv1 =  memoryview(b'test1')
+        mv2 =  memoryview(b'test2')
+        state = {'plain': [0, 'text'], 'x': {'ar': mv1}, 'y': {'shape': (10, 10), 'data': mv1}, 'z': [mv1, mv2], 'top': mv1}
```
jasongrout: Can we add one more item that is nested several levels deep? Perhaps something like:

`'nested': {'a': mv1, 'b': [mv2, {'c': mv1, 'd': [mv1, mv2]}]}`

ipywidgets/widgets/widget.py
maartenbreddels: Nice, setup.py requires > 4.5 as well as the conda-forge recipe

jupyter-js-widgets/src/widget.ts
```diff
@@ -399,23 +406,82 @@ class WidgetModel extends Backbone.Model {
...
+            // on the python side the inverse happens
+            // if we need to remove an object from a list, we need to clone that list, otherwise we may modify
+            // the internal state of the widget model
+            // however, we do not want to clone everything, for performance
+            function seperate_buffers(obj, path) {
+                if(visited_set.indexOf(obj) != -1) { // we already visited this object
+                    return obj;
+                }
+                visited_set.push(obj);
+                if (_.isArray(obj)) {
```
jasongrout: I think all of our supported browsers support [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

jupyter-js-widgets/src/widget.ts
```diff
@@ -399,23 +406,82 @@ class WidgetModel extends Backbone.Model {
...
+                                        is_cloned = true;
+                                    }
+                                    buffers.push(value);
+                                    buffer_paths.push(path.concat([key]));
+                                    delete obj[key]; // for objects/dicts we just delete them
+                                }
+                                else {
+                                    var new_value  = seperate_buffers(value, path.concat([key]));
+                                    if((new_value != value) && !is_cloned) { // only clone when we have to
+                                        obj = _.mapObject(obj, _.identity);
```
jasongrout: `_.extend(obj, {})` would be a more familiar idiom to widget developers.

maartenbreddels: Ah thanks, actually, would [_.clone](http://underscorejs.org/#clone) not be better?

jasongrout: I almost wrote that, but in this case, we know we're not a primitive or an array, right? `_.clone` degenerates to `_.extend` after it checks for other types.

ipywidgets/widgets/widget.py
maartenbreddels: It is less trivial then I thought, this means that the buffers will need to finally finds it's way [here](https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/manager-base.ts#L284) right?

jasongrout: and possibly [here](https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/manager-base.ts#L202). Should we merge your PR and then simplify things in another PR?

jasongrout: well, perhaps we should have a layer in front of both of those functions, right after we get the comm message off the wire, that puts back in the binary buffers, and then we're ready to go.

jasongrout: Perhaps we just insert the binary deserialization at https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/manager-base.ts#L187

jasongrout: (and a few other places, basically we keep the serialized state as binary until *just* before we send it, and we deserialize the binary state *just* after we get the message.)

maartenbreddels: sounds like a good plan, I'll give it a go

jupyter-js-widgets/src/widget.ts
maartenbreddels: I think it reads better.

jasongrout:
We should also probably address the case where we are saving state in the notebook - perhaps by base64-encoding the binary buffers for that.

maartenbreddels:
indeed, i encountered that issue already... i think the strategy can be the same, i'll make the function that extract buffers and put them together in units.ts, so we can reuse them for that.

ipywidgets/widgets/widget.py
jasongrout: Basically, we think of the binary serialization as just a detail of the actual wire format (which it is), so we keep those details as close to the wire as we can.

maartenbreddels:
The only worry I have is that I modify dicts and lists in `_split_state_buffers`, maybe the same strategy is needed as in the .ts code, but I doubt it.

jasongrout:
You'll probably need to do the same thing, right? Why would you not have to modify dicts/lists?

maartenbreddels:
I assume you'd return from a `to_json` method a new dict or list, it doesn't have to do anything with the state of the widget.

jasongrout:
Wow, that ipykernel fix led to this code being cleaned up really nicely!

jasongrout:
> I assume you'd return from a to_json method a new dict or list, it doesn't have to do anything with the state of the widget.

to_json should return a copy of whatever data it is trying to serialize, but that data now might have binary buffers, right? We still need to modify what to_json returns to strip out binary arrays and put it on the wire.

jasongrout:
> Wow, that ipykernel fix led to this code being cleaned up really nicely!

(not just this code, but also the old code!)

maartenbreddels:
Yes, what to_json returns can get modified, maybe we should put that in the docs? Or should I fix it?

jasongrout:
I think it's fine to modify what to_json returns. We're just reformatting it for the wire, and we'll put it right back together on the other side before the corresponding from_json sees it.

jasongrout:
(by the way, once we have this we can change the Image widget to use binary instead of base64 encoding)

jasongrout:
> (by the way, once we have this we can change the Image widget to use binary instead of base64 encoding)

(Well, I guess we could have done it before too, since the image data is a single top-level field.)

jasongrout:
Please let me know when you're to the point to have someone look at it again.

maartenbreddels:
I thought I was 'done'? :)

ipywidgets/widgets/widget.py
```diff
@@ -50,6 +50,47 @@ def _json_to_widget(x, obj):
...
+                if isinstance(v, _binary_types):
+                    substate[i] = None
+                    buffers.append(v)
+                    buffer_paths.append(path + [i])
+        elif isinstance(substate, dict):
+            for k, v in list(substate.items()):  # we need to copy to a list since substate will be modified
+                if isinstance(v, (dict, list, tuple)):
+                    seperate_buffers(v, path + [k], buffer_paths, buffers)
+                if isinstance(v, _binary_types):
+                    substate.pop(k)
```
jasongrout: `del substate[k]` is faster

ipywidgets/widgets/widget.py
```diff
@@ -50,6 +50,47 @@ def _json_to_widget(x, obj):
...
+        if isinstance(substate, (list, tuple)):
+            for i, v in enumerate(substate):
+                if isinstance(v, (dict, list, tuple)):
+                    seperate_buffers(v, path + [i], buffer_paths, buffers)
+                if isinstance(v, _binary_types):
+                    substate[i] = None
+                    buffers.append(v)
+                    buffer_paths.append(path + [i])
+        elif isinstance(substate, dict):
+            for k, v in list(substate.items()):  # we need to copy to a list since substate will be modified
```
jasongrout: Running a few timing tests, it seems that, for example, this

```python
%%timeit
d = {1:1, 2:2, 3:3, 4:4, 5:5, 6:6, 7:7, 8:8, 9:9, 10:10,11:11,12:12,13:13,14:14,15:15}
for k,v in list(d.items()):
    if k % 100 == 0:
        del d[k]
```
is about 30% slower (after subtracting the overhead for creating the dictionary `d`) than
```python
%%timeit
d = {1:1, 2:2, 3:3, 4:4, 5:5, 6:6, 7:7, 8:8, 9:9, 10:10,11:11,12:12,13:13,14:14,15:15}
keys = []
for k,v in d.items():
    if k % 100 == 0:
        keys.append(k)
    pass
for k in keys:
    del d[k]
```

jupyter-js-widgets/src/manager-base.ts
```diff
@@ -185,12 +185,14 @@ abstract class ManagerBase<T> {
      * Handle when a comm is opened.
      */
     handle_comm_open(comm: shims.services.Comm, msg: services.KernelMessage.ICommOpenMsg): Promise<Backbone.Model> {
+        var data =  (msg.content.data as any);
```
jasongrout: delete extra space after =

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,108 @@ function escape_html(text: string): string {
...
 };
+
+/**
+ * Takes an object 'state' and fills in buffer[i] at 'path' buffer_paths[i]
+ * where buffer_paths[i] is a list indicating where in the object buffer[i] should
+ * be placed
+ * Example: state = {a: 1, b: {}, c: [0, null]}
+ * buffers = [array1, array2]
+ * buffer_paths = [['b', 'data'], ['c', 1]]
+ * Will lead to {a: 1, b: array1, c: [0, array2]}
```
jasongrout: `b: {data: array1}`

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,108 @@ function escape_html(text: string): string {
...
+ */
+export
+function put_buffers(state, buffer_paths, buffers) {
+    for (var i=0; i<buffer_paths.length; i++) {
+         // say we want to set state[x][y[z] = buffers[i]
```
jasongrout: `[x][y][z]`

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,108 @@ function escape_html(text: string): string {
...
+        var obj = state;
```
jasongrout: `let obj = state;`

jasongrout: Also, pull `buffer_paths[i]` out of the inner loop, for example, `let buffer_path = buffer_paths[i];` and then use `buffer_path` everywhere below.

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,108 @@ function escape_html(text: string): string {
...
+        for (var j = 0; j < buffer_paths[i].length-1; j++)
+            obj = obj[buffer_paths[i][j]];
+        // and then set: obj[z] = buffers[i]
+        obj[buffer_paths[i][buffer_paths[i].length-1]] = buffers[i];
+    }
+}
+
+/**
+ * The inverse of put_buffers, return an objects with the new state where all buffers(ArrayBuffer)
+ * are removed. If a buffer is a member of an object, that object is cloned, and the key removed. If a buffer
```
jasongrout: Just like on the python side, I think we should assume that the objects and lists given to us at this stage are already copies, and we can mutate them however we wish to get them onto the wire.

jasongrout: Furthermore, I think it's okay to assume that we don't have cyclic references. This translation step should really essentially be the last step before using JSON.stringify. I think these two assumptions will greatly simplify the code here. Also, if we don't have to clone, I think we can delete the underscore import at the top.

jupyter-js-widgets/src/widget.ts
```diff
@@ -398,25 +397,15 @@ class WidgetModel extends Backbone.Model {
...
-                    if (value.buffer instanceof ArrayBuffer
-                        || value instanceof ArrayBuffer) {
-                        buffers.push(value);
-                        buffer_keys.push(key);
-                        delete state[key];
-                    }
-                }
-            }
+            // this function goes through lists and object and removes arraybuffers
+            // they will be transferred seperately, since they don't json'ify
```
jasongrout: sep**a**rately

ipywidgets/widgets/widget.py
```diff
@@ -50,6 +50,47 @@ def _json_to_widget(x, obj):
...
+        # remove binary types from dicts and lists, but keep track of their paths
+        # e.g. {'x': {'ar': ar}, 'y': [ar2, ar3]}, where ar/ar2/ar3 are binary types
+        # will result in {'x': {}, 'y': [None, None]}, [ar, ar2, ar3], [['x', 'ar'], ['y', 0], ['y', 1]]
+        # instead of removing elements from the list, this will make replacing the buffers on the js side much easier
+        if isinstance(substate, (list, tuple)):
+            for i, v in enumerate(substate):
+                if isinstance(v, (dict, list, tuple)):
+                    seperate_buffers(v, path + [i], buffer_paths, buffers)
+                if isinstance(v, _binary_types):
+                    substate[i] = None
```
maartenbreddels: The tuple thing hit me when I was overthinking this in bed... good you catched it :)
I made it reflect the ts/js version more, clone only when needed. It's also reflected in the unittests, if we don't do this, someone will be bit by us modifying their data (see also comments below)

ipywidgets/widgets/widget.py
maartenbreddels: This code will not be the bottleneck, and I think that deleting afterwards makes the code more difficults to read, let me know what you think of the (changed) code.

jupyter-js-widgets/src/utils.ts
maartenbreddels: Thanks for the let/var, learned something new there. And yes, I agree, much cleaner.

jupyter-js-widgets/src/utils.ts
maartenbreddels: Both I disagree with. I've been bit by modifying the data. Say you serialize (from py side), and array like {'data': memoryview(ar), shape: ar.shape}. And on the JS side, you work with this as an object. Then you do not want this js code to delete the data attribute.
Cyclic references exists :( If you don't take this into account, the Controller will trigger this.

ipywidgets/widgets/widget.py
jasongrout: I was also worried about the memory pressure of constantly creating new lists. You're probably right that it won't be the bottleneck.

jupyter-js-widgets/src/utils.ts
jasongrout: I think the serialization step should make a copy of the object being sent (otherwise it's not sending a *snapshot* of the current data). So I think what we're working with here should be our own copy.

The Controller widget has cyclic references in the JSON structure that is stringified? Won't that cause an error in the stringification? That sounds like a problem.

maartenbreddels: Yes, i even saw the widgetmanager with the promises being serialized, didnt look into it further though. Only saw that with Controller.

ipywidgets/widgets/widget.py
jasongrout: > let me know what you think of the (changed) code.

Sure, soon as you push the changes.

jupyter-js-widgets/src/utils.ts
maartenbreddels: I was surprised to see Promises being stringified as well.

maartenbreddels:
Ok, it's more consistent now. Both python and javascript sides have a (\_)remove_buffers/(\_)put_buffers. remove buffers doesn't modify existing dicts/objects (clones when needed), but references non-modifies ancestors. The js side has cyclic protection (needed to support Controller), the py side does not. _put_buffers does modify the state, but that should be fine since it coming off the wire.
I must say, the python unittest gives me a high level of trust in the code now, the fact that the ts/js code is practically the same makes me feel comfortable with it as well.

jasongrout:
I think I figured out why you needed the circular dependency check on the js side. Basically, we use Javascript's toJSON feature (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior) to serialize models, so the models are *not* serialized to "IPY_MODEL_..." strings until *after* the remove_buffers runs, so remove_buffers recurses into these complicated model objects. Looking into it...

maartenbreddels:
Yes, I was surprised to see unpack_models, but not the inverse. But I think if the toJSON can handle it, we should be able to handle it as well.

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,115 @@ function escape_html(text: string): string {
...
+                                is_cloned = true;
+                            }
+                            buffers.push(value);
+                            buffer_paths.push(path.concat([key]));
+                            delete obj[key]; // for objects/dicts we just delete them
+                        }
+                        else {
+                            var new_value  = remove(value, path.concat([key]));
+                            // only assigned when the value changes, we may serialize objects that don't support assignment
+                            if(new_value != value) {
```
jasongrout: `!==`

jupyter-js-widgets/src/utils.ts
```diff
@@ -169,3 +171,115 @@ function escape_html(text: string): string {
...
+                        buffers.push(value);
+                        buffer_paths.push(path.concat([i]));
+                        // easier to just keep the array, but clear the entry, otherwise we have to think
+                        // about array length, much easier this way
+                        obj[i] = null;
+                    }
+                    else {
+                        var new_value  = remove(value, path.concat([i]));
+                        // only assigned when the value changes, we may serialize objects that don't support assignment
+                        if(new_value != value) {
```
jasongrout: `!==`

maartenbreddels:
I agree, good you spotted those bugs in the unittest.

jasongrout:
@maartenbreddels - can you try this out? I think it should work, even with Controller...

maartenbreddels:
With the toJSON the Controller should work fine, happy to see it get merged! Great work.

jasongrout:
You did the great work here! Thanks!

maartenbreddels:
Thanks for the careful and thorough review :+1:

PR #1199, opened 2017-03-15
Clarify in docs that interactive does not display the result.

### LINKED ISSUE #1198: widgets returned by the `interactive` function no longer displays result
With the 6.0.0 update, the widget objects returned by the `interactive` function no longer displays the result of the function that is bound to the input control when display(widget) is called. ie

```python
widget = interactive(func, arg1=val1, arg2=val2)
display(widget)
```

According to the [docs](http://ipywidgets.readthedocs.io/en/latest/examples/Using%20Interact.html#interactive), once `display(widget)` is  called, it should behave as if the `interact` function had been called, but by not actually displaying the return value of func, it is not doing this.

Possibly related to this commit? #23c5

jasongrout:
Fixes #1198.

### DIFF
--- a/docs/source/examples/Using Interact.ipynb
+++ b/docs/source/examples/Using Interact.ipynb
@@ -899,7 +899,7 @@
     "editable": true
    },
    "source": [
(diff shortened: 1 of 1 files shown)


jasongrout:
@ned2 - does this look better? Do you have a suggestion of how else we can make this clearer?

PR #1223, opened 2017-03-22
Removed redundant code

FermiParadox:
The code is not being used in the respective codeblock. I checked both with, and without that "extra" code, and the result seems to be the same, but since I'm not very familiar with ipywidgets, please make sure i didn't remove something needed.


jasongrout:
Looks good to me. Thanks!

PR #1226, opened 2017-03-23
github.com/ipython/ipywidgets -> github.com/jupyter-widgets/ipywidgets

jasongrout:
Reflects the move of ipywidgets from the ipython org to the jupyter-widgets org.


SylvainCorlay:
Merging, but it seems that we need to activate CI services in this org.

jasongrout:
I was just working on that before merging! :)

PR #1228, opened 2017-03-24
Start to unify the comm_open messages in the spec.

jasongrout:
This will address the last part of https://github.com/jupyter-widgets/ipywidgets/issues/1212


jasongrout:
This is ready to go now, I think.

PR #1232, opened 2017-03-28
added _view_count trait to Widget, to count the number of views

maartenbreddels:
added _view_count trait to Widget, to count the number of views in the backend, as discussed here: https://github.com/jupyter-widgets/ipywidgets/issues/1225

There seems to be an issue with the Output widget, it only sends it state once, subsequent calls to `OutputModel.save()` or `OutputModel.save_changes()` do nothing. I'm not sure what is so different about OutputModel.

I also noticed `WidgetView.remove()` was called twice, I'll submit an issue for that, I've worked around that by using `.once(..)`.


ipywidgets/widgets/widget.py
```diff
@@ -247,6 +247,8 @@ def get_view_spec(self):
         help="Name of the view object.").tag(sync=True)
     _view_module_version = Unicode('*',
         help="A semver requirement for the view module version.").tag(sync=True)
+    _view_count = Int(read_only=True,
+        help="The number of views accociated to the model at the front-end").tag(sync=True)
```
jasongrout: "associated". Also, can we indicate this is experimental for now? Maybe "EXPERIMENTAL: The number of views of the model displayed in the frontend. This attribute is experimental and may change or be removed in the future."

jupyter-js-widgets/src/widget.ts
```diff
@@ -623,6 +623,19 @@ abstract class WidgetView extends NativeView<WidgetModel> {
         this.listenTo(this.model, 'change', this.update);
 
         this.options = parameters.options;
+
+        this.once('remove', () => {
+            console.log("remove view count", this.model.get('_view_count'), this.model.get('_view_count') -1)
```
jasongrout: delete the console.log

jupyter-js-widgets/src/widget.ts
```diff
@@ -623,6 +623,19 @@ abstract class WidgetView extends NativeView<WidgetModel> {
...
+            this.model.set('_view_count', this.model.get('_view_count') - 1);
+            this.model.save_changes()
+        });
+
+        this.once('displayed', () => {
+            console.log("add view count", this.model.get('_view_count'), this.model.get('_view_count') +1)
```
jasongrout: delete the console.log

maartenbreddels:
I assume you are ok with merging this, I'll add those comments and remove the logs. It would take some debugging to see why Output is misbehaving though, but that bug/feature can't be in this code I think. I hope to update this soon.

jasongrout:
> I assume you are ok with merging this, I'll add those comments and remove the logs.

Yes, I'm okay merging this as an experimental feature.

jasongrout:
Thanks again for pushing forward on things!

jasongrout:
Thanks!

maartenbreddels:
Looking forward to the next release ! :)

PR #1257, opened 2017-04-05
Introduce a ToggleButtons button_width style attribute.

### LINKED ISSUE #1189: Scale size of the `ToggleButtons` to the text they contain
With the new `ipywidgets 6.0`, the size of the `ToggleButtons` does not scale anymore with the size of the text that they contain. Consider for example the case below:
screen shot 2017-03-07 at 8 10 32 am
Ideally, it would be nice to have the option for the button for "Super Super Super Super Slow" to automatically increase its width (I think this was the default with `ipywidgets 5`). Would it be possible to implement this?

jasongrout:
Fixes #1189 

screen shot 2017-04-05 at 3 10 41 pm

However, after doing this, I'm thinking we should do it a different way. For one, the assumption at https://github.com/jupyter-widgets/ipywidgets/blob/master/jupyter-js-widgets/src/widget.ts#L818 is now not true - the size of the widget could change. Two, perhaps a more useful default is to size the buttons with the content with some margin, with a minimum width so things don't get too small. Or use flex to layout the buttons in the space of the widget, so it's easy to control the total size of the buttons by changing the size of the buttons container.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -16,7 +16,9 @@ from itertools import chain
 from .widget_description import DescriptionWidget
 from .valuewidget import ValueWidget
 from .widget_core import CoreWidget
-from .widget import register
+from .widget_style import Style
+from .trait_types import InstanceDict
+from .widget import register, widget_serialization
 from traitlets import (Unicode, Bool, Int, Any, Dict, TraitError, CaselessStrEnum,
                        Tuple, List, Union, observe, validate)
 from ipython_genutils.py3compat import unicode_type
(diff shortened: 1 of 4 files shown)


jasongrout:
One question: is it more likely for people to want equal-sized buttons, or buttons sized to content?

maartenbreddels:
Sometimes you want an ugly-but-using-all-the-availiable-space solution :)

jasongrout:
which would be equivalent to setting the button's width to `flex: 1` (i.e., flex grow to fill space), right?

maartenbreddels:
I ment as small as possible, to use all the available space for widgets, would that also be possible?

jasongrout:
Okay, that was my question. Should we just change the default to 'size to content', but still provide an option to set a fixed size?

jasongrout:
ping @SylvainCorlay

jasongrout:
> which would be equivalent to setting the button's width to flex: 1 (i.e., flex grow to fill space), right?

Not sure what I was thinking there - we can't set the flex property in a width css attribute.

I'll punt this one to the beta.

jasongrout:
I updated this for the current state of the codebase.

jasongrout:
Let's try this out. The default is the same fixed width as before. Setting the `style.button_width` to `'initial'` makes each button size to its content.

replork:
My toggle button does not have a style.button_width property to set. Has this changed? How can I set the width of my toggle buttons?

PR #1259, opened 2017-04-05
Send a click event from ToggleButtons.

### LINKED ISSUE #763: wishlist: on_click() event for ToggleButtons
It would be useful if the `ToggleButtons` widget would have an `on_click()` method, just like the usual `Button`.

Use case: use this `on_click()` event to re-evaluate an interactive function. This makes sense when the `ToggleButtons` represent actions rather than a selection. These actions might be non-idempotent (clicking the button twice is not the same as clicking it once).

jasongrout:
Fixes #763.

To access this event from python, you can use something like:

```python
class MyToggleButtons(ToggleButtons):
    def __init__(self, **kwargs):
        super(MyToggleButtons, self).__init__(**kwargs)
        self.on_msg(self._handle_button_msg)
    def _handle_button_msg(self, _, content, buffers):
        print(content)
MyToggleButtons(options=list('abc'))
```

CC @jdemeyer - is this sufficient for your usecase?

### DIFF
--- a/jupyter-js-widgets/src/widget_selection.ts
+++ b/jupyter-js-widgets/src/widget_selection.ts
@@ -477,6 +477,9 @@ class ToggleButtonsView extends LabeledDOMWidgetView {
         var value = event.target.value;
         this.model.set('value', value, { updated_view: this });
         this.touch();
+        // We also send a clicked event, since the value is only set if it changed.
+        // See https://github.com/jupyter-widgets/ipywidgets/issues/763
+        this.send({event: 'click'});
     }
 
     private _css_state: any;


alexlenail:
It would be nice if the interface were the same

```
button = widgets.Button(description="Click Me!")
decision = ToggleButtons(options=['a', 'b'])

def on_button_clicked(b):
    print("Button clicked.")

button.on_click(on_button_clicked)
decision.on_click(on_button_clicked)

display(button)
display(decision)
```

> `AttributeError: 'ToggleButtons' object has no attribute 'on_click'`

PR #1262, opened 2017-04-06
Fix _SelectionModel and _MultipleSelectionModel to sync using indices

### LINKED ISSUE #880: Possibility to get key of selected items in Select(Multiple)
I want to add a button that delete selected elements from the option dictionary. To do it I need the selected key, not the value.
Would it be possible to add a `selected_key` or `selected_item` attribute to the select widget?

### LINKED ISSUE #1091: ToggleButtons / _Selection .value in Python side != .value on js side
Say we have
```
toggle = ipywidgets.ToggleButtons(options=[("One", 1), ("Two", 2)])
```
Then clicking on 'One' will reflect that toggle.value == 1.
However, on the js side, value is actually "One", which messes things up if you use jslink.
The problem is I think in SelectView._updateOptions (https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/src/widget_selection.ts#L90)
Which uses _options_labels for both the label and the data.
My guess is that on the Python side, both _options_labels and _options_values in ipywidgets._Selection need sync=True as well.
I unfortunately cannot get the dev-install.sh to work, otherwise I might attempt a PR.

Snippet to reproduce the bug:
```
import ipywidgets
toggle = ipywidgets.ToggleButtons(options=[("One", 1), ("Two", 2)])
label = ipywidgets.Label()
ipywidgets.jslink((label, 'value'), (toggle, 'value'))
ipywidgets.VBox([toggle, label])
```

version info: Python3 / ipywidgets (6, 0, 0, 'beta9')

jasongrout:
Fixes #1091

- [x] Fix javascript side
- [ ] Close #880 when this is merged.

CC @maartenbreddels

### DIFF
--- a/ipywidgets/widgets/tests/test_interaction.py
+++ b/ipywidgets/widgets/tests/test_interaction.py
@@ -12,6 +12,7 @@ except ImportError:
 
 import os
 import nose.tools as nt
+from collections import OrderedDict
 
 import ipywidgets as widgets
 
@@ -153,10 +154,10 @@ def test_list_str():
     d = dict(
         cls=widgets.Dropdown,
         value=first,
-        options=values,
         _options_labels=tuple(values),
         _options_values=tuple(values),
     )
+    d['options'] = tuple(zip(d['_options_labels'], d['_options_values']))
     check_widgets(c, lis=d)
 
 def test_list_int():
@@ -167,10 +168,10 @@ def test_list_int():
     d = dict(
         cls=widgets.Dropdown,
         value=first,
-        options=values,
-        _options_labels=("3", "1", "2"),
+        _options_labels=tuple(str(v) for v in values),
         _options_values=tuple(values),
     )
+    d['options'] = tuple(zip(d['_options_labels'], d['_options_values']))
     check_widgets(c, lis=d)
 
 def test_list_tuple():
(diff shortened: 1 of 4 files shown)


maartenbreddels:
If we sync label too, that would be useful in some cases, what do you think?

jasongrout:
@maartenbreddels, can you give this a try and see how it works?

> If we sync label too, that would be useful in some cases, what do you think?

Possibly, though I don't like having two possibly conflicting sources of truth in the protocol, and it becomes confusing on who sets it (frontend or backend? Both I think have a valid usecase).

jasongrout:
> If we sync label too, that would be useful in some cases, what do you think?

I'd much rather have the label derived from the index independently on the frontend and backend.

jasongrout:
Two design decisions:

(a) in this implementation, the default selection is *no* selection (except for the selection slider, which must have something selected

(b) We *always* coerce the options list to a `((label, value), (label, value)...)` tuple. I think before it kept whatever object was originally assigned. I like this better - it emphasizes the ordered aspect of the options. However, for convenience, we still allow (and even expand) possible input types to include any mapping or iterable.

Thoughts?

jasongrout:
> (a) in this implementation, the default selection is no selection (except for the selection slider, which must have something selected

I went ahead and changed this to always select the first item (if there is a first item) on initialization or when changing the options. That seems to work better with interact, for example, where you just want *some* value by default.

However, I left the multiple select to still unselect all items when changing the options or setting it initially. Thoughts?

jasongrout:
> However, I left the multiple select to still unselect all items when changing the options or setting it initially. Thoughts?

(This is a change - before, if the options changed in a multiple select, it would try to match up the previous selected labels to attempt to select the same items after the change. I think that is very error-prone.)

jasongrout:
Merging...let's iterate in another PR.

PR #1265, opened 2017-04-07
Unicode -> Unicode()

### LINKED ISSUE #1216: Deprecation warning with traitlets 4.3.1: Traits should be given as instances, not types
With traitlets 4.3.1 I get this warning:
```
Traits should be given as instances, not types (for example, `Int()`, not `Int`) Passing types is deprecated in traitlets 4.1.
```
which seems to come from this line: https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget_link.py#L19
Changing that line to use `Unicode()` instead fixes the warning.

jasongrout:
Fixes #1216


jasongrout:
CC @tomdz - thanks for the report and tracking down the problem!

Issue #1044, opened 2017-01-17
Snapshot widget state when syncing


hainm:
# Summary

We should snapshot the widget state when syncing state from the frontend to the kernel, since the sync may happen asynchronously because of the throttling.

---
# Original issue

hi

we have simple code in `nglview` that works fine with `ipywidgets 5.2.2`.
However, there is no effect if using `mode.set(...)` in `6.0.0.beta5` version.

one example: 
```python
this.model.set('camera_str', JSON.stringify(this.stage.viewer.camera))
```

will not update `camera_str` in Python.
https://github.com/arose/nglview/blob/master/js/src/widget_ngl.js#L77-L82

jasongrout:
Can you open up the debugger and see if the widget sync message is getting sent over the websocket? I can help you find where this message should be - let me know if you want help finding it.

hainm:
>let me know if you want help finding it.

yes, please. I am using chrome.

btw, below are screen shots (6.0.0.beta6  vs 5.2.2).
I also clicked the "debug" button in my screen but I think this is not what you meant.

6.0.0.beta6
=========
![debug](https://cloud.githubusercontent.com/assets/4451957/22479924/97fcad8a-e7bd-11e6-9d44-df88329f1c29.gif)

5.2.2
====
![old](https://cloud.githubusercontent.com/assets/4451957/22479945/ad99bf0c-e7bd-11e6-95aa-34903b0df348.gif)

jasongrout:
So now go to the 'network' tab in the debug window, and filter to just find websocket (WS) connections. Take a look at the messages going back and forth on the websocket channel.

jasongrout:
I tried to reproduce, but ran into errors importing pytraj. Do you have a minimal example not using pytraj?

hainm:
either

- install pytraj
`pip install pytraj`

- or use 
```
# cell 1
import nglview as nv
view = nv.demo()
view

# cell 2
# click the atoms
# then
view.picked
```

hainm:
![click](https://cloud.githubusercontent.com/assets/4451957/22486466/902a3428-e7d7-11e6-8107-3ce487301aaa.png)

jasongrout:
(it may be more helpful to try to debug this on chat, as we can communicate more quickly than here. Please ping me or @SylvainCorlay if you're around on https://gitter.im/ipython/ipywidgets)

jasongrout:
[...]

PR #1270, opened 2017-04-07
Make serializers synchronous and guarantee to return a snapshot of the data

jasongrout:
WIP, as this causes problems for binary arrays - basically we’ll need an explicit serializer on attributes that have binary arrays now, which can choose to implement the snapshot however they think best. The default serializer will make a copy.

This is work towards solving fundamental problems exposed at #1044.

CC @maartenbreddels - this affects the binary serialization.

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -307,37 +307,32 @@ class WidgetModel extends Backbone.Model {
      */
     sync(method, model, options: any = {}): any {
         // the typing is to return `any` since the super.sync method returns a JqXHR, but we just return false if there is an error.
-        var error = options.error || function() {
-            console.error('Backbone sync error:', arguments);
-        };
         if (this.comm === undefined) {
-            error();
-            return false;
+            throw 'Syncing error: no comm channel defined';
         }
 
-        var attrs = (method === 'patch') ? options.attrs : model.get_state(options.drop_defaults);
+        let attrs = (method === 'patch') ? options.attrs : model.get_state(options.drop_defaults);
 
         // The state_lock lists attributes that are currently being changed
-        // right now from a kernel message.
-        // We don't want to send these non-changes back to the kernel, so we
-        // delete them out of attrs, (but we only delete them if the value
-        // hasn't changed from the value stored in the state_lock).
+        // right now from a kernel message. We don't want to send these
+        // non-changes back to the kernel, so we delete them out of attrs if
+        // they haven't changed from their state_lock value
         if (this.state_lock !== null) {
-            var keys = Object.keys(this.state_lock);
-            for (var i=0; i<keys.length; i++) {
-                var key = keys[i];
+            for (const key of Object.keys(this.state_lock)) {
                 if (attrs[key] === this.state_lock[key]) {
                     delete attrs[key];
                 }
             }
         }
 
-        if (_.size(attrs) > 0) {
+        let msgState = this.serialize(attrs);
+
+        if (Object.keys(msgState).length > 0) {
 
             // If this message was sent via backbone itself, it will not
             // have any callbacks.  It's important that we create callbacks
             // so we can listen for status messages, etc...
-            var callbacks = options.callbacks || this.callbacks();
+            let callbacks = options.callbacks || this.callbacks();
 
             // Check throttle.
             if (this.pending_msgs >= (this.get('msg_throttle') || 1)) {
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Can you explain a bit more what the problem is?

jasongrout:
This is buried in #1044. Because we have message throttling, state sync messages are not always sent immediately when save_changes is called. The problem is that we serialize the state attributes we are syncing when we send the message, rather than when we request the save. This means that if we sync some attributes, then change the object, the sync message could be sent after the change, which means we sent the wrong values.

What this PR does is serialize the object to get a snapshot of the synced attributes when the save is requested. That means that the serializers must return a snapshot of the state that won't change. It's easy to accomplish this with JSONable values - just stringify and parse to get a cheap deep copy. However, if there are binary arrays in the json state, there's no way to get a snapshot. I imagine that some libraries will not do a snapshot of binary data, but will just sync the current state of the binary array and hope for the best.

I suppose another approach is to make the state sync call explicitly just be a request that the *current* state of the object be synced at some point in the future. This means that we then have to be careful to never leave the state of the object in some inconsistent intermediate state, which sounds tenuous at best.

maartenbreddels:
Actually, can't we juse the code that I wrote for extracting the buffers to do this? I spend effort not to clone the whole thing, but if we rip that part out, it clones everything (except the arrays..)

jasongrout:
Sure, the JSON parse/stringify is just an easy way to deep copy an object. We could use a different deep copy implementation instead, though I it's likely to be much slower (maybe we should time it, though). I think the underlying issue with using it to do a deep copy is that then we are condoning that this sort of async bug is expected for binary data.

(I wish we had simple copy-on-write ArrayBuffers, so we could get a real snapshot of the binary data.)

maartenbreddels:
Is it an option to remove the throttling, or do it differently?

jasongrout:
> Actually, can't we juse the code that I wrote for extracting the buffers to do this? I spend effort not to clone the whole thing, but if we rip that part out, it clones everything (except the arrays..)

and officially sanction this async problem for binary arrays?

(I wish we had some sort of copy-on-write binary view of an array. I just looked into the Chrome source, and if I understand it correctly, they don't implement copy-on-write copies: https://cs.chromium.org/chromium/src/v8/src/builtins/builtins-arraybuffer.cc?l=281 )

jasongrout:
> Is it an option to remove the throttling, or do it differently?

Good question. I think it makes sense to have some sort of throttling implemented somewhere. A typical usecase is dragging a slider which may cause a lot of computation to be done, maybe drawing a plot. The nice thing about the pending messages throttle is that it dynamically adjusts for the delay - sync messages are sent exactly as often as the kernel can handle them.

What is the problem with snapshotting the data the widget wants to sync?

maartenbreddels:
Actually, that is something I raised a while ago https://github.com/jupyter-widgets/ipywidgets/issues/663 and recently updated with some code (on the python side). Maybe it's sth that needs to go into the interact decorator, and we just handle this purely on the python side.

maartenbreddels:
> What is the problem with snapshotting the data the widget wants to sync?

We could, it will take up memory, but maybe we should accept that.

jasongrout:
> We could, it will take up memory, but maybe we should accept that.

That's where we maybe look the other way if someone chooses to not hand us back a snapshot of the state. Or rather, widget authors can decide if they *really* need to hand back a copy or not. But by default, we make a copy.

jasongrout:
The end result is that if you have an attribute that has binary data, and you don't care about this snapshotting business, you just declare that attribute to have the identity serializer, i.e., hand back the attribute as-is.

jasongrout:
1. Actually, thinking about it, I don't think there really will be that much change after all. Already, if you have binary data, you'll probably need a custom serializer to understand how to package up that binary data, so the automatic copying is not done with the current PR. Whether you actually make a copy or not is on your head (it's correct to make a copy, but maybe it's okay to fudge it a bit...).

2. This is the sort of thing I was talking about on the original binary PR, that we should just treat every message we split up as a copy. However, your implementation will work even if people fudge things a bit and don't return copies. That's maybe a good thing to be robust against that.

jasongrout:
Ping @maartenbreddels - I think we should be okay with merging this after all. If you didn't care about making a copy, you could just give an identity function as a serializer.

maartenbreddels:
Hmm, I don't quite follow when `send_sync_message` and when `serialize` are being used, can you explain this a bit.

jasongrout:
Great question. When someone calls for a sync (e.g., with `save_changes()`), the `sync` method gets called. Inside the sync method, we first serialize a message (which *should* snapshot the current state of the attributes were are going to sync). Then we check to see if there are any pending sync messages still in flight (i.e., a sync message that was sent, but we haven't had a corresponding execution status idle message). If there is no pending sync message still being processed by the kernel, then we directly send the message (calling `send_sync_message`). If there is a pending sync message still being processed by the kernel, then we store the serialized snapshotted attributes in `this.msg_buffer`. When we get back a status idle message for the pending message, we call `send_sync_message` directly to send the message we've been saving up: https://github.com/jasongrout/ipywidgets/blob/20e00848139a83659f80cf027d313cdf6ac00adf/jupyter-js-widgets/src/widget.ts#L243.

Does that answer your question?

maartenbreddels:
Ok, thanks that explains a lot. I think I understand the issue now. If you don't give a serializer, it will try to clone your data, and fail if there is a binary buffer (that's fine I guess). In that case you have to give a serializer, which could just do nothing. So in effect the default serializer is a clone using stringify/JSON.parse, if you don't want cloning you need to specify so. I think that is good default behaviour.

jasongrout:
Exactly! Do you mind looking this over and/or trying out this PR and giving me thumbs up/down? It's a fundamental enough change that I'm hesitant to merge it without someone else looking at it.

SylvainCorlay:
Also looking at it today

maartenbreddels:
Will do, give me 24h

(from mobile phone)

Op 10 apr. 2017 23:31 schreef "Jason Grout" <notifications@github.com>:

Exactly! Do you mind looking this over and/or trying out this PR and giving
me thumbs up/down? It's a fundamental enough change that I'm hesitant to
merge it without someone else looking at it.

—
You are receiving this because you were mentioned.
<https://github.com/jupyter-widgets/ipywidgets/pull/1270#issuecomment-293085221>,
or mute the thread

jasongrout:
ping @SylvainCorlay, @maartenbreddels :)

maartenbreddels:
Give me a 4 more hours! 😊

(from mobile phone)

Op 12 apr. 2017 17:14 schreef "Jason Grout" <notifications@github.com>:

jasongrout:
> Give me a 4 more hours! 😊

No pressure!

jupyter-js-widgets/src/widget.ts
```diff
@@ -347,61 +342,71 @@ class WidgetModel extends Backbone.Model {
...
+        for (const k of state) {
+            if (serializers[k] && serializers[k].serialize) {
+                state[k] = (serializers[k].serialize)(state[k], this);
+            } else {
+                // the default serializer just deep-copies the object
+                // TODO: this won't work if the object is a primitive object with binary buffers!
+                // How should we handle those? One way is to declare a serializer for fields
+                // that could have binary that copies what fields it can, and makes a decision
+                // about whether to copy the ArrayBuffer
+                state[k] = JSON.parse(JSON.stringify(state[k]));
```
maartenbreddels: Can we maybe have a warning here, say when an exception occurs, give a hint what may have gone wrong?

jasongrout:
There will be an error from the json engine. However, we can catch it and
give it some context.

maartenbreddels:
Controller works, code looks, ipyvolume doesn't break. The issue in https://github.com/jupyter-widgets/ipywidgets/issues/1195 is still present though. I can kind of reproduce it, not easy, but we should fix that before 7.0, it is large performance penalty since it sends back (for some reason) especially binary buffers it seems.

maartenbreddels:
Yes, exactly, say console.error('serializing failed for ...') and they rethrow the exception would be helpful already.

jasongrout:
Interesting: JSON.stringify already works for ArrayBuffers: 

```
> JSON.stringify(new Int16Array(4))
"{"0":0,"1":0,"2":0,"3":0}"
```

jasongrout:
I added a message for context for serialization errors.

maartenbreddels:
Excellent 👍

Issue #1272, opened 2017-04-08
Make output renderer for classic notebook


jasongrout:
Let's try using something like @gnestor's geojson render extension to render widgets as outputs in the classic notebook: https://github.com/jupyterlab/jupyterlab_geojson/tree/master/nbextension

This would finally do away with the widget area as a special area in the cell.

jasongrout:
Fixed by #1274.

Issue #1237, opened 2017-03-29
Cell Toggle Scrolling & Clear options do not work on figures created by interact() functions


iyer:
If I draw figures through an interact() function, the figures do not respond to the Jupyter|Cell|Current Output|Toggle Scrolling or Clear menus.

I have avoided using matplotlib inline based on the suggestion https://github.com/ipython/ipython/issues/10376

Below is the version info

```python
import ipywidgets
ipywidgets.version_info
(6, 0, 0)
```

Below is the information on jupyter notebook
```
The version of the notebook server is 4.4.1 and is running on
Python 3.5.3 |Anaconda 2.4.0 (64-bit)| (default, Mar  6 2017, 11:58:13) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
```

Below is the code that demonstrates this behavior

```python
import matplotlib.pyplot as plt
from ipywidgets import interact
import pandas as pd

def f(n):
    nrows = 7
    figsize = (7, 5*nrows)
    fig, axes = plt.subplots(nrows=nrows, ncols=1)
    for idx in range(nrows):
        pd.Series(range(n)).plot(ax=axes[idx], figsize=figsize)
    plt.show(fig)

interact(f,n=(5,20));
```

jasongrout:
Great question. We have a new way of doing interacts in ipywidgets 6 that makes the output actually part of the interact. This means you can, for example, embed an interact in another widget, put it beside other controls, have two interacts in the same cell, etc.

Clearing output will work (and will clear the whole interact, not just the output) in notebook 5.0 (due out in a few days).

I'll look into what it would take to have the scrolling toggle working. That would work in JupyterLab, but may take another change in the notebook to enable correctly.

iyer:
Nice! Looking forward to notebook 5.0

Also it would be great to get a fix in for the scrolling behavior which can make reading figures a little tedious.

jasongrout:
This will be fixed by #1274.

jasongrout:
#1274 was merged, so closing this as fixed.

PR #1274, opened 2017-04-10
Make widgets render using output areas in classic notebook.

jasongrout:
This is work towards fixing #1272, and makes the rendering consistent between jupyterlab and the classic notebook.

Essentially, this gets rid of the special widget area above cells, and instead renders widgets using the widget display_data output messages.

This also finishes fixing #1237.


gnestor:
@jasongrout Let me know if you have any questions about this stuff 👌

jasongrout:
Thanks! I'm working through it right now. You may notice that I'm using a slightly different approach (I think it's a bit cleaner) to get the output area, etc.

jasongrout:
(for example, I'm directly requiring the output area, and using the output area prototype to register the mime type, rather than searching for an instance among the cells.)

widgetsnbextension/src/extension.js
```diff
@@ -46,7 +50,7 @@ var handle_cell = function(cell) {
     }
 };
 
-function register_events(Jupyter, events) {
+function register_events(Jupyter, events, outputarea) {
```
gnestor: @jasongrout Yep, this is cleaner. Before we unwebpackified the notebook, doing this didn't work. I will update the mimerender extensions to reflect this :+1:

jasongrout: Yeah, there are other kludgy workarounds I can now delete because we don't have webpack too, like https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/src/widget_output.js#L93

jasongrout:
CC @SylvainCorlay, @maartenbreddels - it would be great if someone looked at/tested this out too, to see what I might have missed.

maartenbreddels:
Any idea of the issue raised here (about Output not syncing is related to, or fixed by this) https://github.com/jupyter-widgets/ipywidgets/pull/1232

jasongrout:
> Any idea of the issue raised here (about Output not syncing is related to, or fixed by this) #1232

Oh, interesting. This might break #1232 - we'll probably need to do something like @gnestor does and hook into the clearing signals to remove views.

jasongrout:
Note to self: @gnestor's clearing function is at https://github.com/jupyterlab/jupyterlab_geojson/blob/master/nbextension/src/renderer.js#L21, triggered at https://github.com/jupyterlab/jupyterlab_geojson/blob/master/nbextension/src/renderer.js#L71.

@gnestor - why did you also handle the output_added.OutputArea event?
 https://github.com/jupyterlab/jupyterlab_geojson/blob/master/nbextension/src/renderer.js#L75

jasongrout:
Do we have a test for the _view_count that would have failed here?

maartenbreddels:
Do you mean a unittest? Or sth automated? Or you mean just an example to demonstrate it?

maartenbreddels:
Btw, for ipyvolume, I've started doing testing in the notebook, as an integration test. Do you have experience or know someone who knows how to set up a full blown test with headless browser etc?

jasongrout:
CC @blink1073 - I think JLab probably currently has the nicest infrastructure for testing currently.

At various times, we've had headless browser testing for widgets using CasperJS. We don't have a good story for widgets right now.

jasongrout:
This *almost* calls the `.remove()` when appropriate (though it's totally kludgy). Ideally, we should have the output area itself sending a signal just *before* it empties itself that we could latch on to.

The only cases I've seen where something messes up now is something like this:

```python
from ipywidgets import *
out = Output()
x = IntSlider()
out
```
```python
with out:
    display(x)
    display(x)
```
```python
x._view_count
```
returns 1 when it should return 2, and then evaluating the last two cells again we get `_view_count` as the correct 4. However, clearing (`out.clear_output`) does not correctly decrement the view count because it's only listening for changes at the cell level, not the output area level.

jasongrout:
Okay, the last commit takes care of clearing an output area. However, it doesn't address the problem of the first _view_count returning 1. Not sure what's happening there yet.

jasongrout:
@maartenbreddels - okay, _view_count looks like it is updating correctly now. It required a few fixes to the recent syncing, so I made those PRs first, and then rebased this on the current master.

I think this is ready to go in.

jasongrout:
Feedback from @maartenbreddels:

1. output widgets don't seem to be incrementing _view_count.

2. Executing a blank cell doesn't seem to decrement _view_count

maartenbreddels:
I don't have time to look into it now, but for the record, it seems like if a list of buffers is send, they are all the same.

maartenbreddels:
Ok, sorry for the noise, that was a bug in ipyvolume.

jasongrout:
@maartenbreddels - the output widget view counts should be fixed now, and if you run https://github.com/jupyter/notebook/pull/2411, the view count should now be updated even if an empty cell is run.

PR #1278, opened 2017-04-12
upgrade setuptools, pip on Travis

minrk:
and drop use of deprecated travis-wheels repo, which hasn’t been updated in some time now that manylinux wheels are on PyPI.


minrk:
I noticed that manylinux wheels are being ignored on some travis envs that have out of date pip. Install should be a bit quicker after this.

jasongrout:
Thanks!

PR #1290, opened 2017-04-17
Disambiguate models and views

### LINKED ISSUE #1256: Multiple widgets have the same model/view information
A ramification of #1228 is that the FloatSlider and FloatRangeSlider are confused in the kernel, since both have exactly the same model and view implementation. So we basically have different presentations of the same model to the user, and have to disambiguate between them.

I think this is a problem with the models, not with the idea of #1228. I think we should probably have separate models for float slider and float range slider, for example.

jasongrout:
This fixes #1256.

Changelog text:

* A refactoring of the text, slider, slider range, and progress widgets in resulted in the `BoundedIntText` and `BoundedFloatText` losing their `step` attribute (which was previously ignored), and a number of these widgets changing their `_model_name` and/or `_view_name` attributes [(#1290)](https://github.com/jupyter-widgets/ipywidgets/pull/1290).

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -98,7 +98,7 @@ class BoundedFloatText(_BoundedFloat):
         description displayed next to the textbox
     """
     _view_name = Unicode('FloatTextView').tag(sync=True)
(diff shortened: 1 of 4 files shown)


jasongrout:
CC @SylvainCorlay for review.

Issue #986, opened 2016-12-14
Make an Output model and view for the embedded widget manager


stsievert:
# Summary

Figure out how to save the state of the output widget (i.e., the output messages) and make an embedded widget manager output view.

---
# Original issue

#16 is much appreciated! Thanks for this :+1:

I have (tried) to embed a widget in a static webpage but haven't had any luck. I follow the documentation and select "Widgets > Embed widget" and copy that snippet into an HTML page. When I open that HTML page, I see a blank screen and can inspect to find a Javascript error:

<img width="791" alt="screen shot 2016-12-14 at 9 52 57 am" src="https://cloud.githubusercontent.com/assets/1320475/21189084/23d85c92-c1e3-11e6-8e67-aebd64f5c5f2.png">

This is with 1b9af7e0693756bb2cde8ca3b2e5acda2fff6f6c and `python setup.py install` in a virtualenv. It's with a simple widget:

``` python
from ipywidgets import interact

def(x):
    return x

interact(f, x=10)
```

stsievert:
Issues shouldn't go on beta software, closing.

SylvainCorlay:
@stsievert this feature requires the current state of master

stsievert:
I am reopening this issue; I still can't embed widgets into a static HTML page (I am on beta6). My process for trying to embed widgets:

1. I define a (simple) function and develop a widget around it.
3. I click on "Embed Widgets" and copy/paste the HTML into `widget.html`.
4. I open `widget.html` in my browser. I expect to see the widget in the browser, however a blank screen shows (and the console shows no errors).

It's not clear from the documentation if I'm supposed to define my own HTML code for the widget. That's the reason the documentation goes into such detail on the JSON schema?

SylvainCorlay:
Hi @stsievert Can you provide the generated html file?

stsievert:
Here's a dropbox link: https://www.dropbox.com/s/b85fath9kiqxxv8/widget.html?dl=0

It implements the function above.

SylvainCorlay:
You are missing the `<html>` and `<body>` tags. The snippet of code that is generated by the embed widgets is meant to be pasted in an already valid html file.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Your Title</title>
  </head>
  <body>
    <!-- page content -->
  </body>
</html>
```

stsievert:
D'oh.

[...]

PR #1294, opened 2017-04-17
Make the output widget store its state.

jasongrout:
This is work on #986.


jasongrout:
@SylvainCorlay - I think this is ready for review. Refreshing a page with an output widget now works. The output messages should be stored when saving the widget state as well.

Unfortunately, it means that we have to double-transfer any output (once as a display_data message, another time as the output widget state).

Issue #1200, opened 2017-03-15
Test setting overflow:auto on containers


jasongrout:
We overflow `overflow:visible` on jupyter widgets, which overrides phosphor's overflow:hidden. We should check to see how overflow:hidden (or even overflow:auto) works, since that would make for cleaner containment.

cf. #1117

PR #1297, opened 2017-04-18
Make container widgets have overflow auto by default.

jasongrout:
This makes the containers a little more predictable regarding layout overflowing the bounds of the container.

cf. #1200.


jasongrout:
CC @ssunkara1.

jasongrout:
We didn't get much feedback on this as a PR. Let's merge this and try it out in the alpha.

PR #1298, opened 2017-04-18
Don’t buffer attributes that are being set from the kernel unless they changed

### LINKED ISSUE #1195: Data send multiple times to backend
While working on #1194, I noticed what I think is a bug. Take the following code:
```python
import ipywidgets
s = ipywidgets.IntSlider(description="hello", min=0, max=2, value=0)
v = ipywidgets.IntText(value=1)
ipywidgets.jslink((v, 'value'), (s, 'value'))
ipywidgets.HBox([s,v])
```
Now in a next cell, you execute:
```python
s.description = "world"
```
All is fine, the js logic with `state_lock` will avoid that this gets send to the python backend again, however, in the js side, `s.model.change` still recorded that `description` was changed. If you now enter a digit in the IntText, it will trigger a save_changes in the IntSlider, and it will again send {description: 'world', ..} to the backend. This happens in many other cases. For instance, in my case, this is an array with a ~million entries, not something you'd like to send around multiple times.
I am not sure how to resolve this, although happy to try.

PS: Is there any way to get a reference to a widgets in the js console, in ipyvolume is just set window.last_scatter = this; for debugging, but maybe there is a general way.

jasongrout:
Fixes #1195

CC @maartenbreddels.

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -276,13 +276,27 @@ class WidgetModel extends Backbone.Model {
 
         // Backbone only remembers the diff of the most recent set()
         // operation.  Calling set multiple times in a row results in a
-        // loss of diff information.  Here we keep our own running diff.
+        // loss of change information.  Here we keep our own running diff.
         //
         // We don't buffer the state set in the constructor (including
         // defaults), so we first check to see if we've initialized _buffered_state_diff.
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Confirmed this fixes https://github.com/jupyter-widgets/ipywidgets/issues/1195

PR #1302, opened 2017-04-19
Serialize/deserialize binary values to the serialization format

jasongrout:
CC @maartenbreddels.

@SylvainCorlay - the version 1 specification was too restrictive - it didn't allow for any additional properties. I bumped the specification to version 2, added the extra fields for binary buffers, and removed the additionalProperties: false keys so we can add optional fields in the future.

### DIFF
--- a/jupyter-widgets-schema/v2/state.schema.json
+++ b/jupyter-widgets-schema/v2/state.schema.json
@@ -11,7 +11,9 @@
         },
         "version_minor" : {
             "description": "Format version (minor)",
-            "type": "number"
+            "type": "number",
+            "minimum": 0,
+            "maximum": 0
         },
         "state": {
             "description": "Model State for All Widget Models - keys are model ids, values are model state",
--- a/jupyter-widgets-schema/v2/state.schema.json
+++ b/jupyter-widgets-schema/v2/state.schema.json
@@ -39,24 +39,29 @@
                         "description": "Binary buffers in the state",
                         "type": "object",
                         "properties": {
-                            "paths": {
-                                "description" : "Array of paths in the state for the corresponding buffers",
+                            "buffers": {
                                 "type": "array",
                                 "items": {
-                                    "description": "A path for a binary buffer value.",
-                                    "type": "array",
-                                    "items": {
-                                        "description": "An object key or array index",
-                                        "type": ["string", "number"]
-                                    }
+                                    "type": "object",
+                                    "properties": {
+                                        "path": {
+                                            "description": "A path for a binary buffer value.",
+                                            "type": "array",
+                                            "items": {
+                                                "description": "An object key or array index",
+                                                "type": ["string", "number"]
+                                            }
+                                        },
+                                        "data": {
+                                            "description" : "A binary buffer encoded as specified in the 'encoding' property",
+                                            "type": "string"
+                                        }
+                                    },
+                                    "required": ["path", "data"]
                                 }
                             },
-                            "data": {
-                                "description" : "Array of encoded binary buffers, encoded as specified in the 'encoding' property",
-                                "type": "array"
-                            },
                             "encoding": {
-                                "description": "The encoding of each item in the data property",
+                                "description": "The encoding of each buffer data property",
                                 "type": "string",
                                 "oneOf": [
                                     {"enum": ["hex"], "description": "Base 16 encoding, as specified in RFC 4648, section 8 (https://tools.ietf.org/html/rfc4648#section-8)"},
@@ -65,7 +70,7 @@
                                 ]
                             }
                         },
-                        "required": ["paths", "data", "encoding"]
+                        "required": ["buffers", "encoding"]
                     }
                 },
                 "required": [ "model_name", "model_module", "state" ]
--- a/jupyter-widgets-schema/v2/state.schema.json
+++ b/jupyter-widgets-schema/v2/state.schema.json
@@ -1,0 +1,64 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "Jupyter Interactive Widget State JSON schema.",
+    "type": "object",
+    "properties" : {
+        "version_major" : {
+            "description": "Format version (major)",
+            "type": "number",
+            "minimum": 2,
+            "maximum": 2
+        },
+        "version_minor" : {
+            "description": "Format version (minor)",
+            "type": "number"
+        },
+        "state": {
+            "description": "Model State for All Widget Models",
+            "type": "object",
+            "additionalProperties" : {
+                "type": "object",
+                "properties": {
+                    "model_name": {
+                        "description" : "Name of the JavaScript class holding the model implementation",
+                        "type": "string"
+                    },
+                    "model_module": {
+                        "description" : "Name of the JavaScript module holding the model implementation",
+                        "type": "string"
+                    },
+                    "model_module_version": {
+                        "description" : "Semver range for the JavaScript module holding the model implementation",
+                        "type": "string"
+                    },
+                    "state": {
+                        "description" : "Serialized state of the model",
+                        "type": "object"
+                    },
+                    "buffer_paths": {
+                        "description" : "Array of paths in the state for the corresponding buffers",
+                        "type": "array",
+                        "items": {
+                            "description": "A path for a binary buffer value.",
+                            "type": "array",
+                            "items": {
+                                "description": "An object key or array index",
+                                "anyOf": [{"type": "string"}, {"type": "number"}]
+                            }
+                        }
+                    },
+                    "buffers": {
+                        "description" : "Array of base64-encoded binary buffers",
+                        "type": "array",
+                        "items": {
+                            "description": "A base64-encoded binary buffer",
+                            "type": "string"
+                        }
+                    }
+                },
+                "required": [ "model_name", "model_module", "state" ]
+            }
+        }
+    },
+    "required": [ "version_major", "version_minor", "state" ]
+}
--- a/jupyter-widgets-schema/v2/state.schema.json
+++ b/jupyter-widgets-schema/v2/state.schema.json
@@ -14,7 +14,7 @@
             "type": "number"
         },
         "state": {
-            "description": "Model State for All Widget Models",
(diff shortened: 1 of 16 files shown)


jasongrout:
@SylvainCorlay - What do you think of the schema change (I'd like to get your thoughts before publishing the new schema package).

jasongrout:
Travis is failing because I haven't published the schema package yet.

jasongrout:
Restarting the build now that we published the jupyter-widgets-schema package

jasongrout:
I'll plan on merging this tomorrow. @SylvainCorlay, @maartenbreddels, please let me know if you have any comments.

jupyter-widgets-schema/v2/state.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                            }
+                        }
+                    },
+                    "buffers": {
+                        "description" : "Array of base64-encoded binary buffers",
+                        "type": "array",
+                        "items": {
+                            "description": "A base64-encoded binary buffer",
+                            "type": "string"
+                        }
```
maartenbreddels: I'd say, include the encoding(s) in the description, since if at some moments we decide to support compression, we may have to change the schema again. If we put the encoding in the state file it's more flexible.

jasongrout: So have one more piece of metadata in the field saying how the buffers are encoded? That makes sense.

maartenbreddels: I'd say a list of encoding, say, to make it more concrete with a python code:
```python
import codecs
data_encoded = data # data is the original data we'd like to encode
encodings = ["bz2", "base64"] # this should come from the json
for encoding in encodings:
    data_encoded = codecs.encode(data_encoded, encoding)
```
And for the decoding:
```python
data_decoded = data_encoded
for encoding in encodings[::-1]:
    data_decoded = codecs.decode(data_decoded, encoding)
```
so if we store the list of used encodings, and specify what we support (now just base64), we are a bit more future proof I think, what do you think?

jasongrout: @SylvainCorlay? With what I just pushed, we have a single encoding for each model's binary buffers (base64 or hex), rather than an encoding per buffer. Since the state serialization happens at the model level right now, and the buffer splitting and encoding is completely independent of knowledge of what is in the buffer, I don't think it makes sense to support a more fine-grained per-buffer encoding specification. I mean, how would a model specify the encoding of a single buffer somewhere in its serialized state?

jasongrout:
(When the next json schema spec is published and tools adopt it, we can use the new `"const"` property instead of single-valued enums for the buffer encoding. See https://github.com/json-schema-org/json-schema-spec/pull/139, which is in draft 6 of the spec (https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-6.24).

maartenbreddels:
So just a single encoding, not a list of, to support compression? Maybe it is overkill, and it can be just a single encoding, and later on we can include a compression option.

SylvainCorlay:
Are you guys available for a quick chat?

SylvainCorlay:
https://appear.in/ipywidgets

jasongrout:
From the appear.in conversation between @SylvainCorlay, @maartenbreddels, and I: the buffer information should be zipped, so that each buffer is represented by an object of the form `{path: [...], data: '...'}`.

jasongrout:
I updated so that the following is a valid state:

```
{
  version_major: 2,
  version_minor: 0,
  state: {
    "id-1": {
      model_name: "asdf",
      model_module: "asdf",
      state: {
        'a': [0,null],
        'b': [{}, {}]
      },
      buffers: {
        buffers: [
          {path: ['a', 1], data: "0123BEEF"},
          {path: ['b', 1, 'c'], data: "AFEF"}
        ],
        encoding: "hex"
      }
    }
  }
}
```
I'd rather not have the same name repeated twice, `buffers.buffers`, but there is precedence for it here, `state.state`. Thoughts?

jasongrout:
(I published this jupyter-widgets-schema as 0.3.0-beta.0)

jasongrout:
Another thing that would be backwards-incompatible changes, but now might be the right time to do it: how about we have the version as a single field:

`version: [2,0]` or `version: {major: 2, minor: 0}`. That's quite a bit cleaner, I think.

jasongrout:
Another thing to consider: I think it might type better to have the widget manager state be an array of objects, each of which has an 'id' key, and absorb the model name/version/module back into the widget state, where it exists anyway...

```
state: [
  {id: 'uuid', {_module_name: 'name', _model_module_version: '1', _model_module: 'module', other_state: 'some other state'}, 
]
```
Still thinking about these, though.

jasongrout:
to avoid the confusion of having buffers.buffers, we could have 

```
buffers: {
        content: [
          {path: ['a', 1], data: "0123BEEF"},
          {path: ['b', 1, 'c'], data: "AFEF"}
        ],
        encoding: "hex"
      }
```
or 
```
buffers: {
        data: [
          {path: ['a', 1], buffer: "0123BEEF"},
          {path: ['b', 1, 'c'], buffer: "AFEF"}
        ],
        encoding: "hex"
      }
```
Thoughts?

jasongrout:
or even
```
buffers: [
          {path: ['a', 1], data: "0123BEEF", encoding: "hex"},
          {path: ['b', 1, 'c'], buffer: "AFEF", encoding: "hex"}
      ]
```
though this is probably the least efficient spacewise, since we are repeating the encoding each time, when it is likely the same for all buffers in the entire state.

maartenbreddels:
For compression, we may want to have a per buffer compression, since one may attempt RLE and bz2 and decide on the fly which compression works better. For encoding it's less obvious, but you could imagine one of the buffers being an external file (in the future, not now), which may have no or a different encoding, while the rest may be inlined, and have base64. So I think the overhead is small but may be worth it.

jasongrout:
Or you could imagine in a hosted situation, storing the buffer in an internal database with an api for getting the contents out.

jasongrout:
Okay, I think I updated everything for the option that looks like:

```
buffers: [
          {path: ['a', 1], data: "0123BEEF", encoding: "hex"},
          {path: ['b', 1, 'c'], data: "AFEF", encoding: "hex"}
      ]
```

Thoughts?

maartenbreddels:
Awesome work! 👍

jasongrout:
Does that mean you're good with merging this?

maartenbreddels:
I'm testing it, but it doesn't seem to work with this line (in embed-webpack.ts):
```
var widget_state_schema = require('jupyter-widgets-schema').v2.state;
```
only v1 is defined, any idea what is going on?

jasongrout:
I'll check. I'm also working on some simple tests in jupyter-js-widgets

jasongrout:
> only v1 is defined, any idea what is going on?

Should hopefully be fixed now.

maartenbreddels:
I don't think I can modify or do a PR to this PR, in fact I did, but it went [here](https://github.com/jasongrout/ipywidgets/pull/1)

maartenbreddels:
embedding is quite broken now
If I do
```python
import ipywidgets as widgets
widgets.Button(description="click")
```
The the generated code gives me this:
```html
<script src="https://unpkg.com/jupyter-js-widgets@~3.0.0/dist/embed.js"></script>
<script type="application/vnd.jupyter.widget-state+json">
{
    "version_major": 2,
    "version_minor": 0,
    "state": {
        "e496e580048740cfb6203a9d4c08d9cf": {
            "model_name": "LayoutModel",
            "model_module": "jupyter-js-widgets",
            "model_module_version": "3",
            "state": {
                "_view_module_version": "3",
                "_model_module_version": "3",
                "_view_count": 1
            }
        },
        "82db7d86f07e4eb783c40dd249c7ad4a": {
            "model_name": "ButtonStyleModel",
            "model_module": "jupyter-js-widgets",
            "model_module_version": "3",
            "state": {
                "_view_module_version": "3",
                "_model_module_version": "3",
                "_view_count": 1
            }
        },
        "1d5f7f75499049c98375918b8cfd3d2a": {
            "model_name": "ButtonModel",
            "model_module": "jupyter-js-widgets",
            "model_module_version": "3",
            "state": {
                "_view_module_version": "3",
                "description": "click",
                "style": "IPY_MODEL_82db7d86f07e4eb783c40dd249c7ad4a",
                "_view_count": 1,
                "layout": "IPY_MODEL_e496e580048740cfb6203a9d4c08d9cf",
                "_model_module_version": "3"
            }
        }
    }
}
</script>
```
For testing I just copy the embed.js to the directory where the html file where I put this code in resides , and change the src tag to 'embed.js'. What I missing are I believe the views, and the script tags for the visible views.

jasongrout:
Yes, it looks like we're missing some script tags that look like this in the old version:

```
<script type="application/vnd.jupyter.widget-view+json">
{
    "model_id": "21150b295bd84591adc9724e747e7f9f"
}
</script>
```

maartenbreddels:
Ok, couldn't resist, seems to work now using [this PR](https://github.com/jasongrout/ipywidgets/pull/3). This doesn't really seem like a good workflow, where I do a PR to your repo, is there a proper way to do this in github?

jasongrout:
You should be able to push directly to my embeddedoutput2 branch, since the checkbox up above to the right is checked (" Allow edits from maintainers.")

`git push git@github.com:jasongrout/ipywidgets.git embeddedoutput2` I think should work.

maartenbreddels:
I don't think I am a maintainer right?

maartenbreddels:
Exciting to see this progress btw, I can clean up quite some code in ipyvolume soon, using binary transfer only from now on.

jasongrout:
> I don't think I am a maintainer right?

Fixed!

jasongrout:
> seems to work now

Feel free to merge when it looks good to you. I'm working on the package version PR, and then I think it will be time to release a beta (or alpha?) and shift focus to testing and docs.

jasongrout:
> seems to work now

Great. I'll merge this and we can iterate.

maartenbreddels:
Ok, I wanted to add some changes, but I'll do that in a separate PR, the Python version of this needs some changes.

PR #1303, opened 2017-04-19
Update to most recent JupyterLab

jasongrout:
This updates the jupyterlab_widgets package to the most recent version of JupyterLab. This uses https://github.com/jupyterlab/jupyterlab/pull/2038.


jasongrout:
In order for this to work, while we still don't have jupyter-js-widgets published, you'll need to do ```shell
cd jupyterlab_widgets
npm install --save ../jupyter-js-widgets
jupyter labextension install .
jupyter lab
```

SylvainCorlay:
@jasongrout if you are around, can you hop on gitter real quick?

PR #1309, opened 2017-04-25
Correct use of `save_changes` and `touch` methods

Ricyteach:
I have been trying for the past couple hours to `save_changes()` using the view object as shown on this page, but kept getting an error. Apparently this part of the API has been changed per this discussion: 

[https://github.com/jupyter-widgets/ipywidgets/issues/1052#issuecomment-274836038](url)

I have attempted to correct the explanation for use of the `save_changes` method (it is to be used on the model, not the view) and have added a line with an explanation for use of the view `touch` method. Hopefully everything is accurate.


jasongrout:
This looks great; thanks!

PR #1310, opened 2017-04-25
Adding type='password' support for Text in TextView class.

jakhani:
Currently there is no way to use Text widget for Password. This code will allow to pass type while using Text widget which will allow to set type='password' for Text.


jasongrout:
Thanks! Up to now, we've made different types of html widgets different things (e.g., the checkbox, color, radio, etc.). Following that pattern, we'd have a `Password()` widget. What do you think about making a separate `Password()` widget that inherits from the text widget? On the other hand, I can see this just being basically just tweaking a textbox, so perhaps making a password=True/False attribute on a textbox would work well.

jakhani:
@jasongrout This is to follow the way html works. In HTML, when you pass type="password", it doesn't show it's value. If there is any other way already available, can you please share some sample code with me?

jasongrout:
@jakhani - I agree it would be great to make this available to users, and there's no way to do it right now. What I'm suggesting is that we either (a) make a new `Password()` that acts almost exactly like a text widget, except that in the view we set the type to password, or (b) make a single `password` attribute to the `Text()` widget that controls this switch. Each of those options makes it easy to expose this one thing (a password box) in a pythonic api.

jasongrout:
(while exposing the input `type` exposes all of the sorts of inputs HTML has to offer: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)

jakhani:
@jasongrout Sure. I have added new widget Password. I have tested it and it's working fine. Can you please review this code?

ipywidgets/widgets/widget_string.py
```diff
@@ -107,3 +107,40 @@ def on_submit(self, callback, remove=False):
             Whether to unregister the callback
         """
         self._submission_callbacks.register_callback(callback, remove=remove)
+
+
+@register
+class Password(_String):
```
jasongrout: Would it be easier to just inherit from Text?

```
@register
class Password(Text):
    _view_name = ...
    _model_name = ...
```
Then you inherit all of the text behavior automatically.

jupyter-js-widgets/src/widget_string.ts
```diff
@@ -385,3 +385,121 @@ class TextView extends LabeledDOMWidgetView {
...
+    defaults() {
+        return _.extend(super.defaults(), {
+            _view_name: 'PasswordView',
+            _model_name: 'PasswordModel'
+        });
+    }
+}
+
+export
+class PasswordView extends LabeledDOMWidgetView {
```
jasongrout: Again, here how about just inheriting from `TextView`, and doing something like:

```
render() {
    super.render();
    this.textbox.setAttribute('type', 'password');
}
```
and let everything else be inherited?

jasongrout: (the only issue I could see here is if the element is appended to the DOM before the input is set to password, the default string might flash up on the screen exposed)

jasongrout: (i.e., that's something we should check to see if it happens)

jasongrout: In fact, why don't we change `TextView` to have a new protected attribute (at the bottom of the class), the `inputType`, which for `TextView` is `'text'` and is used in the `render()` method, but the `PasswordView` subclass just sets this to `'password'`. Then the PasswordView just becomes something like:

```typescript
export
class PasswordView extends TextView {
    inputType: string = 'password';
}
```
or something like that.

jakhani:
@jasongrout I have incorporated all your comments and submitted changes. Can you please have a look at it?

jupyter-widgets-schema/jupyterwidgetmodels.md
```diff
@@ -661,6 +661,23 @@ Attribute        | Type             | Default          | Help
 `placeholder`    | string           | `'\u200b'`       | Placeholder text to display when nothing has been typed
 `value`          | string           | `''`             | String value
 
+### Jupyter.Password
```
jasongrout: Can you delete the changes to this file? This file is documenting the widgets as of 2.1.0 and is automatically generated. We'll generate another listing when we release the next version.

jasongrout:
> @jasongrout I have incorporated all your comments and submitted changes. Can you please have a look at it?

I made a few small changes (reformatting, and changing the inputType to also be readonly), and left one more comment. I'll test this tomorrow, but the code looks good to me now. Feel free to squash in my changes if you want, or leave them as separate commits - either way is fine.

Thanks again!

jakhani:
I have removed schema file changes. One check is failing and I am not able to get logs for the same. Can you please share detail on that in case you find anything?

jasongrout:
It seems like there is some problem with the logs on Travis right now. Hopefully my fix fixes whatever the error was.

jasongrout:
Thank you very much!

jakhani:
@jasongrout Thanks for helping and merging it quickly.

jasongrout:
And congratulations on your first contribution to ipywidgets!

maartenbreddels:
Indeed, thanks for this contribution! 👍

jasongrout:
I thought this was in the alpha version - can you use our alpha release?

We are working hard on finishing up what is needed for the 7.0 final release. If you need this widget before 7.0 is released, you can make a new python package with just this widget and use that in the meantime, right?

Issue #1242, opened 2017-03-30
Versioning changes


jasongrout:
The model specification for a widget (i.e., the list of model traits) is implemented at least two places: the kernel and the frontend. Because we want different frontend and kernel-side implementations to exist, the model state specification should probably not be tied to any particular package name or version. Also, I think the view should be defined by the model version it supports and the name/module. (i.e., the version should essentially be '*'). In order for that to happen, we probably need to get rid of custom messages (which are sent from the view to the kernel) and instead route all of that information through the state.

The model version should only support the major part of the version number (i.e., essentially be an integer). Any change in the actual attributes of the models (whether adding, deleting, or even changing intent of the attribute) should be a breaking change in the model specification.

CC @SylvainCorlay

SylvainCorlay:
Thanks for opening this. I want to give this the attention it deserves but cannot give a thorough response today. Will get back to this tomorrow.

jasongrout:
One problem we have the current versioning system is that we really have at least three different things we need to version in the jupyter-js-widgets package:

1. The version of the widget protocol that the widget manager speaks.
2. The version of the model specification for the core widgets (i.e., the attributes and what they mean).
3. The version of the views (this also impacts the models, since the views can send a custom message that will be received on the python side).

We could simplify the view version to just be the model version the view understands. In order to do that, we really need the model to encapsulate *all* messages that can go between the kernel and frontend, which we could do by making everything go through the state attributes (for example, for button clicks, a new button state attribute that is number of times clicked).

jasongrout:
See #1235 for more comments.

SylvainCorlay:
I am ready to discuss this now!

SylvainCorlay:
So from the in-person discussions with @jasongrout about the protocol specifications

**Asymmetry**

There are a few intrinsic reasons why the widget protocol will remain asymmetrical.

 - The back-end can be long-running, while the front-end is transient, and may connect to an already existing kernel. This is why we have the `request_state` message in the protocol, which is only sent by the front-end to the back-end.

 - We have a necessity of language agnosticism in the back end, so for the widget creation messages, the `class_name` has be be meaningful for any language. Having a registry in the back end becomes necessary so that we don't depend on any specific language syntax for addressing modules. On the other side, we have not met a use case for another front end than a web front end.

[...]

PR #1313, opened 2017-04-26
Make widget managers completely in charge of loading widgets

jasongrout:
This is an iteration on fixing #1242.

The basic idea is that we now consider the module/version to be semantic namespace (i.e., the spec for the model or view implementation), and the package/version to be a concrete implementation that can be loaded. The original thought was that if the package/version was not specified, it would default to the module/version. However, since package/version will be set on basic widgets classes, they must be overridden in custom widgets to be consistent.

### DIFF
--- a/ipywidgets/widgets/tests/test_interaction.py
+++ b/ipywidgets/widgets/tests/test_interaction.py
@@ -728,7 +728,7 @@ def test_get_interact_value():
     from ipywidgets.widgets import ValueWidget
     from traitlets import Unicode
     class TheAnswer(ValueWidget):
-        _model_module = Unicode('jupyter-js-widgets')
+        _model_name = Unicode('TheAnswer')
         description = Unicode()
         def get_interact_value(self):
             return 42
--- a/ipywidgets/widgets/valuewidget.py
+++ b/ipywidgets/widgets/valuewidget.py
@@ -4,10 +4,13 @@
 """Contains the ValueWidget class"""
 
 from .widget import Widget
+from traitlets import Any
 
 
 class ValueWidget(Widget):
     """Widget that can be used for the input of an interactive function"""
+
+    value = Any(help="The value of the widget.")
 
     def get_interact_value(self):
         """Return the value for this widget which should be passed to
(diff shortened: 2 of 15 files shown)


jasongrout:
> However, since package/version will be set on basic widgets classes, they must be overridden in custom widgets to be consistent.

Most custom widgets inherit from DOMWidget or just Widget. A hack to get around the above issue would be to just give the package on the core widgets. It's not quite clear where it puts Widget, DOMWidget, and LabeledDOMWidget.

jasongrout:
Okay, new thought: no package name stuff. The widget manager is responsible for loading correct versions of the models/views, full stop. Right now I'm thinking of three widget managers:

1. The notebook - widget libraries are installed and those are the ones available. It's up to the user to make sure the correct version of the library is installed.
2. JupyterLab - again, specific widget extensions are registered. It's up to the user to make sure the correct version of the library is installed.
3. The embedded widget manager - this is the only one that really uses the module_version as an npm package version. For many widgets, this will be correct. Any where it's not, the widget manager is responsible for translating module versions to package versions.

For each of these cases, the version of jupyter-js-widgets is special-cased to be the one we are extending.

So I'll strip this PR down to some of the cleanup I've done, and then the major change will be:

1. Make each widget manager completely in charge of loading the correct class for views and models. The default implementation will *not* assume require. I'm still undecided if the manager-base will provide code to load widgets from the jupyter-js-widgets package (leaning against it so that in the future, it's easier to separate out the base manager from our core widget implementation).
2. Change the documentation of the `_module_*` attributes to not refer to npm packages, but rather versions of the models/views (which can coincide with npm package names/versions by convention/coincidence).

maartenbreddels:
If that means that views will not be loaded via require by default (in the notebook) I think that is a big plus for development, since that makes js 'reloading' via actually work, like what i do [here](https://github.com/maartenbreddels/ipyvolume/blob/master/examples/dev.ipynb).

jasongrout:
Views would still be loaded using require -that's how things are done in the classic notebook

jasongrout:
@SylvainCorlay, @maartenbreddels - Any thoughts on this before merging?

PR #1314, opened 2017-04-26
fixes embedding for jupyter-sphinx

maartenbreddels:
and code is a bit more general to enable custom embedding such as in ipyvolume


jasongrout:
Looks good to me.

Issue #1320, opened 2017-04-28
DatePicker not setting value


JoshuaC3:
I am unable to get the DatePicker to return a value. I have tried setting this using the click, typed and arrow keys. Nothing helps.

I have put in the text example to show that this works. Here is my code.

```python
import ipywidgets as w

text_box = w.Text(placeholder='Type Contract Name')
start_date_widget = w.DatePicker(description='Start Date')
display(text_box, start_date_widget)

start_date = start_date_widget.value
contract = text_box.value

print(contract, start_date)

'hello' None
```

Chrome browser, using jupyter notebook through jupyterhub, ipywidgets 7.0.

Any debugging for system info needed then please let me know. Thanks.

jasongrout:
I can reproduce this. Looking at the websocket messages, I notice that setting the value gives back an error:
```
TypeError                                 Traceback (most recent call last)
/<snip>/ipywidgets/widgets/widget.py in [0;36m_handle_msg(self, msg)
598                 if 'buffer_paths' in data:
599                     _put_buffers(state, data['buffer_paths'], msg['buffers'])
--> 600                 self.set_state(state)
601 
602         # Handle a state request.

/<snip>/ipywidgets/widgets/widget.py in [0;36mset_state(self, sync_data)
479                     from_json = self.trait_metadata(name, 'from_json',
480                                                     self._trait_from_json)
--> 481                     self.set_trait(name, from_json(sync_data[name], self))
482 
483     def send(self, content, buffers=None):

/<snip>/ipywidgets/widgets/trait_types.py in [0;36mdatetime_from_json(js, manager)
 66     else:
 67         return dt.datetime(
---> 68             js['year'],
 69             js['month'] + [0;36m1,  # Months are 1-based in Python
 70             js['date'],

TypeError: string indices must be integers"
```
So apparently we have a bug in the python code.

This also points out that we need some place for any unhandled output to go.

jasongrout:
The python bug here should be pretty straightforward to fix, so setting as sprint-friendly. I'll open another issue about providing a place for unhandled output to show.

jasongrout:
> I'll open another issue about providing a place for unhandled output to show.

This is now https://github.com/jupyter/notebook/issues/2455. However, we can maybe mitigate this issue by having a single OutputWidget that is created by default when widgets are instantiated, and always syncing when in the scope of that output widget. If you want to see the log, just display that output widget.

jasongrout:
[...]

PR #1331, opened 2017-05-03
Fix a bug where custom serializers are not used

pbugnion:
Prior to this commit, the `WidgetModel$serialize` method silently ignored custom serializers. Now fixed.

To reproduce:

```python
import ipywidgets as widgets
d = widgets.DatePicker()
d
# this displays the date picker
d.value
#  => None (while the date picker is not set)
# change the value in the date picker manually
d.value
# => Without the fix, this is still None. With the fix, it's the correct date.
```

This addresses issue #1320 , but stretches beyond date pickers: any widget that uses custom serialisers would have not worked (I think).

TODO:

 - [x] use `Object.keys()` instead of `for ... in`
 - [x] unit test for null value in state
 - [x] unit test that verifies that custom serializers are used

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -409,7 +409,7 @@ class WidgetModel extends Backbone.Model {
                     // the default serializer just deep-copies the object
                     state[k] = JSON.parse(JSON.stringify(state[k]));
                 }
-                if (state[k].toJSON) {
+                if (state[k] && state[k].toJSON) {
                     state[k] = state[k].toJSON();
                 }
             } catch (e) {
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -401,7 +401,7 @@ class WidgetModel extends Backbone.Model {
      */
     serialize(state) {
         const serializers = (this.constructor as typeof WidgetModel).serializers || {};
-        for (const k in state) {
+        for (const k of Object.keys(state)) {
             try {
                 if (serializers[k] && serializers[k].serialize) {
                     state[k] = (serializers[k].serialize)(state[k], this);
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -401,7 +401,7 @@ class WidgetModel extends Backbone.Model {
      */
     serialize(state) {
         const serializers = (this.constructor as typeof WidgetModel).serializers || {};
-        for (const k of state) {
+        for (const k in state) {
             try {
                 if (serializers[k] && serializers[k].serialize) {
                     state[k] = (serializers[k].serialize)(state[k], this);
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -399,7 +399,7 @@ class WidgetModel extends Backbone.Model {
      * primitive object that is a snapshot of the widget state that may have
      * binary array buffers.
      */
-    serialize(state) {
+    serialize(state: {[key: string]: any}) {
         const serializers = (this.constructor as typeof WidgetModel).serializers || {};
         for (const k of Object.keys(state)) {
             try {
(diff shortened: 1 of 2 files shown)


pbugnion:
Is it worth adding a test for this? If so, some initial pointers would be great!

pbugnion:
Hmm -- this has broken the unit tests (presumably because it exercises code that wasn't exercised before). I'll look into why.

pbugnion:
The error in the test comes from the fact that `manager-base` expects the model to define attributes like `_view_name` in the state, but the test model doesn't provide those. They then get serialised to `null`, which causes the code in `WidgetModel$serialize` to fail because `null` doesn't have a `toJSON` property.

AFAICT, there are two ways to make the test pass.

1. Allow models to not pass in required properties like `_view_name` by changing this:

```javascript
                if (state[k].toJSON) {
                    state[k] = state[k].toJSON();
                }
```

to this:

```
                if (state[k] && state[k].toJSON) {
                    state[k] = state[k].toJSON();
                }
```

I'm not sure what the implications of having an incomplete model are.

2. Give the test-widget a non-null value for `_view_name`.

Some guidance here would be great.

maartenbreddels:
Hi @pbugnion,

thanks for your (first) contribution!
I'd rather keep fix the unittest, to fix the unittest, you can edit `test/src/widget_test.ts` to include the `_view_name`:

```javascript
        return this.manager.new_widget({
            model_module: 'jupyter-js-widgets',
            model_name: 'WidgetModel',
            model_id: this.modelId,
            widget_class: 'ipywidgets.Widget'
        }, {
            _view_module: 'jupyter-js-widgets',
            _view_name: 'WidgetView',
        }).then(model => {
```
Then you can skip the check you just added. I can also change your PR (I think I can).

pbugnion:
@maartenbreddels

> thanks for your (first) contribution!

Thanks! And thanks for the feedback!

> I'd rather keep fix the unitest

I've done a bit more research, and I don't think it's enough to change the test to add `_view_name` to the test widget. I think we need the change to the `serialize` method to handle `null` values in the state (unless I'm missing something?). This wasn't a problem prior to this PR, because the entire `for` loop over the state (line 402) wasn't being run.

I still want to add two unit tests to exercise:
 - custom serialisers
 - null values in the model state

maartenbreddels:
Why isn't adding `_view_name` enough? It passed the unittest here. Also, we need to make sure that state[k] being null is valid? I think @jasongrout should know, he's been actively working on part of the code. We might be hiding bugs if we add thing check.

pbugnion:
> Why isn't adding `_view_name` enough?

This makes the unit test pass, but presumably there *may* be values in the state which are null. Presumably, if I write a widget like:

```python
# Python
class MyWidget(widgets.DOMWidget):
    # normal _view_name things etc.
    _my_null_trait = Int(None, allow_none=True).sync(true)
```

```javascript
// JavaScript
class MyWidgetModel extends DOMWidgetModel {
    defaults: {
        _my_null_trait: null;
    }
}
```

... This results in a model with state (as far as I can tell):

```javascript
{
    _my_null_trait: null
// other key-value pairs
}
```

That's why I want to write a unit test that exercises null values in the model state.

> Also, we need to make sure that state[k] being null is valid?

Agreed, I'm not 100% sure about this -- some help here would be useful.

maartenbreddels:
Yes, you are right, we do need a null check there.
@jasongrout: I checked this fix , and it fixes the issue with the datepicker, if you think the null check should be there (I think it should), I think we can merge this.

jasongrout:
> Yes, you are right, we do need a null check there.

Definitely +1. It was an oversight to not consider a null state value.

jasongrout:
@pbugnion - can you add a unit test serializing a null value in the state?

pbugnion:
Yes -- I think this PR still needs:

 - [ ] use `Object.keys()` instead of `for ... in`
 - [ ] unit test for null value in state
 - [ ] unit test that verifies that custom serializers are used

I'm busy this afternoon, but I should be able to work through this over the next couple of days.

maartenbreddels:
excellent 👍

jasongrout:
Awesome, thanks!

pbugnion:
I think this is ready for another look.

pbugnion:
I think this is ready for another look.

jasongrout:
It looks great to me. I'll test later and merge. Thanks!

jasongrout:
Thanks!

pbugnion:
Great! Thanks a lot for the help.

maartenbreddels:
Excellent work 👍 !

PR #1332, opened 2017-05-04
Change tuple by list in order to avoid ValueError's when Dropdown is …

kikocorreoso:
Hi all,

The narrative in the example of "Using interact" is wrong as if a tuple of strings is used it results in a `ValueError`. The docs say that a list or a dict should be used to create a dropdown widget using `interact`.


jasongrout:
Thanks!

PR #1334, opened 2017-05-05
WIP Don’t call the view general update method if just `_view_count` changes.

jasongrout:
This may solve the issues we're having with _view_count changes in pythreejs causing infinite event trigger loops: https://github.com/jovyan/pythreejs/issues/95

CC @maartenbreddels, @SylvainCorlay

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -651,7 +651,14 @@ abstract class WidgetView extends NativeView<WidgetModel> {
      * Public constructor.
      */
     initialize(parameters) {
-        this.listenTo(this.model, 'change', this.update);
+        this.listenTo(this.model, 'change', () => {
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Ok, if this solves it I'm ok with it. But I did get annoyed by all the websocket msg'es flying around, making debugging difficult. So do you think we should have an opt-in option, where by default it shouldn't keep track of the _view_count, but only when a new property (say `_record_view_count `) is set to True?

jasongrout:
> But I did get annoyed by all the websocket msg'es flying around, making debugging difficult.

True, it doesn't solve that issue.

> but only when a new property (say _record_view_count) is set to True?

I guess we get some weird semantics if the record attribute is set after views are created. Another thing we could do is to let _view_count be None, and if it is None (i.e., null), then *don't* update it.

SylvainCorlay:
This looks good to me.

PR #1335, opened 2017-05-05
Default _view_count to be null, which signifies that counting won’t be done.

jasongrout:
CC @maartenbreddels 

- [x] Test by hand


maartenbreddels:
👍  makes me really happy!

PR #1342, opened 2017-05-11
Fixes the dependency of IPython version when Python version is < 3.3

ribamar-santarosa:
IPython 6.0+ does not support Python 2.7, 3.0, 3.1, or 3.2, but accordingly
to the current specification, it is a valid requirement, even if Python < 3.3 is installed.

This patch makes IPython 6.0+ valid requirement only for Python >= 3.3. Otherwise, it requires IPython to be between 4 and 6. 

Without this solution, I am not being able to install ipywidgets:

````
# pip install jupyter
# Collecting jupyter
#   Downloading jupyter-1.0.0-py2.py3-none-any.whl
# ...
# Collecting ipython>=4.0.0 (from ipywidgets->jupyter)
#   Downloading ipython-6.0.0.tar.gz (5.1MB)
#     100% |████████████████████████████████| 5.1MB 164kB/s 
#     Complete output from command python setup.py egg_info:
#     
#     IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.
#     When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
````

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,6 @@ if 'develop' in sys.argv or any(a.startswith('bdist') for a in sys.argv):
 
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
-    'ipython>=4.0.0',
     'ipykernel>=4.5.1',
     'traitlets>=4.3.1',
     # Requiring nbformat to specify bugfix version which is not required by
@@ -122,6 +121,8 @@ install_requires = setuptools_args['install_requires'] = [
 ]
 
 extras_require = setuptools_args['extras_require'] = {
+    ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
+    ':python_version>="3.3"': ['ipython>=4.0.0'],
     'test:python_version=="2.7"': ['mock'],
     'test': ['nose'],
 }


jasongrout:
I'm not very familiar with extras_require - doesn't this change mean that ipython is not required unless the test optional extra is requested? In other words, this seems like it is essentially removing the ipython requirement.

jasongrout:
Perhaps it is better to dynamically calculate the ipython version depending on the current python version, but still use it in the `install_requires` statement?

ribamar-santarosa:
. I am not an expert on pip packaging, so you do good about being careful. 

. Accordingly to the wheels documentation [1] `extra_requires` is the recommended way to define conditional dependencies. 

. Your question about "is a conditional dependency (as per the definition in this context) any lesser requirement than a normal `install_requires`?" is relevant."
.. The documentation at [1] suggests that they are actual requirements, but can be argued not to be definitive (that's no specification)
.. I searched and couldn't find any manifestation about it -- neither positive nor negative statement. 
..  For me it makes sense that a conditional requirement is no lesser than any requirement, but the ideal would have to find the opinion of the original designers, because they may decide something that does not make much sense anytime. 

. I wouldn't suggest at all coding that dynamically. It makes difficult for packagers, parsers and humans to calculate or debug the dependency tree. This issue at spot is a very normal use case that package managers need to deal with, because very often a package is updated but is not backwards compatible, or introduces a bug.  If there is nothing in the current package definition language that provides the desired functionality, I would go behind the language specifiers and propose it. But it is 2017, I don't think that is necessary. 

[1] https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies

jasongrout:
ah, here is the key from that wheel documentation:

> The extra named ‘’ signifies a default requirement, as if it was passed to install_requires.

In your PR, you are giving the requirement on the `test` extra. Try doing `':python_version...'` instead of `'test:python_version...'`.

ribamar-santarosa:
Sure. Ammended.

jasongrout:
I checked and it works for both python 2.7 and 3.6. Thanks!

ribamar-santarosa:
Great. If this brings up any issues, you can count on me or the guys at github.com/Bright-Computing/ to help you!

jasongrout:
Awesome. Thanks for the contribution, and we hope to collaborate more!

PR #1343, opened 2017-05-11
Install python ipywidgets before jLab

maxnoy:
Install python ipywidgets before jupyter-lab which may fail if there is an issue with jLab install (wrong version, etc).


jasongrout:
Great idea. Thanks!

Issue #1322, opened 2017-04-28
DatePicker Typing in date has odd behaviour


JoshuaC3:
If I try to type 01012017 into the DatePicker I get the following:

![image](https://cloud.githubusercontent.com/assets/11645712/25530114/f6bff7c0-2c1c-11e7-8315-9ab8a07e8fed.png)

It all goes well up until the 0 in 2017. The does the reset on the day and month to dd mm and the 20 at the start of the year to 00.

I can go back, once 2017 is typed and retype 0101

JoshuaC3:
I will add that this problem occurred before the ipywidget 7.0 update

jasongrout:
I'm seeing the same issue on basically the current alpha of 7.0. The input does seem to work on a plain html page. Setting as sprint-friendly because it *should* be straightforward to track down...

pbugnion:
As far as I can tell, the problem goes away if one removes conversions of the date to/from UTC. 

Instead of this:

```javascript
    private _update_value() {
        var value = this.model.get('value');
        this._datepicker.valueAsDate = createDateAsUTC(value);
    }

    private _picker_change() {
        this.model.set('value', convertDateToUTC(this._datepicker.valueAsDate));
        this.touch();
    }
```

having this:

```javascript
    private _update_value() {
        var value = this.model.get('value');
        this._datepicker.valueAsDate = value;
    }

    private _picker_change() {
        this.model.set('value', this._datepicker.valueAsDate);
        this.touch();
    }
```

I think that the underlying problem with `convertDateToUTC` is caused by the Javascript Date constructor having slightly different meanings attached to two digit years. For instance, in the chrome console:

```javascript
> var d = new Date("0005-04-28")
> d
Thu Apr 28    5 01:00:00 GMT+0100 (BST)
> d.getFullYear()
5
> new Date(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDay())
Tue Apr 04 1905 00:00:00 GMT+0100 (BST) // Check the year!!
```

I presume the rationale for converting the dates to and from UTC is to avoid having to pass timezone information to the backend? 

I can see the following possible fixes:

 - Also serialize timezone information in the JSON object that gets serialised and pass this into the datetime constructor in Python.
 - instead of passing a date that's decomposed into elements, pass either an ISO string representation or a number of millis since the epoch. This is nice since (as far as I can tell), both of these minimise the amount of date time manipulation we have to do ourselves.

[...]

PR #1348, opened 2017-05-12
Fix datepicker reset on early dates

pbugnion:
This addresses issue #1322.

Prior to this PR, all the characters were reset when a user started typing the year in the datepicker. This arose because `new Date(year, ...)` interprets a number below 100 as a two-digit year (i.e. 5 gets mapped to 1905), whereas the datepicker interprets it as a 4-digit year (i.e. 5 gets mapped to 5AD). Every time the user typed a character, the view forced a roundtrip to a timestamp and back to convert the date to UTC.

The fix implemented here removes that roundtrip: dates in the view and the JS model are stored as `Date` object without trying to force them to UTC -- there is no loss of information here since dates are time-zone aware anyway. The conversion to UTC happens just before serialization: the date gets passed to the Python layer as a UTC date.

This PR also:
 - changes the trait in the Python model to a `datetime.date`, rather than a `datetime.datetime`. The datepicker object doesn't let us set the time, so it seems a bit strange to pretend we can give that information back to the user. This change is orthogonal to the fix, so we can remove it if you don't think it's the right choice.
 - implements tests for the JS view.

What's left to do:

 - [x] unit tests for {de,}serializers on both the JS and Python side

### DIFF
--- a/ipywidgets/widgets/tests/test_traits.py
+++ b/ipywidgets/widgets/tests/test_traits.py
@@ -27,21 +27,42 @@ class TestColor(TraitTestBase):
 class TestDateSerialization(TestCase):
 
     def setUp(self):
-        self.from_json = date_serialization['from_json']
         self.to_json = date_serialization['to_json']
         self.dummy_manager = None
 
     def test_serialize_none(self):
         self.assertIs(self.to_json(None, self.dummy_manager), None)
 
-    def serialize_date(self):
-        date = dt.date(1900, 2, 5)
+    def test_serialize_date(self):
+        date = dt.date(1900, 2, 18)
         expected = {
             'year': 1900,
             'month': 1,
-            'date': 4
+            'date': 18
         }
-        self.assertIs(self.to_json(date, self.dummy_manager), expected)
+        self.assertEqual(self.to_json(date, self.dummy_manager), expected)
+
+
+class TestDateDeserialization(TestCase):
+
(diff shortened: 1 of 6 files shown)


pbugnion:
This is ready for another look -- besides the bug fix, the only major change is to switch the traitlet in the widget to a `datetime.date` rather than a `datetime.datetime`. This does break backwards compatibility, so I'm happy to revert that change if need be.

jasongrout:
I tweaked things a bit and pushed to your branch. I agree with the change to using simple dates. Since we are using simple dates, though, I don't think there's any need to deal with UTC at all - just use local/naive dates. I also tweaked some of the documentation.

jasongrout:
@pbugnion - your changes look great. If my changes look good to you, let's merge. Do you approve my changes?

jasongrout:
Actually, it looks like we need to normalize to UTC dates in javascript - that seems to be what the input is giving us.

jasongrout:
> Since we are using simple dates, though, I don't think there's any need to deal with UTC at all - just use local/naive dates.

Actually, it looks like we need to normalize to UTC dates in javascript - that seems to be what the input is giving us.

pbugnion:
Thanks for the help! I've added a test for dates before 100AD. I think this is ready for merging.

jasongrout:
Great, thanks!

Issue #1321, opened 2017-04-28
Create a Log output widget to log any output from sync messages


jasongrout:
In #1320, we have the problem that a sync message is causing an error that has no place to be displayed in the front end. Until the front end has a place to display unhandled output, perhaps we should create an output widget that stores any output from sync handling.

I *think* this should be fairly straightforward - create an output widget when widgets are imported, and make sure we are using the output widget when we actually set the state of any widget.

jasongrout:
By the way, this may make calling `this.touch()` on widget views unnecessary to save the changes - maybe now we should *always* call `this.model.save_changes()` (but leave in `this.touch()` for backwards compatibility).

maxnoy:
This is a temporary solution that would allow capturing widget output until a dedicated output tab for all orphan output was implemented in jupyter-lab and/or jupyter-notebook.

The complexity is that the amount of output could be quite large (messages/errors can be generated for each widget change). 

This can be managed by either clearing output for each widget message or by having an adjustable-depth buffer.  The former would not work as the output widget utilizes the same widget messaging mechanism, resulting in the messages being cleared immediately by the update on the output widget. And the latter is a substantial amount of work/ux that may as well be redirected to the long-term solution.

jasongrout:
Agreed! Closing this in favor of the notebook issue!

PR #1353, opened 2017-05-14
Regression handling callback output

pbugnion:
In ipywidgets 6.0.0 (and all the way through to the merge of PR #1274), the outputs of widget callbacks would be printed to the widget output area:

screen shot 2017-05-14 at 07 50 50

This stops working in PR 1274. I presume this is accidental? (if not, the rest of this PR is irrelevant).

As far as I can tell, the issue arises because callback messages do not have the correct `iopub.output` and `iopub.handlers` handlers any more. This is because we don't explicitly bind the handlers of the view to the handlers of the output area.

This PR creates a temporary fix by re-enabling the `WidgetManager$callbacks(view)` method which defines the `iopub` callbacks. Prior to PR #1274, this looked for the cell in which the view is defined. We don't have access to this now, but we do have access to the output area when the widget is defined, so this PR just injects a reference to the output area into the view's options. 

This fix is just a proof of concept, it's not really meant to be the ultimate solution, but some guidance here would be great:

 - Is this even a regression? Or was it intentional?
 - Where should we re-bind the output callbacks? We could do it directly when the widget is rendered, rather than relying on a method on the manager, unless the 
 - Am I missing places where views are defined?

### DIFF
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -196,12 +196,10 @@ class WidgetModel extends Backbone.Model {
                 this.trigger('msg:custom', msg.content.data.content, msg.buffers);
                 return Promise.resolve();
             case 'display':
-                console.log('disp with output: ', this.widget_manager.displayWithOutput)
                 if (this.widget_manager.displayWithOutput) {
                     return;
                 }
                 this.state_change = this.state_change.then(() => {
-                    console.log('display!')
                     this.widget_manager.display_model(msg, this);
                 }).catch(utils.reject('Could not process display view msg', true));
                 return this.state_change;
--- a/jupyter-js-widgets/src/widget.ts
+++ b/jupyter-js-widgets/src/widget.ts
@@ -196,10 +196,12 @@ class WidgetModel extends Backbone.Model {
                 this.trigger('msg:custom', msg.content.data.content, msg.buffers);
                 return Promise.resolve();
             case 'display':
+                console.log('disp with output: ', this.widget_manager.displayWithOutput)
                 if (this.widget_manager.displayWithOutput) {
                     return;
                 }
                 this.state_change = this.state_change.then(() => {
+                    console.log('display!')
                     this.widget_manager.display_model(msg, this);
                 }).catch(utils.reject('Could not process display view msg', true));
                 return this.state_change;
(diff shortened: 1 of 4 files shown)


jasongrout:
It wasn't an accident, but I viewed it as an unfortunate consequence of #1274. For example, what if a widget observe callback called clear_output? That would clear the view that sent the message. Also, now that we can have output and widgets intermixed, it seemed confusing to have output being appended below all of the other widgets and output. However, especially given how complicated #1321 turned out to be, perhaps this can be reconsidered. I still think the proper behavior is now to create an Output widget, but this workaround at least allows errors to appear to the user instead of being hidden.

maartenbreddels:
Yes, I'd say better have 'odd' behaviour/output that can be fixed with using Output, than no msg at all.

jasongrout:
To reproduce:

```
x = IntSlider()
def onchange(change):
    print(change)
x.observe(change, 'value')
x
```

jasongrout:
@pbugnion - can you review my changes to this PR? Basically, I simplified the callback construction, and used the display_model options argument.

jasongrout:
> Where should we re-bind the output callbacks? We could do it directly when the widget is rendered, rather than relying on a method on the manager, unless the

We could set the iopub callback dictionary in the view options when we create the view options. That would save creating the bound function each time a message is sent. Thoughts? I'm not sure it's worth it.

pbugnion:
Thanks -- your changes make the code a lot cleaner. I hadn't realised the options passed into the model also get propagated to the views.

The output still wasn't appearing if the callback was bound to a child model. I've added a fix for passing the output area to children of the outermost widget (I copied the right method from pre PR 1274). To reproduce:

```python
import ipywidgets as widgets

slider = widgets.IntSlider()
box = widgets.HBox([slider])
outer_box = widgets.VBox([box])
def on_change(change):
    print('hello')
slider.observe(on_change, names='value')
outer_box
```

pbugnion:
> We could set the iopub callback dictionary in the view options when we create the view options. That would save creating the bound function each time a message is sent. Thoughts? I'm not sure it's worth it.

I've had a go at doing this. Let me know what you think. It seems cleaner to just bind the iopub callbacks once, but I'm could see benefits either way.

jasongrout:
> The output still wasn't appearing if the callback was bound to a child model. I've added a fix for passing the output area to children of the outermost widget (I copied the right method from pre PR 1274).

Oh right, good catch. I'm pushing an update to the comment + small simplification.

jasongrout:
I'll wait for tests to pass before merging this. Thanks!

PR #1354, opened 2017-05-16
Delete the clear_state commlessOnly argument

jasongrout:
This is broken, and doesn't seem to be used.


jasongrout:
(It was originally introduced by @jdfreder a long time ago to deal with loading widget state on the page, but isn't used currently.)

jasongrout:
I also fixed some date time tests that were failing on my computer (presumably because my computer is not set to UTC.)

PR #1361, opened 2017-05-18
Add a tutorial for using widgets asynchronously, waiting for widget changes from the frontend

### LINKED ISSUE #1287: Pulling updated widget values from front-end
# Summary:

Add the examples in https://github.com/jupyter-widgets/ipywidgets/issues/1287#issuecomment-301982976 and https://github.com/jupyter-widgets/ipywidgets/issues/1287#issuecomment-301990689 to the docs as an advanced example for having a request/reply pattern with widgets using generator-like syntax

# Original issue

I can check for updates to a widget value using `observe`:

```python
import ipywidgets as widgets
import time

w = widgets.Dropdown(
    options=['Addition', 'Multiplication', 'Subtraction'],
    value='Addition',
    description='Task:',
)

def on_change(change):
    print("changed to %s" % change['new'])

w.observe(on_change)

display(w)
```

But how can I do that during another operation, for example:

```python
for i in range(100):
    time.sleep(1)
    # pull for changes to w here.
    # if w.has_changed:
    #     print(w.value)
```

I tried doing something like 

```python
import IPython
ipython = IPython.get_ipython()
ipython.kernel.do_one_iteration()
```
to pull for updates. But is seems unreliable. Is there a better way to do this?

jasongrout:
Fixes #1287.

### DIFF
--- a/docs/source/examples/Index.ipynb
+++ b/docs/source/examples/Index.ipynb
@@ -37,7 +37,8 @@
     "- [Widget Events](Widget Events.ipynb)  \n",
     "- [Widget List](Widget List.ipynb)  \n",
     "- [Widget Styling](Widget Styling.ipynb)    \n",
-    "- [Widget Custom](Widget Custom.ipynb)"
+    "- [Widget Custom](Widget Custom.ipynb)\n",
+    "- [Widget Asynchronous](Widget Asynchronous.ipynb): how to pause and listen in the kernel for widget changes in the frontend."
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


jasongrout:
@hannorein, can you see if the document here (https://github.com/jasongrout/ipywidgets/blob/f51a7badd28fd8510b0e74b4858e840b2ca3a87c/docs/source/examples/Widget%20Asynchronous.ipynb) works well? Any suggestions on wording or code would be appreciated too.

hannorein:
This is super helpful and would have saved me many hours! Thanks for writing it up!

PR #1367, opened 2017-05-19
Avoid resetting the date on invalid keyboard input

### LINKED ISSUE #1115: date picker may empty the value using keyboard navigation
Steps to reproduce:

* open a date picker widget or use one from the [docs](http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Date-picker)
* set the date to 01/31/2017
* Increment the month by 1 using keyboard (arrow up)

Expected behavior: the date changes  to 02/31/2017 and the widget border becomes red because the value is invalid or the day gets reset or the day becomes 28.

Observed behavior: the date changes to mm/dd/yyyy (blank).

pbugnion:
Prior to this commit, the datepicker value was passed to the user for every change, even when it was invalid. This caused errors when the date was partly filled in: it would get reset to empty, losing the partial (invalid) state.

To reproduce:

 - create a datepicker (in Chrome) and set the date to 31st January 2000
 - click on the months field and cycle through the months -- prior to this PR, the input would get reset every time we encountered a month with fewer than 31 days. Now, the datepicker border just goes red and the input is not pushed to the backend.

Fixes #1115 .

It'd be great to get feedback on:

 - giving the user clues that their date is not getting transmitted to the backend. At the moment, I've just added a small orange border. Is that too subtle? Is there another way we do this for other widgets?
 - testing: it's surprisingly hard to inject a bad string into the datepicker programatically, as far as I can tell. I think we can either simulate the key presses directly, which seems error prone and brittle, or somehow mock out the datepicker element so we can give it arbitrary state, or assume that the changes are sufficiently simple that they don't need testing. I'm in favour of the latter, I think, but happy to implement a test if we think it's useful.

### DIFF
--- a/jupyter-js-widgets/src/widget_date.ts
+++ b/jupyter-js-widgets/src/widget_date.ts
@@ -104,8 +104,10 @@ class DatePickerView extends LabeledDOMWidgetView {
     }
 
     private _picker_change() {
-        this.model.set('value', this._datepicker.valueAsDate);
-        this.touch();
+        if (!this._datepicker.validity.badInput) {
+            this.model.set('value', this._datepicker.valueAsDate);
+            this.touch();
+        }
     }
 
     private _datepicker: HTMLInputElement;
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -819,6 +819,10 @@
     border-color: var(--jp-widgets-input-focus-border-color);
 }
 
+.widget-datepicker input[type="date"]:invalid {
+    border-color: var(--jp-warn-color1);
+}
+
 /* Play Widget */
 
 .widget-play {


jupyter-js-widgets/src/widget_date.ts
```diff
@@ -104,8 +104,10 @@ class DatePickerView extends LabeledDOMWidgetView {
     }
 
     private _picker_change() {
-        this.model.set('value', this._datepicker.valueAsDate);
-        this.touch();
+        if (!this._datepicker.validity.badInput) {
```
jasongrout: Wow, I had no idea `.validity` existed. That's great!

jasongrout:
> testing: it's surprisingly hard to inject a bad string into the datepicker programatically, as far as I can tell. I think we can either simulate the key presses directly, which seems error prone and brittle, or somehow mock out the datepicker element so we can give it arbitrary state, or assume that the changes are sufficiently simple that they don't need testing. I'm in favour of the latter, I think, but happy to implement a test if we think it's useful.

I think the latter is also fine for now.



> giving the user clues that their date is not getting transmitted to the backend. At the moment, I've just added a small orange border. Is that too subtle? Is there another way we do this for other widgets?

I think this is fine. The only somewhat similar thing is the readout gains a dropshadow when it is cutting off input, but that's different.

jasongrout:
> I think the latter is also fine for now.

I think that as we develop our testing infrastructure, particularly in light of headless Chrome 59, we may revisit this.

jasongrout:
Tested manually, and it seems to work great. Thanks!

pbugnion:
Great! Thank you.

PR #1368, opened 2017-05-19
Make WidgetModel.set docstring more specific

vidartf:
Sometimes you might want to set several attributes of a model at the same time, with only one "change" notification. `WidgetModel.set()` can handle this directly, so this PR updates the docstring to be more explicit about this.


jasongrout:
Thanks!

PR #1369, opened 2017-05-19
Make a disconnected icon for widgets that are not live

### LINKED ISSUE #1275: Move the disconnected state display into base widget manager/widget view
In order to have a more consistent UI for widgets not currently connected with a kernel, I think we should move the notion of a disconnected widget into the base widget manager. Right now in the classic notebook, we have a small broken link icon indicating a set of widgets not connected to a kernel. We should have some similar notion for other widget views, or at least the top-level containers.

The widgets formed by a widget manager should be considered broken when a kernel is disconnected, killed, restarted, or dead.

jasongrout:
Fixes #1275.

### DIFF
--- a/jupyter-js-widgets/css/widgets-base.css
+++ b/jupyter-js-widgets/css/widgets-base.css
@@ -11,6 +11,7 @@
 
 :root {
     --jp-widgets-color: var(--jp-content-font-color1);
+    --jp-widgets-disconnected-color: #d9534f;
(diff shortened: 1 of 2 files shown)


jasongrout:
Restarting a kernel should also set this disconnected icon...

jasongrout:
It seems that the ways to know if a widget changes connected state are extremely limited. So let's merge this and work on that in another PR.

PR #1370, opened 2017-05-19
Fix widget saving in the classic notebook.

jasongrout:
1. Disable the automatic widget state saving
2. Delete the callbacks for loading and saving state
3. Automatically restore any state saved in notebook metadata.

Also, pass to the set_state method the appropriate metadata (before, it was assuming an old structure of the metadata key).

This may not render widgets in a notebook if it was saved using older code.

### DIFF
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -24,6 +24,11 @@ var WidgetManager = function (comm_manager, notebook) {
     this.keyboard_manager = notebook.keyboard_manager;
     this.comm_manager = comm_manager;
 
+    var widget_md = notebook.metadata.widgets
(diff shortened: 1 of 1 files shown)


jasongrout:
We have the possibility of continuing cruft, like in this situation:

1. A notebook is saved with widget state
2. The notebook is loaded and some widgets are created. Some widget state came from the metadata, some came from live kernel execution
3. Widget state is saved - *all* state, including state saved in the previous notebook.

I provided a Clear command to prune state, but to really clear state, you'd need to clear the notebook state, then *reload* the page to get a fresh clean widget manager, and *then* save state.

maartenbreddels:
excellent work, and you beat me to it 👍

PR #1371, opened 2017-05-19
Change the documentation inline math from $x$ to \(x\) to support docs rendering.

### LINKED ISSUE #1267: documentation on readthedocs fails to show rendered LaTeX
I would have expected the captions on the attached screenshot to be rendered as equations

screen shot 2017-04-07 at 10 09 40

jasongrout:
Changes are a bit more because of some churn thanks to notebook saving.

Fixes #1267


jasongrout:
Thanks!

PR #1372, opened 2017-05-19
Updates to use jupyter_alabaster_theme

charnpreetsingh:
It's really nice that this subproject is using the new theme!

To use the theme `captions` cannot be used in the `toctree` and there needs to be one `toctree`.
This PR consolidates User Guide and Developers Docs into their own sections within the toctree. There is a single Table of Contents on index now.

Leaving `captions` in the `toctree` leads to this issue in the mobile version
screen shot 2017-05-19 at 2 09 42 pm

Also:
- removed `jupyter_alabaster_theme` import in conf.py
- updated copyright text to match jupyter.org website

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -5,7 +5,6 @@
 import os
 import subprocess
 import sys
-import jupyter_alabaster_theme
 import recommonmark.parser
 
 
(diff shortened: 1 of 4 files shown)


jasongrout:
Thanks!

Issue #1373, opened 2017-05-20
Label width style attribute


jasongrout:
One of the main reasons we need a Label widget is to set the width of the label of a control. Instead, we should just make the label width a style attribute of all labeled widgets (and control the indent property of the checkbox).

- [ ] Move LabeledDOMWidget to a new file
- [ ] Make a new LabeledDOMStyle class that has an attribute for the label width
  - [ ] Should we make minwidth/maxwidth/width properties?
  - [ ] Should we just make a width property, and change the current labels to just have a set width, rather than a min/max width?
- [ ] Make labeled dom widget subclasses styles inherit from the labeleddomstyle

I think we may be able to deprecate the label widget then? Most other uses for the label widget should be shifted to the HTML or HTMLMath widget.

We might also address the discrepancy between "description" and "label". In the code we call it label (that's the html element), but the widget attribute is called description. Does anyone have a preference for one or the other? Label is much shorter, but I'm happy to call it description for backwards compatibility. For example, should we call the style property `description_width` or `label_width`?

CC @SylvainCorlay, @maartenbreddels.

PR #1376, opened 2017-05-22
Make label width configurable

jasongrout:
This implements the ideas behind #1373. The idea is that the current label width is still too brittle - we have a min and a max width, but it's too hard to get something much smaller or larger. This introduces a description_width style attribute which lets you easily set the width. As part of this, I eliminated the max-width on the label, since now it is very easy to extend the width of a label. For example, `style=dict(description_width='initial')` makes the label be exactly as long as the text.

This also affects the checkbox indent. @maxnoy - what do you think about `description_width` affecting the checkbox `indent` width? It seems a *bit* confusing that `description_width` affects the width of the indent, not the text.

```python
from ipywidgets import *
style=dict(description_width='50px')
widgets = []
widgets.append(IntSlider(description='slide', style=style))
widgets.append(Dropdown(description='drop', style=style))
widgets.append(Checkbox(description='check', style=style))
VBox(widgets)
```

screen shot 2017-05-22 at 11 48 04 am


ipywidgets/widgets/domwidget.py
```diff
@@ -37,8 +38,15 @@ def remove_class(self, className):
         return self
 
 
+class DescriptionStyle(Style, Widget):
+    """Button style widget."""
+    _model_name = Unicode('SliderStyleModel').tag(sync=True)
```
blink1073: misnamed?

jasongrout: Yep, thanks.

jasongrout:
Opening another issue for discussion on the label width: #1377.

maxnoy:
@jasongrout, this seems like very useful functionality. Maybe if there was a notion of description_width vs content_width, it would be a bit clearer?

jasongrout:
I think we usually have the widgets automatically expand the content width, so there is a correlation between widget width and content width (i.e., to make the content wider, just make the widget wider).

Can we just handle this in documentation, with the explanation that the checkbox description is on the right?

PR #1384, opened 2017-05-23
Override the prompt area correction for widths of output areas.

### LINKED ISSUE #1311: How to define content width of Output widget
```python
    from ipywidgets import Output

    out = Output()
    with out:
        print(''.join('{:.>5}'.format(n) for n in range(5, 150, 5)))

    out.layout.border = '1.5px solid'
    #out.layout.width = '100%'    # no influence
    #out.layout.margin = '0px 0px 0px 0px'    # no influence

    display(out)
```
grafik

How can I use the whole width of the Output widget which should be approximately 115 characters in this case?

jasongrout:
Fixes #1311

### DIFF
--- a/widgetsnbextension/src/widget_output.css
+++ b/widgetsnbextension/src/widget_output.css
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+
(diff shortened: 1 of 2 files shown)


jasongrout:
After: 
screen shot 2017-05-23 at 6 13 13 pm

jasongrout:
`output_subarea` also has a padding of 0.4em, which is applied twice since we are nested inside of an output area. I'm leaving that on for now, but we could examine taking it off in the future as well, maybe.

(with padding of 0.4em gone...)
screen shot 2017-05-23 at 6 22 05 pm

PR #1387, opened 2017-05-24
Embed code generation from Python

### LINKED ISSUE #1241: build static-export from Python
It would be very useful if an offline Python script could produce a working static-widget export. That is, be able to get the result of Widgets > Embed Widgets from ipywidgets directly, run outside IPython / Jupyter.

This would allow standalone scripts to generate pages with bqplot or ipyvolume figures, for example.

How much work would this be? The model information should all be available, right? Is there anything more?

vidartf:
Adds functions to the Python side API for generating embeddable javascript/HTML representation of the widgets. Based on code from [ipyvolume](https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/embed.py).

Resolves #1241.

### TODOs

 - [x] Documentation
 - [x] Tests
 - [x] Figure out source of error: `Could not create model: Model name: DescriptionStyleModel` that appears regularly for full saves.


jasongrout:
I just added that description style model. Happy to help if I can.

vidartf:
Ah, then I'm guessing it is simply the fact that it is not available in the version on the CDN yet.

jasongrout:
Right, it isn't in the cdn yet

maartenbreddels:
Excellent work, glad to see this going upstream. I recommend being able to override the embed url for these kinds of things. Also, would be nice the the html template can be passed as argument to `embed_minimal_html`.

vidartf:
@maartenbreddels : I figured if someone wanted more custom HTML, they could simply use the snippet code, and then wrap it themselves (`embed_minimal_html` is mostly intended as a dev tool).

Having the url overridable sounds like a decent idea.

vidartf:
Regarding the javascript embed parser: Does the embedded widget views need to be specified in dependency order, or does the order not matter?

jasongrout:
> Does the embedded widget views need to be specified in dependency order, or does the order not matter?

I'm not quite sure what you mean. What dependencies do you see between views?

maartenbreddels:
Order doesn't matter afaik.

vidartf:
>  What dependencies do you see between views?

@jasongrout Not entirely sure, but I figured there might be some dependencies e.g. with the layout views. It was mostly catalyzed by the warnings I see for the exported minimal HTML: `Style not applied because a parent view does not exist (widget_layout.js:116)`, although that might be related to the `DescriptionStyleModel` things(?).

Either way I see that the views need a bit further attention, e.g. for the following case:
```python
import ipywidgets
import ipywidgets.widgets.embed as embed

w1 = ipywidgets.IntSlider(min=0, max=100)
w2 = ipywidgets.IntText()
box = ipywidgets.HBox(children=[w1, w2])

embed.embed_minimal_html('test.html', widgets=box)
```

Here, we definitely want to save the state of `w1` and `w2`, but I don't think it makes sense to add all three views to the export. We should probably only include the views of the widgets specified to the function?

jasongrout:
Are all three views added? I would think that just one view should be added, i.e., the top-level widget that is displayed.

vidartf:
The last two commits (15bfcb1 + 01f7923) have fixed this, but see the note in the commit message of 01f7923.

maartenbreddels:
Should we have some unittest with it, it doesn't have to do really fancy, but to make sure there is code coverage for everything. Like exporting a single widget, two widgets, and check if a complicated dependency graph is handled correctly?

vidartf:
@maartenbreddels Definitly, it is already in the TODO list on top! It won't happen today though, as I am heading home. If you have an interesting dependency graph to test for, feel free to share!

maartenbreddels:
I see, 👍  on the tests and documentation. If you are 'ready', let me know, I can see if I can totally remove embed.py from ipyvolume and rely on this, that would be nice!

maartenbreddels:
@vidartf just want to make you aware of what's happening in https://github.com/jupyter-widgets/ipywidgets/pull/1394 . It would make sense this PR would also include a `get_embed_url()` function right? Then we can also use that instead of what I suggested https://github.com/jupyter-widgets/jupyter-sphinx/pull/13/files

vidartf:
@maartenbreddels Yes, there is a `# TODO` note for that, so that would be very nice 👍

vidartf:
@maartenbreddels: I removed the code to add referring widgets. The comment in the code mentioned this being useful for layouts, but widgets refer to the layouts, not the other way around, so those should already be included. Instead, this would add widgets to the state that didn't really need to be included. The only exception I could find was links, which I made a special case for (any link between widgets included in the state are also included).

Am I missing any obvious cases where this would make sense to keep?

vidartf:
Alright, it should be ready for a full review now. Note that I made the default behavior to include the state of all widgets in order to match the behavior of the notebook interface (and be conservative).

maartenbreddels:
Yes, I doubt that comment made sense, and indeed it's needed make jslink work, but there could be other widgets that act similarly. Say a 3rd party link-like-widget.

ipywidgets/widgets/embed.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    if views is not None:
+        try:
+            views[0]
+        except (IndexError, TypeError):
+            views = [views]
+    if include_all:
+        state = Widget.get_manager_state(drop_defaults=drop_defaults, widgets=None)['state']
+    else:
+        state = dependency_state(views, drop_defaults)
```
SylvainCorlay: I don't think that we should try to walk the widgets dependency graph. We had a similar thing at some point to restrict to widgets that were displayed and spent a lot of time tracking corner cases.

Although I really like your implementation with generators :)

There are some cases where you might miss widgets such as a widget that you are not especially looking at that emits events (e.g. a controller widget).

maartenbreddels: For ipyvolume it can be the difference between a 100mb file and a 10mb file (simply because you've created 10 widgets with 10mb of data). So for ipyvolume it's a must have, but maybe don't do this by default? Or have a argument 'include_all=True/False' orso?

jasongrout: I think we need a much better story about cleaning up widgets. It's hard to do automatic garbage collection with the object split across frontend and backend, though.

vidartf: @SylvainCorlay How about this:
 - Replace `include_all=True` with an argument `state=None`.
 - If `state is None`, include state of all widgets known to manager.
 - Else if state is a dict, simply use that. This way API consumers can traverse dependencies for their widgets to generate state.
 - Possibly add some helpers for consumers to build such a state dict on their own?

SylvainCorlay: I like the `state=None` strategy.

(The issue with the `include_all=False` it seems to claim that one would always get a functional set of widgets, which we me have missed some that have e.g. a `jsdlink` to a widget of interest.)

maartenbreddels: But we *do* find `jsdlink`'s right, at least that was what I did in ipyvolume, and @vidartf  copied this behaviour right? So it will work in 99.9% of the cases I think. I actually cannot think of a case it wouldn't work.

vidartf: @maartenbreddels As long as I include the current dependency walker code as helpers (with an updated docstring detailing its limitations), I think you should be able to use the following (will include as example in docs):
```python
from ipywidgets.widgets.embed import embed_snippet, dependency_state
snippet = embed_snippet(my_widgets, state=dependency_state(my_widgets))
```

SylvainCorlay: @maartenbreddels You don't include the jsdlink from a model not in store to a model in store.

maartenbreddels: I'd vote for the inverse of this. `state is None` should be smart (it seems to work fine right?). Otherwise, try explaining to someone that ALL widgets are by default serialized, when you only want the snippet to show my_widget. If for some reason our code fails, we should fix it, and in the meantime you can bypass it by passing `state=Widget.get_manager_state()`.
I don't like regresssion, let's progress! :)

maartenbreddels: @SylvainCorlay  The current code does that actually, if that jsdlink is in `Widgets.widgets` (and every widget should right), it will find it.

SylvainCorlay: It seems to only be included if both source and target are in store.

SylvainCorlay: While if only the target of the link is in store, the link should be added to the store, as well as the source of the link...

Then we must start over again with the rexursive wall from the source...

vidartf: I've committed the outlined change, feel free to play around with it if you want to discover corner cases. Some of the most obvious omissions of the automatic dependency finder are now more explicitly stated in the docstring of the `dependency_state` function.

vidartf:
Alright, I cleaned up the code after the reviews. Given that there will definitely be corner cases that the dependency finder will not find (not without a lot of effort on our part), I think it makes sense to leave the default as simply including all state (at least for this PR).

For now, those confident that the simple dependency walker is sufficient, are free to use that to generate a smaller state. Alternatively, users can build their own custom state, if they have particularly difficult cases to handle, and want to avoid the full state.

PS: Some of the JS tests are failing CI, no idea why.

ipywidgets/widgets/embed.py
```diff
@@ -0,0 +1,262 @@
...
+                store[widget.model_id] = widget._get_embed_state(drop_defaults=drop_defaults)
+
+
+def dependency_state(widgets, drop_defaults=True):
+    """Get the state of all widgets specified, and their dependencies.
+
+    This uses a simple dependency finder, including:
+     - any widget directly referenced in the state of an included widget
+     - any widget in a list/tuple attribute in the state of on an included widget
+     - any widget in a dict attribute in the state of on an included widget
```
jasongrout: delete 'on'

ipywidgets/widgets/embed.py
```diff
@@ -0,0 +1,262 @@
...
+        views = [w for w in Widget.widgets.values() if isinstance(w, DOMWidget)]
+    else:
+        try:
+            views[0]
+        except (IndexError, TypeError):
+            views = [views]
+
+    if state is None:
+        # Get state of all known widgets
+        state = state = Widget.get_manager_state(drop_defaults=drop_defaults, widgets=None)['state']
```
jasongrout: delete one of the `state =`

ipywidgets/widgets/embed.py
```diff
@@ -0,0 +1,262 @@
...
+    data = embed_data(views, drop_defaults=drop_defaults, state=state)
+
+    widget_views = u'\n'.join(
+        widget_view_template.format(**dict(view_spec=json.dumps(view_spec)))
+        for view_spec in data['view_specs']
+    )
+
+    if embed_url is None:
+        # TODO: Get widgets npm version automatically:
+        embed_url = u'https://unpkg.com/jupyter-js-widgets@~3.0.0-alpha.0/dist/embed.js'
```
jasongrout: Perhaps this default should be defined at the top of the module so it's easy to change. Also, it should be updated now that we split the packages up.

ipywidgets/widgets/embed.py
```diff
@@ -0,0 +1,262 @@
...
+    values = {
+        'embed_url': embed_url,
+        'json_data': json.dumps(data['manager_state'], indent=indent),
+        'widget_views': widget_views,
+    }
+
+    return snippet_template.format(**values)
+
+
+def embed_minimal_html(fp, views, **kwargs):
```
jasongrout: document that the file pointer can be a writable file-like object, or can be a string.

docs/source/embedding.md
```diff
@@ -50,6 +50,57 @@ The second option, `Download Widget State`, triggers the downloading of a JSON
...
+## Python interface
+
+Embeddable code for the widgets can also be produced from the Python side.
+The following functions are available in the module `ipywidgets.widgets.embed`:
+
+```rst
+.. embed_snippet::
+
+    s1, s2 = IntSlider(max=200, value=100), IntSlider(value=40)
+    print(embed_snippet(views=[s1, s2]))
```
jasongrout: Since the embed commands are not exported at the package level by default, we should have some import statements noted somewhere showing how to get the commands.

maartenbreddels: :+1:  on this,
```python
import ipywidgets as widgets
widgets.embed.something(...)
```
looks neat

ipywidgets/widgets/embed.py
vidartf: Yes, there is a TODO here. Previously there were talk about making a function somewhere to automatically get the right URL. I can put it as a module global if we're not doing that.

jasongrout: Perhaps the `_version.py` is the right place to put the frontend version(s). Right now, we store the widget spec version there and the protocol version. Adding another attribute there for the package versions of @jupyter-widgets/base and @jupyter-widgets/controls could be the natural place.

maartenbreddels: based on [this](https://github.com/jupyter-widgets/ipywidgets/blob/07e89d8359dbfb048863e8d6e25cf24f0589b8f5/ipywidgets/_version.py), I'd say:
```python
embed_url_default = u'https://unpkg.com/jupyter-js-widgets@~%s-%s.%s/dist/embed.js' % (ipywidgets.__jupyter_widget_version__, ...)
```
Where the ... would need to include alpha.0 somehow, so I guess that also needs to go int _version.py?

vidartf: Also: What is the current URL to use? I haven't been able to track those changes perfectly.

jasongrout: You'll probably need to use two packages: `@jupyter-widgets/base` and `@jupyter-widgets/controls`. Right now the version numbers of those packages have nothing to do with the `__jupyter_widget_version__` (they are both at 0.x versions).

vidartf: PS: We could probably have a function (in `_version.py`?) that peeks in the `package.json` files of the packages.

jasongrout: Oh, never mind. Actually, you just want the html manager: `@jupyter-widgets/htmlmanager`. @pbugnion has been doing related work on this.

vidartf: And how would the two packages work vs an overridden URL? Would an override replace both, or just one of them? Or would we need it to be a dict/sequence now?

jasongrout: (see https://github.com/jupyter-widgets/ipywidgets/pull/1428, and in particular, see the comment https://github.com/jupyter-widgets/ipywidgets/pull/1428#issuecomment-309653571)

jasongrout: > And how would the two packages work vs an overridden URL? Would an override replace both, or just one of them? Or would we need it to be a dict/sequence now?

Right now there is just the html manager package, which defines the `/base` and `/controls` versions, so there is just one include. In the future, I imagine that maybe the html manager does *not* include the `/controls`, and that would be specified separately.

docs/source/embedding.md
vidartf: It does say right above this block that "the functions are available in the module `ipywidgets.embed`". I was hoping this would be enough?

jasongrout: oh, right. I was just cutting and pasting. Well, it wouldn't hurt to have the one import line in the first example.

jasongrout:
@maartenbreddels - just to keep the license situation simpler, and since you are also a Jupyter widgets contributor - I notice that we quote your entire MIT license in one of the files for some code of yours that we are using. Would it be possible for you to consider the code we are using from ipyvolume as a separate contribution to Jupyter Widgets under the jupyter bsd license, as a Jupyter contributor, and thus just consider it under standard Jupyter contributors copyright?

(It's fine either way, but I'm just trying to simplify the license situation where convenient...)

ipywidgets/widgets/embed.py
vidartf: I'll wait with this until package renaming is done #1447.

maartenbreddels:
Yes, I agree, consider all code in this PR as a new contribution or consider the code to be relicensed under bsd.

maartenbreddels:
I'm not sure if there are pieces of the code that are very similar or exact copies of my code, if that is the case, maybe it's better to explicitly put this in the header (# Part of the code is based on ipyvolume, but relicensed by the author (Maarten Breddels) under BSD).

jasongrout:
> I'm not sure if there are pieces of the code that are very similar or exact copies of my code, if that is the case, maybe it's better to explicitly put this in the header (# Part of the code is based on ipyvolume, but relicensed by the author (Maarten Breddels) under BSD).

I think it's also important to consider it copyright by Project Jupyter contributors (as stated at the top of the file) - otherwise the attribution requirements require us to cite your BSD license in full. Personally, I think all we need is your permission as documented here to just use the code as-is in these commits. But of course, happy to do as you wish.

maartenbreddels:
I'm fine with that, just that if someone says 'hey, this code is copied!' they don't need to worry.

vidartf:
Is the current notice OK? As far as I can tell that is the only remaining issue here + the CDN url, and I'd love to get this to purple soon :)

maartenbreddels:
I'm fine with it :+1:

jasongrout:
@maartenbreddels - I tweaked it to say:

> +# Parts of this code is from IPyVolume (24.05.2017), used here under
> +# this copyright and license with permission from the author
> +# (see https://github.com/jupyter-widgets/ipywidgets/pull/1387)

to make it clearer that the copyright and the license matches the rest of the file.

vidartf:
I've just updated the embed CDN url to the same as what the JS code uses (https://unpkg.com/@jupyter-widgets/html-manager@*/dist/index.js), with the same TODO pointing out that this can be made more robust. With this, I think all outstanding issues commented above should be addressed.

jasongrout:
Thanks!

PR #1398, opened 2017-05-31
Add repr strings for widgets

vidartf:
Add a base repr that is a constructor call with the following keyword arguments:
 - All traits with `sync=True` (via `self.keys`), except:
   - When trait name start with an underscore.
   - When value is equal to the default value.

Also adds a few overriden versions as appropriate.

### DIFF
--- a/ipywidgets/widgets/domwidget.py
+++ b/ipywidgets/widgets/domwidget.py
@@ -36,3 +36,9 @@ class DOMWidget(Widget):
         if className in self._dom_classes:
             self._dom_classes = [c for c in self._dom_classes if c != className]
         return self
+
+    def _repr_keys(self):
+        # We also need to include _dom_classes in repr for reproducibility
+        yield from super(DOMWidget, self)._repr_keys()
+        if self._dom_classes:
+            yield '_dom_classes'
--- a/ipywidgets/widgets/domwidget.py
+++ b/ipywidgets/widgets/domwidget.py
@@ -39,6 +39,7 @@ class DOMWidget(Widget):
 
     def _repr_keys(self):
         # We also need to include _dom_classes in repr for reproducibility
-        yield from super(DOMWidget, self)._repr_keys()
+        for key in super(DOMWidget, self)._repr_keys():
+            yield key
         if self._dom_classes:
             yield '_dom_classes'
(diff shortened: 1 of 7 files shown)


jasongrout:
CC @SylvainCorlay

vidartf:
@jasongrout I've implemented the requested changes, and rebased on master (there were some conflicts).

jdemeyer:
This looks pretty good.

One thing I liked about my patch (#1031) is that the attributes come in a "logical" order: with this patch, I find it strange to see keys ordered as `max, min, step, value`. At the very least, I would like to see the `value` attribute first. I would like to see the `value` attribute unconditionally, even if it's equal to the default or not synced.

jdemeyer:
`ToggleButtons` always shows `tooltips=[]` and `icons=[]` in the repr. Could these trivial values be skipped?

jdemeyer:
Here is an idea: in the default `_repr_keys()`, have an explicit list of keys (for example `["value", "min", "max"]`) which would always be shown. Moreover, they would be shown in that order before other keys.

jasongrout:
I think the appropriate place for hardcoding the printing of the value attribute is the value widget base class.

vidartf:
> ToggleButtons always shows tooltips=[] and icons=[]

Maybe all equality checks should be done on the repr string of the value vs the repr string of the default? That would solve such issues, and also for the special cases of the `InstanceDict`s added previously.

> [...] an explicit list of keys [...] which would always be shown

What is the motivation for this? I'm open to making the change, but I struggle to see a use case where this would be helpful. For me, the `repr` string should be a minimal instruction to reproduce the object. If we want to adapt it for human consumption, we should maybe use some sort of prettyprint routine as well. Maybe this should be the `__str__` version?

jdemeyer:
> For me, the repr string should be a minimal instruction to reproduce the object.

That's not my motivation. My motivation is for debugging and testing (more concretely: see https://github.com/sagemath/sage/blob/develop/src/sage/repl/ipython_kernel/widgets_sagenb.py#L181 for an example). So yes, I care more about "human consumption". I want to look at the `repr()` and verify that I get what I want.

vidartf:
PS: My thought about solving this by `repr` vs `str` doesn't really work, as IPython uses `repr` as default text representation of objects, and standard containers (list, dict, etc) uses `repr` of their children.

jasongrout:
This looks good, thanks! The testing error is fixed in master, so it shouldn't hold up this merge.

jdemeyer:
For SageMath, it would be handy to have a new alpha release including this patch.

jasongrout:
We're planning to have another release this week.

PR #1400, opened 2017-05-31
Separate the base widget infrastructure from the core widgets

jasongrout:
This lets the two pieces (the widget manager and base Widget/DOMWidget, vs core widgets) evolve at different rates and be versioned differently.

- [x] Move disconnected css to base package?


jasongrout:
@SylvainCorlay, @pbugnion - what do you think about merging this, and then I'll split out the rest of jupyter-js-widgets to another package (`@jupyter-widgets/corecontrols`? `formcontrols`? `forms`?).

pbugnion:
This is great!

Should the `htmlmanager` now only inject `jupyter-widgets-base`?

PR #1402, opened 2017-06-02
Missing CSS in notebook extension

pbugnion:
At the moment, `npm i` in `widgetsnbextension` fails on master with the error:

```
ERROR in ./src/extension.js
Module not found: Error: Cannot resolve module 'jupyter-widgets-base/css/index.css' in /Users/pascal/oss/ipywidgets/widgetsnbextension/src
 @ ./src/extension.js 137:12-57
```

It looks like the wrong path was to the CSS is used by accident? Anyway, this PR fixes it.


jasongrout:
Great, thanks!

PR #1403, opened 2017-06-02
Install all packages in CI

pbugnion:
This PR adds a CI script to try and install `widgetsnbextension`. This catches issues like the one fixed in PR #1402 .

Is adding this a good idea? It adds a little bit of safety, but does increase the total runtime of the tests.


jasongrout:
I think it makes sense. It seems that the actual script isn't doing anything, though - the test is basically the npm install, right?

pbugnion:
Yes, exactly -- I thought I would include a test script so we have a placeholder for now, but happy to remove it as well.

jasongrout:
Looks good. Thanks!

PR #1407, opened 2017-06-04
Update JupyterLab plugin

jasongrout:
This works for me with https://github.com/jupyterlab/jupyterlab/pull/2348 (once I installed some of the dependencies with full paths - to get this working really, we'd need to release versions of the base and controls packages).


jasongrout:
This is superseded in master.

Issue #1151, opened 2017-02-24
Big Split(tm) Figure out the development workflow after the split


SylvainCorlay:
Iterating on a development version of `jupyter-js-widgets`, `ipywidgets` and `widgetsnbextension` will probably be quite different.

I propose that the recommended development workflow will be to clone the different repos in the same location and we update the `update` scripts accordingly.

jasongrout:
After talking with @SylvainCorlay, I plan to also split up jupyter-js-widgets into 3 packages: 

1. the base widget manager and base widget classes (i.e., Widget, DOMWidget, LabeledDOMWidget)
2. the core widget classes (basically everything subclassing CoreWidget
3. the embedded widget manager

jasongrout:
Our thought is that the public api of the core widget classes will be the widget schema they implement, *not* the actual exported classes. Thus a change in a class interface will *not* be a breaking change, but a change in the schema they implement *will* be a breaking change. The thought is that there will be many more people implementing kernel-side implementations of the widgets (i.e., needing schema version compatibility), and not very many people extending the actual classes themselves (needing class interface compatibility), and that it would be too confusing to have two separate version numbers to keep track of (the schema compatibility and the javascript class interface compatibility).

jasongrout:
One way of emphasizing this is to make a specific module entry point for the classes the module provides, and *not* export the classes in the module namespace. Then we could more easily say that the classes themselves are not part of the public api. There is questions about the extra friction that imposes, but it does make it nice to have a single entry point to get the widgets a package exposes for registration purposes, etc.

jasongrout:
Here's a tentative outline for the split of `jupyter-js-widgets`:

* `@jupyter-widgets/base` - Base widgets, such as Widget, DOMWidget, LabeledDOMWidget, LayoutWidget. These are meant to be subclassed. This would also include the widget serializer, which is of general use.
  - `nativeview.ts`
  - `widget.ts`
  - `widget_layout.ts`
  - `utils.ts`
  - nominally, the output widget (i.e., the module for the output widget will be this namespace, and widget managers are responsible for providing it and special-casing it.)
* `@jupyter-widgets/manager` - the base widget manager that can be subclassed, depends on `@jupyter-widgets/base`
  - `manager-base.ts`
  - `services-shim.ts`
[...]

PR #1409, opened 2017-06-05
[WIP] Traitlets Default Validation

rmorshea:
Beginning with version 5.0 traitlets **may or may not** cross-validate its default values. The code below is currently valid, but if/when they are cross-validated, infinite recursion would result due to the circular validation dependency between `min` and `max`:

```python
class MaxMin(HasTraits):

    min = Int()
    max = Int()

    @validate('min')
    def _validate_min(self, proposal):
        """Enforce min <= value <= max"""
        minimum = proposal['value']
        if minimum > self.max:
            raise TraitError('setting min > max')
        return minimum

    @validate('max')
    def _validate_max(self, proposal):
        """Enforce min <= value <= max"""
        maximum = proposal['value']
        if maximum < self.min:
            raise TraitError('setting max < min')
        return maximum
```

https://github.com/ipython/traitlets/pull/413

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -46,9 +46,9 @@ class _BoundedFloat(_Float):
     def _validate_min(self, proposal):
         """Enforce min <= value <= max"""
         min = proposal['value']
-        if min > self.max:
+        if self.has_trait_value("max") and min > self.max:
             raise TraitError('Setting min > max')
-        if min > self.value:
+        if self.has_trait_value("value") and min > self.value:
             self.value = min
         return min
 
@@ -56,9 +56,9 @@ class _BoundedFloat(_Float):
     def _validate_max(self, proposal):
         """Enforce min <= value <= max"""
         max = proposal['value']
-        if max < self.min:
+        if self.has_trait_value("min") and max < self.min:
             raise TraitError('setting max < min')
-        if max < self.value:
+        if self.has_trait_value("value") and max < self.value:
             self.value = max
         return max
 
--- a/ipywidgets/widgets/widget_int.py
+++ b/ipywidgets/widgets/widget_int.py
@@ -109,9 +109,9 @@ class _BoundedInt(_Int):
     def _validate_min(self, proposal):
         """Enforce min <= value <= max"""
         min = proposal['value']
-        if min > self.max:
+        if self.has_trait_value("min") and min > self.max:
             raise TraitError('setting min > max')
-        if min > self.value:
+        if self.has_trait_value("value") and min > self.value:
             self.value = min
         return min
 
@@ -119,9 +119,9 @@ class _BoundedInt(_Int):
     def _validate_max(self, proposal):
         """Enforce min <= value <= max"""
         max = proposal['value']
-        if max < self.min:
+        if self.has_trait_value("max") and max < self.min:
             raise TraitError('setting max < min')
-        if max < self.value:
+        if self.has_trait_value("value") and max < self.value:
             self.value = max
         return max
 
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,7 @@ setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
     'ipython>=4.0.0',
     'ipykernel>=4.5.1',
-    'traitlets>=4.3.1',
+    'traitlets>=5.0.0',
     # Requiring nbformat to specify bugfix version which is not required by
     # notebook.
     'nbformat>=4.2.0',


minrk:
I think it's important that we try *extremely* hard to prevent this kind of breakage in traitlets. We should never require an unreleased version of a package in another unless something has gone quite wrong. I think we need to make changes in traitlets before release so that 5.0 doesn't cause this kind of breakage.

minrk:
Whatever fix is made over here, it should be done in a backward-compatible way, such that it doesn't introduce a dependency on unreleased traitlets.

rmorshea:
@minrk this breakage will only take effect when traitlets 5.0 is released, so this was a preemptive pull request. We could add `has_trait_value` as a patch to traitlets 4.3 so that the version bump is no longer required.

minrk:
@rmorshea this PR however *also* makes ipywidgets incompatible with traitlets < 5, which is a problem because then both packages must be released at the exact same time (ipywidgets 6 is incompatible with traitlets 5 and ipywidgets 7 is incompatible with traitlets 4). We should never require coordinated releases across Jupyter projects. Every new release of a given Jupyter package should work with the current stable version of every other package at time of release. To that end, *master* of every package should endeavor to work with stable of every other package.

For instance: if traitlets 5 will break widgets, we need to make a new release of widgets that works with traitlets master *before* releasing traitlets 5. In order for that release of ipywidgets to work, though, it must *also* work with the current stable version of traitlets, otherwise we will be releasing a package that cannot be installed due to unsatisfiable dependencies.

One way to make such a fix: check for the presence of `has_trait_value` and only call it if it's defined. That way, the 4.x behavior remains unchanged, and the extra check is only applied once traitlets is updated.

rmorshea:
@minrk, got it. I'll put some thought into a backward compatible solution here.

I just want to check in and see whether we are all still in favor of cross validating defaults though.

IMO it makes sense to enforce uniform behavior when assigning all new values. If we fail to cross validate defaults, there is a discrepancy between the way we handle default values, and every other value a user might set.

rmorshea:
Something to consider: https://github.com/jupyter-widgets/ipywidgets/issues/1422

rmorshea:
@minrk, could we just scrap the idea of validating `min < max` and `max > min`? Afterall if `max < min` then setting any value would raise since `min < value < max` would be an impossible condition to meet. So long as the error message includes the `max` and `min` limits, the problem should be pretty self explanatory from a user's perspective.

minrk:
@rmorshea possibly. It makes more sense to me to use `@observe` and cause min/max to push each other after assignment, rather than `@validate`, which fires before the assignment actually takes place.

SylvainCorlay:
@minrk @rmorshea we cannot use `@observe` instead of `@validate` for pushing min/max, which was the root of a number of race conditions in the widgets infrastructure between front-end and back-end validations.

SylvainCorlay:
Also, I was :-1: on validating default values. Validating initial state is ok, but default values bring a lot of complications.

minrk:
I'm also okay rolling back default value validation.

SylvainCorlay:
I will take the time to read this through though. Sorry for the rough answer.

rmorshea:
@SylvainCorlay, @minrk, if we decide not to cross-validate default values, then we probably shouldn't validate them at all. One alternative would be to pass default validators a `Bunch` containing the trait. Thus a user could defined:

```python
@default
_my_default(self, init):
    # ... make a default
    return init.trait.validate(self, value)
```

Here it's up to the user to ensure that there are no circular dependencies between they're traits.

ankostis:
A couple of weeks ago I had a confusing bug where the `config` functionality had a nasty interplay with the non-validation of default that lead me to think that it's not a good idea to leave *defaults* unchecked:

1. I was starting my app with some traits having invalid defaults (i.e. `default=None, allow_none=False`),
2. then was using config-files to set them to valid non-null values,
3. but some other, unrelated traits, were receiving invalid values from those files,
4. leading the code in [`configurable._load_config()#160`](https://github.com/ipython/traitlets/blob/master/traitlets/config/configurable.py#L160) to  to rollback all file-values,
5. and receive an exception from the [`hold_trait_notifications()#1151`](https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L1151) resource manager due to those invalid defaults.  Practically the raised exception was hiding the real problem.

Unfortunately, I have not saved any clues from that incident; I tried to reproduce it with a simpler setup, but did not have the necessary time.  The step 5 above is my *probable* conclusion - don't take it for granted.

rmorshea:
Any more thoughts on this? Resolving this sooner rather than later would be beneficial.

rmorshea:
@minrk, @SylvainCorlay updated thoughts on this?

minrk:
Given the problems and races we are seeing caused by cross-validation of defaults, I'd be inclined to continue not cross-validating defaults, despite the problems pointed out above that we have in not doing so. Part of the reason being that introducing cross-validation *introduces* bugs into existing working code, whereas failure to validate merely fails to raise informative errors in already buggy code. I'm going with a principle of "least bad" for now, since nothing seems great.

rmorshea:
@minrk, removing cross validation makes sense. My last question then, is whether we should validate defaults at all. As above, I suggested that going halfway with the validation simply makes things more confusing. However, to give users the ability to enforce validation when they need to, we can provide a bunch to default generators with the keys:

+ *name*: the name of the trait
+ *trait*: the associated trait
+ *default*: the "static" default generated by the trait's internal mechanisms.

IMO, because this makes default validation the responsibility of the user, it strikes a nice balance between the needs of those trying to validate, and the problems that can arise when they do.

minrk:
That makes sense, I'm okay with not validating defaults at all. Passing a new argument to default generators is a bit of a problem because the current signature takes no args and we can't break that, so it would have to be something like a new `@default_validate` decorator to get the new signature. And then, I'm not sure if it's worth it since the default generator already knows the trait it's for, and can already do whatever it wants without any arguments.

rmorshea:
@minrk, while it's true that one can retrieve the trait via `type(self).name` it's a feels bit tedious:

```python
# inside default generator method
trait = type(self).name
default = trait.default(self)
return trait.validate(self, default)
```

However, keeping things backward compatible is important, so instead of `@default_validate`, I would propose `default(args=True)` or something of that nature to specify that the default generator should accept arguments.

PR #1380, opened 2017-05-23
[WIP] Export output widget in embed bundle

pbugnion:
Prior to this PR, the output widget could not be embedded. Addresses issue #986 .

PR #1410, opened 2017-06-05
[WIP] Export output widget in embed bundle

pbugnion:
Prior to this PR, the output widget could not be embedded. Addresses issue #986 .

This is a rewrite of PR #1380, which had become out of date after the htmlmanager was moved to a stand-alone package.

I think this still needs:

 - [x] Unittest widget (e.g. with a mock manager)
 - [x] migrate rendermime to a singleton or to the HTMLManager to avoid re-instantiating it for every output area
 - [x] Expose ability to define new renderers (and override the default ones).
 - [x] Custom CSS for the output area (at the moment, it doesn't look great)


pbugnion:
@jasongrout I'm about to start work on letting users override renderers in the HTML Manager. How were you thinking users would hook into this? 

One way would be to define a top-level method in `embed-webpack` (e.g. bound to `window`) called `configureWidgets` that defines top-level options. The options would be stored in some form of global state.

Alternatively, we could just expose the options as an extra argument to `renderInlineWidgets`. We are then just catering for users who use the htmlmanager as a library and who instantiate the widgets themselves. If we go down that route, I'll add another example demonstrating how to do this.

I think I'm mildly in favour of the second option, since it exposes less magic.

jasongrout:
IIRC, a top-level method won't work since you might have multiple instances of the widget manager on a single page - I think this may happen in some of the documentation scenarios.

Since your rendermime instance is exposed as a public attribute on the manager (i.e., the manager's rendermime is what is passed to the output widget), I think that may be enough - people can access the rendermime directly and add other renderers as needed.

jupyter-widgets-htmlmanager/src/htmlmanager.ts
```diff
@@ -45,11 +54,11 @@ class HTMLManager extends widgets.ManagerBase<HTMLElement> {
      */
     protected loadClass(className: string, moduleName: string, moduleVersion: string) {
         return new Promise(function(resolve, reject) {
-            if (moduleName === '@jupyter-widgets/controls') {
+            if (coreWidgetModules.indexOf(moduleName) >= 0) {
```
jasongrout: The `@jupyter-widgets/base` base classes (like DOMWidgetView, WidgetModel, etc.) shouldn't be exported in `@jupyter-widgets/controls` - I think you'll need two checks here, one for each of your `coreWidgetModules`. See https://github.com/jupyter-widgets/ipywidgets/blob/36dd0aa54b810c60a58d9c18afee590f530dc2a8/widgetsnbextension/src/manager.js#L97

jupyter-widgets-htmlmanager/src/htmlmanager.ts
```diff
@@ -70,4 +79,17 @@ class HTMLManager extends widgets.ManagerBase<HTMLElement> {
...
+    /**
+     * Renderers for contents of the output widgets
+     *
+     * Defines how outputs in the output widget should be rendered.
+     */
+    renderMime: RenderMime =
+        new RenderMime({
+            items: [
+                { mimeType: WIDGET_MIMETYPE, renderer: new WidgetRenderer(this) },
+                ... RenderMime.getDefaultItems()
```
jasongrout: please delete the space after `...`

pbugnion:
I've rebased this off the current master. @jasongrout -- your commits were redundant with changes in master so they just got skipped.

Do we normally prefer merge commits or rebases?

jasongrout:
Depends, but I'd say rebase in this case

jasongrout:
@pbugnion - the title still says WIP - is this ready for review?

pbugnion:
The tests fail on Travis but not locally. I just need to figure out why. Beyond this, I think it's ready for review.

jasongrout:
Also, I changed a bit how testing is done and how the repos are built. There's only a few conflicts, but perhaps resolve those and rebasing or merging master in will help isolate the failures.

jasongrout:
@pbugnion - I was looking into updating this. Are you working on it too?

pbugnion:
@jasongrout I definitely won't be working on this today, and probably won't be able to over the weekend, so go ahead! Thanks.

jasongrout:
@pbugnion - I think it *should* work, but I haven't tested it to see if it actually does yet.

jasongrout:
I just tested by hand (modified the example with some state of an output widget), and the text and an embedded widget worked fine. I'll clean up one last thing with the model_id rename and merge when tests pass.

jasongrout:
Actually, let's make the model_id rename a different PR, since it's really a different issue. It came up here because the widgetsnbextension output widget is broken without it.

pbugnion:
Ah this is great -- thanks for your help!

PR #1417, opened 2017-06-07
Fix Widget view name/module None

### LINKED ISSUE #1412: Tries to create view for Link
If you have a notebook cell that ends in a `ipywidgets.jslink(...)` statement, master gives various console errors. Doing some minor debugging, it seems that it tries to look up the class `''` (empty string) in `@jupyter-widgets/controls` (the view name for `Link` is empty string).

vidartf:
Fixes #1412 and adds test.


jasongrout:
Looks good to me. Thanks for refactoring the test infrastructure too.

PR #1418, opened 2017-06-07
Fix dev-install.sh problem 

mwcraig:
I was seeing [install errors](https://gist.github.com/mwcraig/351a08420e7a2099103a2cb3d2c20515) on a machine with no node stuff installed. Installing `jupyter-widgets-base` before the remaining items seems to fix the problem.


jasongrout:
Thanks!

PR #1419, opened 2017-06-07
Fix accordion styling so that it works with nested accordions

mwcraig:
This makes a minor change to the CSS for accordions so that the styling is appropriate when accordions are nested.

## Before:

accordion-stuff

## After:

accordion-stuff


jasongrout:
Thanks!

PR #1424, opened 2017-06-14
Clean up some details in tutorial notebooks

mwcraig:
This PR cleans up some (mostly) minor details in the tutorial notebooks. The only substantive change so far is in `Widget Basics`, where I replaced a glob-style import `from ipywidgets import *` with a the style used in the other notebooks `import ipywidgets as widgets`.

More coming soon, so this sin't ready for merge yet. Opening now in case some of the changes I'm seeing in these first couple of notebooks that I didn't intentionally make (like the deletions metadata from some of the cells) are some I need to worry about....

### DIFF
--- a/docs/source/examples/Using Interact.ipynb
+++ b/docs/source/examples/Using Interact.ipynb
@@ -392,7 +392,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The widget is a `Box`, which is a container for other widgets."
+    "The widget is an `interactive`, a subclass of `VBox`, which is a container for other widgets."
    ]
   },
   {
@@ -408,7 +408,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The children of the `Box` are two integer-valued sliders produced by the widget abbreviations above."
+    "The children of the `interactive` are two integer-valued sliders and an output widget, produced by the widget abbreviations above."
    ]
   },
   {
@@ -629,7 +629,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.5.2"
+   "version": "3.6.1"
   },
   "widgets": {
    "application/vnd.jupyter.widget-state+json": {
--- a/docs/source/examples/Widget Basics.ipynb
+++ b/docs/source/examples/Widget Basics.ipynb
@@ -473,7 +473,7 @@
     }
    },
    "source": [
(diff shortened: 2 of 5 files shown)


mwcraig:
Since I'm not sure how easy it is to read these diffs, here is a list of the changes made in each notebook:

1. `Using Interact`:
    1. Above cell 19, says result of `interactive` is a box, but is actually a `ipywidgets.widgets.interaction.interactive`
    1. Following cell `w.children` has *three* children, not two: the two slides and an `Output` widget.
1. `Widget Basics`
    1. Change `from ipywidgets import *` to `import ipywidgets as widgets`
    1. Remove this senence: "You can hide all of the widgets in the widget area by clicking the grey x in the margin." *I see no ×*.
    1. Mention that `link` exists and that we will return to it later.
1. `Widget Events`
    1. Change `handler passed to the decorator` to `handler passed to observe`
13. `Widget List`
    1. There are 10 numeric widgets, not 8.
    1. Mention that int/float inputs impose limits on enter, not on fly.
    1.  Remove `color` argument because it is ignored.
    1. Selection widgets:
        1. In the description of options for specifiying the selectable options, add the word "string" to the description of the values: `or simply string values for which the label is derived`
        1. Remove the `slider_color` option from `SelectionSlider` because it is ignored.
        1. Remove (the commented out) `readout_format='i'` from `SelectionSlider` because it has no effect on this particular slider and is explained in more detail later.
    1. `Label`: replaced with example that does **NOT** require `HBox`ing, and include  math in this example. The description mentions that labels can be used to create custom descriptions for other widgets.
    1. "Play (Animation) widget"
        1. Add sentence making explicit that the slider is linked to the player.
    1. "Controller": Add an explanatory sentence.
    1. "Layout" widgets --
        1. Add a few explanatory sentences at the beginning of section.
        1. Renamed section to "Container/Layout widgets" to distinguish them from the `Layout` class later.
        1. Add an example of a `Box`.
        1. Break up the `VBox` example code a bit. I found it hard to parse the heavily nested argument list.
        1. In the Tabs example, set the children after the tab is created, change the name of the strings used to generate contents from `list` to something more sensible, and set the titles.
1. `Widget Styling`
    1. "Natural sizes..." -- broke the up code a bit.
    1. `flex-direction`: re-order list of options to match their order in the figure that goes with it: column-reverse, column, row, row-reverse.

mwcraig:
This is ready for a look now. I've numbered the changes in the comment above to make it easier to refer to any changes you would like reverted.

jasongrout:
> In the description of options for specifiying the selectable options, add the word "string" to the description of the values: or simply string values for which the label is derived

The values can be anything, not just strings - basically `str` will be called on them. So, for example, `[1,2,3]` would work just fine.

jasongrout:
This looks great, thanks for auditing these notebooks! I tweaked a few things:

1. Put the tab title setting in a for loop
2. tweaked wording for the Label widget
3. took care of my comment above.

PR #1430, opened 2017-06-19
[WIP] Fix build errors

pbugnion:
As far as I can tell, master has stopped building correctly due to a dependency update (related to [this](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/17239)?). This is (probably) only visible off a fresh clone, or at least removing the node_modules in *jupyter-widgets-controls*. It also occurred on PR #1410.

I think this is due to an update in the type defintions for JQuery. The module definition now reads:

```typescript
declare module 'jquery' {
    export = factory;
}

declare function factory(window: Window, noGlobal?: boolean): JQueryStatic;

declare const $: JQueryStatic;
```

This means that the only ways of importing jquery are:

```typescript
const $ = require('jquery') // uses the exported module 'jquery'
import 'jquery' // uses the global exports, which injects $ into the global namespace
```

In particular, we were importing jQuery using `import * as $ from 'jquery'`, which is now not allowed.

This is still fairly broken on Travis.

### DIFF
--- a/jupyter-widgets-base/src/tsconfig.json
+++ b/jupyter-widgets-base/src/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "declaration": true,
     //"noImplicitAny": true,
-    "lib": ["dom", "es5", "es2015.promise"],
+    "lib": ["dom", "es5", "es2015.promise", "es2015.iterable"],
     "noEmitOnError": true,
     "module": "commonjs",
     "moduleResolution": "node",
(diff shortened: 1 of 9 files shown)


maartenbreddels:
I cannot install ipywidgets at all, also not previous versions. I'll try this fix.

jasongrout:
This indicates this will be fixed soon: https://github.com/DefinitelyTyped/DefinitelyTyped/commit/844920fa549d0c74b5ac2d2d14858a3674952383#commitcomment-22611629

jasongrout:
The import errors are fixed. If you want to remove your import changes, I'll merge - the es2015.iterable changes still look needed.

pbugnion:
Ah thanks for checking! I'll do this now.

jasongrout:
I added the same fix to jupyter-widgets-base. Thanks!

PR #1439, opened 2017-06-21
Adjust promise rejection wrapping

vidartf:
Allows for contextual messages while still keeping original exception.


jasongrout:
Thanks!

PR #1442, opened 2017-06-21
Use Lerna

jasongrout:
This updates the dev and build process to use lerna to coordinate between packages.

Basically, to install and build all of the npm packages, just do:

```sh
npm install
npm run build
```
from the repo root directory.

There are no doubt more simplifications to come, but this gets us started using lerna, and supersedes the old `npm run update` dance we were doing.


jasongrout:
We can iterate from here...

PR #1444, opened 2017-06-22
Add more explanation to description of links

mwcraig:
This adds just a bit at the end of the `Widget Events` notebook explaining the difference between kernel and client-side links, and includes a link to the static version of the page, which nicely illustrates the difference between the two kins of links.


jasongrout:
@mwcraig - what do you think of the tweaks I made?

mwcraig:
Looks good!

PR #1451, opened 2017-06-24
Rename the widget model .id attribute to .model_id

jasongrout:
In Backbone, it will set the .id attribute automatically based on model data. This also may mean that it will make the .id attribute undefined after we have set it.


jasongrout:
This caused the output widget to not work, for example.

Issue #1449, opened 2017-06-22
Output widget module


jasongrout:
Right now, the browser Output widget model/view is always patched into the controls module by hand by each manager. This is awkward. The only reason we did this is so that there would be one javascript module name for the entire ipywidgets python package. However, now we don't have that anyway with the base and controls module.

How about we make a new package just for the output widget: `@jupyter-widgets/output` or something. Then each manager just has to look for that specific package and substitute its implementation in, rather than looking for the controls package and then checking the module name. This also lets us treat the controls package like any other package, and not *have* to patch something in.

jasongrout:
Furthermore, the `@jupyter-widgets/output` package could be an actual package with an abstract implementation of the Output widget, or at least the typings/interface for one, so to make a new output widget you'd inherit from that base class.

jasongrout:
Related to #1447.

pbugnion:
I think that makes sense -- I came to similar conclusions when working on the embed bundle.

pbugnion:
To clarify, you were thinking that the specific implementations of the output widget would remain in their respective packages (e.g. the implementation for notebooks would remain in `widgetsnbextension`? 

The only visible change would be the `widget_module` attribute in the state would now be `@jupyter-widgets/outputs`?

jasongrout:
I was thinking that essentially the `_view_module_` and `_model_module_` would both be this new `@jupyter-widgets/output` module. And yes, I was thinking that each manager would still have their own implementation of at least the view, and possibly the model as well, which may inherit from some abstract base class in the output package.

jasongrout:
Fixed by #1456.

PR #1456, opened 2017-06-26
[WIP] Separate output widget package

pbugnion:
Prior to this PR, all output widgets lived in the `@jupyter-widgets/controls` package. When loading classes from the `controls` package, managers would check whether the class name was `OutputModel` or `OutputView` and patch in their implementation.

Now, the output widget is in a fully separate package which the managers can load just like they load the controls package.

This addresses issue #1449.

A non-exhaustive list of things left to do:

- [ ] Use output module in jupyterlab extension.
- [ ] Use output module in htmlmanager extension
- [ ] Factor out common behaviour if possible.
- [ ] Typings

At the moment, the dependency on the outputs module is expressed as a file dependency in the `package.json`. After the first release, this should be replaced by a dependency on the released version.


jasongrout:
> At the moment, the dependency on the outputs module is expressed as a file dependency in the package.json. After the first release, this should be replaced by a dependency on the released version.

Does it not just work if you put in the package name/version? My impression of the lerna workflow is that it would automatically take care of making a package a link to the internal package, without it having to be published to npm.

widgetsnbextension/src/widget_output.js
```diff
@@ -5,19 +5,18 @@
...
 
-var OutputModel = widgets.DOMWidgetModel.extend({
-    defaults: _.extend({}, widgets.DOMWidgetModel.prototype.defaults, {
-        _model_name: "OutputModel",
-        _view_name: "OutputView",
+var OutputModel = outputWidgets.OutputModel.extend({
+    defaults: _.extend({}, outputWidgets.OutputModel.prototype.defaults(), {
         msg_id: "",
-        outputs: [],
+        outputs: []
```
jasongrout: `outputs` should be in the base class, since that is a necessary part of the model (it is assumed by the python side).

jasongrout:
Wow, thanks for doing this!

widgetsnbextension/src/widget_output.js
pbugnion: That makes sense.

pbugnion:
Re. the file dependency -- I realised we'd switched to lerna after I'd committed that. I'll try and remove it now.

jasongrout:
Great, let's iterate on this!

pbugnion:
Great -- thanks!

PR #1460, opened 2017-06-28
Fix npm install by building the output widget on install.

jasongrout:


blink1073:
I think this should be [prepare](https://docs.npmjs.com/misc/scripts#deprecation-note).

jasongrout:
Actually, we'll just take a cue from jlab and delete the prepublish/prepare steps across the board.

jasongrout:
I'll iterate on cleaning up the package.json files in another PR.

jasongrout:
> I think this should be prepare.

It looks like lerna doesn't support prepare yet? https://github.com/lerna/lerna/pull/499

blink1073:
Shucks.

PR #1464, opened 2017-06-29
Fix module loading for output widget in notebook

pbugnion:
Commit 7b9ea6d introduced a bug in the notebook extension: instead of loading
the overriden implementation of the output widget, we load the base
implementation (which is mostly empty).


jasongrout:
Thanks!

PR #1467, opened 2017-06-29
Make buttons not overwrite css classes.

### LINKED ISSUE #1459: ipywidgets 6 problem with Tabs and Buttons
I am unable to upgrade to ipywidgets 6 because the the Tab widgets is broken.  It seems fixed in 7.0.0a4.  Can the fix be backported to 6.0?

The problem is including a Button or ToggleButton as a child in a Tab causes it to appear in all tabs.

selection_037

jasongrout:
Fixes #1459.


jasongrout:
This is a PR to fix #1459 for a possible future 6.0.1 release.

jasongrout:
rebased on the current 6.x

PR #1469, opened 2017-06-30
Use output base class in html-manager output widget

pbugnion:
Prior to this PR, the output class in the html-manager still inherited from controls, and the manager did not recognize the output module. Now fixed.

This PR also renames references to the output module as `outputBase`, rather than `outputWidget` -- this avoids confusion with the local implementation of the output widget, which we also load as `outputWidget`.


jasongrout:
Thanks!

Issue #1425, opened 2017-06-14
ipywidgets 6 + jupyter-js-widgets 7 incompatibility


gnestor:
I ran into an issue when installing ipywidgets to a fresh conda environment and jupyter notebook loading jupyter-js-widgets/extension from `/usr/local/share/jupyter` vs. `/Users/grant/anaconda/envs/mycondaenv/share/jupyter`. ipywidgets in the conda environment was 6.0.0 and jupyter-js-widgets/extension was linked to my dev install of ipywidgets (7.0.0). 

The issue looked like:

![image](https://user-images.githubusercontent.com/512354/27140829-61c01388-50db-11e7-9be1-a3587ed700e1.png)

![image](https://user-images.githubusercontent.com/512354/27140792-4d043604-50db-11e7-922c-c832d2529db6.png)

The root of the issue is a breaking change introduced by https://github.com/jupyter-widgets/ipywidgets/pull/1194, specifically https://github.com/jupyter-widgets/ipywidgets/commit/9d896f76d01b23b3b9fb60a2948141f77da769ef#diff-30b37447269c75d7eff843470858ea28, which looks for widget state in `msg.content.data.state` vs `msg.content.data`. 

1. I remember in previous versions, ipywidgets would return an error message if there as a version mismatch between JS and Python? That would've made this issue much easier to debug.
2. Given that this breaks backwards-compatibility, is there a legitimate need to move widget state from `msg.content.data` to `msg.content.data.state`? This shouldn't affect anyone unless they version mismatch between JS and Python, so if we solve #1 then we solve this.
3. Is this the way that jupyter paths are supposed to work? I assumed that an nbextension installed to a conda environment would take precedence over an nbextension installed elesewhere?

jasongrout:
The root of the issue was a mismatch between the python and js versions. The `data.state` was just the first incompatibility you saw. Since we don't implement the version check when the notebook starts up anymore (it had architecture flaws), the ipywidgets 6 that you had installed would have always printed that error message.

First and foremost we should have clear docs about clearing out old widget versions and installing fresh. A question is if we should put in some javascript code that attempts to detect if the installation is corrupted, and if so, what kinds of checks we can make. I'll think about this - open to ideas or PRs too!

hainm:
clear docs is super useful.

jasongrout:
I'm working on this by implementing a version check for the manager protocol version when a comm is opened.

jasongrout:
[...]

PR #1470, opened 2017-06-30
Implement widget protocol version checking in the comm open messages.

jasongrout:


jasongrout:
This should fix #1425 by giving a much clearer message early on that there is a version mismatch.

PR #1471, opened 2017-06-30
Correct the description of where widgets appear

mwcraig:
They are now in the cell output area, not a separate area.


jasongrout:
Thanks!

PR #1486, opened 2017-07-06
Audit disable behavior across the widgets

### LINKED ISSUE #1431: No styling change when widget is disabled
Prior to ipywidgets 6, when a widget was disabled its style was changed so that it looked faded, as shown in the pic below. Starting with ipywidgets 6 when `disabled=True` the widget is disabled but the style does not change.

button-disable

jasongrout:
CC @mwcraig 

Fixes #1431

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -623,7 +623,6 @@
     "widgets.Valid(\n",
     "    value=False,\n",
     "    description='Valid!',\n",
-    "    disabled=False\n",
     ")"
    ]
   },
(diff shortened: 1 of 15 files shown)


jasongrout:
We need to audit the disabled attribute for each widget. It wasn't used for several widgets, and for others, it only works halfway (for sliders, it still allows you to change the overflow).

mwcraig:
Looks like you have caught several instances of broken disabled behavior. Want to me to go through the full list and test it?

jasongrout:
@mwcraig - this is ready for review

PR #1491, opened 2017-07-09
Explain selected_index for tab and accordion

### LINKED ISSUE #1386: Accordion and button widgets doesn't work with observe
Hi!
I wanted to do an interactive plot with filtering based on ValueSliders for one of two columns in dataframe. For that purpose I made two ValueSliders which are stacked in the accordion. Additionally I added third panel to accordion with button which basically is supposed to clear the filters. All of the widgets (sliders, button, accordion) are set to observe and run one function. Unfortunately, the function is run only when value of any of sliders is changed. It is not executed when I click a button or change tab in accordion. 

I've checked the js console when using those widgets and nothing is thrown there.

My code:
https://gist.github.com/heavelock/410134ef382b09855b26619f4da02fc7

Jupyter and ipywidget installed using masters pulled 3 hours ago.
ipywidgets 2940f9f56f63257902b27509f18e872e77c3c014
Jupyter c9d399ace6bb2222b213a7727df49c5c9b2fca9b

mwcraig:
This fixes #1386 by adding an explanation of `selected_index` to the `Widget List` notebook.

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -1578,6 +1578,17 @@
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Accordion and Tab use `selected_index`, not value\n",
+    "\n",
+    "Unlike the rest of the widgets discussed earlier, the container widgets `Accordion` and `Tab` update their `selected_index` attribute when the user changes which accordion or tab is selected. That means that you can both see what the user is doing *and* programmatically set what the user sees by setting the value of `selected_index`.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


pbugnion:
This looks great! Should we also document the important traitlets in docstrings? We do this for the [Button widget](https://github.com/jupyter-widgets/ipywidgets/blob/6c99a866f2f658a4a76a960c6e6d57a786fe3288/ipywidgets/widgets/widget_button.py#L29), for instance. 

I think users are more likely to look first at the docstring, then fall back onto documentation?

jasongrout:
Totally agree. Merging this for now - let's audit all of the widgets' documentation for attributes.

PR #1492, opened 2017-07-10
use np.array_equal install of np.equal

SylvainCorlay:


dmadeka:
@SylvainCorlay this caused a number of bugs in bqplot. Can we have a minor release for this PR?

jasongrout:
Done! 7.0.0a11

dmadeka:
@jasongrout yes! on PyPi using --pre? bqplot's dependencies will need to be updated because a lot of things are breaking right now

jasongrout:
Sure, let's update bqplot dependencies.

Issue #1493, opened 2017-07-10
Selection containers (Tab, accordion) should allow a selected index of None for consistency


jasongrout:
This would make it consistent with the selection widgets.

pbugnion:
The selection widgets explicitly disallow out of range indices. 

Do we want to do that for the selection containers as well? It breaks backwards compatibility (I'm not sure how much user code relies on being able to set / get -1 for `selected_index`), but it would be more consistent and ensures that there's only one way to set the container as having nothing selected.

jasongrout:
Fixed by #1495.

SylvainCorlay:
@jasongrout I wonder if the default should not also be `None`. At the moment, for some reason, the default is 0.

jasongrout:
I think that would be confusing in most cases. Let's carry on discussion on #1822, though.

PR #1495, opened 2017-07-11
Allow `selected_index=none` in selection container widgets

pbugnion:
Prior to this commit, the way to *unset* the selection for a tab or accordion widget was to set `selected_index = -1`. Having to set it to `None` would be better (see issue #1493 ). 

This PR allows setting the index to None. The bulk of the changes are in the `Selection` class, which now allows setting the index to null.

~~This is what's left to do~~:
 - [x] explicitly disable setting selected_index to an out of range value
 - [x] unit tests for Selection class 
 - [x] amend docs

This is now ready for review.

### DIFF
--- a/packages/controls/src/phosphor/tabpanel.ts
+++ b/packages/controls/src/phosphor/tabpanel.ts
@@ -124,7 +124,7 @@ class TabPanel extends Widget {
    */
   get currentIndex(): number | null {
     const currentIndex = this.tabBar.currentIndex;
-    # Phosphor tab bars have an index of -1 if no tab is selected
+    // Phosphor tab bars have an index of -1 if no tab is selected
     return (currentIndex === -1 ? null : currentIndex);
   }
 
--- a/packages/controls/src/phosphor/tabpanel.ts
+++ b/packages/controls/src/phosphor/tabpanel.ts
@@ -122,8 +122,10 @@ class TabPanel extends Widget {
    * #### Notes
    * This will be `null` if no tab is selected.
    */
-  get currentIndex(): number {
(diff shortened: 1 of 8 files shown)


pbugnion:
screen shot 2017-07-11 at 07 30 17

packages/controls/src/phosphor/tabpanel.ts
```diff
@@ -130,7 +130,7 @@ class TabPanel extends Widget {
    * Set the index of the currently selected tab.
    *
    * #### Notes
-   * If the index is out of range, it will be set to `-1`.
+   * If the index is out of range, it will be set to `null`.
```
jasongrout: It looks like the logic here is still missing for out of range values.

packages/controls/src/phosphor/tabpanel.ts
```diff
@@ -120,7 +120,7 @@ class TabPanel extends Widget {
    * Get the index of the currently selected tab.
    *
    * #### Notes
-   * This will be `-1` if no tab is selected.
+   * This will be `null` if no tab is selected.
    */
   get currentIndex(): number {
     return this.tabBar.currentIndex;
```
jasongrout: Logic needs to be added here to return null if the tab bar index is -1.

packages/controls/src/phosphor/tabpanel.ts
jasongrout: Actually, you don't need to do anything here - the tab bar handles the logic to set its current index to -1 if out of range. We'll just handle the translation to null up where we retrieve the index.

jasongrout:
Thanks!

PR #1496, opened 2017-07-11
delay matplotlib import

minrk:
only trigger flush figures if matplotlib has already been loaded

avoids unnecessary load of matplotlib on widget import

follow-up to #1474


jasongrout:
Thanks!

PR #1497, opened 2017-07-11
Update for jlab 0.25

jasongrout:


packages/html-manager/src/htmlmanager.ts
```diff
@@ -15,6 +15,16 @@ import { WidgetRenderer, WIDGET_MIMETYPE } from './output_renderers'
 export
 class HTMLManager extends base.ManagerBase<HTMLElement> {
 
+    constructor() {
+        super();
+        this.renderMime = new RenderMime();
+        // TODO: add default renderers
```
blink1073: cf https://github.com/jupyterlab/jupyterlab/blob/65395fa3d5befccaae0e9d99996dd3a122803fde/examples/notebook/src/index.ts#L98

packages/jupyterlab-manager/src/plugin.ts
```diff
@@ -127,7 +132,8 @@ export default widgetManagerProvider;
 /**
  * Activate the widget extension.
  */
-function activateWidgetExtension(app: JupyterLab, registry: IDocumentRegistry): INBWidgetExtension {
+function activateWidgetExtension(app: JupyterLab, docManager: IDocumentManager): INBWidgetExtension {
+  const { registry } = docManager;
```
blink1073: Should be `app.docRegistry`, no extra token required.

jasongrout: nice, thanks

PR #1503, opened 2017-07-14
Register the output plugin in jlab with the correct module and version.

jasongrout:
Thanks to @OliverEvans96 for finding this.


jasongrout:
There is one more issue in #1504 that needs to be investigated - the StyleView issue at the end.

jasongrout:
Let's take care of the other issue in #1504 in another issue.

PR #1530, opened 2017-07-20
Fix vertical slider in Safari.

### LINKED ISSUE #1511: Float Slider in safari has broken rendering/functionality
Working with @stargaser at Scipy2017 sprints on widget to html export, we discovered that Float sliders seem to improperly render in Safari in both the static html and the live notebook context. 

The slider value is clickable & it is capable of changing the value on the basis of the direction that you pull it, but it does not move. It seems to have no track along which to move, which may just be an issue with how Safari expects dimensions to be specified.

jasongrout:
Fixes #1511


jasongrout:
Tested in Chrome, Safari, and Firefox on OS X.

Issue #1401, opened 2017-06-01
Text.value not updated when pasting using mouse


sbmacdonald:
I have a Text widget, that will have a long uuid pasted into it.  The "on_change" event is not fired if I right-click and paste the value. Below the Text widget I have a Button that relies on the value of the Text widget being set.  Currently the only way to ensure the Text value is set, is for the user to press a navigation key (enter, tab, arrow), which seems to trigger the "on_change" event.

This is rather disappointing user experience, any thoughts ?

jasongrout:
What browser, version, and operating system is this on?

sbmacdonald:
Chrome & Firefox on Ubuntu 14.04

sbmacdonald:
I am happy to generate an AWS hosted sample notebook if that helps.

I LOVE jupyter and ipywidgets!

jasongrout:
Ah, looking at the code it looks like we explicitly look for key events: https://github.com/jupyter-widgets/ipywidgets/blob/d998e902f37a33ad963eb7de7f23961838dbfcb8/jupyter-js-widgets/src/widget_string.ts#L322

Perhaps we should instead listen for the `input` event: https://developer.mozilla.org/en-US/docs/Web/Events/input

jasongrout:
Would you like to work on this? PRs are definitely welcome!

sbmacdonald:
I am not a web developer, and I don't have very good understanding of web events.  If It weren't for these widgets I would be lost.  I am not adverse to helping out, but just not sure you want me mucking around in the code!

For me, (ideally) the value of the `Text` widget would be set/updated  whenever focus is changed away from the `Text` widget.  I would prefer not to have to register an observer to watch the `Text` value for changes.   Ideally a right-click-paste, followed by a change in focus to my "run" button would be enough for the `Text` value to be set.

jasongrout:
I think with the input event, the change would actually happen immediately, without blurring the input.

We have two modes for the text widget - continuous updates (using keypress events) and non-continuous updates (again using keypress events, but listening for the enter key specifically). I think we should experiment with changing these two modes to using the `input` event and `change` event, respectively.

maartenbreddels:
Another option would be to attach an event handler like Button click, so by default the value is always synced, but if you only want it on blur, you'd have to attach a event handler. I think that would be more flexible.

sbmacdonald:
I guess what I'm suggesting is that mouse events should trigger the Text value to be set just like key events.

[...]

PR #1545, opened 2017-07-27
Revamp event handling in text boxes and add continuous_update option

### LINKED ISSUE #584: Add a continuous_update option for text widgets
The sliders have a continuous_update option that only triggers a value change when the user stops dragging a slider: https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget_int.py#L155. Likewise, the text boxes should have an option to only update the value when the text box is blurred, or maybe when the user presses enter in a one-line textbox (sort of the equivalent of the on_submit handler textboxes currently have).

jasongrout:
- [x] Check the width of float/int/bounded text boxes now - they seem to be shorter.
- [x] deprecate the on_submit handler for text boxes, and handle the events similar to how things are done in this PR (with input and change events).

Also, specialize those inputs to type number, and add a step option.

 I default continuous_update to False for the textboxes because I figured if people were typing, that's typically what is wanted.

FIxes #584 and #1401.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -21,7 +21,7 @@ Major user-visible changes in ipywidgets 7.0 include:
   ```
 - Removed the version validation check since it was causing too many false warnings about the widget javascript not being installed or the wrong version number. It is now up to the user to ensure that the ipywidgets and widgetsnbextension packages are compatible. ([#1219](https://github.com/jupyter-widgets/ipywidgets/pull/1219))
 - Play range is now inclusive (max value is max, instead of max-1), to be consistent with Sliders
-- A refactoring of the text, slider, slider range, and progress widgets in resulted in the `BoundedIntText` and `BoundedFloatText` losing their `step` attribute (which was previously ignored), and a number of these widgets changing their `_model_name` and/or `_view_name` attributes ([#1290](https://github.com/jupyter-widgets/ipywidgets/pull/1290))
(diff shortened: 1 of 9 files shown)


jasongrout:
CC @mpacer

jasongrout:
@mpacer (or anyone), if you have a better name than "continuous_update", I'm all ears...

mpacer:
I can give it a spin before the end of the day :) thanks for getting to this so quickly!

mpacer:
I think it would help to include in the docstrings what the default behaviour is (i.e., when `continuous_update` is set to off). E.g., does it send when either you hit enter/return or change focus as asked for in #584?

mpacer:
Also it'd be good to have an example in the docs of both kinds of behaviour. This shouldn't need backend connections as far as I can tell so it should render nicely in the docs with an appropriate example.

jasongrout:
Oh, the issue there is that when you delete, it recognizes it's not a number, so it's patiently waiting for you to type something that *is* a number.

jasongrout:
(that's also why it resets when you click out - if the value is not valid, it resets to the previous value)

jasongrout:
> Oh, the issue there is that when you delete, it recognizes it's not a number, so it's patiently waiting for you to type something that is a number.

What if we put in a placeholder like "Type a number" or something? That would show if you deleted everything in the box. Would that be enough of a hint?

jasongrout:
@mpacer, I updated the docs. Thoughts?

jasongrout:
Merging - @mpacer, feel free to open another issue if you see something we missed.

Issue #1544, opened 2017-07-26
readout_format fails silently


adwasser:
IPython version 6.1.0

Any improper string input for the `readout_format` keyword argument of `FloatSlider` is associated with the interactive failing to display, without any indication that the `readout_format` input is bad.  

Is there a way to warn users about this?

![widgets](https://user-images.githubusercontent.com/4566243/28648049-d583a4d2-721f-11e7-892c-82407261acc1.png)

pbugnion:
Thanks for raising an issue. For reference, this is still broken on current master. The view throws when it renders.

For a minimal example:

```python
import ipywidgets as widgets
widgets.FloatSlider(1.0, readout_format='broken')
```

[This](https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_float.ts#L66) is the line that causes the exception. 

@jasongrout Do you think the solution to this is to define a more specific traitlet for the readout format, or to somehow expose the exception thrown by the JS client in the Python layer?

jasongrout:
> define a more specific traitlet for the readout format

I like this more than trying to push errors back to python. Is there a nice regex for these formats? What if we had a string traitlet that validated based on a regex?

jasongrout:
(Marking as Future, but if someone submits a PR in time for 7.0, that's great too)

pbugnion:
> What if we had a string traitlet that validated based on a regex?

[This](https://github.com/d3/d3-format/blob/master/src/formatSpecifier.js) seems useful. The bulk of the parsing is done by a regex.

> Marking as Future, but if someone submits a PR in time for 7.0, that's great too

Happy to take a look at this over the weekend, unless someone else gets there first.

jasongrout:
Fixed in #1550.

PR #1550, opened 2017-07-29
Add traitlet for validating number format

pbugnion:
Prior to this PR, the `readout_format` trait was validated as a unicode in the `{Float,Int}{,Range}Slider` widgets. If the format was not a valid d3 format, this lead to an exception being thrown client-side, but that exception wasn't visible to the user. This lead to issue #1544 .

This PR adds a trait that validates that the format string is correctly formatted. The code for this PR is very much inspired by the [format specifier parsing code](https://github.com/d3/d3-format/blob/master/src/formatSpecifier.js) in d3-format.

```python
>>> import ipywidgets as widgets
>>> widgets.FloatSlider(readout_format='..3f')
# [...] stack trace

TraitError: The 'readout_format' trait of a FloatSlider instance must be a valid number format, but a value of u'..3f' <type 'unicode'> was specified.
```

### DIFF
--- a/ipywidgets/widgets/__init__.py
+++ b/ipywidgets/widgets/__init__.py
@@ -5,7 +5,7 @@ from .widget import Widget, CallbackDispatcher, register, widget_serialization
 from .domwidget import DOMWidget
 from .valuewidget import ValueWidget
 
-from .trait_types import Color, Datetime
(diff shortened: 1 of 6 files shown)


jasongrout:
Thanks. Looks good to me, and thanks for adding tests too!

Issue #1551, opened 2017-07-29
htmlmanager does not seem to work with 3rd party widgets


maartenbreddels:
The output for this html is fine:
```python
import ipywidgets as widgets
import ipywidgets.embed
slider = widgets.FloatSlider()
slider
widgets.embed.embed_minimal_html('slider.html', slider)
```
However, using other libraries such as ipyvolume or pythreejs does not seem to work.
```python
from pythreejs import *
import numpy as np
from IPython.display import display
from ipywidgets import HTML, Text
from traitlets import link, dlink
ball = Mesh(geometry=SphereGeometry(radius=1), 
                                    material=LambertMaterial(color='red'),
                                    position=[2, 1, 0])

scene = Scene(children=[ball, AmbientLight(color='#777777')])

c = PerspectiveCamera(position=[0, 5, 5], up=[0, 0, 1],
                      children=[DirectionalLight(color='white', 
                                                 position=[3, 5, 1], 
                                                 intensity=0.5)])
renderer = Renderer(camera=c, 
                    scene=scene, 
                    controls=[OrbitControls(controlling=c)])
display(renderer)
```

Chrome's screenshot for network gives this:

![screen shot 2017-07-29 at 19 23 17](https://user-images.githubusercontent.com/1765949/28746913-1cf337da-7494-11e7-8cac-0bff7976841e.png)

When I put jupyter-threejs.js in the same folder, it gives this:
![screen shot 2017-07-29 at 19 24 35](https://user-images.githubusercontent.com/1765949/28746929-4a729606-7494-11e7-9f10-aa1e802fad55.png)

I tried to debug it, but I'm not sure what's going on here. Not that in the last case, it even tried to fetch `file:///Users/maartenbreddels/src/ipywidgets/@jupyter-widgets/base.js`
(pythreejs master and ipywidgets master)

pbugnion:
Good catch. I don't think that threejs has been migrated to ipywidgets 7.0.0. As far as I can tell, the migration would involve:
 - setting the `_model_module_version` traitlet explicitly -- at the moment it defaults to the version on the base widget (hence why it tries to look for jupyter-threejs@0.3.0 on unpkg).
 - changing references from `jupyter-js-widgets` to `@jupyter-widgets/base`.

We should probably write a migration guide for widget libraries and make an effort to help widget library developers with the migration?

[...]

PR #1556, opened 2017-07-31
WIP Export the @jupyter-widgets/base and @jupyter-widgets/controls modules in html manager

jasongrout:
Still very much a work in progress, but pushing here so there is a backup and people can see what I'm working on.

See #1551

### DIFF
--- a/examples/web4/README.md
+++ b/examples/web4/README.md
@@ -13,12 +13,12 @@ replace
 
 with
 
-`<script src="https://unpkg.com/@jupyter-widgets/html-manager/dist/index.js"></script>`
+`<script src="https://unpkg.com/@jupyter-widgets/html-manager/dist/embed.js"></script>`
 
 If you need a specific version of the HTML widget manager, you can include a
 semver range. For example:
 
-`<script src="https://unpkg.com/@jupyter-widgets/html-manager@^2.0.0/dist/index.js"></script>`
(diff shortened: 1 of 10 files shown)


jasongrout:
This is a bit more complicated: The html manager has several 'entry points'. It's not completely clear where to put the code that loads require on the page if it isn't already there. The code that exposes the htmlmanager's base and controls library of course should be *after* the code that ensures requirejs is on the page.

  - A library (`index`) that exports the html manager and a simple function for generating an embed script
  - A script (`embed-webpack`) that can just be included on a page, which automatically goes through every script tag representing widget state/views and renders them. This script also puts require on the page if it's not there.

jasongrout:
I ended up:

1. Webpacking a new html manager package into index.js
1. Making the embed script (now called embed.js) responsible for loading all of the scripts (base, controls, html manager) from unpkg.com using require.

stonebig:
does it mean jupyterlab is unsuable without internet connection ?

jasongrout:
> does it mean jupyterlab is unsuable without internet connection ?

This PR only deals with embedding widgets on a webpage, it has nothing to do with JupyterLab. This is only about enabling the workflow of 'get the minimal widgets infrastructure up on a page on the internet'. As a convenience, the embed.js provided here loads all of the widget infrastructure from unpkg.com, so you don't need anything installed locally.

jasongrout:
To really test this out, I'm going to have to merge and release a package. We'll iterate in another PR if needed.

PR #1558, opened 2017-08-01
Fix invalid readout_format in docs

pbugnion:
The widget list page currently displays a traceback because the readout_format
for some of the sliders is set to `'i'`, which is not a valid [d3 format](https://github.com/d3/d3-format#locale_format) (it's
also not a valid [Python format type](https://docs.python.org/3.4/library/string.html#format-specification-mini-language)).

Prior to PR #1550, this looked like it was working because d3-format doesn't throw
an exception when the wrong type gets passed in -- it just silently converts it
to the type `''`, corresponding to 'either decimal or exponent notation, but trim
insignificant zeros'.


jasongrout:
Thanks!

Issue #1552, opened 2017-07-30
Selection widgets docstrings not helpful


pbugnion:
It's not particularly easy to discover how to use selection widgets (e.g. select, dropdown etc.). To pass in options, one has to use the `options=` kwargs argument, but this isn't visible in the docstrings. The [docs](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Selection-widgets) are very detailed, but I think it would be useful to expand the docstring. I note that the docstrings for `Button` is very detailed, for instance.

I'm more than happy to write docstrings for the Selection widgets if we think it's useful.

jasongrout:
Definitely!

jasongrout:
Fixed by #1559.

PR #1559, opened 2017-08-01
Selection widgets docstrings

pbugnion:
This PR adds docstrings for the selection widgets.

Addresses issue #1552.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -370,7 +370,15 @@ class _MultipleSelection(DescriptionWidget, ValueWidget, CoreWidget):
 
 @register
 class ToggleButtonsStyle(Style, CoreWidget):
-    """Button style widget."""
+    """Button style widget.
+
+    Parameters
+    ----------
+
+    button_width: str
+        The width of each button. This should be a valid CSS
+        width, e.g. '10px' or '5em'.
+    """
     _model_name = Unicode('ToggleButtonsStyleModel').tag(sync=True)
     button_width = Unicode(help="The width of each button.").tag(sync=True)
 
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -23,7 +23,8 @@ from traitlets import (Unicode, Bool, Int, Any, Dict, TraitError, CaselessStrEnu
                        Tuple, List, Union, observe, validate)
 from ipython_genutils.py3compat import unicode_type
 
-_selection_parameters_docs = """
+_doc_snippets = {}
+_doc_snippets['selection_params'] = """
     options: dict or list
         The options for the dropdown. This can either be a list of values, e.g.
         ``['Galileo', 'Brahe', 'Hubble']`` or ``[0, 1, 2]``, a list of
@@ -58,7 +59,7 @@ _selection_parameters_docs = """
 """
 
 
-_multiple_selection_parameters_docs = """
+_doc_snippets['multiple_selection_params'] = """
     options: dict or list
         The options for the dropdown. This can either be a list of values, e.g.
         ``['Galileo', 'Brahe', 'Hubble']`` or ``[0, 1, 2]``, a list of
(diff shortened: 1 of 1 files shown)


ipywidgets/widgets/widget_selection.py
```diff
@@ -23,6 +23,40 @@
                        Tuple, List, Union, observe, validate)
 from ipython_genutils.py3compat import unicode_type
 
+_selection_parameters_docs = """
+    options: dict or list
+        The options for the dropdown. This can either be a
```
jasongrout: Can we 'demote' the dictionary representation and put in a note about order? Also, add an example where the values are not just strings?

> This can either be a list of values, e.g. ``['Galileo', 'Brahe', 'Hubble']`` or ``[0, 1, 2]``, a list of (label, value) pairs, e.g. ``[('Galileo', 0), ('Brahe', 1), ('Hubble', 2)]``, or a dictionary mapping the labels to the values, e.g. ``{'Galileo': 0, 'Brahe': 1, 'Hubble': 2}`` (in the case of a dictionary, the order of options is not guaranteed).

jasongrout:
This looks great. I had one comment inline. Perhaps we should also take this opportunity to document the options for the multiselect controls?

pbugnion:
> Perhaps we should also take this opportunity to document the options for the multiselect controls?

Yes, definitely.

maartenbreddels:
Just out of curiosity, why the `__doc__ = ...`?

pbugnion:
> Just out of curiosity, why the `__doc__ = ...`?

To avoid explicitly the same options in every docstring, I use a format string, e.g.

```
"""Group of toggle buttons that represent an enumeration.

    Only one toggle button can be toggled at any point in time.

    Parameters
    ----------
    {selection}
""".format(selection=_selection_parameter_docs)
```

AFAIK, you are allowed to use string interpolation in normal function / class docstrings, hence the need to explicitly assign it to `__doc__`. I'm not wedded to this implementation, though. I was planning on spending a bit of time seeing how we'd done this in other parts of the codebase and emulating that.

maartenbreddels:
Ah didn't know that. In vaex I use a [decorator](https://github.com/maartenbreddels/vaex/blob/2790103ad510ce8574f61856b38edcc173792a06/vaex/dataset.py#L773)

jasongrout:
> In vaex I use a decorator

That's a very interesting approach!

pbugnion:
We already use a decorator to inject docstrings in [widget_int](https://github.com/jupyter-widgets/ipywidgets/blob/master/ipywidgets/widgets/widget_int.py), so I ended up using decorators to inject the snippets, rather than manually modifying `__doc__` in every class. The implementation of the decorator is pretty much copied from vaex.

pbugnion:
This is now ready for review.

ipywidgets/widgets/widget_selection.py
```diff
@@ -406,8 +564,17 @@ class SelectionSlider(_SelectionNonempty):
         help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
 
 @register
+@_doc_subst
 class SelectionRangeSlider(_MultipleSelectionNonempty):
-    """Slider to select a single item from a list or dictionary."""
+    """
+    Slider to select multiple contiguous items from a list.
```
jasongrout: Can you add a note that the index, value, and label attributes indicate the first and last items in the range, rather than every element in the range?

ipywidgets/widgets/widget_selection.py
```diff
@@ -23,6 +23,99 @@
                        Tuple, List, Union, observe, validate)
 from ipython_genutils.py3compat import unicode_type
 
+_doc_snippets = {}
+_doc_snippets['selection_params'] = """
+    options: dict or list
```
jasongrout: Can you change this to "list or dict" to put the preferred value first?

jasongrout:
Two small comments, and then I think it's ready to merge.

jasongrout:
I went ahead and tweaked things like I suggested above. Thanks again!

pbugnion:
Thanks!

PR #1567, opened 2017-08-02
Typings fixes + fix save_changes callbacks

vidartf:
Some minor fixes I discovered while digging around in the code.

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -490,7 +490,7 @@ abstract class ManagerBase<T> {
      * @param data The initial data for the comm
      * @param metadata The metadata in the open message
      */
-    protected abstract _create_comm(comm_target_name: string, model_id: string, data?: any, metadata?: any): Promise<any>;
+    protected abstract _create_comm(comm_target_name: string, model_id?: string, data?: any, metadata?: any): Promise<any>;
     protected abstract _get_comm_info();
 
     /**
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -450,11 +450,12 @@ class WidgetModel extends Backbone.Model {
      *
      * This invokes a Backbone.Sync.
      */
-    save_changes(callbacks?) {
+    save_changes(callbacks?: { success: (model, response, options) => void, error: (model, response, options) => void }) {
         if (this.comm_live) {
-            let options: any = {patch: true}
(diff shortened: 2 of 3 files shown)


packages/base/src/widget.ts
```diff
@@ -450,11 +450,12 @@ class WidgetModel extends Backbone.Model {
      *
      * This invokes a Backbone.Sync.
      */
-    save_changes(callbacks?) {
```
jasongrout: The callbacks here are Jupyter callbacks, and don't have to do with Backbone. See where we use them in https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/widget.ts#L358

packages/base/src/manager-base.ts
```diff
@@ -490,7 +490,7 @@ abstract class ManagerBase<T> {
      * @param data The initial data for the comm
      * @param metadata The metadata in the open message
      */
-    protected abstract _create_comm(comm_target_name: string, model_id: string, data?: any, metadata?: any): Promise<any>;
+    protected abstract _create_comm(comm_target_name: string, model_id?: string, data?: any, metadata?: any): Promise<any>;
```
jasongrout: Why did you indicate that model_id is optional?

vidartf: Because it is being called with optional string e.g. here: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/manager-base.ts#L261

vidartf: Note, the docs of `WidgetOptions` say that either `comm` or `model_id` should be supplied, but it is used without it e.g. here: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_controller.ts#L276  (and it still seems to work).

packages/base/src/widget.ts
vidartf: Oh, I see. I'll change update the PR with typings to help indicate this.

vidartf:
I started working on improving the typings, but I think those could benefit from a more complete overhaul, so I'll close this for now.

vidartf:
Yes, that would be very nice. If there's time, we could possibly do a sprint on JupyterCon 👍

jasongrout:
Yes, good idea. We should also talk about the future plans after ipywidgets 7.

PR #1573, opened 2017-08-03
Fix the examples

### LINKED ISSUE #1473: Fix web3 and web-tmpnb examples
Currently these examples don't work because we don't register a renderer when we execute the code. We may just move them over to using the html manager (in which case, we should copy some of the changes we've been making in their widget manager back to the html manager, such as using async in loadClass, etc.)

jasongrout:
tmpnb will need to be upgraded to ipywidgets 7 before we can proceed with fixing the tmpnb example. Or we can switch to binder to load a specific kernel configuration (see https://github.com/jupyterhub/binderhub/issues/13).

Fixes #1473.

### DIFF
--- a/examples/web-tmpnb/README.md
+++ b/examples/web-tmpnb/README.md
@@ -1,11 +1,14 @@
-# Using jupyter-js-widgets in non-notebook, web context + tmpnb backend
+# Using Jupyter Widgets in non-notebook, web context with a tmpnb backend
+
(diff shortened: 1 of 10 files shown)


jasongrout:
I think this fixes the web3 and tmpnb examples as much as we can for now. The tmpnb example will need more work after they upgrade to ipywidgets, or be converted to a binder example.

PR #1583, opened 2017-08-04
Allow falsy view info other than undefined

vidartf:
Adjust the widget manager's view info check to allow falsy values other than undefined. This should be the case as `None`/`null` and empty strings (for the version) are allowed values (which indicates that the widget does not have a view).

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -248,7 +248,7 @@ abstract class ManagerBase<T> {
         let commPromise;
         // we check to make sure the view information is provided, to help catch
         // backwards incompatibility errors.
-        if (!options.view_name || !options.view_module || !options.view_module_version) {
+        if (_.contains([options.view_name, options.view_module, options.view_module_version], undefined)) {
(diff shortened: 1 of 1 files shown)


packages/base/src/manager-base.ts
```diff
@@ -248,7 +248,7 @@ abstract class ManagerBase<T> {
         let commPromise;
         // we check to make sure the view information is provided, to help catch
         // backwards incompatibility errors.
-        if (!options.view_name || !options.view_module || !options.view_module_version) {
+        if (_.contains([options.view_name, options.view_module, options.view_module_version], undefined)) {
```
jasongrout: Would you mind using the more standard Javascript construct:

```javascript
[options.view_name, options.view_module, options.view_module_version].indexOf(undefined) > 0
```
We only have one other place where we use underscore in this file, and I think it's better to use standard javascript these days where it isn't too much more work.

jasongrout: (if you have strong feelings about using underscore instead, that's fine too)

jasongrout: Merging as-is.

PR #1585, opened 2017-08-04
In interact_output, clear with wait=True to prevent flicker.

### LINKED ISSUE #1582: `interactive_output` flickers
Similar to the problem described [here](https://github.com/jupyter-widgets/ipywidgets/issues/910), I want to customize the layout of my interactive widgets. The kludgy approach is to call `interactive` then get the children out, put them in a specific layout, and put them back in. Like so:

```
import numpy as np
from ipywidgets import *
import matplotlib.pyplot as plt

def plot_oscillation(wavelength, magnitude, phase):
    x = np.linspace(-20, 20, 1000)
    plt.plot(x, magnitude*np.sin((x+phase)/(wavelength/(2*np.pi))))
    plt.show()
    
w = dict(wavelength = FloatSlider(2*np.pi, min=0.1, max=20, step=0.1, description='wavelength'),
         magnitude = FloatSlider(1, min=0.1, max=10, step=0.1, description='magnitude'),
         phase = FloatSlider(0, min=0, max=10, step=0.1, description='phase'))

int_w = interactive(plot_oscillation, **w)

output = int_w.children[-1] # last child is the output

int_w.children = [HBox([VBox([*int_w.children[:-1]]), output])]
display(int_w)
```
This approach works and does *not* flicker, but it's pretty clunky. The `interactive_output` function was created to provide a better way to do this sort of thing. It works but flickers. See here:

```
import numpy as np
from ipywidgets import *
import matplotlib.pyplot as plt

def plot_oscillation(wavelength, magnitude, phase):
    x = np.linspace(-20, 20, 1000)
    plt.plot(x, magnitude*np.sin((x+phase)/(wavelength/(2*np.pi))))
    plt.show()
    
w = dict(wavelength = FloatSlider(2*np.pi, min=0.1, max=20, step=0.1, description='wavelength'),
         magnitude = FloatSlider(1, min=0.1, max=10, step=0.1, description='magnitude'),
         phase = FloatSlider(0, min=0, max=10, step=0.1, description='phase'))

output = interactive_output(plot_oscillation, w)
box = HBox([VBox([*w.values()]), output])
display(box)
```
It seems like it might be related to [this](https://github.com/jupyter-widgets/ipywidgets/issues/145) now-fixed issue with `interact`.

jasongrout:
Fixes #1582

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -67,10 +67,10 @@ def interactive_output(f, controls):
 
     out = Output()
     def observer(change):
-        out.clear_output()
         kwargs = {k:v.value for k,v in controls.items()}
         show_inline_matplotlib_plots()
         with out:
+            clear_output(wait=True)
             f(**kwargs)
             show_inline_matplotlib_plots()
     for k,w in controls.items():


jasongrout:
Testing this does indeed fix the problem in #1582.

DohunLee:
@jasongrout 
If you want to share your notebook, but you have applied the "wait=True" changes to the package in your environment, does this fix the flickering in the shared environment?

If not how could you resolve that issue?

PR #1595, opened 2017-08-06
Handle buffers in `Widget._should_send_property`

### LINKED ISSUE #1592: Receiving binary buffers from frontend fails
If receiving buffers from the front end, the following line fails: https://github.com/jupyter-widgets/ipywidgets/blob/master/ipywidgets/widgets/widget.py#L601

The error is `TypeError: <memory at 0x000001E2CA4F8708> is not JSON serializable`.

An outline of the trace is:
 - Comm message with buffer received. Calls `set_state`.
 - `set_state`:
   - Locks the properties.
   - Deserializes the data (to numpy array in this case).
   - Calls `set_trait`.
 - `notify_change` calls `_should_send_property`.
 - `to_json` (my serializer) gives back a dict with a memoryview.
 - `jsondumps` chokes on the memoryview.

vidartf:
Previous check in `_should_send_property()` did not handle buffers. After this change, the comparison is performed in the serialized state which also prevents creation of new widgets/instances during comparison.

Fixes #1592.


jasongrout:
@maartenbreddels - the last commit (7d56205) changes the definition of what is a binary object in python - basically it adds `bytearray` and gets rid of the python 2 `buffer` (which is superseded by `memoryview`, even on python 2).

What do you think?

jasongrout:
@vidartf - what do you think of my changes? I feel like this is ready to merge now, so if you and @maartenbreddels approve it, let's merge.

maartenbreddels:
LGTM 👍

ipywidgets/widgets/widget.py
```diff
@@ -597,10 +623,16 @@ def _should_send_property(self, key, value):
...
-        if (key in self._property_lock
-            and jsonloads(jsondumps(to_json(value, self))) == self._property_lock[key]):
-            return False
-        elif self._holding_sync:
+        if key in self._property_lock:
+            # model_state, buffer_paths, buffers
+            split_value = _remove_buffers({ key: to_json(value, self)})
+            split_lock = _remove_buffers({ key: self._property_lock[key]})
+            # Compare state and buffer_paths
+            if (jsonloads(jsondumps(split_value[0])) == jsonloads(jsondumps(split_lock[0]))
```
vidartf: I would rather pass `sort_keys=True` to `dumps` rather than pass it back through `loads`, but it is not critical and I have not profiled the two.

ipywidgets/widgets/widget.py
```diff
@@ -483,7 +483,7 @@ def _compare(self, a, b):
         else:
             return a == b
 
-    def _compare_buffers(self, a, b, key):
+    def _buffer_list_equal(self, a, b):
```
vidartf: If we have this as an extension point, keeping the key could be very informative as to which kind of comparison to perform. If anything, it's a problem with the current `_compare` that it does not have it.

vidartf:
Went through it, and only found some minor points commented above. Neither are sticking points for me, so feel free to make the final decision on them as you see fit and merge.

ipywidgets/widgets/widget.py
vidartf: E.g. for certain keys you might want to do a floating point comparison with a given precision, while for another you need exact equality. Without the key, you will not be able to determine which comparison to apply.

jasongrout: Can you elaborate on this being an extension point? I don't understand how you're thinking of this as an extension point.

vidartf: I mean extension point as in "something an inheriting class might want to override".

ipywidgets/widgets/widget.py
jasongrout: Done. I also factored out this comparison into a top-level function.

ipywidgets/widgets/widget.py
jasongrout: Actually, this comparison depends on the top-level put/remove buffer functions, so I factored out the comparison to also be a top-level function.

jasongrout: > I mean extension point as in "something an inheriting class might want to override".

Ah. Since it depends on the implementation of the top-level functions, I think it makes more sense to not override it, and instead put it at the top level too.

vidartf: The current implementation works for me, but if someone uses a serializer/deserialize set that might introduce noise when data is passed through a loop (`a != serializer(deserialize(a))`), e.g. floating point inaccuracies, they will want to override this comparison.

It is still possible to do that by overriding the `_should_send_property` function, but it will be more brittle as it needs to take into account the internals of that method (e.g. `_property_lock`), making it more brittle.

vidartf: Actually, I might be over-engineering this one. Other fields in the state other than buffers might also have this issue, so there is no obvious reason to give buffers preferential treatment.

jasongrout:
One more iteration to try to avoid copying if possible (in python 3, or python 2 with byte memoryviews). @maartenbreddels - does 1c4d8a651bc3f6df361620e9bdfe041800fc65ec look good to you?

jasongrout:
@vidartf said on gitter: 

> @jasongrout Does memoryview.tobytes() actually copy the buffer? I think your changes are ok either way, but just FMI

PR #1604, opened 2017-08-08
Access widgets like a dict

draperjames:
So this can be used to change the attributes of the child widgets themselves but I can't seem to update the children attribute itself. For example if a widget `w` has a children `widgets.HTML()` with title of `'hello'` then this works`w.as_dict["hello"].value = "Change"` works just fine but `w.as_dict["hello"] = widgets.Button(description="hi")` will not update. From what I gather one does not simply make a setter for a traitlet...

### DIFF
--- a/ipywidgets/widgets/widget_selectioncontainer.py
+++ b/ipywidgets/widgets/widget_selectioncontainer.py
@@ -59,6 +59,36 @@ class _SelectionContainer(Box, CoreWidget):
             return self._titles[index]
         else:
             return None
+    
+    @property
+    def as_dict(self):
+        """Access the children of the selection container like a dictionary."""
+        # Test to make sure every section has a child widget.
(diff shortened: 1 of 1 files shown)


draperjames:
[This gist](https://gist.github.com/draperjames/f777dcc7d1871dd3cd139f40b3d40b38) gives you a preview of the behavior in a notebook without having to do all the git business.

draperjames:
The `as_dict` function should probably be rewritten to use the eventful traits from #1071 once it is merged. I think that some type of eventful dict is necessary to have getter and setter properties. 

## To do list for setter method
- [ ] Wait for merger of #1071.
- [ ] Rewrite method to call the eventful dict as a property.
- [ ] Wire that to children with a observer that children is updated on any change in the eventful dict.

Issue #1598, opened 2017-08-07
Error initializing output widget in notebook when comm does not exist


pbugnion:
To trigger the bug, 

1. create an output widget with, e.g.:

```python
from ipywidgets import interact
def f(x):
    return x
interact(f, x=10)
```

2. save the widget state (using the *Widgets* > *Save Notebook Widget State* menu entry).

3. restart the kernel

4. reload the browser client by reloading the page

This triggers an error `Uncaught (in promise) TypeError: Cannot read property 'kernel' of undefined`. AFAICT, the error is caused [here](https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/src/widget_output.js#L24). We dereference `this.comm`, but `this.comm` can be null. 

Happy to look into this this week, unless someone gets there first.

jasongrout:
> Happy to look into this this week, unless someone gets there first.

+1

jasongrout:
Fixed by #1605.

PR #1605, opened 2017-08-09
[WIP] Avoid error initializing output widget in notebook when comm is missing

pbugnion:
Prior to this commit, the output widget would crash the extension if there was no comm available when it was instantiated. This might happen if the kernel is halted and the page is reloaded. See issue #1598.

Now it just displays a 'disconnected' sign like the other widgets:

screen shot 2017-08-09 at 07 47 33

Missing:
 - [ ] unit tests


jasongrout:
This works great for me. Thanks!

PR #1608, opened 2017-08-09
Expose API on jlab extension

### LINKED ISSUE #1607: Add registerWidget to the INBWidgetExtension interface
@vidartf  pointed out that we should add the  registerWidget signature to INBWidgetExtension 
 at https://github.com/jupyter-widgets/ipywidgets/blob/cdb088d0a3c2b7a0f246d6dffb8a3a2534d8c5f3/packages/jupyterlab-manager/src/plugin.ts#L59

vidartf:
Exposes the `registerWidget` function on the extension interface, so that the correct typings are declared and exported.

Fixes #1607.

### DIFF
--- a/packages/jupyterlab-manager/src/plugin.ts
+++ b/packages/jupyterlab-manager/src/plugin.ts
@@ -56,7 +56,7 @@ const INBWidgetExtension = new Token<INBWidgetExtension>('jupyter.extensions.nbW
  * The type of the provided value of the plugin in JupyterLab.
  */
 export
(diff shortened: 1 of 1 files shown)


jasongrout:
Travis has an error in the js tests: 

>   SyntaxError: expected expression, got ')'
>   at test/build/bundle.js:17889

vidartf:
The Travis error doesn't make sense to me: it's in the tests for the base package? Can also not reproduce locally. Would you mind restarting it to check if it was a fluke?

vidartf:
I added a pure formatting commit, and now the tests pass. I blame cosmic rays.

PR #1611, opened 2017-08-09
Downgrade the js in the webpack config to work with older node versions.

jasongrout:


blink1073:
How did you `let` this happen?  😜

jasongrout:
*groan*

PR #1615, opened 2017-08-10
Use require.js to load the embed script

jasongrout:
This allows us to easily automatically use the local base, controls, and html manager next to the embed script, and also makes it much clearer what is going on (for example, you *have* to have require on the page to get the module).

The first commit in this PR takes a different approach - separate `embed-cdn` and `embed-local` files that hardcode the destination. It's a bit easier to put on the page, but it obfuscates that we are using require and how the functions are run after page load.


examples/web4/README.md
```diff
@@ -13,12 +13,12 @@ replace
 
 with
 
-`<script src="https://unpkg.com/@jupyter-widgets/html-manager/dist/embed.js"></script>`
+`<script src="https://unpkg.com/@jupyter-widgets/html-manager/dist/embed-cdn.js"></script>`
```
jasongrout: TODO: fix these links.

jasongrout:
I've settled on a different approach. We still have the `dist/embed.js` that lets you embed standard widgets, and does *not* use requirejs, so it can't load third-party widgets. We also have a requirejs module, `dist/embed-require.js`, which *does* use require.js and lets you load custom widgets.

jasongrout:
Merging this so we can release a new version of html-manager, test, and iterate.

PR #1619, opened 2017-08-11
Make the inner html widget expand to fill the available space.

### LINKED ISSUE #1613: Question on HTML widget width
Trying to set the max_width on an HTML widget, but there is a div in the way. Referring to this picture:

screenshot from 2017-08-09 19-11-25

You'll see that I want the "Hello, World!" text and blue background to span the width of the cell (this is just an example of a larger HTML problem involving in iframe, but is a simple example). If I give a layout object, you can see that goes to the div above. The text has its own width. But the div between them (or the label div) could either be made to expand to 100% to allow the Text to span the region. I added a style= width: 100% and it did work. 

How can I access either the content or the description width to make room? Or is there another way?

jasongrout:
Fixes #1613

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -1065,3 +1065,14 @@
 .widget-html, .widget-htmlmath {
     font-size: var(--jp-widgets-font-size);
 }
+
+.widget-html > .widget-html-content, .widget-htmlmath > .widget-html-content {
+    /* Fill out the area in the HTML widget */
+    align-self: stretch;
+    flex-grow: 1;
+    flex-shrink: 1;
+    /* Makes sure the baseline is still aligned with other elements */
+    line-height: var(--jp-widgets-inline-height);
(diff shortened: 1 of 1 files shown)


jasongrout:
CC @dsblank

jasongrout:
@dsblank - can you check if these CSS changes are sufficient for you?

dsblank:
Thanks for taking a look at this! (Is there an easy way to make such a change in a non-development setup of jupyter/ipywidgets?)

dsblank:
Here is the cell, and it works perfectly! Thanks!
```
%%html

<style>
.widget-html > .widget-html-content, .widget-htmlmath > .widget-html-content {
     /* Fill out the area in the HTML widget */
     align-self: stretch;
     flex-grow: 1;
     flex-shrink: 1;
     /* Makes sure the baseline is still aligned with other elements */
     line-height: var(--jp-widgets-inline-height);
     /* Make it possible to have absolutely-positioned elements in the html */
     position: relative;
}
</style>
```

PR #1621, opened 2017-08-11
Set default layout to {} at frontend

maartenbreddels:
fixes https://github.com/jupyter-widgets/ipywidgets/issues/1620

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -281,7 +281,7 @@ abstract class ManagerBase<T> {
             options_clone.comm = comm;
             let widget_model = this.new_model(options_clone, serialized_state);
             return widget_model.then(model => {
-                model.sync('create', model, {drop_defaults: true});
+                model.sync('create', model);
                 return model;
             });
         }, () => {
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -281,7 +281,7 @@ abstract class ManagerBase<T> {
             options_clone.comm = comm;
             let widget_model = this.new_model(options_clone, serialized_state);
             return widget_model.then(model => {
-                model.sync('create', model);
+                model.sync('create', model, {drop_defaults: true});
                 return model;
             });
         }, () => {
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -544,7 +544,7 @@ class DOMWidgetModel extends WidgetModel {
 
     defaults() {
         return _.extend(super.defaults(), {
-            layout: null,
+            layout: {},
             style: null,
             _dom_classes: []
         });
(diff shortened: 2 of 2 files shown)


jasongrout:
Doesn't style have the same issue?

maartenbreddels:
That's a bit confusing, DOMWidget doesn't have a style attribute (on the backend), but some subclasses do. I don't feel comfortable removing it so close before a release.

jasongrout:
Oh yeah - the fact that domwidget has style in the browser is an implementation detail, not part of the spec. We likely will have the same issue on any widgets that actually have a (nontrivial) style attribute.

jasongrout:
Can you also add a default of `style = {}` for ButtonModel, DescriptionModel, FloatSliderModel, FloatProgressModel, FloatRangeSliderModel, IntSliderModel, IntProgressModel, IntRangeSliderModel, ToggleButtonsModel?

jasongrout:
In general, I'm not sure new_widget should be transferring the defaults.

jasongrout:
@maartenbreddels - what if we don't sync the defaults? i.e., what if we change https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/manager-base.ts#L284 to have the options.drop_defaults (see https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/widget.ts#L335)

jasongrout:
This would rely on the defaults being *exactly* the same between the frontend and backend, otherwise there will be inconsistent state, of course. Or perhaps the kernel defaults would be sent over, since they wouldn't be in the property_lock. That might be better anyway.

jasongrout:
@maartenbreddels - can you try not syncing the defaults instead of the change you made to the layout default?

maartenbreddels:
That's a much better plan I think. I'll first try to confirm that #1622 introduces the bug seen in #1621.

maartenbreddels:
It works, I wouldn't be surprised if this will show up other bugs though where defaults are not properly set on both sides. But otoh, otherwise they only show up when doing embedding, this will make them show up earlier.

jasongrout:
Actually, thinking more about it, this may introduce a subtle race condition. Suppose I create a widget from the js side, which goes over basically without attributes to python. Then I start modifying it in js, and thus send a message about the update. I get a message back from python with the default values, which then overwrite my new values. Now I have the new values in python, but the default values in js.

Here's another idea - just delete the default value altogether for layout and style on the js side, and make sure the js code can deal with it. It should be created in the python side and automatically synced back.

jasongrout:
(i.e., make sure the views can deal with those two values being undefined)

jasongrout:
@maartenbreddels - can you check to see if my latest commit here also solves the problem? This way, the state messages still follow the spec (which doesn't allow for an empty dictionary).

jasongrout:
I was trying to reproduce your error, and I was getting confused that I was not seeing the error message. Then I happened to find it in the websocket messages, even though it wasn't displaying. But then I realized that of course it wouldn't display - the callbacks are not set for that `new_widget` sync message - in fact, where should it display? Typically new_widget is called from a model, so there is no clear place to display.

It just reminded me yet again that we should have a clear notebook-wide display for messages that don't find a home, especially error messages.

Issue #1624, opened 2017-08-12
Audit documentation


jasongrout:
I noticed a few places where the documentation headings were off, for example in the Async Widgets page: http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Asynchronous.html - some of the headings seem to show up in the table of contents.

Someone should just look at all of the latest documentation at http://ipywidgets.readthedocs.io/en/latest/ and verify that it looks okay.

- [ ] Noticed this in the last example in Using Interact.ipynb - set just the output to have a specific size.
- [ ] layout attributes in widget styling don't look very good - break into headings instead of a nested list.

mwcraig:
I'll have a PR open in a bit with some really minor fixes. Noticed this in the last example in `Using Interact.ipynb`:

<img width="267" alt="using_interact_ _jupyter_widgets_7_0_0b12_documentation" src="https://user-images.githubusercontent.com/1147167/29257445-aa214932-8077-11e7-82d1-106529ee67bf.png">

mwcraig:
Widgets are missing in the `Widget Events.ipynb` notebook.

mwcraig:
css properties are not layed out correctly (though the source notebook looks fine): http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Exposed-CSS-properties

mwcraig:
Also in `Widget Styling`, a link which looks fine in the source but not in the output: http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Acknowledgement

mwcraig:
Further down in the section on layout in `Widget Styling`, this (again looks fine in the source):

<img width="893" alt="layout_and_styling_of_jupyter_widgets_ _jupyter_widgets_7_0_0b12_documentation" src="https://user-images.githubusercontent.com/1147167/29257625-1858fb2e-8079-11e7-9d67-ad60e1f82e40.png">

jasongrout:
> I'll have a PR open in a bit with some really minor fixes. Noticed this in the last example in Using Interact.ipynb:

This is because the height is set on the whole interact, but the image expands to fill space with a given aspect ratio, for some reason squishing the sliders. We should maybe set the image size as a fixed size?

jasongrout:
@mwcraig - I think I fixed each of the issues you pointed out above. Thanks again for going through those.

jasongrout:
[...]

PR #1625, opened 2017-08-14
Fix some minor documentation issues

mwcraig:
This should fix the issue with heading in the asynchronous notebook, and makes a couple other pretty minor fixes. The other oddities I noticed (but which don't seem to be problems in the notebooks themselves) are noted in the comments on #1624.


jasongrout:
Looks great, thanks!

PR #1629, opened 2017-08-14
Make a script you can load which implements the amd embedding.

jasongrout:
CC @maartenbreddels.

The goal is to enable this for the amd embedding:

```html
<script src='https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js'></script>
<script src='https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embedlib-amd.js'></script>
<script src='https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js'></script>
```


jasongrout:
Perhaps it should be `libembed` rather than `embedlib`?

jasongrout:
Yeah, I like libembed better - implemented in the last commit.

jasongrout:
Let's let tests finish, merge, release, and iterate.

maartenbreddels:
Great, that means that jupyter-sphinx needs this right:
```python
        app.add_javascript('https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js')
        app.add_javascript('https://unpkg.com/@jupyter-widgets/html-manager@%s/dist/libembed-amd.js' % __html_manager_version__)
        app.add_javascript('https://unpkg.com/@jupyter-widgets/html-manager@%s/dist/embed-amd.js' % __html_manager_version__)
```

jasongrout:
Yes, I think so.

jasongrout:
Iterating on this - this may not work. Requirejs is getting the wrong location for picking up the local amd modules. We may need to have a few more script tags.

jasongrout:
Even better idea - bundle all of the defined requirejs modules into one file, which we'll load at once rather than in separate require statements. In the next PR.

maartenbreddels:
If you manage to come up with just a single file, that loads require when not present, and supports 3rd party that would be a big win btw.

jasongrout:
> Even better idea - bundle all of the defined requirejs modules into one file, which we'll load at once rather than in separate require statements. In the next PR.

This is now #1630

PR #1630, opened 2017-08-15
Concat embed amd files together to have one file to include

jasongrout:
This concatentates all of the amd embedding scripts together so that we just have one file to include as a script tag. This ensures that there is only one request, and that all of the version numbers are consistent.

Now a custom widget embedding just requires:

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.js" integrity="sha256-zPseknSmTN790o9fjoPFR/gHHmF2APMkKLgHke4RTM4=" crossorigin="anonymous"></script>
<script src="https://unpkg.com/@jupyter-widgets/html-manager/dist/embed-amd.js"></script>
```


jasongrout:
CC @maartenbreddels

PR #1633, opened 2017-08-15
Resolve instead of return Promise

maartenbreddels:
followup of jupyter-widgets/ipywidgets#1630
```javascript
var p = new Promise((resolve, reject) => {
    return Promise.resolve(1) // this does not resolve
}).then((value) => {
    console.log('got value', value)
})

var p = new Promise((resolve, reject) => {
    resolve(Promise.resolve(1)) // should be this or maybe .then(resolve, reject) ?
}).then((value) => {
    console.log('got value', value)
})
```
Is this pattern maybe present in more places?


jasongrout:
Great catch!

PR #1637, opened 2017-08-15
Fix the output interact height example, regenerate widgets.

jasongrout:
See #1624


jasongrout:
I also redid the flex tutorial. I'm still not convinced we need a flex tutorial when the standard one is more comprehensive.

jasongrout:
Seems to be failing in an unrelated way (with html manager).

PR #1643, opened 2017-08-16
Fix binary synchronization and make Image use binary values

### LINKED ISSUE #1642: Page refresh for notebook does not work for widgets with buffers
Using latest ipyvolume on pypi:
```python
import ipyvolume as ipv
ipv.examples.klein_bottle()
```
Refresh, and one of my deserializers is missing a .buffer property.

jasongrout:
Fixes #1642 by working around an underscore bug.

This also makes the image use a binary value so that we have at least one core widget using the binary values and messages.

This means that custom deserializers must not return a DataView or ArrayBuffer (other typed arrays like Uint8Array are fine, though).

Test code:

```python
from ipywidgets import Image
from io import BytesIO
import numpy as np
from scipy.misc import imsave
x = np.zeros((255, 255), dtype=np.uint8)
x[:] = np.arange(255)
b = BytesIO()
imsave(b, x, format='png')

im = Image(layout=dict(width='100px', height='100px', border='1px solid red'))
im
```
Then
```
im.value=b.getvalue()
```
and also make sure the image is displayed when the page is refreshed.
CC @maartenbreddels


packages/controls/src/widget_image.ts
```diff
@@ -43,9 +44,13 @@ class ImageView extends DOMWidgetView {
          * Called when the model is changed.  The model may have been
          * changed by another view or by a state update from the back-end.
          */
-        var image_src = 'data:image/' + this.model.get('format') + ';base64,' + this.model.get('_b64value');
-        this.el.setAttribute('src', image_src);
-
+        var blob = new Blob([this.model.get('value')], {type: `image/${this.model.get('format')}`});
+        var url = URL.createObjectURL(blob);
+        var oldurl = this.el.src;
+        this.el.src = url;
```
maartenbreddels: I think the createObjectURL is deprecated (firefox gives you a warning on that).
https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL
The proper way would be to use `this.el.srcObject = blob` I think. We may want to polyfill this with https://github.com/webrtc/adapter , which is what I do in https://github.com/maartenbreddels/ipywebrtc

maartenbreddels: Sorry for the noise, that was only for HTMLMediaElement, not HTMLImageElement

jasongrout:
@maartenbreddels - our problem is more serious than my simple workaround. Since underscore is doing the deep compare, *any* arraybuffer or dataview object anywhere nested in the JSON of a deserialized trait is going to have this issue.

I'm tempted to just override the backbone function and insert a special-case call to our own isequal.

maartenbreddels:
I guess both Backbone's set and changedAttributes?

jasongrout:
I think we override set and call our own isEqual.

We're slowly completely reimplementing backbone...

maartenbreddels:
Or we fork and publish underscore with a fixed _.isEquals, maybe less work.

packages/controls/src/widget_image.ts
jasongrout: I took some of this code from https://developer.mozilla.org/en-US/docs/Web/API/Blob

maartenbreddels: Yes, it's completely correct.

jasongrout:
Note that lodash has specific references to typed arrays, indicating they handle that case: https://github.com/lodash/lodash/blob/8e631dfcd496bc355ee7ceeb959421b0788b9bbc/.internal/baseIsEqualDeep.js#L56

jasongrout:
> Or we fork and publish underscore with a fixed _.isEquals, maybe less work.

And then we have to fork backbone and update its dependency. I really don't want to maintain an underscore fork, especially when lodash seems to work in this case.

jasongrout:
[This](https://github.com/lodash/lodash/wiki/Migrating#backbone) indicates that we can replace underscore with lodash for backbone, but requires a special webpack config. We can't do that in jlab, though. Too bad.

jasongrout:
@maartenbreddels - I think this is ready for review and testing. I could also remove underscore from controls, but that's not pressing.

PR #1644, opened 2017-08-16
Comment out an unnecessary css import.

jasongrout:
See https://github.com/maartenbreddels/ipyvolume/issues/41#issuecomment-322870907

### DIFF
--- a/packages/controls/css/labvariables.css
+++ b/packages/controls/css/labvariables.css
@@ -30,7 +30,8 @@ all of MD as it is not optimized for dense, information rich UIs.
 /*
  * Optional monospace font for input/output prompt.
  */
-@import url('https://fonts.googleapis.com/css?family=Roboto+Mono');
+ /* Commented out in ipywidgets since we don't need it. */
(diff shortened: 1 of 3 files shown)


packages/controls/css/labvariables.css
```diff
@@ -30,7 +30,8 @@ all of MD as it is not optimized for dense, information rich UIs.
 /*
  * Optional monospace font for input/output prompt.
  */
-@import url('https://fonts.googleapis.com/css?family=Roboto+Mono');
+ /* Commented out in ipywidgets since we don't need it. */
+/* @import url('https://fonts.googleapis.com/css?family=Roboto+Mono'); */
```
vidartf: The comment seems strange to me ("in ipywidgets"). Also: Why is it commented out and not simply deleted?

vidartf: Oh, sorry for the drive-by comments: I see now that is file is a simple copy from the JLab source.

jasongrout: Yes - I wanted to give some indication for when the copy is updated what should happen.

Issue #1650, opened 2017-08-17
Offline font-awesome use


chrisjsewell:
As discussed in maartenbreddels/ipyvolume#41 and maartenbreddels/ipyvolume#55, I can currently embed a custom widget, such that it can be rendered offline, with the following scripts:

```html
<link href="script_dir/font-awesome/css/font-awesome.min.css" rel="stylesheet">
<script src="script_dir/require.min.js" crossorigin="anonymous"></script>
<script src="script_dir/embed-amd.js" crossorigin="anonymous"></script>

``` 

and directory structure:

```
my_embedded.html
ipyvolume.js
script_dir/
    embed-amd.js    
    font-awesome/
    require.min.js
```

Is it expected that I need a local copy of the font-awesome directory, or is this something embed-amd.js should handle?

To illustrate the issue, if I remove https://github.com/chrisjsewell/chrisjsewell.github.io/blob/master/3d_atomic/converted/3D%20Atomic%20Visualisation_files/ivol_example1.html#L11 in my test file and try to open it offline (having cleared my cache), I get:

<img width="528" alt="image" src="https://user-images.githubusercontent.com/2997570/29387887-256fc974-82da-11e7-939e-45e3c64abbd8.png">

jasongrout:
The fonts are stripped out from the webpack bundle, and webpack probably tries to load the from the best guess of where they are. Let me try something...

jasongrout:
@chrisjsewell, @maartenbreddels - can you see the solution in #1652 and see if that works for your usecase?

chrisjsewell:
Cheers. Just to clarify, for my example directory structure, would this mean that I would use:

```html
<script>
        window.__jupyter_widgets_assets_path__ = './script_dir/';
</script>
<script src="script_dir/require.min.js" crossorigin="anonymous"></script>
<script src="script_dir/embed-amd.js" crossorigin="anonymous"></script>
```

I assume I still need a local font-awesome folder present,  and I'll still need a require.config call if I wished to have ipyvolume.js in script_dir?

jasongrout:
> I would use:

Yes



> I still need a local font-awesome folder present

No, not for the base widgets. If ipyvolume is also including fontawesome, I don't know.

> I'll still need a require.config call if I wished to have ipyvolume.js in script_dir

Yes. I think.

jasongrout:
[...]

PR #1652, opened 2017-08-17
Upgrade webpack and add a way to give the path for assets

jasongrout:
Here's how to use the new `__jupyter_widgets_assets_path__` to give the location of assets, like the fontawesome font:

```html
<script>
        window.__jupyter_widgets_assets_path__ = './dist/';
</script>
<script src="./dist/embed.js" crossorigin="anonymous"></script>
```
This provides a workaround for #1650.


jasongrout:
Perhaps better would be to default to wherever the script is loaded from, but apparently finding the path of the currently executing script is a very tricky thing.

PR #1658, opened 2017-08-18
instantiate IPython when testing _ipython_display_

minrk:
if IPython isn't instantiated, display calls print (IPython master)


jasongrout:
Thanks!

PR #1659, opened 2017-08-18
Widget manager state in classic notebook

jasongrout:
This PR does two main things:

1. Provides better logic for showing and updating a disconnected state
2. Makes it clearer in the classic notebook what is live and what is not, and fixes an error for embedding where it used to always look at old widget state rather than current widget state.

Actually, it looks like (2) is never really triggered - I'm not sure in what cases a completely new kernel is created. It seems that restarting, shutting down, and switching kernels all keeps the same kernel object with the same widget manager attached. Only refreshing the page, of course, creates a new javascript kernel object.

CC @minrk, who may know when a completely new `Jupyter.notebook.kernel` object is created.


jasongrout:
A sensible case could be made that restarting a kernel should behave like killing and starting, but that requires a lot more code changes this close to release, and it's not clear that that is always what is wanted.

minrk:
I just did a bit of reading, and I'm pretty sure the Kernel is never recreated on a classic nb page once it exists. Every path I found that could start a kernel or session has something like:

```javascript
if (it_already_exists) {
  update_it()
} else {
  get_a_new_one()
}
```
It would be reasonable to instantiate a new Kernel when doing things like switching kernelspecs, but instead the existing object is modified rather than replaced.

jasongrout:
Okay, that explains my confusion, thanks! I was starting to wonder if I was missing something big.

I did notice when I switched from a python kernel to an R kernel, it seemed to keep the same kernel, but when switching *back* to a python kernel, it actually refreshed the page, which of course then instantiates a new kernel object.

minrk:
> but when switching back to a python kernel, it actually refreshed the page

That could be if the R kernel has a `kernel.js`. Since that can arbitrarily modify the page, when switching away from a kernel with kernel.js, the page is refreshed to avoid any undesirable effects lingering.

jasongrout:
That makes sense, @minrk. Thanks!

PR #1674, opened 2017-08-26
ENH: Give a clearer error for widget js.

ssanderson:
Adds a new 'text/html' entry in the mimebundle for widgets, which points
users to the installation instructions in the widgets docs.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -758,7 +758,7 @@ _FALLBACK_HTML_TEMPLATE = """\
 <p>Failed to display Jupyter Widget of type <code>{widget_type}</code>.</p>
 <p>
   If you're reading this message in Jupyter Notebook or JupyterLab, it probably means that
-  the widgets Javascript library is either not installed or not enabled. See
+  the widgets JavaScript library is either not installed or not enabled. See
   the <a href=https://ipywidgets.readthedocs.io/en/latest/user_install.html>Jupyter Widgets Documentation</a>
   for setup instructions.
 </p>
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -757,10 +757,11 @@ class Widget(LoggingHasTraits):
 _FALLBACK_HTML_TEMPLATE = """\
 <p>Failed to display Jupyter Widget of type <code>{widget_type}</code>.</p>
 <p>
-  If you're reading this message in Jupyter Notebook or JupyterLab, it probably means that
-  the widgets JavaScript library is either not installed or not enabled. See
-  the <a href=https://ipywidgets.readthedocs.io/en/latest/user_install.html>Jupyter Widgets Documentation</a>
-  for setup instructions.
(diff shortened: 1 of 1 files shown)


ssanderson:
cc @jasongrout

pbugnion:
Nice! That's much clearer.

jasongrout:
+1. @pbugnion, can you review and merge if it looks good to you?

ipywidgets/widgets/widget.py
```diff
@@ -748,3 +749,22 @@ def _gen_repr_from_keys(self, keys):
...
+
+    def _fallback_html(self):
+        return _FALLBACK_HTML_TEMPLATE.format(widget_type=type(self).__name__)
+
+
+_FALLBACK_HTML_TEMPLATE = """\
+<p>Failed to display Jupyter Widget of type <code>{widget_type}</code>.</p>
+<p>
+  If you're reading this message in Jupyter Notebook or JupyterLab, it probably means that
+  the widgets Javascript library is either not installed or not enabled. See
```
pbugnion: Should this be JavaScript, rather than Javascript -- that's how [Wikipedia](https://en.wikipedia.org/wiki/JavaScript) spells it.

ipywidgets/widgets/widget.py
```diff
@@ -748,3 +749,22 @@ def _gen_repr_from_keys(self, keys):
...
+  the <a href=http://ipywidgets.readthedocs.io/en/latest/user_install.html>Jupyter Widgets Documentation</a>
```
pbugnion: We can use https rather than http for the protocol of the URL.

pbugnion:
If you reopen a notebook that contains lots of widgets, or some widgets that take time to render, this message will be displayed until the widgets have finished rendering. The user might think their installation is broken, whereas they actually just need to wait.

Maybe we can change the copy to include this? e.g.

If you're reading this message in Jupyter Notebook or JupyterLab, your widget is either still loading, or the widgets Javascript library is either not installed or not enabled.

ssanderson:
> If you reopen a notebook that contains lots of widgets, or contains some widgets that take a long time to render, this message will be displayed until the widgets have finished rendering. The user might think their installation is broken, whereas they actually just need to wait.

Interesting.  Is there any way to distinguish between these cases on the frontend side and show a different representation (e.g. a loading spinner) when the javascript is still loading?  Ideally I'd want to avoid giving the user too many different possible cases to think about if we can.

ssanderson:
@pbugnion updated with your feedback:

image

I still think it'd be nice to try to distinguish the cases mentioned in https://github.com/jupyter-widgets/ipywidgets/pull/1674#issuecomment-325406977, but that's probably a larger piece of work and shouldn't block this improvement IMO.

pbugnion:
> Is there any way to distinguish between these cases on the frontend side and show a different representation (e.g. a loading spinner) when the javascript is still loading? Ideally I'd want to avoid giving the user too many different possible cases to think about if we can.

I'll open a new issue to discuss showing a loading icon while the widget manager is rebuilding its state after loading.

> that's probably a larger piece of work and shouldn't block this improvement IMO

I definitely agree.

PR #1675, opened 2017-08-26
Fix minor issue with repr strings

vidartf:


maartenbreddels:
LGTM, got bitten by this bug actually :(

PR #1676, opened 2017-08-26
Add filename and URL support for `Image`

### LINKED ISSUE #435: `Image` does not accept URLs, only data
Hey,

I think I found a bug with `Image`.

The ipywidget version of `Image` does not display an image, i.e., you get the broken link image symbol.

``` python
from ipywidgets import Image
Image(url='http://history.nasa.gov/ap11ann/kippsphotos/5903.jpg')
```

Whereas the IPython `Image` version does work.

``` python
from IPython.display import Image
Image(url='http://history.nasa.gov/ap11ann/kippsphotos/5903.jpg')
```

I also have the same problem when showing local images `Image(filename="image.jpg")`.

I'm using the following version

``` python
>>> import ipywidgets
>>> ipywidgets.__version__
'4.1.1'
```

pganssle:
Fixes #435.

This adds two convenience methods for pulling images directly from filenames (one alternate constructor and one "setter"-style method). It also adds support for the "url" format (per @SylvainCorlay's advice), which interprets the `value` as a URL.

I would also like to automatically detect when `value` has been set to a `six.text_type` (`str`/`unicode`) and automatically set `format` to `'url'`, but apparently validators fire *after* the type checking. Is there a hook in traitlets that fires before type checking?

I added some very basic tests, but I was not entirely sure how to mock the front-end and capture the messages - if someone can point me in the right direction I can improve these tests. I did *not* add any tests on the JS side - any advice on how to write tests for the front end would be appreciated.


pganssle:
It seems the test failures are because I was assuming that `ipywidgets/widgets/tests/data` would be available at test time, but it seems that ipywidgets packages the tests directory as a submodule under `widgets`, and the data directory doesn't come along with it, because it doesn't have a `__init__.py` in it.

I was under the impression that generally a tests directory is not included as part of the package (part of the source distribution, but not available from the library), is there a specific reason it was done this way?

For the moment I guess I can add the image as part of the package_data, but it might be worth fooling around with the way tests are handled so that the actual installed package can be slimmer?

maartenbreddels:
I actually forgot during the sprint that I've done sth 'similar' with the `VideoStream` in [ipywebrtc](https://github.com/maartenbreddels/ipywebrtc/blob/9f0e79d2a7e1f00d78f492f44dcd59f425da6628/ipywebrtc/webrtc.py#L26). Note that having a VideoStream doesn't mean we should not have a Video widget, but I'll touch on that in a different issue/PR.

(cc @jasongrout )
Maybe it's useful to have a similar API or Image, VideoStream (and if we want a Video widget), if we VideoStram to go upstream. I believe you two were also discussed downloading the image. Something like this is useful to have to the VideoStream as well, since cross origin videos are 'tainted' and not possible to say stream over webrtc, or will taint other canvas'es or streams if you start mixing them. If you download them and send them over as blobs you won't run into these issues.
would an API like this make sense:
```
Image.from_file(...)
Video(Stream).from_file(...)

Image.from_url(...)
Video(Stream).from_url(...)
```
Although the `from_url` may not make it clear it will download it.

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +33,40 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
     width = CUnicode(help="Width of the image in pixels.").tag(sync=True)
     height = CUnicode(help="Height of the image in pixels.").tag(sync=True)
     value = Bytes(help="The image data as a byte string.").tag(sync=True)
+
+    @classmethod
+    def from_filename(cls, filename, **kwargs):
```
maartenbreddels: I'd say `from_file`

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +33,40 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+        """
+        Create an `Image` from a local filename.
```
maartenbreddels: `... from a local file.` ?

ipywidgets/widgets/widget_image.py
pganssle: That's a reasonable change. The main reason I didn't go with `from_file` was that it seemed like `from_file` would take a stream buffer object (like an open `file`), rather than a `filename`.

I suppose I could change it so that it's `from_file` and try to infer whether it's already an open file or a file path.

maartenbreddels: You're right, that will be expected, and a nice feature :)

packages/controls/src/widget_image.ts
```diff
@@ -44,11 +44,21 @@ class ImageView extends DOMWidgetView {
...
+            url = URL.createObjectURL(blob);
+        } else {
+            url = String.fromCharCode.apply(null, new Uint8Array(value.buffer));
+        }
+
+        // Clean up the old objectURL
         var oldurl = this.el.src;
         this.el.src = url;
-        if (oldurl) {
+        if (oldurl && typeof oldurl !== 'string') {
```
maartenbreddels: quite tricky, but createObjectURL does create a string object. Maybe sth like:
```javascript
if (oldurl && oldurl.startsWith('blob:')) {
 ...
}
```

pganssle: Hm... If it's just a string does it still need to be freed? Or is it just cast to a string when accessed through `this.el.src`?

maartenbreddels: It is a string of the form, `'blob:<someid>'`, see https://github.com/jupyter-widgets/ipywidgets/pull/1685 how I solved it for the videostream.

pganssle: Yeah, I played around with it in the console and I see what's going on. The `src` attribute is a string that points to the location of the resource, which in this case is an in-memory blob.

I think the only downside of this `startsWith('blob:')` approach is that a user might create *another* blob that they *don't* want revoked, and pass that one's resource locator to the model, and when they change the image, we'll be freeing their blob, not *our* blob. This seems like an extreme edge case, though, and I'm sure there are simple enough workarounds that dealing with this edge case is not worth adding additional complexity.

I imagine that the "right" way to do this (which avoids this edge case) would be to set a flag on the model that indicates whether the current URL was allocated by `update`, and if so free it - rather than trying to infer this from the value at all.

maartenbreddels: I don't think users have access to the `blob:... ` strings, so that would indeed be rather unlikely.
The [docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) are a big vague, but it seems to be its reference counting behind the scenes, so I wouldn't worry about it.

maartenbreddels:
I think it's better to include the tests and (provided the data for testing is small) to include this as well. In case of issues you can always ask users to run the tests.
Another option than may work is to install with -e in [here](https://github.com/jupyter-widgets/ipywidgets/blob/master/scripts/travis_install_py.sh)

pganssle:
@maartenbreddels I'm not sure it particularly helps if the user has the tests in the *module* unless there's a specific mechanism for test discovery I don't know about. I think this is not common, but it's less work for me to do it that way so I guess I'm not complaining too much ;)

jasongrout:
@maartenbreddels, @pganssle - where are we on this?

jasongrout:
@pganssle - I think I resolved the conflict with master correctly, but it wouldn't hurt to look at it.

maartenbreddels:
As mentioned in #1685 it would be nice to have the API for Image and VideoStream with the static methods similar:
 * `from_url` will put the url in value 
 * `from_download` will download and put the content in value
 * `from_file` will read the file and put the content in value
Does that sound good?

jasongrout:
> Does that sound good?

Those are convenience methods? That sounds all right to me.

pganssle:
Last time I changed anything on this, I ended up spinning my wheels again trying to mess around with the tests. I guess I'll just ship the test data with the tests, but I still think it's not a good idea for the test directory itself to be importable from the package. I means you have to ship a bunch of unnecessary data with your package for something that's not useful for end-users anyway.

I'll clean this up soon, and maybe make a separate PR for review that moves the tests outside of the package.

maartenbreddels:
Why do you think it is not a good idea? I haven't made up my mind yet
though. An advantage I see is that you can ask users to run the tests if
you ship it. What is the downside?

(from mobile phone)

pganssle:
I think for one thing the abstraction is wrong - tests are not user-facing code, they are part of development and deployment. They don't serve to benefit the *user* at all, except indirectly in that it might aid the developers to diagnose or fix a problem. Some of the problems of shipping the tests are basically problems mainly because they are burdens on the user (albeit possibly minor) with no benefit for the user.

Some downsides I see:

1. To the extent that your tests are intended for use by the users, they become part of the public API, meaning changes to the tests and test organization might be a backwards-incompatible change (though probably most people aren't actually relying on the tests for anything).
2. As you add more data, the fact that the tests are there at all is an increasing burden - you may end up shipping 100 MB of images, videos, etc for your tests, when the library/app itself is only 10kb.
3. You either end up introducing unnecessary runtime dependencies on libraries you only need for tests (e.g. pytest, freezegun, etc), or you're shipping something where the tests don't even work in a standard installation out of the box *anyway*.

Most of these are not huge show-stoppers in most situations, but the benefit you get (users can run tests even from a wheel install) is not really worth it. As long as the tests are included in the source distribution, users who really want to run the tests can do so pretty easily, though I think this won't be something that is of general interest anyway, since most of the time a user has problems with something specific that your library is failing at (i.e. they already *have* a failing test case). Unit tests are more useful for saying, "is this library failing, and if so, how?" As long as the tests have already been run upstream, the only reason to run them again once you have them is to check that they work on your particular platform. That's something that is useful for people running package managers (e.g. debian, arch linux), but they're probably building from source or repo anyway.

jasongrout:
+1 for moving tests out of the distributed package - I think you've eloquently spoken to reasons for doing that. That change should be a different PR as you mention, of course.

maartenbreddels:
Thanks Paul, I've made up my mind I think. I'm currently splitting vaex up into multiple packages (but monorepo), and then it becomes a bit blurry which test belongs to which package, so it would be more natural to not include it in the source tree.

For the testing framework, my vote goes to py.test, nosetests is maintained but dead, and I'm really positive about py.test (fixtures are great, capturing stdout keeps the noise down, just running failed tests, jumping to the debugger on fail).

jasongrout:
Ping on this. We thought it would be nice to go in to the 7.2 release, which we're hoping to do by early next week.

packages/controls/src/widget_image.ts
```diff
@@ -44,11 +44,20 @@ class ImageView extends DOMWidgetView {
          * Called when the model is changed.  The model may have been
          * changed by another view or by a state update from the back-end.
          */
-        let blob = new Blob([this.model.get('value')], {type: `image/${this.model.get('format')}`});
-        let url = URL.createObjectURL(blob);
-        let oldurl = this.el.src;
+        let url;
+        let format = this.model.get('format');
+        let value = this.model.get('value');
+        if (format != 'url') {
```
jasongrout: `!==` please

packages/controls/src/widget_image.ts
```diff
@@ -44,11 +44,20 @@ class ImageView extends DOMWidgetView {
...
+            var blob = new Blob([value], {type: `image/${this.model.get('format')}`});
+            url = URL.createObjectURL(blob);
+        } else {
+            url = String.fromCharCode.apply(null, new Uint8Array(value.buffer));
```
jasongrout: Does this work for utf8 urls? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode#Getting_it_to_work_with_higher_values.

It's probably better to use the TextDecoder api (and shim for IE?): https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder

jasongrout: In fact, apparently the notebook itself uses TextDecoder to get the JSON parts of a binary message: https://github.com/jupyter/notebook/blob/179bb24fbf79d153812858126127a91431da3319/notebook/static/services/kernels/serialize.js#L20

jasongrout: Weird that it is used in the notebook, especially since it is not supported in IE 11 and in development in Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/

pganssle:
@jasongrout Sorry this has totally fallen off my radar. Depending on how early my son falls asleep, I'll either make these changes tonight, tomorrow night or this weekend.

vidartf:
> Because I don't use any sort of task management software or anything, I'm going to leave the tab open, nagging at me until I get to it

At this point, I consider open tabs as my task management system 😅

packages/controls/src/widget_image.ts
jasongrout: Looks like the notebook already includes a polyfill: https://github.com/jupyter/notebook/issues/3457

I think we should assume the TextDecoder api works on the page. The notebook has a polyfill, and lab doesn't support IE/Edge. People embedding widgets, if the widgets use binary messages, will need to put it on the page.

pganssle:
@jasongrout and @maartenbreddels OK, so I believe this is ready to go.

One thing I don't *love* about this implementation is that because ipython/traitlets#451 is unresolved, we're stuck between choosing to force `value` to be `Bytes` on input to the constructor (in which case if you have a URL that is some unicode object, you must decode it before passing it to `value`), or erroneously setting it as a `Union[Bytes,Unicode]` trait (or something like that).

I think having a dedicated `from_url` parameter that accepts `unicode` is the best compromise until traitlets has support for a preprocessor hook.

ipywidgets/widgets/widget_image.py
```diff
@@ -12,6 +12,17 @@
 from .widget import register
 from traitlets import Unicode, CUnicode, Bytes, validate
 
+def _text_type():
```
pganssle: So this is a weird one. `setup.py` does not declare a dependency on `six`, but several of `ipywidgets`' dependencies do, including `traitlets`.

We can drop this whole `_text_type()` charade by just adding `six` to the dependency list, but I wasn't sure if that was something you wanted to do in a minor version, despite the fact that it will have no consequences whatsoever. You can change that in this PR or just do it in a second PR that replaces this `_text_type()` convenience function.

ipywidgets/widgets/widget_image.py
```diff
@@ -6,13 +6,22 @@
...
-import base64
-
 from .widget_core import CoreWidget
 from .domwidget import DOMWidget
 from .valuewidget import ValueWidget
 from .widget import register
-from traitlets import Unicode, CUnicode, Bytes, observe
+from traitlets import Unicode, CUnicode, Bytes, validate
+
+def _text_type():
```
pganssle: So this is a weird one. `setup.py` does not declare a dependency on `six`, but several of `ipywidgets`' dependencies do, including `traitlets`.

We can drop this whole `_text_type()` charade by just adding `six` to the dependency list, but I wasn't sure if that was something you wanted to do in a minor version, despite the fact that it will have no consequences whatsoever. You can change that in this PR or just do it in a second PR that replaces this `_text_type()` convenience function.

pganssle:
@jasongrout I'm by no means a TypeScript expert, the build seems to be [choking on `TextDecoder`](https://travis-ci.org/jupyter-widgets/ipywidgets/jobs/359169942#L601). I basically copied the serialize example you linked to. In the Chrome console it seems to work OK. Any clue as to what's wrong?

jasongrout:
Do this to fix the ts error:

https://github.com/jupyterlab/jupyterlab-github/pull/41/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R56


See the line ` "@types/text-encoding": "^0.0.32",`

ipywidgets/widgets/widget_image.py
jasongrout: I'm thinking that the next major version of ipywidgets will be python 3 only, and it's probably not worth it at this point to add a dependency on six for 4 lines of code.

jasongrout:
ping @maartenbreddels for review

jasongrout:
This looks good to me. I'll leave it open for the European working hours for any comments from @SylvainCorlay or @maartenbreddels. If there is no objection, I'll merge.

Thanks @pganssle!

ipywidgets/widgets/widget_image.py
```diff
@@ -6,13 +6,22 @@
...
+    # six is not a direct dependency of this module
+    # This replicates six.text_type
+    import sys
+    if sys.version_info < (3, 0):
+        return unicode
+    else:
+        return str
+
+_text_type = _text_type()
```
maartenbreddels: I'd say a more pythonic way would be
```python
_text_type = str
try:
    _text_type = unicode
except:
    pass
```

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +44,66 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+            img = Image(value=url, format='url')
+
+        But both unicode and bytes arguments are allowed for ``url``.
+
+        Parameters
+        ----------
+        url: [str, bytes]
+            The location of a URL to load.
+        """
+        if isinstance(url, _text_type):
```
maartenbreddels: This would fail for a str type in Python 2, you could do (avoiding six, and what is done in widget.py):
```python
from ipython_genutils.py3compat import string_types
if isinstance(url, string_types):
 ....
```

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +44,66 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+        filename: str
+            The location of a file to read into the value from disk.
+
+        **kwargs:
+            The keyword arguments for `Image`
+
+        Returns an `Image` with the value set from the filename.
+        """
+        value = cls._load_file_value(filename)
+
```
maartenbreddels: What about guessing format:
```python
base, ext = os.path.splitext(filename)
if ext and 'format' not in kwargs:
    kwargs = dict(**kwargs, format=ext[1:])  # copy dict and remove leading .
```

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +44,66 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+        filename: str
+            The location of a file to read into value from disk.
+        """
+        value = self._load_file_value(filename)
+
+        self.value = value
+
+    @classmethod
+    def _load_file_value(cls, filename):
+        if getattr(filename, 'read', None) is not None:
```
maartenbreddels: Any reason not to use `if hasattr(filename, 'read')` ?

ipywidgets/widgets/widget_image.py
pganssle: It's not intended to match `str`, it's intended to match `unicode`. `str` in Python 2 does not need to be decoded.

ipywidgets/widgets/widget_image.py
pganssle: Yes. It is mainly because [it's dangerous in Python 2](https://hynek.me/articles/hasattr/).

ipywidgets/widgets/widget_image.py
maartenbreddels: Ok, makes sense (maybe a comment?)

ipywidgets/widgets/widget_image.py
maartenbreddels: Thanks, always learning!

ipywidgets/widgets/widget_image.py
pganssle: @maartenbreddels  Good idea, though I think the implementation needs to be a bit more complicated - it should try to map the extension to [the MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types), which is not always the same thing.

I'll fix that up.

maartenbreddels: Yes, we can choose maybe from a whitelist (all the ones starting with image/). Maybe give a warning if it's not known?

jasongrout: You could just use the python library for mapping filenames to formats: https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type

pganssle:
@maartenbreddels OK, added in the MIME type inference.

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +44,102 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+            with open(filename, 'rb') as f:
+                return f.read()
+
+    @classmethod
+    def _guess_format(cls, filename):
+        # file objects may have a .name parameter
+        name = getattr(filename, 'name', None)
+        name = name or filename
+
+        try:
```
jasongrout: Can we use the python mimetypes library? That seems cleaner...

jasongrout: https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type

pganssle: @jasongrout Good call, didn't know about that.

pganssle: Updated. I think this is *much* cleaner.

ipywidgets/widgets/widget_image.py
```diff
@@ -32,3 +44,87 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
...
+        if getattr(filename, 'read', None) is not None:
+            return filename.read()
+        else:
+            with open(filename, 'rb') as f:
+                return f.read()
+
+    @classmethod
+    def _guess_format(cls, filename):
+        # file objects may have a .name parameter
+        name = getattr(filename, 'name', None)
```
maartenbreddels: Didn't know that!

jasongrout:
@maartenbreddels - feel free to merge when this looks good to you!

maartenbreddels:
Excellent work Paul!
It's all green light, happy to merge this, but now we're all together, what about PIL Images and numpy arrays?
What's i'm doing a lot is sth like this:
```python
    # I is numpy arrays with 'intensities' of shape (height, width)
    colormap = matplotlib.cm.get_cmap(colormap)
    rgba = colormap(I, bytes=True)
    width, height = rgba.shape[1], rgba.shape[0]
    f = StringIO()
    img = PIL.Image.frombuffer("RGBA", (width, height), rgba, 'raw', 'RGBA', 0, 0)
    img.save(f, "png")
    return f.getvalue()
```
Do we want?
 * `from_pil_image(img)
 * `from_ndarray(image_array)` # takes (height, width, channels(3, 4)) of type byte or floatX type as argument
It feels a bit out of scope, but I think many people would use it.

I think we can add support for this without explicitly importing PIL/Pillow, and happy to do a separate PR for this is Paul feels like it enough :)

jasongrout:
> happy to do a separate PR for this is Paul feels like it enough :)

I think definitely separate PR - this one has evolved enough.

pganssle:
@maartenbreddels Probably best not to let the best be the enemy of the good. It's always possible to add features in later PRs or later releases.

One thing to note is that if you are going to have a lot of variant forms of roughly the same thing, for explicit dispatch I have written a [library called variants](https://github.com/python-variants/variants) that provides some syntactic sugar for adding variant methods.

Unfortunately it [does not currently work with classmethods](https://github.com/python-variants/variants/issues/1), but that's definitely coming (plus there are other ways to implement the same general concept without the variants library that *will* work with classmethods). [Here are the slides](https://pganssle.github.io/lightning-talks/python-variants/#/) for a lightning talk I gave about this, if you are interested.

maartenbreddels:
@pganssle you seemed to be on roll, didn't want to hold you back if you had energy left ;)

sruthiiyer:
Can someone tell how ro add hyperlink on image in ipywidgets

PR #1677, opened 2017-08-28
Add traits to Image widget for mouse position and clicks

mwcraig:
This adds four new traits to the image widget, all pixel positions relative to the upper left corner of the image.

+ `x_click` and `y_click`, the pixel position of the most recent mouse click.
+ `x_mouse` and `y_mouse`, the pixel position of the mouse (changes triggered by the `mousemove` event on the front end).

@ejeschke -- this moves in the direction of making a ginga widget easier. I think it might be better to add a `Canvas` widget for that (looks relatively straightforward).

Whether this is the way to go for ginga or not this seems like a nice improvement to the `Image` widget.

I'm happy to change trait names or rethink the approach.

edit: add to do list:

- [ ] Correctly handle calculation of raw pixel position when border and/or padding are non-zero. Currently pixel position is measured relationship  to upper left of the border.
- [ ] Make click an event with a custom payload.
- [ ] Catch keypresses using suggestions from https://github.com/jupyter-widgets/ipywidgets/pull/1677#issuecomment-325587748.
- [ ] Make mouse monitoring optional.
- [ ] Factor mouse interactivity out of the `Image` widget.

### DIFF
--- a/ipywidgets/widgets/widget_image.py
+++ b/ipywidgets/widgets/widget_image.py
@@ -12,7 +12,7 @@ from .widget_core import CoreWidget
 from .domwidget import DOMWidget
 from .valuewidget import ValueWidget
 from .widget import register
-from traitlets import Unicode, CUnicode, Bytes, observe
+from traitlets import Unicode, CUnicode, Bytes, observe, Float
 
 
 @register
@@ -32,3 +32,5 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
     width = CUnicode(help="Width of the image in pixels.").tag(sync=True)
     height = CUnicode(help="Height of the image in pixels.").tag(sync=True)
     value = Bytes(help="The image data as a byte string.").tag(sync=True)
+    x_click = Float(help="x position of most recent mouse click in pixels").tag(sync=True)
+    y_click = Float(help="y position of most recent mouse click in pixels").tag(sync=True)
--- a/ipywidgets/widgets/widget_image.py
+++ b/ipywidgets/widgets/widget_image.py
@@ -34,3 +34,5 @@ class Image(DOMWidget, ValueWidget, CoreWidget):
     value = Bytes(help="The image data as a byte string.").tag(sync=True)
     x_click = Float(help="x position of most recent mouse click in pixels").tag(sync=True)
     y_click = Float(help="y position of most recent mouse click in pixels").tag(sync=True)
+    x_mouse = Float(help="x position of pointer in pixels").tag(sync=True)
+    y_mouse = Float(help="y position of pointer in pixels").tag(sync=True)
--- a/packages/controls/src/widget_image.ts
+++ b/packages/controls/src/widget_image.ts
@@ -23,6 +23,8 @@ class ImageModel extends CoreDOMWidgetModel {
             value: new Uint8Array(0),
             x_click: 0,
             y_click: 0,
+            x_mouse: 0,
+            y_mouse: 0
         });
     }
 }
@@ -80,6 +82,7 @@ class ImageView extends DOMWidgetView {
         return {
             // Dictionary of events and their handlers.
             'click': '_handle_click',
+            'mousemove': '_handle_mousemove'
         }
     }
 
@@ -105,6 +108,12 @@ class ImageView extends DOMWidgetView {
         this.touch();
     }
 
+    _handle_mousemove(event) {
+        var mouse_position = this._get_position(event);
+        this.model.set('x_mouse', mouse_position.x, {updated_view: this});
+        this.model.set('y_mouse', mouse_position.y, {updated_view: this});
+        this.touch();
+    }
     /**
      * The default tag name.
      *
--- a/packages/controls/src/widget_image.ts
+++ b/packages/controls/src/widget_image.ts
@@ -20,7 +20,9 @@ class ImageModel extends CoreDOMWidgetModel {
             format: 'png',
             width: '',
             height: '',
-            value: new Uint8Array(0)
+            value: new Uint8Array(0),
+            x_click: 0,
+            y_click: 0,
         });
     }
 }
@@ -72,6 +74,35 @@ class ImageView extends DOMWidgetView {
             URL.revokeObjectURL(this.el.src);
         }
         super.remove()
+    }
+
+    events(): {[e: string]: string} {
+        return {
+            // Dictionary of events and their handlers.
+            'click': '_handle_click',
+        }
+    }
+
+    _get_position(event) {
+        var bounding_rect = this.el.getBoundingClientRect();
+        var y_offset = bounding_rect.top;
+        var x_offset = bounding_rect.left;
+        return {
+            'x': event.clientX - x_offset,
+            'y': event.clientY - y_offset
+        }
+    }
+    /**
+     * Handles and validates user input.
+     *
+     * Calling model.set will trigger all of the other views of the
+     * model to update.
+     */
+    _handle_click(event) {
+        var mouse_position = this._get_position(event);
+        this.model.set('x_click', mouse_position.x, {updated_view: this});
+        this.model.set('y_click', mouse_position.y, {updated_view: this});
+        this.touch();
     }
 
     /**


mwcraig:
An animation of this in action:

ipw-image-clicks

pbugnion:
This looks great!

For the click behaviour, is the expectation that the user would observe the `x_click` and `y_click` traitlets? This wouldn't work if, for instance, the user clicks twice in the same place.

Another strategy would be to emulate the behaviour of the button widget. In the frontend, we dispatch a custom message in response to click events (see e.g. [here](https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_button.ts#L122)). We could attach the coordinates of the click to the payload. On the Python side, we just register callbacks on click events and call these callbacks when we receive the custom message from the frontend. Intuitively, this seems a bit cleaner, unless I'm missing something?

maartenbreddels:
Nice idea! What do you think about making this into a separate widget, so we can reuse this for canvas/webgl, video or any other DOMWidget? It may be a bit tricky with multiple views, it needs to attach event handers to all, but it would be more flexible I'd think.

ejeschke:
> this moves in the direction of making a ginga widget easier. I think it might be better to add a Canvas widget for that (looks relatively straightforward).

@mwcraig, nice work!  This looks great and provides a couple of the most important event callbacks (if we could get keystrokes as well this would probably handle 95% of the needs!).  I agree with @maartenbreddels that this would work best as a new kind of widget, possibly just subclassed from `Image`.  Is that what you mean by a canvas widget ?

My suggestion for the ginga integration would be to follow the example for the [Tk back end](https://github.com/ejeschke/ginga/blob/master/ginga/tkw/ImageViewTk.py), which is similar in that it inherits from the PIL rendering base and also uses a `set_widget()` type method for connecting it to the image widget and setting up callbacks for the available events.  This approach would allow the most flexibility for the user to build up all the GUI with jupyter widgets and then just connect the ginga rendering (and now event mapping!) to the new widget with a simple `set_widget()` call at the end.

What this would mean is that we would add a proper `ImageViewJpy` *class* (to the module of the same name) and then an `ImageViewEvent` subclass to that.  You can see the progression in the Tk example.

jasongrout:
I think this looks great. I think tracking mouse movement should default to off, given how much information that's going to send over the wire. If you want to generalize it to other widgets, that's fine too, but I'm fine merging this for Image if tracking mouse movements is optional and off by default.

mwcraig:
@pbugnion -- I can add a click event easily enough, thanks for the suggestion.

mwcraig:
@maartenbreddels -- do you mean factoring out the mouse location stuff so it can be used in other widgets, or making a widget separate from the current `Image` iwdget? I'm happy to do either as long as I can eventually click on images 😀

mwcraig:
@jasongrout -- can you point me to an example of turning event monitoring on and off? Also, I've only checked that this works in Chrome...I would not be at all surprised if it turns out that it doesn't work in some browsers.

mwcraig:
@ejeschke -- I *think* keypresses won't be too bad, will give it a try.

By "canvas" widget I meant a widget represented as an HTML Canvas element. There are a couple of out-of-date partial implementations turned up by google. 

That is definitely not a replacement for the Image widget, but seems like the natural way to go if one wants to draw shapes on the image.

mwcraig:
Safari and Firefox work (at least on mac)...

maartenbreddels:
I think it can be a separate widget, that can attach to a DOMWidget. There will be other widgets in the pipeline that could benefit from this. I can imagine something like this..
```
mouse = widgets.MouseListener(source=image_widget)
keyboard = widgets.KeyboardListener(source=image_widget)
```
MouseListener could have `mouse_x` and `mouse_y` traits, it could also do things like lasso selections (on a key modifier) and have a `lasso` traint (list of x, y positions). Some of this code is in ipyvolume's FigureView btw, would be good to refactor this out. The only issue with ipyvolume now would be that the root dom element of the view is now what needs to be tracked (we could do sth like `source_element = view.mouseListenenElement || view.el`).

About keyboard events, what I found the most natural (what I do in ipyvolume) is to bind global keylisteners [see ipyvolume](https://github.com/maartenbreddels/ipyvolume/blob/0ed4885d62396f4a51e7e0910d6e5edd39e41368/js/src/figure.js#L138) but track if we're [hovering over the element](https://github.com/maartenbreddels/ipyvolume/blob/0ed4885d62396f4a51e7e0910d6e5edd39e41368/js/src/figure.js#L550), and only [handle them if we are hovering over it](https://github.com/maartenbreddels/ipyvolume/blob/0ed4885d62396f4a51e7e0910d6e5edd39e41368/js/src/figure.js#L604).

mwcraig:
There are still some basic pixel calculations I need to work on, e.g. when padding is non-zero, should have some updates in a day or two.

ejeschke:
> That is definitely not a replacement for the Image widget, but seems like the natural way to go if one wants to draw shapes on the image.

That's fine, but just letting you know that if it is subclassed on the PIL (as the Tk one is, and also the currently Ginga notebook viewer) then all the drawing happens server-side and the implementation of the widget is greatly simplified, because all it has to do is provide an image widget with event callbacks for clicks, cursor movement, scrolling and keystrokes.

mwcraig:
> the implementation of the widget is greatly simplified

After several hours of working through the basics of the display size of an image in a widget (#1689) I'm all for simple on the widget side.

Does the widget need to return mouse location in image coordinates (i.e. taking into account any scaling done on the screen)? I assume yes (and is not too hard to implement)

ejeschke:
> Does the widget need to return mouse location in image coordinates (i.e. taking into account any scaling done on the screen)? I assume yes (and is not too hard to implement)

Yes.  I assume that the (x, y) coordinate system is based in the upper left, but there is easy option if it is based in the lower right.  Basically we just need to know which pixel in the image the cursor is over.  

If we want full functionality, we would need the following events:
* cursor down (with button)
* cursor motion
* cursor up (with button)
* widget enter
* widget leave
* widget focus
* key down
* key up
* scrolling (wheel or touchpad)
* pinching

But like I said, if we could get 
* cursor down
* cursor move
* cursor up
* key pressed

it would handle about 95%+ of the functionality.

mwcraig:
@maartenbreddels -- I'm struggling a bit with how to attach, e.g. `MouseListener` to a widget. I know how to write the typescript to actually handle events. Should I look at the way the `Layout` widget is implemented as an example?  Or is the idea that there would be a a front-end property of the `MouseListener` called `source` and then loop over all of the events (for move, click, double-click, etc) adding a listener to the `source` for each one?

This is my first real foray into the JS side of things and I still find it a little mystifying...

maartenbreddels:
that `events()` thingy is something that backbone gives you I think (not 100% sure). The 'normal' way would be to use [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) like what I do in [ipyvolume](
https://github.com/maartenbreddels/ipyvolume/blob/master/js/src/figure.js#L381)
You'd have to iterate over all views of the model (WidgetModel.views object), but also catch the creation of new models (can't recall if that is possible)). Then for each view you add mouse listeners.

mwcraig:
This is what I *think* I know how to do (based very much on the way layouts are handled):

+ Add a `MouseListener` widget with options for choosing which events (e.g. click, dblclick, mousemove) that that listener instance will pay attention to.
+ Add a `listener` or `listeners` attribute to `DOMWidgetModel` and the corresponding view, which adds the listener(s) to the widget (e.g. `Image`).

so that something like this is possible:

```python
clicks_only = MouseListener(click=True, dbclick=True, remaining_events=False)
image_widget.listeners = [clicks_only]
```

and then your image widget is clickable.

If I turn that around, to try to make this happen:

```python
clicks_only.source = image_widget
```

then I see two issues:

1. I don't know how make the JS side do anything with MouseListener because there is no view of it. I could force a view by creating the view in the model but that seems....wrong.
2. It seems like the python object ought to know what things it is listening to, so even with this second approach a property will need to be added to the widget (even if it is a hidden property) that keeps track of the listeners.

maartenbreddels:
I prefer that widgets don't know about listeners.

> 1. I don't know how make the JS side do anything with MouseListener because there is no view of it. I could force a view by creating the view in the model but that seems....wrong.

No, a view shouldn't be created. You can access a model's views (`model.views`), the only issue now is that there is no event being generated when a new view is created.
Lets assume for the moment the views have all been created, then you can do sth like this: 
```javascript
Promise.all(_.values(model.views)).then((views) => {
  _.each(views, (view) => {
    view.el.addEventListener('click', () => {
      console.log('click')
    })
  })
})
```

> 2. It seems like the python object ought to know what things it is listening to, so even with this second approach a property will need to be added to the widget (even if it is a hidden property) that keeps track of the listeners.

Why does the Python side need to know?

mwcraig:
> No, a view shouldn't be created. You can access a model's views (model.views), the only issue now is that there is no event being generated when a new view is created.

Let me try to reframe the question I had...the whole JS side of this is new enough to me I think I'm still missing something basic (`Promise` was a word I had to lookup yesterday 😮 ).

Imagine I create a `MouseListener` widget on the python side (presumably subclassing from `CoreWidget` or `Widget`, not `DOMWidget` because the listener isn't part of the DOM). 

As I understand it, that means creating a model on the Javascript side that essentially just holds properties, so in this case a list of events that listener can listen to and a source widget (or list of widgets).

In the rest of the widgets, changes on the front-end are triggered when something in the model changes; views are listening for changes to the model and update something in response, so my first thought was, imitating the `Layout` widgets, to have a view that isn't rendered but acts on another view (or group of views).

I assume I could just, on initialization of the front-end model, add the listeners from the model itself.

The other thing I don't quite understand is how to pass information about the widget to which the listener is being attached to the JS side. In other words, if I do `mouse = widgets.MouseListener(source=image_widget)` on the python side, how does `image_widget` show up on the JS side?

The upshot of this might be that I'm not ready to do the more general case yet....though it would be easy enough to factor out the stuff for generating the proper relative coordinates (and the additional calculations needed for  images).

mwcraig:
> Why does the Python side need to know?

Partly because I want to deal with as much as possible with the events on the python side, partly because I'm still thinking about this primarily in the context of an `Image` widget. So I think about the events generated on the front by click as showing as events (or changes in the properties of) a widget like the `Image` widget.

I gather the alternative would be to look for events (python side) coming from the `MouseListener`, which would include in the event message which (python) widget (`Image` or whatever) the mouse event occurred in.

mwcraig:
Some progress to report since last week...I now understand enough about backbone and the front end to understand how to implement something like what @maartenbreddels was talking about, i.e. a `MouseListener` widget that on the js side attaches listeners to all of the views 

With that I can (on the js side) attach a click listener to any widget and generate  message in the console when a view of that widget is clicked on.

I *thought* it would be easy to send a custom event back to the python side of `MouseListener` as a click event (sort of like button does).

The problem I'm running into is that the message isn't being sent to python....any suggestions? I do know the code reaches `_handle_click` and steps into `send`...once I have that I'll work up a couple possible implementations for feedback.

```javascript
export
class MouseListenerModel extends WidgetModel {
    static serializers = {
        ...WidgetModel.serializers,
        source: {deserialize: unpack_models},
    }
    defaults() {
        return _.extend(super.defaults(), {
            _model_name: 'MouseListenerModel',
            _view_name: 'MouseListenerView',
            source: null
            });
    }
    initialize(attributes, options: {model_id: string, comm?: any, widget_manager: any}) {
        super.initialize(attributes, options);
        this.on('change:source', this.attach_listeners, this)
    }

    attach_listeners() {
        console.log("Attaching listeners")
        var current_source = this.get('source')
        _.each(current_source.views, (view_promise) => {
            Promise.resolve(view_promise).then((view: DOMWidgetView) => {
                view.el.addEventListener("click", dom_click.bind(this, view))
            })
        })
    }

    _handle_click(event) {
        console.log('Sending to python!')
        this.send({event: 'click'}, {})
    }
}
```

jasongrout:
Are you sure the message isn't being sent? There's no message on the websocket channel (you can see the websocket messages in the Networks tab of the Chrome debugger. Filter for "WS" at the top, click on the websocket connection in the left, and you'll see the "frames" (i.e., websocket messages) on the right.

mwcraig:
You are right, a message is being sent over the websocket, but the python side doesn't respond, so the code below generates nothing in the notebook.

```python
@register
class MouseListener(CoreWidget):
    _model_name = Unicode('MouseListenerModel').tag(sync=True)
    source = InstanceDict(DOMWidget).tag(sync=True, **widget_serialization)

    def __init__(self, **kwargs):
        super(MouseListener, self).__init__(**kwargs)
        self._click_handlers = CallbackDispatcher()
        self.on_msg(self._handle_mouse_msg)

    def _handle_mouse_msg(self, foo, content, buffers):
        raise RuntimeError
        print('foo: ', foo)
        print('content: ', content)
        print('buffers: ', buffers)
```

mwcraig:
I can get the same kind of behavior for a button if I modify https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/widget.ts#L143 to replace the `callbacks` argument with `null` or `{}`, which is what I'm effectively doing with the `MouseListenerModel`.

By "same behavior" I mean doing this produces nothing on the python side when the button is clicked:

```python
def print_click(change):
    import pdb; pdb.set_trace()
    print(change)
    raise RuntimeError
    assert 0
b = Button(description='Hello')
b.on_click(print_click)
b
```

mwcraig:
Should also mention that `MouseListenerModel` does not have an associated view...

maartenbreddels:
Look at the websocket frames, but the frame after the custom msg is send (or a few after that), you may see an exception being caught and send to the frontend, but not appearing anywhere (I remember having sth like that).

mwcraig:
Thanks for the help -- turns out the issue, fundamentally, was that I didn't really understand how notebooks work. The click handlers I was trying were in fact being executed in python and the results sent back to the server side over the websocket, but they weren't displayed in the notebook because the MouseHandler has no view.

ejeschke:
> The click handlers I was trying were in fact being executed in python and the results sent back to the server side over the websocket, but they weren't displayed in the notebook because the MouseHandler has no view.

@mwcraig, I don't know how generalized you want to make these additions, but at least for the purpose of adding event handling to the ginga jupyter implementation this is sufficient, because the python side will generate new image frames in response to the events and push these up via the image trait, which is already there.

mwcraig:
@ejeschke -- I *think* I'm at the point now that I understand enough about how events work in the browser, what events are available, and how to get information about those events back and forth to python that I can put together something fairly general, in two sense:

+ The full list of events you [listed above] should be doable. In principle, much of the javascript code is the same for attaching listeners for each type of event,  and the event handler on the javascript side is pretty simple, just passing a message to the python side.
+ The mouse/keyboard event handlers should be attachable to any kind of widget (not just `Image`). I'm mostly interested in making `Image` work, but the code I'd add to make just image work would be virtually the same as the code needed to make the same handling work on any widget. The details of attaching the listeners is different somewhat if you are writing them for a single widget type (backbone handles the setup and teardown of listeners in that case) vs for the general case, but not really much easier or harder.

`Image` widgets need a little extra magic to get a useful (i.e. location in the original array) x,y for a mouse event because that translation from click x,y to array x,y depends on the dimensions (and padding, etc) of which *view* of the image you click and on the dimensions of the array.

A `Canvas` widget (which I'm not proposing to write at this point) allows for much more complicated transformations of the image on the javascript side (rotation and scaling), which would add another layer of complexity to getting the array x,y out.

My ETA for a draft of the implementation of a couple of event types (say click and dblclick) is a couple of days, planning to work up what it would look like in a couple arrangements:

+ As events added only to the `Image` widget.
+ As a separate flexible `MouseListener` widget that can be hooked up to a source widget (Image or Label or whatever) as suggested by @maartenbreddels. 
+ Adding a `listeners` property to the core DOMWidget (i.e. to all widgets) to which one can attach a `MouseListener`.

mwcraig:
I'm closing this in favor of #1727, which is much more general.

PR #1679, opened 2017-08-28
Add instructions on setting `_{model,view}_module_version` client-side in the migration guide

pbugnion:
If you don't include `_model_module_version` and `_view_module_version` in the defaults in custom widget models, embedding with `embed_minimal_html` does not work. `embed_minimal_html` gets the state without defaults, so unless the versions are explicitly specified client-side, they will not be available in the embedded widget state.


jasongrout:
Thanks!

PR #1685, opened 2017-08-29
[WIP] get mediastream widgets upstream from ipywebrtc

maartenbreddels:
This is a polished version of what I had in ipywebrtc, this is the start, I think there are still a few widgets that need to be added, but this is really the core.

The only issue I see now is that VideoStream's value trait needs allow_none=True, otherwise I get:
```bash
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/maartenbreddels/src/ipywidgets/ipywidgets/widgets/tests/test_mediastream.py", line 21, in test_video_stream_from_file
    v = VideoStream.from_file(LOGO_MP4)
  File "/Users/maartenbreddels/src/ipywidgets/ipywidgets/widgets/widget_mediastream.py", line 61, in from_file
    return cls(value=f.read(), **kwargs)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 958, in __new__
    inst.setup_instance(*args, **kwargs)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 986, in setup_instance
    super(HasTraits, self).setup_instance(*args, **kwargs)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 977, in setup_instance
    value.instance_init(self)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 522, in instance_init
    v = self._validate(obj, self.default_value)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 591, in _validate
    value = self.validate(obj, value)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 2026, in validate
    self.error(obj, value)
  File "/Users/maartenbreddels/anaconda3/lib/python3.5/site-packages/traitlets/traitlets.py", line 625, in error
    raise TraitError(e)
traitlets.traitlets.TraitError: The 'value' trait of a VideoStream instance must be a bytes object, but a value of None <class 'NoneType'> was specified.
```
the VideoStream.from_file/url/download should agree fully with what is done here:
https://github.com/jupyter-widgets/ipywidgets/pull/1676
I think we can factor these classmethods out in some util base class.

Maybe it seems a bit strange to have a VideoStream and an Image widget, but I think what is needed as well is a Video and ImageStream widget. Not just for symmetry reasons, but because they are useful.
Video should be different from VideoStream in that it will not show a mediastream, but a real source/video file, and thus can have meaningful controls, play/pause etc.
I think we also could use an image stream, which would draw its image into a canvas, which then can be used to stream to anything else (texture, webrtc, name it..).

The `webrtc-adapter` js package is a polyfill for `getUserMedia`, `.captureStream` and `.srcObject`.


jasongrout:
ping @maartenbreddels - I'm going through the open PRs. Do you have interest in pushing this through?

maartenbreddels:
Yes, needs some more thinking though...

maartenbreddels:
Discussed this with @vidartf and @SylvainCorlay, and this is gonna stay as it is now, although I still not like it as it is.

My issue with this is (after renaming VideoStream to Video) that `Video` inherits from `MediaStream`, but in the browser API, a `VideoHTMLElement` is not a `MediaStream`, but has a captureStream method that would result in a MediaStream. 
To model this (if you want to stay close to the browser API) you could create a `MediaStream` widget, and to signal that a widget has this interface, you could have a `HasMediaStream` interface on both the kernel and frontend side. 
We couldn't agree on the details, and I don't think @vidartf saw the point of it, so we decided to keep it as it is for now, and let it evolve in the ipywebrtc repo.

So I think we can close this for now, unless there are strong reasons to merge this.

jasongrout:
> So I think we can close this for now, unless there are strong reasons to merge this.

Sounds good.

PR #1686, opened 2017-08-29
allow for no crossorigin tag

maartenbreddels:
Some browers don't like this when loading from local disk.

### DIFF
--- a/ipywidgets/embed.py
+++ b/ipywidgets/embed.py
@@ -237,7 +237,7 @@ def embed_snippet(views,
 
     load = load_requirejs_template if requirejs else load_template
 
-    use_cors = 'crossorigin="anonymous"' if cors else ' '
+    use_cors = ' crossorigin="anonymous"' if cors else ''
     values = {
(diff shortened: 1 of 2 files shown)


pbugnion:
I think the spaces in the interpolated template aren't quite right:

```python
import ipywidgets.embed
cors = True
use_cors = 'crossorigin="anonymous"' if cors else ' '
ipywidgets.embed.load_template.format(use_cors=use_cors, embed_url=ipywidgets.embed.DEFAULT_EMBED_SCRIPT_URL)

# => '<script src="https://unpkg.com/@jupyter-widgets/html-manager@^0.10.0/dist/embed.js"crossorigin="anonymous"></script>'
```

The `crossorigin` attribute is squashed up against `src`. I think that your definition of `use_cors` should be `' crossorigin="anonymous"' if cors else ''` (or the template should include a space).

maartenbreddels:
Good catch!

jasongrout:
@pbugnion - if this looks good to you, can you merge?

pbugnion:
Looks great!

PR #1687, opened 2017-08-29
Close all widgets at once

maartenbreddels:
Convenience function when you have lots of widgets open that draw resources and want to close them all.
```python
widgets.Widget.close_all()
```


jasongrout:
LGTM

PR #1696, opened 2017-09-02
Box widget docstrings and unittests

pbugnion:
This PR adds docstrings and unittests (Python-side) for the box widget.

The docstrings use the decorator pattern copied from ipyvolume. Since we're repeating this in a few places, it may be useful to factor it out into a separate source file (docutils?), but I think that's out of scope of this PR.

This also adds very basic unit tests for the Python side of the box widget. The unittests are constructed such that they verify the state given to the client is what the client expects.


jasongrout:
@pbugnion - is this ready?

pbugnion:
Yes! It makes no functional changes, but should probably still get reviewed.

Issue #1691, opened 2017-09-01
ToggleButtons description width doesn't change with `style` kwarg


hinnefe2:
Following the docs [here](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Description) I'm trying to change the width of the description of a `ToggleButtons` by passing in a `style` kwarg. This works for an `IntSlider` as in the docs, but not for a `ToggleButtons`.

If someone can point me towards the relevant parts of `IntSlider` and `ToggleButtons` I'm happy to tag a stab at fixing this myself.

MWE:

```
from ipywidgets import IntSlider, ToggleButtons
from IPython.display import display

style = {'description_width': 'initial'}

# description reads 'A too long ...'
display(IntSlider(description='A too long description'))

# description reads 'A too long description'
display(IntSlider(description='A too long description', style=style))  

 # description reads 'A too long ...'
display(ToggleButtons(description='A too long description', options=['a','b'], style=style)) 
```

Versions:
```
{'commit_hash': 'd86648c5d',
 'commit_source': 'installation',
 'default_encoding': 'UTF-8',
 'ipython_path': '/Users/hhinnefeld/miniconda3/lib/python3.5/site-packages/IPython',
 'ipython_version': '6.1.0',
 'os_name': 'posix',
 'platform': 'Darwin-16.6.0-x86_64-i386-64bit',
 'sys_executable': '/Users/hhinnefeld/miniconda3/bin/python',
 'sys_platform': 'darwin',
 'sys_version': '3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, '
                '17:52:12) \n'
                '[GCC 4.2.1 Compatible Apple LLVM 4.2 (clang-425.0.28)]'}

ipykernel==4.6.1
ipython==6.1.0
ipython-genutils==0.2.0
ipywidgets==7.0.0
jupyter-client==5.1.0
jupyter-console==5.1.0
jupyter-core==4.3.0

```

jasongrout:
[...]

PR #1714, opened 2017-09-13
WIP Make ToggleButtons respect passed `style` kwarg

hinnefe2:
This PR is a (work in progress) attempt to solve #1691.

Things that ~~I've tried so far (which haven't worked)~~ this includes:

- [x] make `ToggleButtonStyle` inherit from `DescriptionStyle` in both the `.py` and `.ts` definitions.
- [x] modify the `styleProperties` in the `.ts` definition.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -50,6 +50,7 @@ Gabriel Becker <gmbecker@ucdavis.edu> gmbecker <gmbecker@ucdavis.edu>
 Gael Varoquaux <gael.varoquaux@normalesup.org> gael.varoquaux <>
 Gael Varoquaux <gael.varoquaux@normalesup.org> gvaroquaux <gvaroquaux@gvaroquaux-desktop>
 Gael Varoquaux <gael.varoquaux@normalesup.org> Gael Varoquaux <>
+Henry Hinnefeld <henry.hinnefeld@gmail.com> hinnefe2 <hhinnefeld@civisanalytics.com>
 Ingolf Becker <ingolf.becker@googlemail.com> watercrossing <ingolf.becker@googlemail.com>
 Jake Vanderplas <jakevdp@gmail.com> Jake Vanderplas <vanderplas@astro.washington.edu>
 Jakob Gager <jakob.gager@gmail.com> jakobgager <jakob.gager@gmail.com>
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -13,7 +13,7 @@ except ImportError:  #python3.x
     izip = zip
 from itertools import chain
 
-from .widget_description import DescriptionWidget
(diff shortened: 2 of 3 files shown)


jasongrout:
Ah, you'll also need to modify the styleProperties like the int slider: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_int.ts#L59

jasongrout:
To explain more: the error is here in the Travis log (which you can find by clicking on the "Details" link above, or on the red X:

```
lerna ERR! execute src/widget_selection.ts(370,7): error TS2417: Class static side 'typeof ToggleButtonsStyleModel' incorrectly extends base class static side 'typeof DescriptionStyleModel'.
lerna ERR! execute   Types of property 'styleProperties' are incompatible.
lerna ERR! execute     Type '{ button_width: { selector: string; attribute: string; default: any; }; }' is not assignable to type '{ description_width: { selector: string; attribute: string; default: any; }; }'.
lerna ERR! execute       Property 'description_width' is missing in type '{ button_width: { selector: string; attribute: string; default: any; }; }'.
```
The last line indicates that the styleProperties does not properly include a description_width value, which leads to the above suggestion to augment the styleProperties with the DescriptionStyleModel styleProperties.

Isn't typescript great? In normal javascript, we wouldn't have found out about this omission until we hit a probably hard-to-debug bug where the style properties wouldn't actually update. In this case, the typings force us to fix the bug before things even compile.

jasongrout:
@hinnefe2 - thanks again for this PR. I should make a release tonight (i.e., in the next 10 hours or so) - please let me know if you have time to look at the above comment, or alternatively I can look into making the change.

jasongrout:
Also, one more thing: the name/address on the commit is: "hinnefe2 <hhinnefeld@civisanalytics.com>". If you'd like your real name to show up in our changelog, can you either: 

(a) Tell git your name: `git config --global user.name "John Doe"`, OR
(b) add an entry to our `.mailmap` file which matches git commit information to current names/emails: https://github.com/jupyter-widgets/ipywidgets/blob/master/.mailmap (add a line that looks like:

```
Correct Name <correct email> hinnefe2 <hhinnefeld@civisanalytics.com>
```

jasongrout:
(I could also make the release tomorrow morning, i.e., within 24 hours, if the extra time helps...)

hinnefe2:
Hi @jasongrout - I think this is good to go. On my local machine the description widths for toggle buttons respect the `style` argument now.  

Thanks for helping me through this - java/typescript is a bit outside my wheelhouse but you made it easy to contribute. I'll definitely be back next time I find something that could use fixing.

jasongrout:
Awesome, thanks!

jasongrout:
This looks great to me too. I checked it and it works as well. Thanks again!

PR #1717, opened 2017-09-18
Factor out `doc_subst` decorator

pbugnion:
We use the `doc_subst` pattern in several source files. Each time, the decorator was re-implemented as a private function. This PR factors out the decorator into a new source file, `docutils`. This seems like a useful refactoring before adding many new docstrings.

I'm not wedded to the name `docutils`, if anyone has a better suggestion.


maartenbreddels:
:+1: great!

PR #1723, opened 2017-09-20
IE 11 croaks on `string`. Use '' instead.

aowi:
The change in 7062add7 introduced `` which IE doesn't like, causing the extensions to not load properly.


jasongrout:
You're right! Thanks!

PR #1727, opened 2017-09-25
Add DOMListener widget

mwcraig:
This PR replaces #1677. It adds a new `DOMListener` widget that can be used to watch for DOM events on any widget. Briefly, the usage is that suggested by @maartenbreddels in #1677:

```python
from ipywidgets import Label, DOMListener

lab = Label('Click me!')

dom = DOMListener()
dom.source = lab
dom.watched_events = ['click',  'keyup']

def on_event(event):
    lab.value = event['type']

lab
```

and then clicking or releasing a key generates an event.

I've written a tutorial notebook in `docs/source/examples/Widget DOM Event` so you can kick the tires fairly easily.

Some unresolved things:

+ [ ] Decide whether callbacks also take a reference to either the DOMListener or to the widget to which they are attached (it is easy to do either or both).
+ [ ] Decide whether it is really a good idea to have the keys of the returned even depend on the type of event (e.g. event keys are different for a click than for a keydown).

**edit 2017-09-24**:
+ [ ] Keyboard events capture the keypress and do not pass it along to the notebook....yea or nay? (If this doesn't happen then pressing `x` has a nasty side effect).
+ [ ] If the `wheel` event is being watched then the scroll is not passed to the notebook. 
+ [ ] This uses the `_view_count` property (only on the javascript side) to attach listeners to views created after the listener is set up. That is currently labeled experimental so not sure about using it.

I'm sure there will be other things that come up. Comments welcome!


mwcraig:
@ejschke -- the list of supported events is at https://github.com/jupyter-widgets/ipywidgets/pull/1727/files#diff-fb98af03d9e08e2e4706ab9635870c4dR14 and is fairly easy to extend (i.e. just requires adding the name of the event to the list).

mwcraig:
Quick demo from the documentation notebook:

ipywidgets_dom_events

pbugnion:
I think this is a really nice approach for handling DOM events in a general way.

ejeschke:
@mwcraig, this looks like the ticket!  I'll kick the tires on it.

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,275 @@
...
+
+    _add_key_listener(event_type, view) {
+        // Key listeners should:
+        //     + Only fire when the mouse is over the element.
+        //     + Not propagate up to the notebook (because imagine you
+        //       press 'x' on your widget and cut the cell...not what
+        //       you probably want the user to experience)
+        //
+        // The approach here is to add a key listener on mouseenter and remove
+        // it on mouseleave.
```
ejeschke: @mwcraig, does this approach mean that we would miss a key release if the cursor moves out of the element?

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,275 @@
...
+            if (this.get('ignore_modifier_key_events') &&
+               _.contains(this.get('_modifier_keys'), event.key)) {
+                // If the key event is supposed to be ignored, then skip it.
+                return
+            }
+            this._send_dom_event(event)
+            // Need this (and useCapture in the listener) to prevent the keypress
+            // from propagating to the notebook.
+            event.stopPropagation()
+        }
```
ejeschke: I think we need a `stopPropagation()` or possibly a `preventDefault()` in the case of button 1 press and drag (if the user is listening for mousedown/mousemotion/mouseup events (otherwise the browser tries to drag the element).  Also for right mouse button mousedown/click, otherwise the browser tries to pop up a context menu.

ejeschke:
@mwcraig, see the branch [jupyter-backend-events](https://github.com/naojsoft/ginga/tree/jupyter-backend-events) for a version of ginga that uses your `DOMWidget` to provide events.  Example notebook [here](https://github.com/naojsoft/ginga/blob/jupyter-backend-events/ginga/examples/jupyter-notebook/Jupyter%20Widget%20Ideas.ipynb) (it's in the examples/jupyter-notebook folder).  (On github the widget from the saved notebook does not seem to be rendered properly--not sure why, since it should just be a VBox containing an Image and an HTML element).

ejeschke:
@jasongrout, is there a way to register for a python callback for a resize of the `Image` widget, or is that something that could/should be added to the `DOMWidget`?

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,275 @@
...
+        // Get coordinates relative to the
+        let relative_xy = _get_position(generating_view, event)
+        event['relativeX'] = relative_xy.x
+        event['relativeY'] = relative_xy.y
+        if ('_array_xy' in generating_view) {
+            console.log("Hey, nice Image!")
+            let array_coords = generating_view['_array_xy'](event)
+            event['arrayX'] = array_coords.x
+            event['arrayY'] = array_coords.y
+        }
```
ejeschke: If we expand this from just "wheel" to include "mousedown", "mousemove" and "mouseup" it solves the problem of the first mouse button events getting processed by the browser.

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,275 @@
...
+        let key_handler = (event) => {
+            // console.log('Key presses FTW!', event)
+            if (this.get('ignore_modifier_key_events') &&
+               _.contains(this.get('_modifier_keys'), event.key)) {
+                // If the key event is supposed to be ignored, then skip it.
+                return
+            }
+            this._send_dom_event(event)
+            // Need this (and useCapture in the listener) to prevent the keypress
+            // from propagating to the notebook.
```
ejeschke: If we add a `event.preventDefault()` here it prevents the quote keys, slash key, home, end , page up and page down from being processed by the browser.

mwcraig: Are these keypresses that you would prefer be passed on to the browser while over the widget?

packages/controls/src/widget_dom_events.ts
mwcraig: That is correct

packages/controls/src/widget_dom_events.ts
ejeschke: No, they are currently being processed by the widget (which is listening for them) and then also affecting the browser.  If the widget is listening for them they should not be also processed by the browser, IMHO.  For example, if you implement a scrolling function bound to the `home` key then not only does it scroll your viewer (in the widget) but also the browser will scroll the whole notebook back to the top of the notebook.

mwcraig:
@ejeschke -- I just pushed a few changes:

+ Added an option `prevent_default_action` for `DOMListener`.
+ Added several more events, including `contextmenu`.

AFAICT in chrome, a right-click generates the event `contextmenu`, so that is now a watchable event. If the `DOMListener` instance has `prevent_default_action=True` then the default action (e.g. context menu, or dragging) will not occur.

The snippet below shows how to prevent the default and disable dragging. I put the `dragstart` watcher into a separate listener because I don't really care about its events....I just want to prevent the default action. It could be included in the `im_events` listener instead.

```python
image = Image(value=value, format='png')

im_events = DOMListener()
im_events.source = image
im_events.watched_events = ['click', 'contextmenu']
im_events.prevent_default_action = True

no_drag = DOMListener(source=image, watched_events=['dragstart'], prevent_default_action = True)
```

mwcraig:
@jasongrout @maartenbreddels -- this is ready for review, I think...

packages/controls/src/widget_dom_events.ts
mwcraig: Ah, I'll try to get that fixed tomorrow. Any idea which key(s) one of those things is bound to on a mac keyboard?

ejeschke:
@mwcraig, thanks, that fixed the mouse event sharing issues!

I'm still having a little trouble with the browser-recognized keystrokes, see my response in the inline commentary.

packages/controls/src/widget_dom_events.ts
ejeschke: After your recent updates, the remaining problem keys seem to be forward slash, single quote (these two trigger the browser quick find box), and pageup/pagedown/home/end (which scroll the browser page).  I'm testing with Chrome on Linux.

Seems like if you just test in the right spot in the key handler for `prevent_default_action` = `True` and do an `event.preventDefault()` it should fix this.  I had all the keys working with a modified version of `widget_dom_events.ts` in which I invoked that in the key handler.

mwcraig:
@ejeschke -- can you please try the browser-recognized keystrokes?

ejeschke:
> can you please try the browser-recognized keystrokes?

Working perfectly!

ipywidgets/widgets/widget_dom_events.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @validate('watched_events')
+    def _validate_watched_events(self, proposal):
+        value = proposal['value']
+        bad_events = [v for v in value if v not in
+                      self._supported_mouse_events
+                      + self._supported_key_events]
+        if bad_events:
+            raise ValueError('The event(s) {} are not '
+                             'supported.'.format(bad_events))
```
maartenbreddels: I always like it when error msg'es say what the option are, for instance, in matplotlib if I don't know the marker type, i just type nonsense and i'll pick something from the error msg, beats reading documentation :)

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,276 @@
...
+    // the listener is attached. This is essentially what layerX and layerY
+    // are supposed to be (and are in chrome) but those event properties have
+    // red box warnings in the MDN documentation that they are not part of any
+    // standard and are not on any standards tracks, so get what we need here.
+    var bounding_rect = view.el.getBoundingClientRect();
+    var y_offset = bounding_rect.top;
+    var x_offset = bounding_rect.left;
+    return {
+        'x': Math.round(event.clientX - x_offset),
+        'y': Math.round(event.clientY - y_offset)
```
maartenbreddels: What is the reason for the rounding?

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,276 @@
...
+        if (current_model.name ==  "DOMWidgetModel") {
+            // We never actually listen to a bare DOMWidgetModel. However,
+            // the InstanceDict trait type initializes to a basic version
+            // of the required class...so effectively, we should treat
+            // this model as undefined.
+            return
+        }
+        if (! (typeof(current_model.get('_view_count')) === "number")) {
+            // Sorry, but we need the view count...
+            current_model.set('_view_count', 0)
```
maartenbreddels: I think that is ok, i don't see this disappearing, unless we get feedback that this idea isn't good. Although for your use case, we may want to have events in the model that are send when views are created and destroyed.

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,276 @@
...
+                case "mouse":
+                    let prevent_default = this.get('prevent_default_action')
+                    let handler = this._dom_click.bind(this, view)
+                    view.el.addEventListener(event, handler)
+                    // Keep track of the listeners we are attaching so that we can
+                    // remove them if needed.
+                    this._cache_listeners(event, view, handler)
+                    break
+                default:
+                    console.log('Not familiar with that message source')
```
maartenbreddels: console.error would be appropriate for this case (not important though)

packages/controls/src/widget_dom_events.ts
maartenbreddels: You can imagine when outside the notebook, you'd want this behaviour to change maybe (global keys).

packages/controls/src/widget_dom_events.ts
```diff
@@ -0,0 +1,276 @@
...
+        this._cache_listeners('mouseleave', view, disable_key_listen)
+    }
+
+
+    _dom_click(generating_view, event) {
+        // Get coordinates relative to the
+        let relative_xy = _get_position(generating_view, event)
+        event['relativeX'] = relative_xy.x
+        event['relativeY'] = relative_xy.y
+        if ('_array_xy' in generating_view) {
```
maartenbreddels: I find the names _array_xy/arrayX a bit unclear. A natural name would maybe be natural_x and natural_y (no pun indented 😃 ) . Also would be useful to have more coordinates, I would say x, y are wrt the element, page_x,page_y would be relative to the page (including scrolling), window_x/y would be wrt upper left corner (excluding scolling) (I'm not 100% sure that is the same semantics as in js/css/html).
Also, I'd think normalize versions (between 0 and 1) would be useful.

maartenbreddels:
For debugging (not just now, but also when people use it) It may be useful to have a custom view for these widgets. For a mouse listener the view can show the status, like which buttons are pressed, the x/y coordinate etc.

packages/controls/src/widget_dom_events.ts
mwcraig: The bounding rectangles return fractional pixels but it doesn't seem correct to pass that directly on to the user. To me it implies sub-pixel accuracy that isn't really there.

packages/controls/src/widget_dom_events.ts
mwcraig: So an option that captures the key events whether or not the mouse is over the element?

packages/controls/src/widget_dom_events.ts
mwcraig: > Also would be useful to have more coordinates, I would say x, y are wrt the element, page_x,page_y would be relative to the page (including scrolling), window_x/y would be wrt upper left corner (excluding scolling) 

I think almost all currently returned in the event (full list is https://github.com/jupyter-widgets/ipywidgets/pull/1727/files#diff-6856e80b321629aaabf83b82a9939b43R23)

>  A natural name would maybe be natural_x and natural_y (no pun indented 😃 ) .

I'm fine with a different name...I also thought about `_image_x/imageX` but `natural` works (and puns are always in-scope).

> Also, I'd think normalize versions (between 0 and 1) would be useful.

Normalized versions of the natural coordinates, I assume?

packages/controls/src/widget_dom_events.ts
ejeschke: Ah, but it might be correct if the user has scaled the widget, and a pixel in the image is taking more than one pixel in the display.  At least it would be nice to get the floating point values.

jasongrout:
@mwcraig: FYI, I merged master into this and resolved a conflict

jasongrout:
I've been looking at this and thinking about this. Taking a step back - I know we talked about having mouse clicks in the image widget. @maartenbreddels - it sounds like you have a few other usecases in mind. What are they?

I want to make sure we're not overextending ourselves in the core to the point of creating a huge maintenance burden in the future. We went down this sort of path early on with CSS, exposing the entire CSS capabilities to the user, and it created problems moving forward in the widgets architecture, so we rolled it back to a more limited set of styling properties that we have now.

I also want to make sure we're encouraging best practices. Unless we have a compelling usecase for trapping key events, I think it's probably better to leave those out for now. Taking over key handling, especially without a visual indication of focus, can be really confusing for a user. Not to mention that the key handling in browsers is still rather primitive and inconsistent.

But again, what are several usecases we really want to address here? Clicking on an image widget to get the coordinates in the image? What else?

ejeschke:
If you want to implement an application that has key bindings, it is impossible unless you can register for and handle key events.  It's really that simple.  In the case that @mwcraig and I are discussing here it is an image viewer that modifies the display based on key, mouse, scrolling, etc. events in the window.

I think the DOMWidget was a clever way to allow one to capture events in the Image widget without having to add those event hooks in the Image widget itself.

jasongrout:
> an image viewer that modifies the display based on key, mouse, scrolling, etc. events in the window.

Ah, I didn't realize (or forgot?) that you were aiming for an image viewer with much more control than just picking a point.

At some point, it makes sense to implement a full-featured widget in javascript itself. Is that what you have currently? Seems like I remember @mwcraig showing me a pretty fancy image viewer.

Are there cases besides an image viewer where we'd like to take control of the user interaction outside of the widget?

mwcraig:
> At some point, it makes sense to implement a full-featured widget in javascript itself.

FWIW, the stuff shown below  is now possible with no javascript beyond what is in this PR (at least I think there is no additional javascript). It is based on a branch of ginga that @ejeschke has been working on, so it is plugging into the framework he had already developed for other front-ends.

That said, we could get to the same functionality if the DOMListener were a separate widget package (or if the key handling part were separate).

@maartenbreddels will need to address his specific use cases but I think the argument he made early on was that the same functionality needed to make an image interactive also works to make any widget interactive and that such functionality was likely to be useful beyond the core set of widgets.

I agree that for most of the core (i.e. in `ipywidgets`) widgets mouse interaction doesn't make much sense.

image_viewer

mwcraig:
> Taking over key handling, especially without a visual indication of focus, can be really confusing for a user. 

Yeah, this hit me several times while working on the demo above 😮 , and a similar thing happens every time I use the `notebook` interface in matplotlib. Shift-Enter stopped working every time I moused over.

This could conceivably be addressed by visually indicating the focus has shifted when the mouse is over the widget (if there are key listeners attached).

The issue isn't specific to key presses, though -- when moused over the image, scrolling zooms the image instead of moving the browser content.

maartenbreddels:
What I would like to see in addition to this is to make lasso or rectangular selections possible. For ipyvolume I need a lasso to send to the Python side to do fancy selections in 3d. I do this myself now, but I think it would be great if we can solve all of this at once.

ejeschke:
> What I would like to see in addition to this is to make lasso or rectangular selections possible. For ipyvolume I need a lasso to send to the Python side to do fancy selections in 3d. I do this myself now, but I think it would be great if we can solve all of this at once.

Once you have mouse events (with down and up separate) it's pretty straightforward to generate a lasso function.  Maybe a separate module that utilizes `DOMWidget` to make a library of reusable user interactions?  DOMWidget right now delivers a straightforward minimal "register for events" functionality, which makes it very clean, IMO.

ejeschke:
> At some point, it makes sense to implement a full-featured widget in javascript itself. Is that what you have currently? Seems like I remember @mwcraig showing me a pretty fancy image viewer.

Yes, but that is a different web backend.  the advantage of using the ipywidgets `Image` widget as the canvas is that the user building an application can stay entirely in ipwidgets-land, and all the UI is just standard ipywidgets.

jasongrout:
> Yes, but that is a different web backend. the advantage of using the ipywidgets Image widget as the canvas is that the user building an application can stay entirely in ipwidgets-land, and all the UI is just standard ipywidgets.

What I'm suggesting is that it may make more sense to have a super-charged Image ipywidget with user interactions such as mousewheel, click, selection, etc.

One of my main concerns with pushing all of the browser DOM events over to Python is that it encourages people to write all of the event handling code in Python, and I think it is generally a better practice and experience to have the first level of event handling in js, and expose a more semantic interface back to python (which presumably is lower bandwidth communication and an api that has a smaller scope for easier maintenance going forward). For example, for a lasso selection, I think it's a better practice to do the mouse handling in js, and just communicate the final selection (semantically as a selection) to python.

ejeschke:
> For example, for a lasso selection, I think it's a better practice to do the mouse handling in js, and just communicate the final selection (semantically as a selection) to python.

Agree in the general case, but presumably people are using ipywidgets because they are interested in coding on the python side of things.  DOMWidget is more of a widget accessory tool and probably will be useful for people like me that don't want to have to create a whole new widget with all the javascript bits, just to get a few events that are not exposed through the regular widget interface.

ejeschke:
> don't want to have to create a whole new widget with all the javascript bits, just to get a few events that are not exposed through the regular widget interface.

Having written one javascript interface for all these events (different back end) and thinking how much time it took to figure it out across N browsers, and how much time @mwcraig probably spent figuring out the js for this, and how much the next M people will spend re-figuring out how to add those events for their custom widgets...  it just seems like having a little accessory widget that encapsulates all of that separate from all the other widgets is a really clean solution.

jasongrout:
> Agree in the general case, but presumably people are using ipywidgets because they are interested in coding on the python side of things. DOMWidget is more of a widget accessory tool and probably will be useful for people like me that don't want to have to create a whole new widget with all the javascript bits, just to get a few events that are not exposed through the regular widget interface.

Yes, I also see DOMListener as being a valuable escape hatch/prototyping tool of sorts.

I'm not debating having a DOMListener widget available somewhere (though I would caution against its overuse). What I am concerned about is adding and maintaining a broad, changing, and sometimes inconsistent (between browsers) DOM event interface as part of core widget controls.

As a way forward, how about we continue to develop and refine the DOMListener widget as a separate python package (like the other widget nbextensions), which lets us experiment and iterate about how much of the DOM event interface to expose to the user and in what way without worrying about the stronger backwards compatibility assumptions core widgets now has grown into. As we gain experience in this, we incorporate selected, widely-used parts of the event interface into widgets core.

mwcraig:
> As a way forward, how about we continue to develop and refine the DOMListener widget as a separate python package (like the other widget nbextensions), which lets us experiment and iterate about how much of the DOM event interface to expose to the user and in what way without worrying about the stronger backwards compatibility assumptions core widgets now has grown into. 

I'm fine with something like this...have a couple thoughts about the eventual direction of the Image widget too (think the current one is broken in a couple of ways) but won't be able to follow up until Friday at the earliest.

mwcraig:
fyi, this functionality now exists in a separate package:  https://github.com/mwcraig/ipyevents

jasongrout:
As such, closing this PR. Thanks again for working on this!

PR #1729, opened 2017-09-28
Use python compare in `_should_send_property`

vidartf:
Partial revert of 6445ea9c43282a081dc2da0a76e5844721685e56.

Main reason for reversion: float/int comparison.
 1. On the JS side the value of a CFloat gets set to 1.0. It is serialized as `'1'` when sending update to kernel.
 2. The python deserializer treats this as an int `1`, and sends this for validation.
 3. CFloat coerces this to a float `1.0`.
 4. `_should_send_property` detects this as a change (`'1' != '1.0'`).

This was causing issues where multiple jslinked CFloats were entering infinite loops.

This PR goes back to comparing in JSON-able form, in which `1.0 == 1`.

Also, the original reason for changing this code was based on flawed logic (no widgets are created since they are compared in JSON-able state, and not passed through deserializers). Sorry about that. The buffer separation part is unaffected by this.


jasongrout:
Huh. `JSON.stringify(1.0)` gives "1". I guess that's a relic of js only having floating point numbers (so 1 is literally the same as 1.0).

jasongrout:
@vidartf - can you write a test which ensures we cover this case?

vidartf:
@jasongrout I added some tests. I also added one test that currently fails as a discussion point:

Currently, the JS `JSON.stringify(1.0)` gives "1" as you've noted, and so does `JSON.stringify(1)`. However, if some client were to serialize all numbers as floats, i.e. `JSON.stringify(1) === '1.0'`, any `Int` traits would fail. Should we just let this be?

vidartf:
Note: I pushed another commit that simply defines this as "expected behavior", leaving behind the test as disabled for future reference. If that is the way to go, this can simply be merged.

jasongrout:
Looks good, thanks!

PR #1733, opened 2017-10-02
WIP: Remove implicit any from TypeScript files

cnishina:


cnishina:
Please review the `//TODO(cnishina)`. I am unsure of the return type for `window.require` and the `options` parameter since it is not used by the method.

cnishina:
Also in `examples/web-tmpnb/src/manager.ts`, the return type to display_view is a `Promise<HTMLElement>`; however, returning the view is of type `base.WidgetView`. To resolve these type diffs, I had to cast the view to any then to HTMLElement.

cnishina:
Also going to close this PR. I'll revisit this after lint efforts are complete

PR #1739, opened 2017-10-04
Fix dependency issues.

jasongrout:


blink1073:
We use this [script](https://github.com/jupyterlab/jupyterlab/blob/1a66a342895ea3d9d46138701e9b71f248c1b5ba/scripts/package-integrity.js) in JupyterLab to make sure all of the packages have the correct dependencies.

jasongrout:
Thanks!

blink1073:
`mathjax` should also be removed [here](https://github.com/jupyter-widgets/ipywidgets/blob/81931fecb88324e02bb361748132de2755bf29ad/packages/jupyterlab-manager/src/tsconfig.json#L7), cf https://github.com/jupyterlab/jupyterlab/issues/3050#issuecomment-334226702.

jasongrout:
Thanks. I also notice the tool doesn't handle imports that are purely typings (for example, at https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/jupyterlab-manager/src/manager.ts#L4 we just import backbone for the typings, so we depend on the @types/backbone package - the integrity script flags that we're missing backbone and the @types/backbone package is unused...

I peeked at the source of the integrity checking code, and it doesn't look it distinguishes between imports for types and runtime imports.

blink1073:
You only need `@types/` in `dependencies` if you are re-exporting an interface.  Also, you don't need to declare `types` in `tsconfig`, it will automatically use `node_modules/@types`.

blink1073:
In the case of a re-exported `@types`, you'd add it to the `IGNORE` packages in the script.

cnishina:
I just noticed that `@jupyterlab/coreutils` was not there. 👍

jasongrout:
I'm working through a number of dependency issues pointed out be Steve's script.

cnishina:
Also found that we need the `@types/base64-js`. This probably belongs in this PR. See https://github.com/jupyter-widgets/ipywidgets/pull/1733/commits/eb65e3de6870b1a38d586a198654c860c6a6a9e0#diff-1e1cb1264372c8fecbcf22d8d8f3af30

jasongrout:
I rebased on master and I think this is good now. I'll wait until it passes checks before merging.

Issue #1724, opened 2017-09-22
Convenience methods on output widget


jasongrout:
We should have convenience methods for appending stdout, stderr, and display_data messages to an output widget, so that this:

```python
out = Output()
out.outputs = out.outputs + ({'name': 'stdout', 'output_type': 'stream', 'text': 'hi\n'},)
```
becomes something like this:
```python
out.append_stdout('hi\n')
```
Likewise, also have an `out.append_stderr` and `out.append_display_data`.

jasongrout:
(see https://github.com/jupyter-widgets/ipywidgets/issues/565#issuecomment-331522316)

jasongrout:
Fixed by #1752.

Issue #1722, opened 2017-09-20
race conditions, output widgets, and matplotlib


mheilman:
I ran into some strange behavior when trying to show matplotlib plots when background threads complete.  IIUC, matplotlib maintains global state and doesn't work well with multiple threads, but I'm still seeing some strange behavior even with a lock around code that renders plots to particular output widgets: some plots never show up, even though the IPython kernel seems to be idle.

Also, without locks, the `with output_widget: ...` functionality (which is super cool btw!) doesn't work well: things like plots and printed messages end up all over the place if different threads are trying to produce output.  It'd be nice if users didn't have to put locks in their code.

It's very possible I'm missing something here, but I didn't see any answers in a quick search on GitHub and stackoverflow (a few semi-related issues are linked below).

This is pretty complicated, so here's as minimal an example as I could come up with:

```python
from concurrent.futures import ThreadPoolExecutor
import matplotlib.pyplot as plt
import matplotlib
import numpy as np
import ipywidgets
from IPython.display import display
from threading import RLock

print("Widgets: {}".format(ipywidgets.__version__))
print("MPL: {}".format(matplotlib.__version__))

# some data to plot
x = np.arange(1000) + 1
y = np.log(x)

# number of plots to make
n = 10

# make some output widgets.
widgets = [ipywidgets.Output() for _ in range(n)]
# I also noticed that if the output widgets aren't initialized with some text or plot,
# then the plots end up in arbitrary order (e.g., try commenting out the for loop here).
for w in widgets:
    with w:
        print('loading')
display(*widgets)


# a function that takes a little while to run, after which plotting will happen.
def foo():
    for _ in range(1000):
        np.random.rand()

# a lock since matplotlib isn't thread safe.
# without this lock, the plots go all over the place (to the wrong output widgets?)
lock = RLock()

# callback to render each plot when foo is done
def make_render_func(i, widget):
    def render_plot(_):
        with lock:
            widget.clear_output()
            with widget:
                fig, ax = plt.subplots()
                ax.plot(x, y, color='blue')
                ax.set_title("plot %d" % i)
                plt.show()
    return render_plot

# make a future for each plot
ex = ThreadPoolExecutor(max_workers=n)
for i in range(n):
    f = ex.submit(foo)    
    f.add_done_callback(make_render_func(i, widgets[i]))

```

Here's an example of what happens (plot 8 never shows up):

[...]

PR #1752, opened 2017-10-12
added convenience methods on output widget

mheilman:
This adds some convenience methods for adding output items to the Output widget in order to address #1724.

It also should make it easier to work around the apparent race condition discussed in #1722.

### DIFF
--- a/ipywidgets/widgets/widget_output.py
+++ b/ipywidgets/widgets/widget_output.py
@@ -12,6 +12,7 @@ from .widget_core import CoreWidget
 
 import sys
 from traitlets import Unicode, Tuple
+from IPython.core.interactiveshell import InteractiveShell
 from IPython.display import clear_output
 from IPython import get_ipython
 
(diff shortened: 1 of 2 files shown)


ipywidgets/widgets/widget_output.py
```diff
@@ -68,3 +69,36 @@ def _flush(self):
...
+    def append_display_data(self, display_object):
+        """Append a display object as an output.
+
+        Parameters
+        ----------
+        display_object : IPython.core.display.DisplayObject
+            The object to display (e.g., an instance of
+            `IPython.display.Markdown` or `IPython.display.Image`).
+        """
+        fmt = InteractiveShell.instance().display_formatter.format
```
mheilman: This is based on [this line in IPython.core.display.py](https://github.com/ipython/ipython/blob/36ab971deab9c2822a1558caaa3ba99a49deb789/IPython/core/display.py#L291).  It seemed easier to follow how the `display` function works, where a user can pass, e.g., an `Image` object rather than asking users to provide the "raw" output format that this format function provides.  If there's a better way to get this formatting functionality, let me know.

jasongrout:
This looks good to me. Thanks!

rgbkrk:
Ah great, thanks for this.

jasongrout:
Hopefully it will be released this afternoon.

PR #1755, opened 2017-10-12
set display to block for Tab widget tab bar

### LINKED ISSUE #1754: tab contents and title overlap with vertical scrolling
If one sets the layout height for a `Tab` widget, then the contents overlap with the title after scrolling down:

screen shot 2017-10-12 at 4 17 48 pm

Here's some code to reproduce this:

```python
import ipywidgets

out1 = ipywidgets.Output()
tab = ipywidgets.Tab([out1], layout=ipywidgets.Layout(height='400px'))

with out1:
    for _ in range(40):
        print("foo")

tab.set_title(0, "bar")
tab
```

Some version info:

* Python 3.6.2
* ipywidgets 7.0.1
* notebook 5.1.0
* Chrome 61
* Mac OS 10.12.6

mheilman:
I think this fixes #1754. Here's a screenshot related to the ones in #1754, but with the change proposed here:

screen shot 2017-10-12 at 5 03 49 pm

Perhaps there's a better way, but I figured I'd suggest this before going too far down the CSS rabbit hole.

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -931,6 +931,7 @@
 
 .jupyter-widgets.widget-tab .widget-tab-bar.p-TabBar {
     font: var(--jp-widgets-font-size) Helvetica, Arial, sans-serif;
+    display: block;
 }
 
 .jupyter-widgets.widget-tab .p-TabBar-tab {


jasongrout:
This seems to fix things for the cases I tried, and I think it's definitely better than what is there. Thanks!

sccolbert:
Note that this solution is only going to be okay if you *aren't* using a `BoxPanel` to layout the tab bar above the content. If you *are* doing that, then you should set a `min-height` on the tab bar, and retain the default CSS behavior.

This fix will also break down if you ever add any extra content to the `TabBar.node`.

sccolbert:
This is also likely to fail if you ever change the tab bar orientation to `vertical`

sccolbert:
If I've included something in the core Phosphor CSS, that usually means it's correct for 99% of cases, and should not be changed lightly.

jasongrout:
Thanks for the tips!

> Note that this solution is only going to be okay if you aren't using a BoxPanel to layout the tab bar above the content.

We're directly using a PanelLayout in a custom phosphor widget.

> This is also likely to fail if you ever change the tab bar orientation to vertical

This isn't supported in widgets, but thanks - noted if we ever do support that.

PR #1763, opened 2017-10-20
Fix tabs CSS

jasongrout:
We had diverged quite a bit from the PhosphorJS CSS. Fixing those things also ended up fixing #1754 in a more comprehensive way than #1755.


jasongrout:
CC @sccolbert - essentially we refresh our PhosphorJS CSS, and fix our CSS so that it actually is up to date with the code.

jasongrout:
screen shot 2017-10-19 at 5 21 40 pm

jasongrout:
CC @mheilman too

mheilman:
Cool.  Updating phosphor seems like a better solution. :-)

I checked out the branch, and it seems to work locally for me to resolve #1754.

jasongrout:
Thanks for checking!

PR #1766, opened 2017-10-20
Hide prompt area in classic notebook output widget.

### LINKED ISSUE #1765: hide out_prompt_overlay in Output widget
The Output widget shows the overlay on the left side of the output area that lets people collapse an output cell in a notebook (see screenshot below).  It doesn't seem like this is behavior that one would want when using an Output widget.  Would it be possible to hide this overlay by default or make it optional?

screen shot 2017-10-20 at 12 49 13 pm

Would it make sense to add some CSS like `.jupyter-widgets-output-area > .out_prompt_overlay {display: none}`?

jasongrout:
This works around for the bug fixed in https://github.com/jupyter/notebook/pull/2961

Thanks to @mheilman for noting the problem and suggesting a fix.

Fixes #1765

### DIFF
--- a/widgetsnbextension/src/widget_output.css
+++ b/widgetsnbextension/src/widget_output.css
@@ -7,3 +7,8 @@
 .jupyter-widgets-output-area div.output_subarea {
     max-width: 100%;
 }
+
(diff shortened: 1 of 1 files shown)


jasongrout:
Published in widgetsnbextension 3.0.6

PR #1768, opened 2017-10-21
Automatic fixes from tslint

jasongrout:
Thanks to @cnishina getting tslint working, we get some automatic style fixes for free with `tslint --fix`. I'm pushing and merging these since there are no outstanding PRs on the examples.

### DIFF
--- a/examples/web-tmpnb/src/index.ts
+++ b/examples/web-tmpnb/src/index.ts
@@ -10,12 +10,12 @@ import {
     Kernel, ServerConnection
 } from '@jupyterlab/services';
 
-document.addEventListener("DOMContentLoaded", function(event) {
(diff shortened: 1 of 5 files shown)


cnishina:
```
Package integrity failures:
{
 "examples/web3": [
 "Unused package: @jupyter-widgets/controls"
 ]
}
```

I also ran into this when I did a rebase for the first attempt at tslint.

jasongrout:
Thanks, I'm working on it.

jasongrout:
Basically, the web3 example was improperly including only the widget controls css - that should be provided by the html manager. Unwinding that chain is going to take a bit more time than I have today.

jasongrout:
Actually, let's make the dependency issue a different issue to fix, and work around it for now.

PR #1779, opened 2017-10-27
Updates for compatibility with JupyterLab

blink1073:
- Bumps the Phosphor and JupyterLab dependencies to match those of JLab.
- Adds `noUnusedLocals` to `jupyterlab_manager`
- Allows us to run `npm run addsibling <path-to-jupyterlab_manager>`:

<image src="https://user-images.githubusercontent.com/2096628/32102436-e8995b06-bae1-11e7-9b1e-e25f57c85550.png" width=400>


blink1073:
I stopped short of adding a test that uses `addsibling` to the Travis build, but could add it based on the [google drive extension test](https://github.com/jupyterlab/jupyterlab-google-drive/blob/1a45a562a1801bb1b7186a505d0a871fd22f78de/.travis.yml).

blink1073:
An alternative to this would be to move the `jupyterlab_manager` package to `JupyterLab`, since all of its `@jupyter-widgets` dependencies are >1.0.

blink1073:
We should discuss this at the meeting today.

blink1073:
For reference, adding this to JLab incurs a 1.42 MB delta (7.66 -> 9.08 MB), but we can put the `@jupyter-widgets` dependencies in a separate bundle using a [vendor chunk](https://webpack.js.org/plugins/commons-chunk-plugin/#explicit-vendor-chunk).

maartenbreddels:
Do I see a 'mirrored' cell?

blink1073:
You do indeed: https://github.com/jupyterlab/jupyterlab/pull/3159.

jasongrout:
Thanks!

Issue #1781, opened 2017-10-29
Simplified interface for JupyterLab extension?


blink1073:
Do third party widgets need more than the [registerWidget](https://github.com/jupyter-widgets/ipywidgets/blob/c6520c644eb7bae477e3c8dc559d0771c42709d4/packages/jupyterlab-manager/src/plugin.ts#L115) function?  If not, a simplified token could be in `@jupyter-widgets/base` and third party widgets could be stable already. 

Survey, all using just that function:
[ipyvolume](https://github.com/maartenbreddels/ipyvolume/blob/a24d5ece55c4b36ebd962fe062daecc80dbece99/jslab/index.js)
[bqplot](https://github.com/bloomberg/bqplot/blob/a73a7e5929e194272350fd09b2958973f1745890/jslab/index.js)
[jupyter-matplotlib](https://github.com/matplotlib/jupyter-matplotlib/blob/1d50775d9d25ce19c261fa2f75042be8c3d0b7a3/js/src/labplugin.js)
[nglview](https://github.com/arose/nglview/blob/1c63952b0ab9293593f8d6cea76a218cb806dea2/jslab/index.js#L10)
[pythreejs](https://github.com/jovyan/pythreejs/blob/28193ffce96eb85f522209dd3c33772c11bc8caf/jslab/lib/index.js)
[ipyleaflet](https://github.com/ellisonbg/ipyleaflet/blob/e9061c53f9f80658c020214eb5d8c799d72c2ef2/jslab/index.js#L9)
[ipytalos](https://github.com/qzchenwl/ipytalos/blob/7549f34cce7420e3d1071d6a973c4e1c6120d29a/jslab/index.js)
[ipyscales](https://github.com/vidartf/ipyscales/blob/19e578084c9989c8713dbb962a859facbf905126/packages/jlextension/src/index.ts#L53)
[qgrid-jupyterlab](https://github.com/quantopian/qgrid/blob/7fda48b4abe65a7594641bd0ee0487f5eb1f70c2/jslab/index.js#L12)

These are the ones I knew about plus: https://www.npmjs.com/browse/depended/@jupyter-widgets/jupyterlab-manager

blink1073:
cc @SylvainCorlay @maartenbreddels @hainm

blink1073:
That same registry could be used for both notebooks and consoles (once we had an API for console extensions).

blink1073:
cc @vidartf @chenwl @TimShawver

[...]

PR #1789, opened 2017-11-01
Move the JupyterLab registry token to the base package.

jasongrout:
Addresses #1781 

I looked into moving the token to the base package. I realized a few things:

1. There is a lot of confusion right now in what the jlab plugin exports. It exports the actual jlab document widget, but it shouldn't - it should just export the registration function. This should definitely be untangled.
2. The registry is specific to the jlab widget manager - the classic notebook widget manager does not have the concept of a registry (it uses require as a default registry). So it's a little weird to have the concept of a registry in the base package and/or the base widget manager.

This is definitely still a work in progress.

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -57,6 +57,7 @@
   },
   "dependencies": {
     "@jupyterlab/services": "^0.50.2",
+    "@phosphor/coreutils": "^1.2.0",
     "@phosphor/messaging": "^1.2.1",
     "@phosphor/widgets": "^1.3.0",
     "@types/backbone": "^1.3.33",
--- a/packages/base/src/index.ts
+++ b/packages/base/src/index.ts
@@ -8,3 +8,4 @@ export * from './widget_style';
 export * from './services-shim';
 export * from './viewlist';
 export * from './utils';
+export * from './registry';
--- a/packages/base/src/registry.ts
+++ b/packages/base/src/registry.ts
@@ -1,0 +1,43 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import {
+  Token
+} from '@phosphor/coreutils';
+
+/**
+ * A runtime interface token for a widget registry.
+ */
+export
+const IJupyterWidgetRegistry = new Token<IJupyterWidgetRegistry>('jupyter.extensions.jupyterWidgetRegistry');
+
+/**
+ * A registry of Jupyter Widgets.
+ *
+ * This is used by widget managers that support an external registry.
+ */
+export
+interface IJupyterWidgetRegistry {
+  /**
+   * Register a widget module.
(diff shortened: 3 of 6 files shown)


blink1073:
We could.define the interface and token I base.for a jlab manager, and have the implementation in a separate package. That is what we do for the statedb for example.

blink1073:
Actually the default statedb class is in the same file but the point stands. 😁

jasongrout:
> We could.define the interface and token in base for a jlab manager, and have the implementation in a separate package.

Yes, that's essentially what the current PR state does. But that's what I meant by it being weird to have the concept (i.e., interface) for a registry in the base package, because the most prevalent widget manager does not have the concept of a registry (it punts to require to do that).

blink1073:
We could call it a JupyterLab registry.  It doesn't tie the package to JupyterLab, but makes it explicit what it is intended to be used for.

jasongrout:
Yes. It still feels weird to have jlab specific stuff in the base package, but I think we'll probably do it as a matter of practicality for widget library authors.

blink1073:
Doesn't seem weird enough to have a completely separate package, given that it doesn't (in effect) add any dependencies and to the base package and is only an interface and a runtime representation of an interface.

blink1073:
It could be made more generic:

```typescript
/**
 * A runtime interface object for a widget registry.
 */
export
const IJupyterWidgetRegistry = new Token<IJupyterWidgetRegistry>('jupyter.extensions.jupyterWidgetRegistry');


/**
 * A registry of Jupyter Widgets.  Used by widget managers that support
 * an external registry.
 */
export
interface IJupyterWidgetRegistry {
  /**
   * Register a widget module.
   */
  registerWidget(data: WidgetManager.IWidgetData): void;
}
```

jasongrout:
> It could be made more generic:

That's almost exactly what is in this PR: https://github.com/jupyter-widgets/ipywidgets/pull/1789/files#diff-121363ec153b2a74f364564bd76b8fd5

I like how your docs indicate "for those widget managers that use a registry".

blink1073:
Yeah I cribbed it from the PR 😉

jasongrout:
@blink1073 - I rebased and squashed the changes. Let me know if it looks good to you. If it does, I'll merge and release a new version of the jlab manager.

jasongrout:
Marking as 7.0.x since it just changes the experimental jlab interface.

blink1073:
Looks great, thanks!

jasongrout:
tests were failing - your dependency-checking script caught the no-longer-necessary dependency. Thanks again for that!

blink1073:
👊

jasongrout:
Just released `@jupyter-widgets/jupyterlab-manager` 0.29 and `@jupyter-widgets/base` 1.1 with this change.

PR #1796, opened 2017-11-04
WIP React experiments

jasongrout:
Don't merge. This is just some experiments in using react.


cnishina (commented):
If these comments are not applicable to TypeScript because the tsx file or the usage, let's discuss them.

jasongrout:
Thanks for the helpful comments. I'm still in the experimental phase, just trying to get the patterns right, and then will go back through and clean up the code.

What do you think of the React pattern here?

cnishina:
No problem. I need to really read up on React to be helpful. _Disappears for a week to read up on it..._

PR #1800, opened 2017-11-07
Lazy control loading in jlab

jasongrout:
This enhances the jlab registry to allow the controls to be loaded lazily (or really any module to be loaded lazily).

There are two approaches. One creates a "LazyPromise" which takes a function, and only calls the function if the promise is actually invoked with `.then` or `.catch`. This is a bit esoteric, but allows us to have a nice api only dealing with promises. The second approach allows someone to register a function that returns the module exports, however, things are not automatically cached - the function is called every time the module exports are needed. If the user does automatic caching (for example, a requirejs call, or import should cache the module load), then this is probably not a huge issue.

Thoughts on which approach is better?


jasongrout:
CC @blink1073, CC https://github.com/jupyterlab/jupyterlab/issues/3196

blink1073:
I think it is fair to leave it up to the provider about the caching, so the latter approach.  In the case of `() => import('foo')`, they can use the magic comments to specify the loading type: https://webpack.js.org/api/module-methods/#import-.

Issue #1749, opened 2017-10-12
TypeScript formatting


cnishina:
We could improve the TypeScript quality of code by having a clang format. The  [clang-format](https://www.npmjs.com/package/clang-format) tool could standardize the format automagically:

* We can set the clang style for line length, style guide (like Google's style guide for JavaScript)
* Add to the `build:src` step to autoformat code
* Add test to check the format in step in travis tests

cnishina:
I will have a PR on this.

pbugnion:
While I think that adopting a convention for laying out both Python and JS code is good, I am very wary of code auto-formatters that reformat your code on build. I personally would prefer to use tslint. Beyond personal preferences, I note that this is what JupyterLab uses.

If we run the linting as part of the Travis tests, we could guarantee the quality of code going into the repo.

cnishina:
Not a problem. I'll look into using tslint.

pbugnion:
Great! I think improving consistency in the JS layer is tremendously useful.

cnishina:
Alright how about something like this https://github.com/jupyter-widgets/ipywidgets/pull/1757

If this looks good, I'll continue with the rest of the files and then get back to the implicit any issue.

cnishina:
From our offline discussion, we will break these up into smaller reviewable chunks. The first commit should be to add the tslint.json into the directories with the `lint` script. Next commits will break down to a list of commits to clean up via tslint as a cross between the types fixes:

- use fat arrow over function calls
- use let over var
- add missing semicolons / remove extra semicolons
- format 100 column width
- format spacing indentation

and the folder with the edits:

- examples
- packages/base/src
- packages/base/test
- etc.

So an example commit would be:

> tslint to use fat arrows over function for packages/base/src
> 
> - reference issue #1749
> - override tslint rule: foobar

As an approach, overriding the tslint rule should be better than using inline `/* tslint:disable */` `/* tslint:enable*/` Overall, this will make the commits smaller, easier to review, and rebasing would be easier / avoid making Travis unhappy.

cnishina:
other PRs: #1785, #1786, #1786

PR #1804, opened 2017-11-07
tslint curly, one variable per declaration, triple equals fixes for packages/base

cnishina:
- reference issue #1749

### DIFF
--- a/packages/base/src/backbone-patch.ts
+++ b/packages/base/src/backbone-patch.ts
@@ -44,7 +44,7 @@ import * as utils from './utils';
 // This *MUST* be called with the model as the `this` context.
 export
 function set(key, val, options) {
-    if (key === null) {
+    if (key == null) {
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


cnishina:
- Made fixes to missing indentation.

Note to reviewer:

- Changes made from == to === should be verified that it is okay to compare with ===

packages/base/src/backbone-patch.ts
```diff
@@ -44,7 +44,9 @@ import * as utils from './utils';
 // This *MUST* be called with the model as the `this` context.
 export
 function set(key, val, options) {
-    if (key == null) return this;
```
jasongrout: This is a direct patch of the backbone source, so I'd rather keep the `==` here for better compatibility.

jasongrout:
One inline comment - other than that, I think it looks fine. Thanks!

jasongrout:
Thanks again! I fixed the one comment I made, and will merge when tests pass.

PR #1805, opened 2017-11-07
Simplify the imports

jasongrout:
In particular, just use the parts of lodash we actually use.

### DIFF
--- a/packages/controls/src/utils.ts
+++ b/packages/controls/src/utils.ts
@@ -1,65 +1,15 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+export {
+    uuid, WrappedError, resolvePromisesDict
+} from '@jupyter-widgets/base';
+
+import {
+    WrappedError
+} from '@jupyter-widgets/base';
+
 import * as _ from 'underscore';
-
-/**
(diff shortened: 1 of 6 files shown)


blink1073:
Nice!

packages/controls/src/utils.ts
```diff
@@ -1,65 +1,15 @@
...
-    let s = [];
-    const hexDigits = '0123456789ABCDEF';
-    for (let i = 0; i < 32; i++) {
-        s[i] = hexDigits.substr(Math.floor(Math.random() * 0x10), 1);
-    }
-    s[12] = '4';  // bits 12-15 of the time_hi_and_version field to 0010
-    s[16] = hexDigits.substr((s[16] & 0x3) | 0x8, 1);  // bits 6-7 of the clock_seq_hi_and_reserved to 01
+import {
+    WrappedError
+} from '@jupyter-widgets/base';
```
vidartf: Note: This is now imported twice.

jasongrout: Thanks. See #1808

jasongrout: Oh wait. it is not imported twice. It's exported once, and imported once.

vidartf: Oh! Sorry for the noise then!

PR #1807, opened 2017-11-08
Update README.md

orpheus92:


jasongrout:
Thanks! This looks great. I'll wait for the tests to pass before merging.

jasongrout:
I notice that in the commit message, git only knows your name as "Di". We usually give credit to contributors in our release notes, and most people prefer to have their full name (see https://groups.google.com/d/msg/jupyter/j40P78_h3U4/K3vdg5VGCgAJ, for example). Do you want us to refer to you as "Di", or would you like us to mention you by full name?

orpheus92:
Sure, Di Wang is my full name.

jasongrout:
Great, I'll add an entry into our mailmap file, which maps the name in the commit to a full name. You can also change your Github name so that it shows up in future commits.

jasongrout:
Thanks again!

PR #1808, opened 2017-11-08
Delete duplicate import

jasongrout:


jasongrout:
It's imported once and exported once, so it's not a duplicate. This is invalid, so closing.

PR #1816, opened 2017-11-12
update widgets fallback message

akhmerov:
the fallback message also shows up in a jupyterlab console, where it is currently confusing to the user because it suggests that the widgets are incorrectly installed

image

I suggest reformulating the image being specific that it is only supposed to work in the jupyterlab notebooks.


jasongrout:
Thanks! I tweaked the wording just slightly too, and will merge once tests finish passing again.

akhmerov:
Ah, good catch!

PR #1817, opened 2017-11-13
Add buffers argument to comm open

vidartf:
After https://github.com/jupyter/notebook/issues/2730 and https://github.com/jupyterlab/jupyterlab/issues/2556 were fixed, the comm open functions now support the buffers argument. This PR updates the different shims and utilities to reflect this. It also adds a an interface definition for the comm object used in widgets (the classic notebook one, shimmed in jlab).

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -292,6 +292,15 @@ abstract class ManagerBase<T> {
         });
     }
 
+    register_model(model_id: string, modelPromise: Promise<WidgetModel>): void {
+        this._models[model_id] = modelPromise;
+        modelPromise.then(model => {
+            model.once('comm:close', () => {
+                delete this._models[model_id];
+            });
+        });
+    }
+
     /**
      * Create and return a promise for a new widget model
      *
@@ -319,7 +328,7 @@ abstract class ManagerBase<T> {
         }
 
         let modelPromise = this._make_model(options, serialized_state);
-        this._models[model_id] = modelPromise;
+        this.register_model(model_id, modelPromise);
         return await modelPromise;
     }
 
@@ -349,9 +358,6 @@ abstract class ManagerBase<T> {
             comm: options.comm,
         };
         let widget_model = new ModelType(attributes, modelOptions);
-        widget_model.once('comm:close', () => {
-            delete this._models[model_id];
-        });
         widget_model.name = options.model_name;
         widget_model.module = options.model_module;
         return widget_model;
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -9,7 +9,7 @@ import {
 } from './widget';
 
 import {
-    shims
+    IComm
 } from './services-shim';
 
 import {
@@ -111,7 +111,7 @@ interface WidgetOptions {
     /**
      * Comm object associated with the widget.
      */
-    comm?: any;
+    comm?: IComm;
 
     /**
      * The model id to use. If not provided, the comm id of the comm is used.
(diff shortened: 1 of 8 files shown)


jasongrout:
Thanks!

This bumps our notebook requirement to 5.1? If so, then I think it is a change for ipywidgets 8.

vidartf:
That depends: The buffers argument is optional, and if supplied will simply be ignored in comm.open for older versions of notebook. So it will not break any existing behavior. However, it might not work as intended for any who decide to start developing against this functionality (buffers might not be sent as expected, but silently discarded).

As such, I see two possible ways to not have this be a major break:
 - Document that the use of the buffers argument requires any notebook dependency to be > 5.1 (consumer is responsible).
 - Check on load, and polyfill if too old.

jasongrout:
A few more thoughts:

1. I guess we do specify that buffers can occur in the comm_open messages (https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/messages.md#instantiating-a-widget-object-1) - so we're good on that front.
2. The comm_close messages are actually not documented in the message spec. We should document those (and indicate the widget should be deleted when it receives a comm close message)
3. How are you imagining a polyfill would work?
4. Can we rename `IComm` to something like `IClassicComm` to make sure there is a distinction between the comms provided by the jlab services package, and the classic comm interface we are emulating from the classic notebook?

vidartf:
1. We already support buffers in Comm open going from the kernel to the frontend. This PR is simply adding symmetry by ensuring that it is also supported from the frontend to the kernel.

2. The comm_close message itself is specified in the general message spec, but I agree we should specify that widgets should be deleted on close.

3. Instead of calling out to `kernel.comm_manager.new_comm` in `_create_comm`, we copy the implementation of that code, and use it if we detect that buffers are not supported. This is for older versions of notebook, in which the implementation has been stable since 4.0, so there should be little risk of conflicts.

4. Sure

jasongrout:
That `new_com` method (https://github.com/jupyter/notebook/blob/4.x/notebook/static/services/kernels/comm.js#L34) creates a Comm object and calls its open method (https://github.com/jupyter/notebook/blob/4.x/notebook/static/services/kernels/comm.js#L150), which does not have the buffers argument. Are you saying that we also subclass the Comm object to have an `open` method with a buffers argument, and use that? I think that sounds like an okay workaround, so +1.

+1 on your answers to (1) (2) and (4) as well.

vidartf:
@jasongrout On it. Any idea what is the best way to check whether the polyfill is needed? I can check the `.length` field of the `new_comm` method, but I'm not sure if it is the most robust.

vidartf:
I added the polyfill + rename (pt 3 & 4). I'm not entirely happy with the robustness of the polyfill check, but feel free to commit on top of that if you know a better way.

I don't think pt 2 needs to go in this PR.

jasongrout:
> I don't think pt 2 needs to go in this PR.

Agreed.

jasongrout:
> I added the polyfill + rename (pt 3 & 4). I'm not entirely happy with the robustness of the polyfill check, but feel free to commit on top of that if you know a better way.

Given how stable this code is, we may just want to include the notebook 5.2 implementation and use that until widgets depends on 5.2+ (perhaps in the 8.0 release).

jasongrout:
I think this looks great. Thanks!

jasongrout:
> We might possibly want to override the open function for the Comm created in the else block below. Not sure...

I don't think that's necessary. That comm open function should not be called.

vidartf:
Thanks for cleaning it up :)

PR #1822, opened 2017-11-17
Selection widgets fixes

SylvainCorlay:
Minor changes

- `button_style` should not be allowed to be `None`.
- The selected index in the selection container is allowed to be None. I think that the default value should be None (currently, it is 0, which does not make sense if no list of children was passed yet).

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -166,7 +166,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
 
     _options_full = None
 
-    # This being read-only means that it cannot be changed from the frontend!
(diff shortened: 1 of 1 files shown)


jasongrout:
I think in most cases, people will expect that if they create a dropdown, the first item will be selected by default.

Perhaps we should handle the (relatively rare?) case of no children separately.

SylvainCorlay:
The issue is that when we create a tab / accordion widget with no argument, the current default value does not make sense.

SylvainCorlay:
I don't see a simple means to have a consistent behavior other than setting the default to `None`.

SylvainCorlay:
I have rebased and removed the change to the selection container. Will open another separate PR for it for a 7.1.

jasongrout:
Thanks!

PR #1828, opened 2017-11-20
CONTRIBUTING.md did not have relevant info

CleoQc:
ref: https://github.com/jupyter-widgets/ipywidgets/issues/1827

Attempted to centralize all info related to how to become a contributor into one place

PR #1829, opened 2017-11-20
Starting changes in dev_install.sh

jzf2101:
updating `dev_install.sh` commented conda installation instructions based #1828


jasongrout:
Thanks!

PR #1831, opened 2017-11-20
Add group for doc test

fenwickipedia:
Test  .travis.yml changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,7 @@ script:
     - 'if [[ $GROUP == js ]] ; then bash ./scripts/travis_script_js.sh ; fi'
 matrix:
     include:
-        - python: 3.6
-          env: GROUP=doc
-        - python: 3.6
-          env: GROUP=js BROWSER=firefox TRAVIS_NODE_VERSION=6.11.0
+        - python: GROUP="$GROUP"
+          env: GROUP="$GROUP"
 after_success:
     - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
     - 'if [[ $GROUP == doc ]] ; then bash ./scripts/travis_script_doc.sh ; fi'
     - 'if [[ $GROUP == js ]] ; then bash ./scripts/travis_script_js.sh ; fi'
 matrix:
+    include:
         - python: 3.6
           env: GROUP=doc
         - python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,19 +27,11 @@ script:
     - 'if [[ $GROUP == doc ]] ; then bash ./scripts/travis_script_doc.sh ; fi'
     - 'if [[ $GROUP == js ]] ; then bash ./scripts/travis_script_js.sh ; fi'
 matrix:
-    include:
-        - python: 3.4
-          env: GROUP="$GROUP"
-        - python: 3.5
-          env: GROUP="$GROUP"
         - python: 3.6
-          env: GROUP="$GROUP"
+          env: GROUP=doc
+        - python: 3.6
+          env: GROUP=js BROWSER=firefox TRAVIS_NODE_VERSION=6.11.0
         - python: 2.7
-          env: GROUP="$GROUP"
-    exclude:
-        - python: 3.4
-          env: GROUP="docs"
-        - python: 3.5
-          env: GROUP="docs"
+          env: GROUP=doc
 after_success:
     - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: python
 python:
     - 2.7
+    - 3.3
     - 3.4
     - 3.5
     - 3.6
@@ -11,9 +12,7 @@ addons:
   firefox: latest
 cache: yarn
 env:
-    - GROUP=docs
     - GROUP=python
-    - GROUP=js
 sudo: false
 before_install:
     - 'if [[ $GROUP == python ]] ; then bash ./scripts/travis_before_install_py.sh ; fi'
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
   firefox: latest
 cache: yarn
 env:
+    - GROUP=docs
     - GROUP=python
 sudo: false
 before_install:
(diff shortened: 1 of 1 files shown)


jasongrout:
@fenwickipedia - The rationale for having the js and doc groups under just python 3.6 is that we'd like to not have to build those for *every* python version we are testing against.

jzf2101:
FYI I don't think 2.7 works for documentation and also 3.3. doesn't work with travis

jasongrout:
Do you mean the docs don't build in python 2.7? And do you mean that travis no longer supports 3.3?

jzf2101:
Yes

jzf2101:
See https://github.com/jupyter/notebook/blob/master/.travis.yml

jasongrout:
Travis seems to be building these python 3.3 tests just fine. How about we simply add a 2.7 doc group, just like in https://github.com/jupyter-widgets/ipywidgets/blob/master/.travis.yml#L30-L31.

jzf2101:
That's odd I don't know why it's working for us and not for jupyter notebook

jzf2101:
@fenwickipedia i think you need to fix your gitprofile

jasongrout:
We need to test on 3.3, since ipywidgets 7 supports python 3.3. I think probably all you need to check is that docs build on 2.7, by adding just a single block like https://github.com/jupyter-widgets/ipywidgets/blob/master/.travis.yml#L30-L31.

fenwickipedia:
Thanks for the feedback folks! I'll do some scrubbing in Tuesday's sprint
per suggestions.

-tf-

sent from my android phone

On Nov 20, 2017 6:34 PM, "Jason Grout" <notifications@github.com> wrote:

We need to test on 3.3, since ipywidgets 7 supports python 3.3. I think
probably all you need to check is that docs build on 2.7, by adding just a
single block like https://github.com/jupyter-widgets/ipywidgets/blob/
master/.travis.yml#L30-L31.

—
You are receiving this because you were mentioned.
<https://github.com/jupyter-widgets/ipywidgets/pull/1831#issuecomment-345866940>,
or mute the thread

fenwickipedia:
@jasongrout I reverted changes and added 2.7 and it's failing now on the js builds for all versions:
```
error Command "test:unit:" not found. Did you mean "test:unit"?
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

fenwickipedia:
@jasongrout I got your point about js doc builds and took them out of env: section, and added 3.3 back. (js build was failing previously because where I added js to env, I didn't add browser and other params). But builds are green now :D

jasongrout:
Thanks! This looks much simpler :). Why did we delete the `include:` line?

jasongrout:
Looking at it, it looks like deleting the `include: ` line in the `matrix:` section was maybe unintentional. Can you put it back in and then I think this will be ready to merge?

.travis.yml
```diff
@@ -26,10 +26,11 @@ script:
     - 'if [[ $GROUP == doc ]] ; then bash ./scripts/travis_script_doc.sh ; fi'
     - 'if [[ $GROUP == js ]] ; then bash ./scripts/travis_script_js.sh ; fi'
 matrix:
-    include:
```
jasongrout: Can you put this line back in? I think deleting it was unintentional.

fenwickipedia:
Typo, my bad! Pushed the fix.

fenwickipedia:
fail on the doc build stage, maybe just needs to be kicked again.

fenwickipedia:
Just putting a link here for the issue https://github.com/jupyter-widgets/ipywidgets/issues/1830

jasongrout:
Looks great - thanks! I'll wait for the build to pass before merging.

fenwickipedia:
I'm seeing that 2.7 is built when the plan executes for docs, but in docs/source/conf.py we have:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
#
```
which probably explains `No such kernel named python3` error in the log

jzf2101:
@blink1073 says regarding bugs in build of the notebooks in 2.7:

> I think jupyter nbconvert --ExecutePreprocessor.kernel_name=python2 --to notebook --execute mynotebook.ipynb should do it
> https://nbconvert.readthedocs.io/en/latest/execute_api.html#execution-arguments-traitlets
> you should be able to use * syntax to select multiple notebooks as well.
> I'd suggest testing in a local python 2 conda environment rather than poking at Travis

jasongrout:
> I'm seeing that 2.7 is built when the plan executes for docs, but in docs/source/conf.py we have:
> which probably explains No such kernel named python3 error in the log

To elaborate on what @blink1073 is saying - the kernel issue indicates that the notebooks have a python 3 kernel specified, and it's not finding a python 3 kernel (hence the error).

To proceed, it sounds like the thing to do is to get the docs rendering on python 2 first (i.e., set up a conda environment with python 2, and try making the docs). Fixing those errors will resolve these issues.

I don't have a ton of experience with the nbsphinx and jupyter_sphinx plugins, which is what is being used to run the doc notebooks. It doesn't look like nbsphinx has a way to pass options on to nbconvert about what kernel to use (to use the option that @blink1073 pointed out above), so fixing this may involve either writing a something that manually converts the doc notebooks to use python 2 kernels, or extending nbsphinx to be able to pass options on to nbconvert (to run notebooks with a different kernel).

Or we can say that docs build on python 3 only...

jasongrout:
> Or we can say that docs build on python 3 only...

P.S. I don't know of any complaints that our docs only build on python 3. To judge the impact of the effort above - are there situations where building docs on python 3 is a problem?

jzf2101:
https://github.com/jupyter/notebook/pull/2808 has a good reference to building docs only having python 3 available.

jasongrout:
> jupyter/notebook#2808 has a good reference to building docs only having python 3 available.

That brings us back to the status quo - the current repo only builds docs on python 3 anyway. So...no change?

This PR is exposing the fact that building ipywidgets docs only works on python 3 right now. That's really good to know (thanks!). The question is whether it is worth it to fix it to build docs on python 2.

jzf2101:
It appears as the issue in the notebook pr has to do with not being able to call python 3 when testing. It may be useful to figure out how to build the docs if you're a python 2 contributor?

jasongrout:
I'll close this for now. If any developers on python 2 would like to push this through, please comment and we can reopen this or pursue it in a different PR.

PR #1832, opened 2017-11-20
Reference RTD Latest in Dev Documentation

jzf2101:
changing the dev install instructions to reference rtd latest


jasongrout:
I tweaked the wording just a bit to emphasize that this is the 'latest' docs.

jasongrout:
Thanks!

Issue #1827, opened 2017-11-20
Dev Installation Instructions and Contributing.md are confusing


jzf2101:
Dev Installation instructions are in readme not contributing.md. The dev installation instructions in the readme also are different from the dev installation instructions on the website.

peristeri:
The instructions in the `dev_install.sh` provides yet an other way of setting up the dev environment.

CleoQc:
I've copied the intructions that were in `dev_install.sh` but I am not sure what to do with 'it does not work on both a mac and linux environment python 3.6.x' 
(that aside, I'm totally failing at installing on Windows so I can't really test anything)

Doodledun:
Per discussion in Gitter, trying to replace all the places use `npm` with `yarn`  in `dev_install.sh` seems to work for me. I am on linux 16.04 and using python 2.7

peristeri:
I confirm that using `yarn` instead of `npm` worked for me.

PR #1833, opened 2017-11-21
Use `yarn` instead of `npm` in the dev install script

peristeri:
With the dependency to `jupyterlab` which uses `yarn` to build its js files. Switching to `yarn` solved the error of: `base.ts(5,27): error TS2307: Cannot find module '@jupyterlab/services'`

Came across this error when helping with #1827

### DIFF
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -10,13 +10,12 @@
 # cd ipywidgets
 # ./dev-install.sh
 
-echo -n "Checking npm... "
-npm -v
+echo -n "Checking yarn... "
+yarn -v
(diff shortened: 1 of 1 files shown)


jzf2101:
@jasongrout I'm cool with this if you are

jasongrout:
Looks good to me. Thanks!

vidartf:
Does this mean that all projects that depend on Jupyterlab now have to switch from npm to yarn?

blink1073:
No, I think the underlying issue here is that you can't mix `npm install` and `yarn`.  Consumers of jlab packages can use whichever they wish.

Issue #1719, opened 2017-09-18
Introduce a ToggleButtons button_color style attribute


simaoh:
Can we change color and font of the buttons in `ToggleButtons`?
Should this be added to the ToggleButtonsStyle class in `ipywidgets/widgets/widget_selection.py`? (as in https://github.com/jupyter-widgets/ipywidgets/issues/1189) Something like this:

```python 
from .widget_core import CoreWidget
from .widget_style import Style
from .trait_types import InstanceDict, Color

class ToggleButtonsStyle(Style, CoreWidget):
     """Button style widget."""
     _model_name = Unicode('ToggleButtonsStyleModel').tag(sync=True)
     button_width = Unicode(help="The width of each button.").tag(sync=True)
     button_color = Color('#060482', allow_none=True, help="Color of the button").tag(sync=True)
     font_weight = Unicode(help="Button text font weight.").tag(sync=True)
```

pbugnion:
Sounds like a good idea to me!

Doodledun:
@pbugnion I am at PyconCa Sprint, I am very interested in this issue. I am going to try to work on this one.  Thanks for the tips!

jasongrout:
FIxed by #1834. Thanks again!

PR #1834, opened 2017-11-21
1719 support more styles to toggle buttons

Doodledun:


ipywidgets/widgets/widget_selection.py
```diff
@@ -384,9 +385,22 @@ class ToggleButtonsStyle(DescriptionStyle, CoreWidget):
     button_width: str
         The width of each button. This should be a valid CSS
         width, e.g. '10px' or '5em'.
+
+    button_color: str or None
+        The color of each button. This should be a valid CSS color unit,
+        e.g. 'yellow', '#FFFF00', '255,255,0'.
```
jasongrout: The rgb color should be of the form `rgb(255,255,0)`, right? https://developer.mozilla.org/en-US/docs/Web/CSS/color

Doodledun: Yes, you are totally right. Thanks for the catch! I will make updates accordingly.

jasongrout:
For future reference, this is fixing #1719.

jasongrout:
Testing the examples given reveals that the color trait won't take something like `'rgb(255,255,0)'`, so I deleted that from the examples.

jasongrout:
Thanks again!

ankostis:
Is this supposed to be merged in `ipywidgets-7.4.2` (``jupyterlab-v0.35.4`)?
```python
>>> display(widgets.Button().style.keys)
'_model_module',
 '_model_module_version',
 '_model_name',
 '_view_count',
 '_view_module',
 '_view_module_version',
 '_view_name',
 'button_color',
 'font_weight']
>>> display(widgets.ToggleButton().style.keys)
['_model_module',
 '_model_module_version',
 '_model_name',
 '_view_count',
 '_view_module',
 '_view_module_version',
 '_view_name',
 'description_width']
```

jasongrout:
> Is this supposed to be merged in `ipywidgets-7.4.2` (`jupyterlab-v0.35.4`)?

This is answered in https://github.com/jupyter-widgets/ipywidgets/issues/3268#issuecomment-913758351

PR #1841, opened 2017-11-27
Make the basic phosphor css less specific

### LINKED ISSUE #1837: Box Widget - overflow css inconsistency
Hello,

By default the overflow on a Box (css class .widget-box) is working well.
But when added into another widget it getting override by `.jupyter-widgets .p-Widget` which have the `overflow:hidden`
In the meantime I didn't find any way to set his style (via style or layout).
Here a code to reproduce it:
```
import ipywidgets as widgets

items = [widgets.Label(str(i), layout=widgets.Layout(width='auto',min_height='20px', height='20px',display='flex')) for i in range(10)]
styleOverflowAuto = {'overflow': 'auto'}
box_layout = widgets.Layout(width='300px',display='flex',flex_flow='column', height='150px', overflow='auto')
box = widgets.Box(items, layout=box_layout, style=styleOverflowAuto)
text = widgets.Text(layout=widgets.Layout(width='300px'),placeholder='Filter', disabled=False)
component = widgets.VBox([text, box])
#overflow works well when alone
#display(box)
#Overflow doesn't work when embedded
display(component)
```

On other note, it was working before, I updated my packages including ipywidgets, I can't recall on which version I was previously. I tried to downgrade to find when the breaking change appears, but without success

Can i do anything?

jasongrout:
This is so that it doesn’t override widget-specific settings.

Fixes #1837

### DIFF
--- a/packages/controls/css/phosphor.css
+++ b/packages/controls/css/phosphor.css
@@ -32,7 +32,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
  /* The following section is derived from https://github.com/phosphorjs/phosphor/blob/23b9d075ebc5b73ab148b6ebfc20af97f85714c4/packages/widgets/style/widget.css */
 
-.jupyter-widgets .p-Widget {
+ /* The basic widget css is not namespaced so that it isn't more specific than widget classes. */
+.p-Widget {
   box-sizing: border-box;
   position: relative;
   overflow: hidden;


jasongrout:
We namespace most of the phosphor css in case a different version of phosphor is on the page - our embedded bundle's phosphor would then stomp on the global phosphor's css.

However, we now make an exception for the top-level `.p-widget` css, since it is causing problems with specificity in our widget css selectors (i.e., overriding things it shouldn't). We could say that it is up to the widget manager to put this one snippet on the page, but for backwards compatibility, we just do it still. This has the possibility of stomping on possibly different phosphor css already on the page, but I think that conflict unlikely as the `.p-widget` css is very minimal.

PR #1848, opened 2017-12-02
WIP Make JupyterLab top-level widgets draggable

jasongrout:
The idea here is to make it so that it is easy to drag widgets into their own jupyterlab tab.

WIP - don't merge yet.

### DIFF
--- a/packages/jupyterlab-manager/src/draghandle.ts
+++ b/packages/jupyterlab-manager/src/draghandle.ts
@@ -1,0 +1,177 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+/**
+ * A widget that abstracts out the drag operation to create a new panel in a dock area.
+ */
+
+import {
+    Widget
+} from '@phosphor/widgets';
+
+import {
+    Message
+} from '@phosphor/messaging';
+
+import {
+    Drag
+  } from '@phosphor/dragdrop';
+
+import {
+    MimeData
+  } from '@phosphor/coreutils';
+
+/**
+ * The threshold in pixels to start a drag event.
+ */
+const DRAG_THRESHOLD = 5;
+
+
+export
+class DragHandle extends Widget {
+
+  constructor(options: DragHandle.IOptions) {
+    super();
+    this._newWidget = options.newWidget;
+    this.node.textContent = 'Drag';
+  }
+
+  /**
+   * Handle the DOM events for the widget.
+   *
+   * @param event - The DOM event sent to the widget.
+   *
+   * #### Notes
+   * This method implements the DOM `EventListener` interface and is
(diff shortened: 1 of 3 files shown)


jasongrout:
Currently, the mirrored jupyter widget does not correctly set itself as the main area current widget when it is focused. I'm investigating how to do this.

@blink1073 - any hints about how to set a widget as the dock panel currentWidget? For some reason, it seems that my new widget from the drop does not have the focus/blur handlers that seem to be on other things setting them as the dock current widgets.

blink1073:
A dropped widget is [activated](https://github.com/phosphorjs/phosphor/blob/master/packages/widgets/src/dockpanel.ts#L590) by the dock panel, and you've defined an `onActivateRequest` that ensures focus on the panel, is `onActivateRequest` not getting called on your `FocusPanel`?

jasongrout:
onActivateRequest is getting called just fine.

What seems to be missing is that the widget is not getting added to the shell's _tracker (as happens, for example, in https://github.com/jupyterlab/jupyterlab/blob/9f5f0ac2a4620e7e8b4432ed311566d8b547e0ae/packages/application/src/shell.ts#L372 for the right-click cell mirroring). I suppose it's not getting added to the shell's `_tracker` because the drop action is an action at the phosphor dockpanel level (i.e., it doesn't know about the shell's tracker), not the application shell level.

Should the application shell override the drop event in the phosphor dockpanel to also add the new widget to the shell focus tracker?

jasongrout:
The reason dragging from the file listing works is that the drop callback at https://github.com/jupyterlab/jupyterlab/blob/9f5f0ac2a4620e7e8b4432ed311566d8b547e0ae/packages/filebrowser/src/listing.ts#L1159 invokes `DocumentManager.open`, calling `DocumentManager._createOrOpenDocument`, which calls the opener at https://github.com/jupyterlab/jupyterlab/blob/9f5f0ac2a4620e7e8b4432ed311566d8b547e0ae/packages/docmanager-extension/src/index.ts#L89, which adds it to the main area using `addToMainArea` (which then registers the widget with the application shell focus tracker).

From this chain, it appears that a drop callback must add a widget to the main area using the shell `addToMainArea`? That seems awkward - it's probably better to have the shell's main area dockpanel's drop overridden to add the new widget to the focus tracker if it's not already there?

blink1073:
Ha, originally the dock layout *had* a focus tracker.  Yes, I think we need to subclass `DockLayout` to add the linkage.

jasongrout:
I made a PR to JupyterLab for adding dock children to the focus tracker: https://github.com/jupyterlab/jupyterlab/pull/3310

jasongrout:
We now have output mirroring in jlab, which I think solves this in a different way, so I'll close this.

Issue #1858, opened 2017-12-07
npm package error


AugustoYao:
After installed jupyterlab 0.29.2, it works for a while with `ipywidgets`.  Since 6, Dec, 2017, after I re-build jupyterlab with command `jupyter lab clean && jupyter lab build`, jupyterlab behaves a little weird. At first, the page's height is about180 px, it cannot take my full screen. After Checking for a while, I found that @jupyter-widgets/controls seems at wrong order, the css override order is not as it was before. I wait for quite a few hours, then I rebuild jupyterlab. This time,  the page seems work fine, but when executing `import ipywidgets
ipywidgets.IntSlider()`, the output area shows that `Error displaying widget`. I wonder how can I specify npm package version, so that it can only build my own package, without updating packages like @jupyter-wigets/controls and so on? 
P.s. I noticed that @jupyter-widgets/controls have two new versions in last two days. But both don't work well in my environment, so I just want to roll back to 1.0.10. 
@jasongrout  Could you give me some kindly advice, please?

jasongrout:
I just pushed the `@jupyter-widgets/jupyterlab-manager` 0.31.3, which fixes this issue. You can upgrade to that, or you can install a specific version like 1.0.10 by doing something like `jupyter labextension install @jupyter-widgets/jupyterlab-manager@1.0.10` (you might need to uninstall your current extension first).

jasongrout:
@AugustoYao - please let us know if one of the above solutions works for you.

AugustoYao:
After uninstall @jupyter-widgets/jupyterlab-manager with command `jupyter lab uninstall @jupyter-widgets/jupyterlab-manager`, I re-install @jypyter-widgets/jupyterlab-manager@0.31.3, and it fails too, following is the error message while installing @jupyter-widgets/jupyterlab-manager@0.31.3
```
[jhyao@sdd002023 ~]$ jupyter labextension install @jupyter-widgets/jupyterlab-manager@0.31.3
> node node-version-check.js
> npm pack @jupyter-widgets/jupyterlab-manager@0.31.3
jupyter-widgets-jupyterlab-manager-0.31.3.tgz
Traceback (most recent call last):
  File "/disk/sata/sdi/jhyao/Sky4.1/SkyDiscovery/bin/jupyter-labextension", line 6, in <module>
    sys.exit(jupyterlab.labextensions.main())
[...]

PR #1859, opened 2017-12-07
Enable the jlab plugin to work with ipywidgets 7.0.*

jasongrout:
Until we have better semver matching of model spec versions, we’ll register the controls under all appropriate model spec versions.


jasongrout:
Fixes #1858.

Issue #1876, opened 2017-12-14
RTD docs builds timing out


takluyver:
I just noticed that ipywidgets docs builds are failing on Readthedocs with a timeout.

E.g. https://readthedocs.org/projects/ipywidgets/builds/6379796/

jasongrout:
Thanks!

jasongrout:
@SylvainCorlay - would it be possible for you to take a look at this? It appears that RTD has resource limits of 15m and 1G of memory: http://read-the-docs.readthedocs.io/en/latest/builds.html.

jasongrout:
I looked into this. After a series of PRs that tweaked our build configuration (#1884, #1885, #1886, #1887), our build now takes about 400s instead of over 900s.

I also put in some PRs to avoid installing recommonmark from pip, presuming that the conda installation would be faster: https://github.com/conda-forge/recommonmark-feedstock/pull/1, https://github.com/conda-forge/commonmark-feedstock/pull/5.

But for now, our docs are not timing out anymore.

jasongrout:
reopening this, as it seems more fickle than I originally thought. Some of the recent builds have been failing as well. Comparing passing with failing builds, it seems that a sphinx-build with lots of errors is very costly (taking ~240s instead of ~5s). See for example the ` python /home/docs/checkouts/readthedocs.org/user_builds/ipywidgets/conda/latest/bin/sphinx-build -b latex -D language=en -d _build/doctrees . _build/latex ` step of https://readthedocs.org/projects/ipywidgets/builds/6486111/

jasongrout:
My guess is that if we fix those errors showing up in the sphinx-build, the times will drop drastically and we won't time out.

takluyver:
This is a guess, but it might be to do with caching. I've noticed that Sphinx warnings about reading source files are often hidden if you rebuild after Sphinx has cached data. If so, the errors and the slowness are two symptoms of needing to re-read files or regenerate cached data.

jasongrout:
Do you know how to effect caching on RTD? I'm in unfamiliar territory here.

takluyver:
I don't, sorry!

pbugnion:
I'm going to close this, since it no longer seems relevant. Feel free to re-open if I'm missing something.

PR #1886, opened 2017-12-26
More updates to docs build environment.

jasongrout:
This cleans up what looks like unnecessary restrictions now.


jasongrout:
See #1876

Issue #1908, opened 2018-01-17
Make the installation use the notebook 5.3 nbextension conf.d installation approach


jasongrout:
This would make it so that we don't have to use the post-link script in the conda package, and the pip package would not need to also do `jupyter nbextension install ...`.

See https://github.com/jupyter-widgets/widget-ts-cookiecutter/pull/9/files for an example of how to do this.

See jupyter/notebook#3116 for the notebook PR, released in notebook 5.3.

jasongrout:
Another example of how to do this is https://github.com/maartenbreddels/ipyvolume/commit/415f0fb904b65a18cabb6939dc6a586c820ec797

blink1073:
Happy to help with this at the workshop.

jasongrout:
Thanks. I've been accumulating a lot of "good first issue"s particularly for people wanting to tackle things at the workshop.

maartenbreddels:
working on it :)

maartenbreddels:
or maybe I shouldn't, let me know.

jasongrout:
I think it's fine. We definitely have enough "good first issue"s to go around.

jasongrout:
(but if you have other stuff to do, and can wait a few days, you can do it with a new contributor next week and help them come up to speed...)

jasongrout:
Fixed in #1911

PR #1911, opened 2018-01-17
for notebook 5.3 you don't need to enable the extensions anymore, solves #1908

maartenbreddels:
for notebook 5.3 you don't need to enable the extensions anymore, solves #1908


jasongrout:
For backwards compatibility, it's actually:

```
{
  "load_extensions": {
    "jupyter-js-widgets/extension": true
  }
}
```

maartenbreddels:
good catch 👍

jasongrout:
Thanks. Came from trying it out and seeing it didn't work :).

maartenbreddels:
After the third widget I got a bit overconfident :)

jasongrout:
There be dragons in this ancient codebase!

jasongrout:
This works great. Thanks!

Issue #1809, opened 2017-11-09
export the traitlets link function


jasongrout:
It's a little confusing that the link function comes from traitlets, but the jslink one comes from ipywidgets. I think it makes sense to re-export the traitlets link function so that the api is more consistent.

rprimet:
I'll give it a try

jasongrout:
Fixed by #1923.

PR #1923, opened 2018-01-23
re-export link and dlink from traitlets for consistency with the javascript api

rprimet:
…script API

### DIFF
--- a/docs/source/examples/Widget Events.ipynb
+++ b/docs/source/examples/Widget Events.ipynb
@@ -250,18 +250,7 @@
    "source": [
     "### Linking traitlets attributes in the kernel\n",
     "\n",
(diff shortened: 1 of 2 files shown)


rprimet:
(related issue: #1809)

jasongrout:
Thank you! Can we also delete the cell in the example where traitlets is imported?

rprimet:
Good catch, thanks! This should be fixed now.

jasongrout:
Thank you!

SylvainCorlay:
Quick note @rprimet, your git configuration does not match your GitHub user email so that your account is not credited with this commit. Sorry for not catching that earlier.

rprimet:
Ah thanks for that ! I'll update my git config.

PR #1924, opened 2018-01-23
adding option to rename interact button 1870

dharmaquark:
code change for https://github.com/jupyter-widgets/ipywidgets/issues/1870

### DIFF
--- a/ipywidgets/widgets/interaction.py
+++ b/ipywidgets/widgets/interaction.py
@@ -158,8 +158,8 @@ class interactive(VBox):
         The function to which the interactive widgets are tied. The `**kwargs`
         should match the function signature.
     __options : dict
-        A dict of options. Currently, the only supported key is
-        ``"manual"``.
+        A dict of options. Currently, the only supported keys are
+        ``"manual"`` and ``"manual_name"``.
     **kwargs : various, optional
         An interactive widget is created for each keyword argument that is a
         valid widget abbreviation.
@@ -177,6 +177,7 @@ class interactive(VBox):
         self.f = f = __interact_f
         self.clear_output = kwargs.pop('clear_output', True)
         self.manual = __options.get("manual", False)
+        self.button_name = __options.get("button_name", "Run Interact")
         self.auto_display = __options.get("auto_display", False)
 
         new_kwargs = self.find_abbreviations(kwargs)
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
@dharmaquark it seems that you have not configured your local git installation so that the commit is not associated with your GitHub account.

For your GitHub account to be credited with this commit you can either

 - add the email associated with this commit to your GitHub account
 - re-do this commit after you change changed your git configuration to correspond to the github email.

jasongrout:
@dharmaquark - we can work together today to get the right email associated with the commit. Good catch @SylvainCorlay - thanks!

dharmaquark:
thanks! I've added my work email to my personal GitHub account- I think this should fix the issue? If not, I can work with Jason offline to fix my configuration!

SylvainCorlay:
This looks good to me. I approve the PR!

jasongrout:
Thanks! Sorry for the delay in merging this.

wernight:
interact.options aren't well documented. Seems like `interact_manual` should be updated to support this.

davidboles:
I would love to find an example of this in action - I can't seem to get it to work. For example, all of the commented-out attempts below fail, either because it isn't valid Python or the button remains labelled "Run Interact".

```
#@widgets.interact_manual.option(manual_name='Make It So')(color=['blue', 'red', 'green'], lw=(1., 10.))
#@widgets.interact_manual(color=['blue', 'red', 'green'], lw=(1., 10.)).options(manual_name='Make It So')
def plot(freq=1., color='blue', lw=2, grid=True):
    t = np.linspace(-1., +1., 1000)
    fig, ax = plt.subplots(1, 1, figsize=(8, 6))
    ax.plot(t, np.sin(2 * np.pi * freq * t), lw=lw, color=color)
    ax.grid(grid)
    
#w = widgets.interact_manual(plot, color=['blue', 'red', 'green'], lw=(1., 10.)).options(manual=True, manual_name='Make It So')
#w = widgets.interact_manual(plot, color=['blue', 'red', 'green'], lw=(1., 10.), options={'manual_name': 'Make It So'})
#w = widgets.interact_manual(plot, color=['blue', 'red', 'green'], lw=(1., 10.), manual_name='Make It So')

```

I think we'd benefit from having a simple working example.

jasongrout:
Here is one way. I think it still could probably use another iteration making it easier:

```python

interact_calc=interact.options(manual=True, manual_name="Calculate")

@interact_calc(x=(0,10))
def f(x):
    print(x)
```
You're right - this should be in the docs.

jasongrout:
You could also do:
```python
interact_calc=interact_manual.options(manual_name="Calculate")
```

PR #1928, opened 2018-01-23
Added FloatLogSlider

### LINKED ISSUE #719: Custom scales for sliders
Often when using sliders to manipulate some algorithm parameters it is useful to have logarithmic scale, instead of linear. Now I guess the only way is to manually apply the transformation in the handler, e.g. have the slider from -3 to 3 and in handler write `value = 10**value`. One thing is that it's not really convenient (code defining the real slider range becomes split into two parts in entirely different places), and another is that the slider label of course shows the "raw" value, not the real one used in computations.

I think there is a relatively easy solution - add two function arguments for `FloatSlider` to convert back and forth between the slider raw value and the real one (in my example they would be `lambda x: 10**x` and `lambda x: log10(x)`, and/or maybe a boolean argument `log` for convenience, as probably logarithmic scale is needed way more often than other custom ones.

sebasguts:
a slider that uses a logarithmic scale. Closes #719

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -58,6 +58,40 @@ class _BoundedFloat(_Float):
             self.value = max
         return max
 
+class _BoundedLogFloat(_Float):
+    max = CFloat(4.0, help="Max value for the exponent").tag(sync=True)
+    min = CFloat(0.0, help="Min value for the exponent").tag(sync=True)
+    base = CFloat(10.0, help="Base of value").tag(sync=True)
+    value = CFloat(1.0, help="Float value").tag(sync=True)
+
+    @validate('value')
+    def _validate_value(self, proposal):
+        """Cap and floor value"""
+        value = proposal['value']
+        if self.base ** self.min > value or self.base ** self.max < value:
+            value = min(max(value, self.base **  self.min), self.base **  self.max)
+        return value
+
+    @validate('min')
+    def _validate_min(self, proposal):
+        """Enforce base ** min <= value <= base ** max"""
+        min = proposal['value']
+        if min > self.max:
+            raise TraitError('Setting min > max')
+        if self.base ** min > self.value:
+            self.value = self.base ** min
+        return min
+
+    @validate('max')
+    def _validate_max(self, proposal):
+        """Enforce base ** min <= value <= base ** max"""
+        max = proposal['value']
+        if max < self.min:
+            raise TraitError('setting max < min')
+        if self.base ** max < self.value:
+            self.value = self.base ** max
+        return max
+
 
 @register
 class FloatText(_Float):
@@ -140,6 +174,48 @@ class FloatSlider(_BoundedFloat):
         '.2f', help="Format for the readout").tag(sync=True)
     continuous_update = Bool(True, help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
     disabled = Bool(False, help="Enable or disable user changes").tag(sync=True)
+
+    style = InstanceDict(SliderStyle).tag(sync=True, **widget_serialization)
+
+
+@register
+class FloatLogSlider(_BoundedLogFloat):
+    """ Slider/trackbar of logarithmic floating values with the specified range.
+
+    Parameters
+    ----------
+    value : float
+        position of the slider
+    base : float
+        base of the logarithmic scale. Default is 10
+    min : float
+        minimal position of the slider in log scale, i.e., actual minimum is base ** min
+    max : float
+        maximal position of the slider in log scale, i.e., actual maximum is base ** max
+    step : float
+        step of the trackbar, denotes steps for the exponent, not the actual value
+    description : str
+        name of the slider
+    orientation : {'horizontal', 'vertical'}
+        default is 'horizontal', orientation of the slider
+    readout : {True, False}
+        default is True, display the current value of the slider next to it
+    readout_format : str
+        default is '.2f', specifier for the format function used to represent
+        slider value for human consumption, modeled after Python 3's format
+        specification mini-language (PEP 3101).
+    """
+    _view_name = Unicode('FloatLogSliderView').tag(sync=True)
+    _model_name = Unicode('FloatLogSliderModel').tag(sync=True)
+    step = CFloat(0.1, help="Minimum step in the exponent to increment the value").tag(sync=True)
+    orientation = CaselessStrEnum(values=['horizontal', 'vertical'],
+        default_value='horizontal', help="Vertical or horizontal.").tag(sync=True)
+    readout = Bool(True, help="Display the current value of the slider next to it.").tag(sync=True)
+    readout_format = NumberFormat(
+        '.2f', help="Format for the readout").tag(sync=True)
+    continuous_update = Bool(True, help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
+    disabled = Bool(False, help="Enable or disable user changes").tag(sync=True)
+    base = CFloat(10., help="Base for the logarithm").tag(sync=True)
 
 
     style = InstanceDict(SliderStyle).tag(sync=True, **widget_serialization)
--- a/packages/controls/src/widget_float.ts
+++ b/packages/controls/src/widget_float.ts
@@ -12,12 +12,13 @@ import {
 import * as _ from 'underscore';
 
 import {
-    IntSliderView, IntRangeSliderView, IntTextView
+    IntSliderView, IntRangeSliderView, IntTextView, BaseIntSliderView
 } from './widget_int';
 
 import {
     format
 } from 'd3-format';
+
 
 export
 class FloatModel extends CoreDescriptionModel {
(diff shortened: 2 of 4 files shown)


ipywidgets/widgets/widget_float.py
```diff
@@ -58,6 +58,39 @@ def _validate_max(self, proposal):
             self.value = max
         return max
 
+class _BoundedLogFloat(_Float):
+    max = CFloat(10.0, help="Max value").tag(sync=True)
```
jasongrout: Can we clarify that the min/max are *exponents*, while the value is an actual number? That was confusing to me.

ipywidgets/widgets/widget_float.py
```diff
@@ -58,6 +58,39 @@ def _validate_max(self, proposal):
...
+        min = proposal['value']
+        if min > self.max:
+            raise TraitError('Setting min > max')
+        if min > self.value:
+            self.value = min
+        return min
+
+    @validate('max')
+    def _validate_max(self, proposal):
+        """Enforce min <= value <= max"""
```
jasongrout: We should compare value to the `base**min` and `base**max`, right?

ipywidgets/widgets/widget_float.py
```diff
@@ -58,6 +58,39 @@ def _validate_max(self, proposal):
...
+            value = min(max(value, self.base **  self.min), self.base **  self.max)
+        return value
+
+    @validate('min')
+    def _validate_min(self, proposal):
+        """Enforce min <= value <= max"""
+        min = proposal['value']
+        if min > self.max:
+            raise TraitError('Setting min > max')
+        if min > self.value:
```
jasongrout: We should compare value to the `base**min` and `base**max`, right?

ipywidgets/widgets/widget_float.py
```diff
@@ -141,6 +176,46 @@ class FloatSlider(_BoundedFloat):
     continuous_update = Bool(True, help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
     disabled = Bool(False, help="Enable or disable user changes").tag(sync=True)
 
+    style = InstanceDict(SliderStyle).tag(sync=True, **widget_serialization)
```
jasongrout: Thanks for catching that we didn't have the right style!

ipywidgets/widgets/widget_float.py
```diff
@@ -141,6 +176,46 @@ class FloatSlider(_BoundedFloat):
...
+@register
+class FloatLogSlider(_BoundedLogFloat):
+    """ Slider/trackbar of logarithmic floating values with the specified range.
+
+    Parameters
+    ----------
+    value : float
+        position of the slider
+    min : float
+        minimal position of the slider
```
jasongrout: Again, please clarify min/max are exponents

ipywidgets/widgets/widget_float.py
```diff
@@ -141,6 +176,46 @@ class FloatSlider(_BoundedFloat):
...
+    max : float
+        maximal position of the slider
+    step : float
```
jasongrout: Please clarify step is a step in the exponent, not the actual value.

ipywidgets/widgets/widget_float.py
```diff
@@ -141,6 +176,46 @@ class FloatSlider(_BoundedFloat):
...
+    readout : {True, False}
+        default is True, display the current value of the slider next to it
+    readout_format : str
+        default is '.2f', specifier for the format function used to represent
+        slider value for human consumption, modeled after Python 3's format
+        specification mini-language (PEP 3101).
+    """
+    _view_name = Unicode('FloatLogSliderView').tag(sync=True)
+    _model_name = Unicode('FloatLogSliderModel').tag(sync=True)
+    step = CFloat(0.1, help="Minimum step to increment the value").tag(sync=True)
```
jasongrout: help should clarify this is changing the exponent.

packages/schema/jupyterwidgetmodels.latest.md
```diff
@@ -296,6 +296,31 @@ Attribute        | Type             | Default          | Help
...
+`_model_name`    | string           | `'FloatLogSliderModel'` | 
+`_view_module`   | string           | `'@jupyter-widgets/controls'` | 
+`_view_module_version` | string           | `'1.1.0'`        | 
+`_view_name`     | string           | `'FloatLogSliderView'` | 
+`base`           | number (float)   | `2.0`            | Base for the logarithm
+`continuous_update` | boolean          | `true`           | Update the value of the widget as the user is holding the slider.
+`description`    | string           | `''`             | Description of the control.
+`disabled`       | boolean          | `false`          | Enable or disable user changes
+`layout`         | reference to Layout widget | reference to new instance | 
+`max`            | number (float)   | `10.0`           | Max value
```
jasongrout: Changing the help above to clarify min/max/step are talking about the exponent, then regenerating this, should change the docs here too.

sebasguts:
Thanks for the review :).

I changed all the comments as proposed, and updated the PR.

packages/controls/src/widget_float.ts
```diff
@@ -69,6 +70,36 @@ class FloatSliderModel extends BoundedFloatModel {
     readout_formatter: any;
 }
 
+export
+class FloatLogSliderModel extends BoundedFloatModel {
+    defaults() {
+        return _.extend(super.defaults(), {
+            _model_name: 'FloatLogSliderModel',
+            _view_name: 'FloatLogSliderView',
+            step: 1.0,
```
jasongrout: This is a different default than the python side.

packages/controls/src/widget_float.ts
```diff
@@ -85,6 +116,111 @@ class FloatSliderView extends IntSliderView {
...
+     */
+    handleTextChange() {
+        let value = this.stringToValue(this.readout.textContent);
+        let vmin = this.model.get('min');
+        let vmax = this.model.get('max');
+
+        if (isNaN(value as number)) {
+            this.readout.textContent = this.valueToString(this.model.get('value'));
+        } else {
+            value = Math.max(Math.min(value as number, vmax), vmin);
```
jasongrout: This is comparing the actual value to vmin and vmax, which are exponents.

packages/controls/src/widget_float.ts
```diff
@@ -85,6 +116,111 @@ class FloatSliderView extends IntSliderView {
...
+     * model to update.
+     */
+    handleSliderChanged(e, ui) {
+        let base = this.model.get('base');
+        let actual_value = Math.pow(base,this._validate_slide_value(ui.value));
+        this.model.set('value', actual_value, {updated_view: this});
+        this.touch();
+    }
+
+        _validate_slide_value(x) {
```
jasongrout: Please dedent this line.

ipywidgets/widgets/widget_float.py
```diff
@@ -141,6 +174,48 @@ class FloatSlider(_BoundedFloat):
...
+    _model_name = Unicode('FloatLogSliderModel').tag(sync=True)
+    step = CFloat(0.1, help="Minimum step in the exponent to increment the value").tag(sync=True)
+    orientation = CaselessStrEnum(values=['horizontal', 'vertical'],
+        default_value='horizontal', help="Vertical or horizontal.").tag(sync=True)
+    readout = Bool(True, help="Display the current value of the slider next to it.").tag(sync=True)
+    readout_format = NumberFormat(
+        '.2f', help="Format for the readout").tag(sync=True)
+    continuous_update = Bool(True, help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
+    disabled = Bool(False, help="Enable or disable user changes").tag(sync=True)
+    base = CFloat(2., help="Base for the logarithm").tag(sync=True)
```
jasongrout: I know we had this discussion before - what do you think about defaulting to base 10? I think that might be more common in everyday use.

packages/controls/src/widget_float.ts
```diff
@@ -85,6 +116,111 @@ class FloatSliderView extends IntSliderView {
...
+     * the value is to be changed
+     *
+     * if any of these conditions are not met, the text is reset
+     */
+    handleTextChange() {
+        let value = this.stringToValue(this.readout.textContent);
+        let vmin = this.model.get('min');
+        let vmax = this.model.get('max');
+
+        if (isNaN(value as number)) {
```
jasongrout: You don't need the `as number` since the `stringToValue` function above is declared as returning a number.

packages/controls/src/widget_float.ts
jasongrout: Also, you don't need the `as number` here either.

ipywidgets/widgets/widget_float.py
jasongrout: If we did change it to base 10 (I think we probably ought to), we probably ought to default the max to something like 4, so we only go up to 10k.

jasongrout:
Thanks! Just a few more comments above, and then I think we are good to go.

sebasguts:
Thanks for the review again :). I adressed all the comments, so I also changed the base to 10 and the default max to 4.

jasongrout:
Thanks! I noticed one last issue in doing final testing:

screen shot 2018-03-21 at 3 38 33 pm

Just invoking with `FloatLogSlider()` (or `FloatLogSlider(value=0)`) doesn't check against the min, apparently, so we get an invalid value (the value is actually 0, it's not just a readout problem).

jasongrout:
I think the easiest way to correct it is to declare the default `value` for a `_BoundedLogFloat` to be one:

```
    value = CFloat(1.0, help="Float value").tag(sync=True)
```

sebasguts:
Thank you for the review again, and thanks for the catch. I actually did not test it without parameters.

I added the default to both sides, and the bug seems to have disappeared.

This one was quite strange, as it only seem to be appeared if `value=0` was passed, `value=-1` for example would have set the initial value to 1. I assume it comes from the logarithm computation on the JavaScript side, but I am not completely sure.

jasongrout:
Looks good. Thanks!

sebasguts:
Thanks for all the useful comments and for merging this PR :).

PR #1929, opened 2018-01-24
Fix the output widget version.

jasongrout:
This is a separate widget with a separate version number. We accidentally tied it to the controls version.

Fixes https://github.com/jupyter-widgets/ipywidgets/issues/1917 and https://github.com/jupyter-widgets/ipywidgets/issues/1912

### DIFF
--- a/ipywidgets/_version.py
+++ b/ipywidgets/_version.py
@@ -10,6 +10,7 @@ __version__ = '%s.%s.%s%s'%(version_info[0], version_info[1], version_info[2],
 
 __protocol_version__ = '2.0.0'
 __jupyter_widgets_base_version__ = '1.0.0'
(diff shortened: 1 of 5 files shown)


blink1073:
Merci beaucoup!

SylvainCorlay:
> Merci beaucoup!

🇫🇷

SylvainCorlay:
The test failure is due to the fact that it is formally a change in the spec...

jasongrout:
I changed the spec back, considering it an error in the spec, since this broke things.

jasongrout:
I'll release a patch version so we get back to a working state...

jasongrout:
patch release done: ipywidgets 7.1.1, widgetsnbextension 3.1.3

PR #1933, opened 2018-01-24
Adding min_gap property to IntRangeSlider

dharmaquark:
first pass at supporting feature request in: 
https://github.com/jupyter-widgets/ipywidgets/issues/1312
definitely open to suggestions here on how to set default/fallback values (especially when validating a new min/max against existing values/min_gap properties, and when validating a changed value against existing min/max/min_gap properties) - if there's a cleaner/more consistent way to handle those cases (or if there's a bug I've missed while testing :) ), I can clean up that logic!
If it makes sense, I can make a similar addition (when it looks clean/good to merge) to the FloatRangeSlider as part of this PR

### DIFF
--- a/ipywidgets/widgets/widget_int.py
+++ b/ipywidgets/widgets/widget_int.py
@@ -15,7 +15,6 @@ from .trait_types import Color, InstanceDict, NumberFormat
 from traitlets import (
     Unicode, CInt, Bool, CaselessStrEnum, Tuple, TraitError, default, validate
 )
-from warnings import warn
 
 _int_doc_t = """
 Parameters
--- a/ipywidgets/widgets/widget_int.py
+++ b/ipywidgets/widgets/widget_int.py
@@ -15,6 +15,7 @@ from .trait_types import Color, InstanceDict, NumberFormat
 from traitlets import (
     Unicode, CInt, Bool, CaselessStrEnum, Tuple, TraitError, default, validate
 )
+from warnings import warn
 
 _int_doc_t = """
 Parameters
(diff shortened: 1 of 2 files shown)


packages/controls/src/widget_int.ts
```diff
@@ -371,6 +380,40 @@ class IntRangeSliderView extends BaseIntSliderView {
      */
     handleSliderChanged(e, ui) {
```
dharmaquark: for when continuous_update is set to true, I think I may want to also add validation in handleSliderChange (right now, with this current logic, sliding is unrestricted in continuous mode, but will 'snap back' to the min_gap once you release the mouse- that might not be the most intuitive approach?)

ipywidgets/widgets/widget_int.py
```diff
@@ -258,13 +259,50 @@ def _validate_bounds(self, proposal):
...
         lower, upper = super(_BoundedIntRange, self)._validate_value(proposal)
         lower, upper = min(lower, self.max), min(upper, self.max)
         lower, upper = max(lower, self.min), max(upper, self.min)
+        if self.min_gap > 0:
+            value_gap = upper - lower
+            if value_gap < self.min_gap:
+                # if value gap is smaller than min_gap, clamp
+                # values within the min-max range of the slider
+                lower = self.min
+                upper = self.max
```
pbugnion: Should we reset to the slider bounds, or should we just go back to the previous value? (I think `proposal` still contains the old `value`?)

pbugnion:
Thanks for submitting this! I made one comment inline.

ipywidgets/widgets/widget_int.py
dharmaquark: ah, thanks, didn't realize proposal had that info- I agree, resetting to the previous value would cleaner here, I'll test/try to make that change!

jasongrout:
@dharmaquark - thanks again, and sorry for the delay in reviewing this.

I was trying to follow the logic in the validators for what to do in cases where there is a conflict. How does this sound for a design? The assumptions here are that the value is the most likely to change, the min/max second, and the min gap is the least likely to change.

* **Setting min gap**: setting a min gap that is too big (either bigger than value range or max range) fails with an error.
* **Setting min/max**: If this makes the max range smaller than min gap, fail with error. Otherwise, adjust the value to make sure the value is within the min/max.
* **Setting the value in python**: If the values are outside the min/max, clip them. Then check to see if the value_range is larger than the min_gap. If it is, great. If not, fail with error.
* **Setting the value in js**: If the values are outside the min/max, clip them (this could happen if setting the values from the readout). Then check to see if the value range is larger than min gap. If it is, great. If not, reset to old value.

Issue #1846, opened 2017-11-30
Output capture decorator


jasongrout:
It's a little awkward making a callback function who's output should be captured:

```python
from ipywidgets import Output, IntSlider, VBox
from IPython.display import display
out = Output()
slider = IntSlider()
def change(change):
    with out:
        print('value: '+change.new)
slider.observe(change, 'value')
display(VBox([slider, out]))
```

Some people probably would prefer to have a decorator that automatically captures output:

```python
from ipywidgets import Output, IntSlider, VBox
from IPython.display import display
out = Output()
slider = IntSlider()

@out.capture() # new decorator - not sure if this is a good name for it...
def change(change):
    print('value: '+change.new)

slider.observe(change, 'value')
display(VBox([slider, out]))
```

jasongrout:
(this would be similar to https://docs.python.org/3/library/contextlib.html#using-a-context-manager-as-a-function-decorator)

jasongrout:
(Note that I don't think we should necessarily use the ContextDecorator, just that that is one way to implement the idea)

hainm:
```python
@out.capture()
def change(change):
    print('value: '+change.new)
```

nice

jasongrout:
Fixed by #1934. Thanks @pbugnion!

PR #1934, opened 2018-01-26
Add @capture decorator to output widget

pbugnion:
This PR adds a `@out.capture` decorator that pushes all the output of a function to an output widget. It addresses issue #1846 .

For instance, 

```py
out = widgets.Output()

@out.capture
def f(arg):
    print(arg)

f('argument')

out.outputs
# => ({'name': 'stdout', 'output_type': 'stream', 'text': 'argument'},)
```

This is useful for capturing output from widget callbacks:
```py
b = widgets.Button()
b.on_click(f)
# everytime someone clicks on `b`, `out.outputs` gets appended to.
```

To do:

 ~~- [x] Documentation on output widget (probably in a separate PR)~~ (moved to issue #1935)
 - [x] Use `functools.wraps` to avoid losing the function name and docstring (https://docs.python.org/3.6/library/functools.html#functools.wraps)
 ~~- [ ] Is there a better name than `out.capture`? (I'm comfortable with `out.capture`).~~
 ~~- [ ] Should current outputs be cleared on entry? (I think that would be more confusing -- we could add a recipe to the documentation for creating an output widget with a clear button).~~

### DIFF
--- a/ipywidgets/widgets/tests/test_widget_output.py
+++ b/ipywidgets/widgets/tests/test_widget_output.py
@@ -5,11 +5,11 @@ from ipywidgets import Output
 
 
 def _make_stream_output(text, name):
-        return {
-            'output_type': 'stream',
-            'name': name,
-            'text': text
-        }
+    return {
+        'output_type': 'stream',
+        'name': name,
+        'text': text
+    }
 
 
 def test_append_stdout():
--- a/ipywidgets/widgets/widget_output.py
+++ b/ipywidgets/widgets/widget_output.py
@@ -45,7 +45,7 @@ class Output(DOMWidget):
         with out:
             print('prints to output widget')
 
-        @out.decorate
+        @out.capture()
         def func():
             print('prints to output widget')
     """
@@ -60,11 +60,36 @@ class Output(DOMWidget):
     outputs = Tuple(help="The output messages synced from the frontend.").tag(sync=True)
 
     def clear_output(self, *pargs, **kwargs):
+        """
+        Clear the content of the output widget.
+
+        Parameters
+        ----------
+
+        wait: bool
+            If True, wait to clear the output until new output is
+            available to replace it. Default: False
+        """
         with self:
             clear_output(*pargs, **kwargs)
 
     def capture(self, clear_output=True, *outer_args, **outer_kwargs):
-        """Decorator to capture the stdout and stderr of a function"""
+        """
+        Decorator to capture the stdout and stderr of a function.
+
+        Parameters
+        ----------
+
+        clear_output: bool
+            If True, clear the content of the output widget at every
+            new function call. Default: True
+
+        wait: bool
+            If True, wait to clear the output until new output is
+            available to replace it. This is only used if clear_output
+            is also True.
+            Default: False
+        """
         def capture_decorator(func):
             @wraps(func)
             def inner(*args, **kwargs):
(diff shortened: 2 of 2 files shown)


maartenbreddels:
I like it :+1:. I think capture is a good name. An alternative would be redirect, coming from redirecting stdout/stderr, but I like it less than capture.

pbugnion:
Thanks!

> I think capture is a good name

I tend to agree. Jason isn't so keen. An alternative could be `slurp`, but it's pretty ugly.

maartenbreddels:
What about making it callable?
```python
@out
def f(arg):
    print(arg)
```

pbugnion:
> What about making it callable?

So replacing `.capture` with `.__call__`? I personally think that's a bit less explicit, but happy to implement it that way if we think that's the right way.

pbugnion:
Another possible question is -- we could clear the current outputs at each new entrance to `func`. This brings it a bit closer to how an output cell works: the user just sees the result of the last execution of `func`.

maartenbreddels:
Indeed, I think the name `capture` is self documenting, maybe `@out.redirect_output` even more. It's longer, but we have a tab key for a reason :).
Now that I think about it, do we want to redirect stderr/stdout to seperate places maybe, how does the output widget do it now? From a user point of view this may seen nice:
 * `@out.redirect_all`
 * `@out.redirect_stderr`
 * `@out.redirect_stdout`

pbugnion:
> how does the output widget do it now

The `outputs` trait is a tuple of each output type -- you get entries for display_data, error, output streams etc:

```
@out.capture
def f():
    """hello"""
    print('hi')
    display(HTML('<p>hello</p>'))
    raise Exception('bla')

f()

# =>

({'name': 'stdout', 'output_type': 'stream', 'text': 'hi\n'},
 {'data': {'text/html': '<p>hello</p>',
   'text/plain': '<IPython.core.display.HTML object>'},
  'metadata': {},
  'output_type': 'display_data'},
 {'ename': 'Exception',
  'evalue': 'bla',
  'output_type': 'error',
  'traceback': ['some long traceback']})
```

Given the number of output types, I'm not sure there'd be a strong use case for redirecting some outputs but not others. That behaviour would also be substantially different to the current context manager.

pbugnion:
I think this is ready for review and merging as is.

jasongrout:
What if we made it `output.capture(clear=True)` for the automatic clearing case? We could even put in the automatic matplotlib magic that `@interact` does.

We could support both `@output.capture` and `@output.capture(clear=True)` by checking the first argument. We could also change things to be `@output.capture()` for the base case. Or we could do something like what @jdemeyer did for interact, and do 

```python
@output.capture.options(clear=True)
def f():
    ...
```

maartenbreddels:
I don't like the `.option`, but the optional keyword seems pretty clean!

jasongrout:
Once we move to python 3, it will be even cleaner since we can insist that clear=True is a keyword argument.

maartenbreddels:
Is it maybe a plan to put in comments in the code, that we can grep to make that easy? `# PY3: ...` orso

pbugnion:
This now supports `clear_output` -- I went for `clear_output` as the argument name, instead of `clear`, to remain consistent with the `clear_output` method name.

capture-decorator-example

jasongrout:
Thanks!

pbugnion:
Great, thank you!

maartenbreddels:
Yeah, great work!

PR #1943, opened 2018-01-30
Use window.requirejs instead of window.require which is overriden by the RTD theme

SylvainCorlay:


vidartf:
So, this should mean we don't need [this helper script](https://github.com/jupyter-widgets/widget-ts-cookiecutter/blob/ed554d3955b8dbc311f0964078b3df5208f111f0/%7B%7Bcookiecutter.github_project_name%7D%7D/docs/source/_static/helper.js) for widget docs, right?

vidartf:
Also, for inter-linking: https://github.com/rtfd/sphinx_rtd_theme/issues/497

maartenbreddels:
Was wondering about that too. @vidartf this issue only showed up using the the rtd theme locally right?

vidartf:
@maartenbreddels yes

SylvainCorlay:
Right, I did not know about this helper.

vidartf:
Temporary workaround together with the code in doc config.

PR #1945, opened 2018-01-31
'new' is not guaranteed to be in change

rmorshea:
With [eventful traits](https://github.com/ipython/traitlets/pull/466) near completion this should be changed due to the fact that the only keys guaranteed to be included in the change dictionary are `'owner'`, `'type'`, and `'name'`.

cc: @SylvainCorlay, @jasongrout


jasongrout:
Thanks!

PR #1961, opened 2018-02-07
Clean up state restore in nbextension manager

### LINKED ISSUE #1875: Refreshing a live notebook with saved state
I'm not sure how big of an issue this is, but looking at the constructor of [WidgetManager](https://github.com/jupyter-widgets/ipywidgets/blob/a9c4069570aa92368f590296569e411d9c49e123/widgetsnbextension/src/manager.js#L55-L129), I notice that:
- If there is saved widget state, `set_state` is called, which checks for live widgets and connects to them etc, and creates any missing widgets. [here](https://github.com/jupyter-widgets/ipywidgets/blob/a9c4069570aa92368f590296569e411d9c49e123/widgetsnbextension/src/manager.js#L67)
- It checks if there are any live widgets on the kernel, and if so connects to them and creates the JS comms and widgets for them. [here](https://github.com/jupyter-widgets/ipywidgets/blob/a9c4069570aa92368f590296569e411d9c49e123/widgetsnbextension/src/manager.js#L79) and [here](https://github.com/jupyter-widgets/ipywidgets/blob/a9c4069570aa92368f590296569e411d9c49e123/widgetsnbextension/src/manager.js#L106)

This means that if you have a notebook with widgets, then save the state, and refresh the page without restarting the kernel, the JS comms and widgets are created twice. For certain widget configurations this can lead to an indeterminate state, as duplicates of widgets/comms are created with the same id (only the last one is registered with the manager). At least AFAICT.

vidartf:
- Restore live widgets from kernel before widgets from the saved state.
- Afterwards, restore any widgets from saved state that would not overwrite a live widget.
- Also adds some helpful comments.

Fixes #1875.

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -328,6 +328,7 @@ abstract class ManagerBase<T> {
         }
 
         let modelPromise = this._make_model(options, serialized_state);
+        // this call needs to happen before the first `await`, see note in `set_state`:
         this.register_model(model_id, modelPromise);
         return await modelPromise;
     }
@@ -425,6 +426,16 @@ abstract class ManagerBase<T> {
         let models = state.state;
         // Recreate all the widget models for the given widget manager state.
         let all_models = this._get_comm_info().then(live_comms => {
+            /* Note: It is currently safe to just loop over the models in any order,
+               given that the following holds (does at the time of writing):
+               1: any call to `new_model` with state registers the model promise (e.g. with `register_model`)
+                  synchronously (before it's first `await` statement).
+               2: any calls to a model constructor or the `set_state` method on a model,
+                  happens asynchronously (in a `then` clause, or after an `await` statement).
+
+              Without these assumptions, one risks trying to set model state with a reference
+              to another model that doesn't exist yet!
+            */
             return Promise.all(Object.keys(models).map(model_id => {
 
                 // First put back the binary buffers
@@ -444,7 +455,7 @@ abstract class ManagerBase<T> {
                     return this._models[model_id].then(model => {
                         // deserialize state
                         return (model.constructor as typeof WidgetModel)._deserialize_state(modelState || {}, this).then(attributes => {
-                            model.set_state(attributes);
+                            model.set_state(attributes);  // case 2
                             return model;
                         });
                     });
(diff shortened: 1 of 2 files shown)


widgetsnbextension/src/manager.js
```diff
@@ -49,6 +49,18 @@ function new_comm(manager, target_name, data, callbacks, metadata, comm_id, buff
...
+function filter_existing_model_state(manager, state) {
+    var models = state.state;
+    models = Object.keys(models)
+          .filter(function(model_id) {
+              return !manager._models[model_id];
+          })
+          .reduce(function(res, model_id) {
+              return Object.assign(res, { [model_id]: models[model_id] });
+          }, {});
+    return Object.assign({}, state, {state: models});
```
jasongrout: Can you use Backbone's _.extend? Object.assign is not supported in IE, and we support IE in the classic notebook, IIRC.

widgetsnbextension/src/manager.js
```diff
@@ -63,24 +75,29 @@ var WidgetManager = function (comm_manager, notebook) {
...
+    //    any live widgets.
+
+    // Register with the comm manager. (1)
     this.comm_manager.register_target(this.comm_target_name, _.bind(this.handle_comm_open,this));
 
-    // Attempt to reconstruct any live comms by requesting them from the back-end.
+    // Attempt to reconstruct any live comms by requesting them from the back-end (2).
     var that = this;
-    var backed_widgets_loaded = this._get_comm_info().then(function(comm_ids) {
+    var backed_widgets_loaded = Promise.resolve().then(function() {
```
jasongrout: I don't think we need to create an extra promise. `_get_comm_info` returns a promise, so that should work fine, right?

vidartf:
Addressed the comments, should be ok!

packages/base/src/manager-base.ts
```diff
@@ -425,6 +426,16 @@ abstract class ManagerBase<T> {
...
+            /* Note: It is currently safe to just loop over the models in any order,
+               given that the following holds (does at the time of writing):
+               1: any call to `new_model` with state registers the model promise (e.g. with `register_model`)
+                  synchronously (before it's first `await` statement).
+               2: any calls to a model constructor or the `set_state` method on a model,
+                  happens asynchronously (in a `then` clause, or after an `await` statement).
+
+              Without these assumptions, one risks trying to set model state with a reference
+              to another model that doesn't exist yet!
+            */
```
SylvainCorlay: Nice explanation.

vidartf:
I'm going to merge this, unless there are any reservations?

SylvainCorlay:
No reservation. I can even take it upon myself to press the green button.

👍 on the very clear inline comments.

PR #1968, opened 2018-02-10
Fixup binary buffer saving

### LINKED ISSUE #1950: widget Image not working in documentation and nbviewer
Issue #1914 tracks the resolution of the rendering of widgets in the documentation.  
Well, *almost* completely.  
The [widget Image](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Image) still does not get rendered. 

It looks very much like the nbviewer behavior.
See an [example of nb containing widget Images](https://nbviewer.jupyter.org/github/oscar6echo/nbviewer-test/blob/master/demo_core_widgets_1.ipynb) rendered on nbviewer.

And it does work even less in nbconvert which still does not support widgets 7.x in general. Cf jupyter/nbconvert#636.

SylvainCorlay:
Fixes #1950

### DIFF
--- a/packages/controls/src/widget_image.ts
+++ b/packages/controls/src/widget_image.ts
@@ -20,9 +20,16 @@ class ImageModel extends CoreDOMWidgetModel {
             format: 'png',
             width: '',
             height: '',
-            value: new Uint8Array(0)
+            value: new DataView(new ArrayBuffer(0))
         });
     }
+
+    static serializers = {
+        ...CoreDOMWidgetModel.serializers,
+        value: {serialize: (value, manager) => {
+            return new DataView(value.buffer.slice(0));
+        }}
+    };
 }
 
 export


SylvainCorlay:
It seems that if the default value for `value` was `new DataView(new ArrayBuffer())` we would not even need to do the test in the new serializer and simply do `return new DataView(value.buffer)`.

packages/controls/src/widget_image.ts
```diff
@@ -23,6 +23,13 @@ class ImageModel extends CoreDOMWidgetModel {
             value: new Uint8Array(0)
         });
     }
+
+    static serializers = {
+        ...CoreDOMWidgetModel.serializers,
+        value: {serialize: (value, manager) => {
+            return new DataView(value instanceof ArrayBuffer ? value : value.buffer);
```
jasongrout: Since we technically need a copy, wouldn't it be better to do a slice? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice

jasongrout:
> It seems that if the default value for value was new DataView(new ArrayBuffer()) we would not even need to do the test in the new serializer and simply do return new DataView(value.buffer).

I'm happy to change the default here.

packages/controls/src/widget_image.ts
SylvainCorlay: Fixed

SylvainCorlay:
OK, now doing `slice(0)` on the buffer.

jasongrout:
Do you want to change the default so we can simplify the code?

SylvainCorlay:
ok

SylvainCorlay:
Done.

SylvainCorlay:
Ready to go. It works for the image.

SylvainCorlay:
@jasongrout merge?

jasongrout:
Looks good to me. Thanks!

PR #1969, opened 2018-02-10
Save notebook widget state to fix Image rendering

SylvainCorlay:


SylvainCorlay:
I could check on nbviewer that it is now fixed.

PR #1976, opened 2018-02-16
Return a promise so that comm message processing is delayed until we are instantiated

jasongrout:
Otherwise, we miss some messages coming from the kernel, because the comm message handler is not registered.


jasongrout:
CC @SylvainCorlay - this seems to take care of the problems we were seeing in jupyterlab missing message updates on occasion.

SylvainCorlay:
Wow, that was a tricky one. I spent 2 hours trying to find this.

Now catching up on the messages on gitter.

jasongrout:
Well, I had the "advantage" of spending far longer than two hours reasoning through this sort of thing in the classic notebook :).

SylvainCorlay:
> Well, I had the "advantage" of spending far longer than two hours reasoning through this sort of thing in the classic notebook :).

Yeah, been there :) thanks for this.

PR #1983, opened 2018-02-25
Extend unit tests for the Output widget

pbugnion:
This builds on PR #1934, and the diff currently includes some commits from that PR.

While working on PR #1934, I felt that the test coverage on the Python side of the output widget was somewhat sparse. The difficulty is that a lot of the output widget code relies on `IPython.display` and `IPython` methods. I've mocked these out in these tests (using manual monkeypatching, rather than introducing a dependency on a mocking library).

* Is this useful? I personally think this will help developers like me who are less familiar with the codebase, but happy to discard this if maintainers think it adds too much overhead.
* Should we write this differently? For instance, we could use a real mocking library rather than my home-grown attempts at one. There isn't (AFAICT) much mocking in unit tests at the moment, so it's probably worth settling on one way of doing things.

### DIFF
--- a/ipywidgets/widgets/tests/test_widget_output.py
+++ b/ipywidgets/widgets/tests/test_widget_output.py
@@ -124,6 +124,25 @@ class TestOutputWidget(TestCase):
         assert clear_output.calls[0] == clear_output.calls[1] == \
             ((), {'wait': True})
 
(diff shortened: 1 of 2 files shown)


jasongrout:
I think this looks great. Thanks!

PR #1984, opened 2018-02-26
Update typings of widget serializers

vidartf:
Otherwise the `DOMWidgetModel` loses it's key-signature typing.


jasongrout:
Thanks!

Issue #1811, opened 2017-11-09
output widget append_display_data does not display widgets


jasongrout:
It seems there are some problems either with the output widget `append_display_data` method (from https://github.com/jupyter-widgets/ipywidgets/pull/1752), or how we display widgets.

```python
from ipywidgets import Output, IntSlider
out = Output(layout={'border': '1px solid black'})
out.append_display_data(IntSlider())
out
```
should capture the intslider inside of the output widget, but instead we get

<img width="574" alt="screen shot 2017-11-08 at 8 56 40 pm" src="https://user-images.githubusercontent.com/192614/32589107-7c2781ba-c4c7-11e7-86a7-b80f48ff4213.png">

If I use the output widget as a context manager, it works fine:
```python
from ipywidgets import Output, IntSlider
from IPython.display import display
out = Output(layout={'border': '1px solid black'})
with out:
    display(IntSlider())
out
```
<img width="533" alt="screen shot 2017-11-08 at 8 58 43 pm" src="https://user-images.githubusercontent.com/192614/32589132-a4c1ae7a-c4c7-11e7-8d2b-1c1ddd8bc43a.png">

jasongrout:
From the docs http://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html:

> Prior to IPython 6.1, _ipython_display_ was the only way to display custom mime-types without registering a new Formatter.
> IPython 6.1 introduces _repr_mimebundle_ for displaying custom mime-types, so _ipython_display_ should only be used for objects that require unusual display patterns, such as multiple display calls.

We use `_ipython_display_`, in which we manually call `display` to display a widget. When we switch to using `_repr_mimebundle`, I think this issue will resolve itself.

jasongrout:
IPython 6.1 was released 31 May 2017. I think it will be a while before we can mandate IPython 6.1+ (likely not until ipywidgets 8), so we may just have to live with this for now.

jasongrout:
actually, we may be able to register a display using the `for_type` mechanism instead, until we can move to a simple `_repr_mimebundle_`: http://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html#IPython.core.formatters.BaseFormatter.for_type

GaiYu0:
Hi, I am experiencing a similar problem. In my case, the slider widget is displayed properly but the output widget cannot be displayed. Is there any temporary solution? 

Thank you very much!

[...]

PR #1986, opened 2018-02-27
Use _repr_mimebundle_ instead of _ipython_display_ message for initial display

SylvainCorlay:
I guess this is up to debate, but using the `execute_result` message instead of `display_data` enables the display of the prompt number like for other types of outputs.

screen shot 2018-02-27 at 8 16 37 pm

cc @Carreau @jasongrout

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -697,7 +697,7 @@ class Widget(LoggingHasTraits):
         """Convert json values to objects."""
         return x
 
-    def _ipython_display_(self, **kwargs):
+    def _repr_mimebundle_(self, **kwargs):
         """Called when `IPython.display.display` is called on the widget."""
         if self._view_name is not None:
 
@@ -715,9 +715,8 @@ class Widget(LoggingHasTraits):
                     'model_id': self._model_id
                 }
             }
-            display(data, raw=True)
-
             self._handle_displayed(**kwargs)
+            return data
 
     def _send(self, msg, buffers=None):
         """Sends a message to the model in the front-end."""


SylvainCorlay:
cc @Carreau @jasongrout

SylvainCorlay:
(Test failures are expected, will fix if we decide to go this way)

SylvainCorlay:
I wonder if there is a means to specify `raw=True`, or if it is even necessary for `display_mime_bundle`. @Carreau would probably know.

jasongrout:
See https://github.com/jupyter-widgets/ipywidgets/issues/1811#issuecomment-343055374 - we can't do this just yet because of backwards compatibility.

However, see https://github.com/jupyter-widgets/ipywidgets/issues/1811#issuecomment-343056524 for a workaround that might work until we can assume ipython 6.1.

jasongrout:
(fixed the description to match the code :).

jasongrout:
(And yes, I think we should definitely move to `_repr_mimebundle_` when we can)

SylvainCorlay:
OK, I am working on the rich display mechanism for `xeus-cling` and I was wondering why widgets never used the `execute_result` message. Thanks for the pointer.

The `xeus-cling` rich display will only have the counterpart to `_repr_mime_bundle_` from the start.

jasongrout:
> The xeus-cling rich display will only have the counterpart to _repr_mime_bundle_ from the start for all objects that have a rich representation.

Great!

jasongrout:
> However, see #1811 (comment) for a workaround that might work until we can assume ipython 6.1.

This won't work after all.

> 

> (And yes, I think we should definitely move to _repr_mimebundle_ when we can)

Likely in ipywidgets 8.

jasongrout:
This PR is superseded by #2021

PR #1993, opened 2018-03-08
Add utility function resolveUrl on manager

vidartf:
This is needed to be able to resolve file URLs relative to the widget's notebook.

Example: My widget needs to load an image at the path 'img/earth.jpg'. This is implied to be relative to the path of the current notebook (I cannot see a case where this wouldn't be what the user would expect, but I'm sure someone will come along at some point). In classic notebook, this works out of the box, since the url always incorporates the location of the notebook. For jupyter lab, this is not the case. By adding a utility function on the widget manger, we add a frontend-agnostic way of resolving the URL.

### DIFF
--- a/packages/base/src/manager-base.ts
+++ b/packages/base/src/manager-base.ts
@@ -481,6 +481,16 @@ abstract class ManagerBase<T> {
             this._models[i].then(model => { model.comm_live = false; });
         });
     }
+
+    /**
+     * Resolve a URL relative to the current notebook location.
+     *
+     * The default implementation just returns the original url.
+     */
+    resolveUrl(url: string): Promise<string> {
+        return Promise.resolve(url);
+    }
+
     /**
      * The comm target name to register
      */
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
Thanks @vidartf. This should solve https://github.com/ellisonbg/ipyleaflet/issues/120, which is exactly what you described.

maartenbreddels:
@astrofrog I think you also needed this a while back, and I gave you a dirty hack for it, this looks much cleaner

jasongrout:
Thanks!

vidartf:
Should this have been using `getDownloadUrl` instead of `resolveUrl` on `UrlResolver`?

vidartf:
Mainly to be future proof in case anybody adds another Drive with an overridden [getDownloadUrl](https://github.com/jupyterlab/jupyterlab/blob/793dd5ff16ced7329604b9ce13ed3413f32f1824/packages/services/src/contents/index.ts#L1031).

jasongrout:
Good question. Can you open another issue so we can discuss, and possibly target to a 7.2.1?

PR #1995, opened 2018-03-13
if moduleName contains a slash, use a different module/path

maartenbreddels:
This is useful if a package contains multiple widget extensions (I use this for ipysheet).
For instance, use:
`_model_module = Unicode('ipysheet/renderer').tag(sync=True)`
The embed manager will try got get `ipysheet/renderer.js` from the server its running on (which is a valid request, and useful for local testing), if that fails it will fetch:
`https://unpkg.com/ipysheet@semver/dist/renderer.js`

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -19,20 +19,31 @@ let requirePromise = function(pkg: string | string[]): Promise<any> {
     });
 }
 
+function moduleNameToCDNUrl(moduleName: string, moduleVersion: string) {
+    let packageName = moduleName;
+    let fileName = 'index'; // default filename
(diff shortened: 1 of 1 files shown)


vidartf:
Does this need to special case "@my-namespace/mymodule" ?

maartenbreddels:
Indeed, good call. Do you see any other issues with this?

jasongrout:
The only other issue I see with this is that it still inserts the `dist/` - is there a way we don't have to hardcode that portion either? I don't see a way to have it nicely support both the other widget managers, but perhaps you have an idea?

jasongrout:
> I think the 'dist/' part is ok, I see it as a manager specific implementation.

Oh right - it is very specific to how the html manager is fetching things.

jasongrout:
Looks good to me, then. Thanks!

PR #1998, opened 2018-03-17
Allow requireLoader to be configurable

Madhu94:
I'd really like to be able to configure where I look up the widget models and views classes, and not always pick up from unpkg (for inhouse software).

I didn't know if there was a better way to do it, but allowing a configurable loader for htmlManager might do it.

Please let me know if this can be merged or if there is a better solution to my problem

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -52,12 +52,14 @@ function requireLoader(moduleName: string, moduleVersion: string) {
  * Render widgets in a given element.
  *
  * @param element (default document.documentElement) The element containing widget state and views.
+ * @param loader (default requireLoader) The function used to look up the modules containing
+ * the widgets' models and views classes. (The default loader looks them up on unpkg.com) 
  */
 export
-function renderWidgets(element = document.documentElement) {
+function renderWidgets(element = document.documentElement, loader: (moduleName: string, moduleVersion: string) => Promise<any>  = requireLoader) {
     requirePromise(['@jupyter-widgets/html-manager']).then((htmlmanager) => {
         let managerFactory = () => {
-            return new htmlmanager.HTMLManager({loader: requireLoader});
+            return new htmlmanager.HTMLManager({loader: loader});
         }
         libembed.renderWidgets(managerFactory, element);
     });


jasongrout:
Great idea! Can you do two changes to your PR?

1. The type for the loader should be more specific: `(moduleName: string, moduleVersion: string) => Promise<any>`

2. Can you push the default into the signature? Something like:

```typescript
function renderWidgets(element = document.documentElement, loader: (moduleName: string, moduleVersion: string) => Promise<any> = requireLoader) {
```

Madhu94:
Thanks for the feedback ! I improved the type declaration.

jasongrout:
Now with the default in the signature, you don't need the `||`.

Madhu94:
I'm sorry for the back and forth here.

jasongrout:
No problem at all! You're doing great!

Madhu94:
Squashed the commits, let me know if anything else is needed.

jasongrout:
One more thing - can you add a note to the function documentation about the loader, similar to the note for the element parameter? Then I think we're good to go.

jasongrout:
Looks good, thanks! I'll wait for tests to pass before merging.

jasongrout:
Thanks again!

jasongrout:
@Madhu94 - one more thing. In our release notes, we give credit to people who contributed, using their full name (see https://groups.google.com/forum/#!msg/jupyter/j40P78_h3U4/K3vdg5VGCgAJ for example). Your commit right now shows that it is from `nmadhum <nmadhum@deshaw.com>`. If you would like your proper name to appear in the release notes, could you add an entry to the .mailmap file (https://github.com/jupyter-widgets/ipywidgets/blob/master/.mailmap)? You would add a line in alphabetical order that looks like:

```
FirstName LastName <email> nmadhum <nmadhum@deshaw.com>
```
Thanks!

Or if you just want to post your name here, I can add the listing. Of course, if you'd rather be mentioned just using your github username, that's fine too, and we won't add an entry to the mailmap.

And for the future, you can set your name in git so that it shows up in your git commits following the directions at https://help.github.com/articles/setting-your-username-in-git/.

Madhu94:
Thanks for your help @jasongrout, hopefully my next PR will go better as I _won't_ be a Typescript novice by then.

PR #2003, opened 2018-03-20
Fix URL resolver compilation.

jasongrout:
I had an error trying to compile:

```
> @jupyter-widgets/jupyterlab-manager@0.33.2 build:src [snip]/ipywidgets/packages/jupyterlab-manager
> tsc --project src

src/manager.ts(163,25): error TS2339: Property 'resolveUrl' does not exist on type 'IContext<IModel>'.
```
This fixes it.


vidartf:
Yup, the change happened in https://github.com/jupyterlab/jupyterlab/commit/970b57908750514b3f93407aa6c5aac5547dbdb6 . I guess I was working in an outdated dev env.

PR #2004, opened 2018-03-20
Update .mailmap

Madhu94:


jasongrout:
Thanks! The line needs to read `First Last <email> Commit Name <commitemail>` (see https://git-scm.com/docs/git-shortlog#_mapping_authors), so it should be:

```
Madhumitha Natarajan <galt4000@gmail.com> nmadhum <nmadhum@deshaw.com>
```

jasongrout:
Great, thanks!

PR #2005, opened 2018-03-20
Fix Edge select box styling to show selected items

### LINKED ISSUE #1996: Selectmultiple display issue on Microsoft Edge
Hi all, 

I have an issue with using the Selectmultiple ipywidget on Microsoft Edge. When I select items of the Selectmultiple widget I can't see the selection, altough it is selected. Unfortunately I couldn't find any solutions on the internet so far. Has anyone the same issue?

Details:
Microsoft Edge 41.16299.248.0
Microsoft EdgeHTML 16.16299
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:54:25) 
Jupyter  4.4.1
ipywidget 7.0.5

jasongrout:
Fixes #1996

Apparently Edge is the only browser that was paying attention to the checked option’s color and background - no other browser seemed to be affected by this.

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -730,11 +730,6 @@
     border-color: var(--jp-widgets-input-focus-border-color);
 }
 
-.widget-select > select > option:checked {
-    color: var(--jp-ui-font-color0);
-    background: var(--jp-layout-color0) repeat url("data:image/gif;base64,R0lGO...");
-}
-
 .wiget-select > select > option {
     padding-left: var(--jp-widgets-input-padding);
     line-height: var(--jp-widgets-inline-height);
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -693,11 +693,6 @@
     text-shadow: 0 0 0 #000;
 }
 
-.widget-dropdown > select > option:checked {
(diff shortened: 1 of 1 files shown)


vidartf:
Looking at the code that is being removed, the base64 data `url("data:image/gif;base64,R0lGO...")` has an ellipsis, and `R0lGO` just decodes to `GIF8`. Is this some CSS trick?

maartenbreddels:
I think it's a copy paste from something that a browser shortened

vidartf:
> I think it's a copy paste from something that a browser shortened

If so, could this be the reason why it was broken? Or why it only seemed to affect Edge? What was the original intended content of this data?

jasongrout:
My guess is that this logic came from when we had our own custom dropdown control. It was then copied down to the select. It seems that FF, Chrome, and Safari ignore the background for checked options in select boxes, and only Edge actually uses it.

jasongrout:
> I think it's a copy paste from something that a browser shortened

Very likely


> What was the original intended content of this data?

I added it at https://github.com/jasongrout/ipywidgets/commit/b6dc12e5171f7ae5171947650f485b3b020e5aca#diff-d360df7fecb4e07cfe3887abeda7dd15R675 - I have no idea at this point what the intended background was.

jasongrout:
Any objection to merging this?

PR #2007, opened 2018-03-20
Delete html fallback message.

### LINKED ISSUE #1777: mathjax affects dropdown options inconsistently
I noticed some strange behavior where dropdown options get rendered (I think) as equations by mathjax, but only when multiple dropdowns are displayed at the same time.

screen shot 2017-10-25 at 11 13 53 am

As a workaround, is there a way to disable MathJax for a single dropdown?

versions:

* notebook==5.1.0
* ipywidgets==7.0.3
* Python 3.6.2 :: Continuum Analytics, Inc.
* Chrome on MacOS

### LINKED ISSUE #1951: Hide Jupyter "Failed to display Jupyter Widget" on Github
Hello There!

Posting my notebook to GitHub adds:
> Failed to display Jupyter Widget of type HBox.
> 
> If you're reading this message in the Jupyter Notebook or JupyterLab Notebook, it may mean that the widgets JavaScript is still loading. If this message persists, it likely means that the widgets JavaScript library is either not installed or not enabled. See the Jupyter Widgets Documentation for setup instructions.
> 
> If you're reading this message in another frontend (for example, a static rendering on GitHub or NBViewer), it may mean that your frontend doesn't currently support widgets.

To every cell that I use `tqdm_notebook` with. I would like to hide that warning when posted on Github. It's a long warning that takes away from the content of the notebook!

I tried to follow this StackOverflow post: https://stackoverflow.com/questions/9031783/hide-all-warnings-in-ipython 

But alas, I was unable to remove the warning!

Thanks!

jasongrout:
This reverts #1674, since we discovered some problems with the classic notebook and html outputs.

Fixes #1777
Fixes #1951

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -708,7 +708,6 @@ class Widget(LoggingHasTraits):
             # http://www.iana.org/assignments/media-types/media-types.xhtml.
             data = {
                 'text/plain': repr(self),
-                'text/html': self._fallback_html(),
                 'application/vnd.jupyter.widget-view+json': {
                     'version_major': 2,
                     'version_minor': 0,
@@ -749,23 +748,3 @@ class Widget(LoggingHasTraits):
             for key in keys
         )
         return '%s(%s)' % (class_name, signature)
-
-    def _fallback_html(self):
-        return _FALLBACK_HTML_TEMPLATE.format(widget_type=type(self).__name__)
-
-
-_FALLBACK_HTML_TEMPLATE = """\
-<p>Failed to display Jupyter Widget of type <code>{widget_type}</code>.</p>
-<p>
-  If you're reading this message in the Jupyter Notebook or JupyterLab Notebook, it may mean
-  that the widgets JavaScript is still loading. If this message persists, it
-  likely means that the widgets JavaScript library is either not installed or
-  not enabled. See the <a href="https://ipywidgets.readthedocs.io/en/stable/user_install.html">Jupyter
-  Widgets Documentation</a> for setup instructions.
-</p>
-<p>
-  If you're reading this message in another frontend (for example, a static
-  rendering on GitHub or <a href="https://nbviewer.jupyter.org/">NBViewer</a>),
-  it may mean that your frontend doesn't currently support widgets.
-</p>
-"""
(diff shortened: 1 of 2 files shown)


jasongrout:
I'll also add a docs page explaining what to do if you see this, but were hoping to see widgets, basically containing the helpful text here.

vidartf:
Note: The `repr` in the plain text mimetype isn't always nice and clean (it can become quite large for complex widgets). As such, I'm not sure if this is a good fix for #1951.

jasongrout:
> Note: The repr in the plain text mimetype isn't always nice and clean (it can become quite large for complex widgets). As such, I'm not sure if this is a good fix for #1951.

Good point. Can we try it out and see? At least it's informative of what you're missing, instead of boilerplate text that gets old after a while.

vidartf:
> Can we try it out and see?

If this is only intended as a stop-gap measure until #1777 / jupyter/notebook#2980 gets fixed, then sure. My preferred long-term solution would be a short HTML entry with a link to an entry in the docs describing the possible causes of seeing the message.

Use case where the repr is longer than the current message: pythreejs, e.g. https://github.com/jovyan/pythreejs/blob/master/examples/Animation.ipynb

ipywidgets/widgets/widget.py
```diff
@@ -708,7 +708,6 @@ def _ipython_display_(self, **kwargs):
             # http://www.iana.org/assignments/media-types/media-types.xhtml.
             data = {
                 'text/plain': repr(self),
```
mwcraig: If we want to ensure this never gets too big, how about `repr(self)[:160]` (or some other appropriate number).

vidartf: I think it is nice to have the full repr available, I'm just arguing whether that should be the fallback format or not.

jasongrout: Since you can always do `repr()` to get the full repr, I think chopping it off (and appending an ellipsis) is a good compromise for the fallback.

jasongrout:
> My preferred long-term solution would be a short HTML entry with a link to an entry in the docs describing the possible causes of seeing the message.

+1

jasongrout:
I just elided anything over length 110 characters - that's enough to fit on one line in the classic notebook. I think it looks all right for a fallback. What do you think?

screen shot 2018-03-23 at 12 47 27 pm

mwcraig:
LGTM

jasongrout:
@vidartf?

vidartf:
As long as we agree on the long term solution, I'm happy with the suggested solution.

jasongrout:
Great, thanks!

PR #2008, opened 2018-03-21
Add a small changelog for 7.2

jasongrout:


mwcraig:
I gather there are no more commits coming for 7.2? Took a look at this last week and it was good; was holding off on the review until just before release.

jasongrout:
> I gather there are no more commits coming for 7.2? Took a look at this last week and it was good; was holding off on the review until just before release.

I don't have anything else in the queue. I bumped a few things to 7.3 that didn't look like they were close to being resolved. Does this look good? If not, we can follow-up with another PR.

(I did change the changelog a bit after this PR, so look at the master version first...)

mwcraig:
LGTM!

PR #2001, opened 2018-03-19
Update flex-grow example

DougRzz:
Flex grow example has been corrected.  Unfortunately I don't know how to embed the widgets so they are rendered in Github.

PR #2010, opened 2018-03-22
Updated flex-grow example

jasongrout:
Rebase of #2001, stripped down to just the changes in the notebook inputs.

CC @DougRzz.


jasongrout:
@DougRzz - please let us know if there is anything in #2001 that didn't make it into this change. Thanks again!

Issue #1632, opened 2017-08-15
Should "Save Widgets" be a checkbox?


dsblank:
Issue started here: https://github.com/jupyter/notebook/issues/2735

jasongrout:
Thanks - this makes a lot of sense if there are the appropriate notebook hooks.

Madhu94:
I'll look into this.

Madhu94:
I'll take this up.

jasongrout:
Thanks!

oscar6echo:
As requested by @Madhu94 let here is a summary of my suggestion in #1952:

It think the notebook regular button should perform these operations in sequence:
+ clear all past widget states (to avoid piling them up)
+ save current widget state (if any)
+ save notebook

so that widgets
+  are a natural part of the notebook experience to the regular user.
+ can be saved through nbconvert - important to users running/saving/converting notebooks in batch as part of a publishing pipeline.

Madhu94:
Would it be useful to have this information persisted in the user's config files?

oscar6echo:
I am not sure what you mean exactly by user's config file.  
But my initial idea would be to have the info (widget state I suppose you mean) in the notebook itself.  
If this is the case well nothing more is required, right ? Or did I miss something ?

Madhu94:
Oh sorry, I was referring to the addition of a checkbox to save widget state by default. When the user makes a choice to leave this checked or not, should we save this choice somewhere (like in `jupyter_notebook_config.py` ? Or the metadata of the notebook might do

oscar6echo:
Ah ok ! Then I think yes.  
Indeed if the `jupyter_notebook_config.py` should have a variable that determines whether the save widget box is checked upon start. That would be very convenient.

oscar6echo:
Cf. issue jupyter/nbconvert#751 as several people (incl. me) would like to be able to save widget state from nbconvert.

jasongrout:
It looks like there are appropriate hooks in the classic notebook:

https://github.com/jupyter/notebook/blob/2aac713937c0487dd2193c85a69f02e9556421fe/notebook/static/notebook/js/notebook.js#L2729

There is an event that the notebook triggers before saving - perhaps in the widgets menu we have a setting (i.e., menu item with toggle checkmark) that automatically saves the widget state when you save the notebook.

This goes along with #2012.

Juanlu001:
[...]

PR #2012, opened 2018-03-22
Update widget menu items in classic Notebook

### LINKED ISSUE #1866: Clarify how to fully clean widget state when updating versions
I have some example notebooks in my custom widget repository that I want to update to use the latest version of the JS package whenever I release a new version, but I find it difficult to get it right on first try and often end up with old widget states still in my notebooks.

From experimenting I think the required steps are:
1. Open notebook
2. Clear cell output
3. Clear widget state
4. Refresh page
5. Execute cells
6. Save widget state

(this seems consistent with [this comment](https://github.com/jupyter-widgets/ipywidgets/issues/1692#issuecomment-326632419))

Step 4 seems to be required -- perhaps some widget state persists in the client-side JS even after running "clear widget state". Is this intended? If so, can it be made clearer that this step is necessary, because running "clear widget state" sounds like it should be enough?

I wasn't sure what minimal example would help but I can try to put one together if needed.

jasongrout:
Fixes #1866

This changes the menu items to:

```
Close All Widgets
----------------
Save Widgets to Notebook
----------------
Download Widget State
Export Widgets to HTML
```

### DIFF
--- a/widgetsnbextension/src/embed_widgets.js
+++ b/widgetsnbextension/src/embed_widgets.js
@@ -56,7 +56,7 @@ data,
 
                 var mod = dialog.modal({
                     show: true,
-                    title: 'Embed widgets',
+                    title: 'Export Widgets to HTML',
                     body: content,
                     keyboard_manager: Jupyter.notebook.keyboard_manager,
                     notebook: Jupyter.notebook,
--- a/widgetsnbextension/src/manager.js
+++ b/widgetsnbextension/src/manager.js
@@ -206,16 +206,13 @@ WidgetManager.prototype._init_actions = function() {
     };
     Jupyter.menubar.actions.register(this.saveWidgetsAction, 'save-with-widgets', 'widgets');
 
-    this.clearWidgetsAction = {
+    this.closeAllWidgetsAction = {
         handler: (function() {
-            delete Jupyter.notebook.metadata.widgets;
-            Jupyter.menubar.actions.get('jupyter-notebook:save-notebook').handler({
-                notebook: Jupyter.notebook
-            });
-        }),
-        help: 'Clear the widget state information from the notebook'
+            this.clear_state();
+        }).bind(this),
+        help: 'Close all widgets'
     };
-    Jupyter.menubar.actions.register(this.saveWidgetsAction, 'save-clear-widgets', 'widgets');
+    Jupyter.menubar.actions.register(this.closeAllWidgetsAction, 'close-all-widgets', 'widgets');
 };
 
 /**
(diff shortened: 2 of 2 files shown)


jasongrout:
CC @ricklupton

jasongrout:
> close the views.

The views should be deleted when the are removed from the page. We *don't* want to close the widget (i.e., the model) since we may display it again.

jasongrout:
> The widget manager should listen to the clear_all_outputs event and

Clearing all output does not generate an event, so that would need to be added to the notebook first:

https://github.com/jupyter/notebook/blob/2aac713937c0487dd2193c85a69f02e9556421fe/notebook/static/notebook/js/notebook.js#L2028

vidartf:
> We don't want to close the widget (i.e., the model) since we may display it again.

What about for clearing all outputs and restarting the kernel? Is this a separate event?

jasongrout:
Restart and clear all also doesn't have an event: https://github.com/jupyter/notebook/blob/2aac713937c0487dd2193c85a69f02e9556421fe/notebook/static/notebook/js/notebook.js#L2290

jasongrout:
So...moving forward with this - is this PR enough of an improvement to merge, or does it muddle things further and should be reconsidered before merging?

jasongrout:
Actually, let's take a step back and see if we also want to implement #1632 here too.

It looks like there are appropriate hooks in the classic notebook:

https://github.com/jupyter/notebook/blob/2aac713937c0487dd2193c85a69f02e9556421fe/notebook/static/notebook/js/notebook.js#L2729

There is an event that the notebook triggers before saving - perhaps in the widgets menu we have a setting (i.e., menu item with toggle checkmark) that automatically saves the widget state when you save the notebook.

jasongrout:
(I'm not sure if the jlab notebook has such hooks, but we don't have a widget menu in jlab anyway)

jasongrout:
Okay everyone, how about this?

```
Close All Widgets
----------------
[ ] Save Live Widgets when Saving Notebook
[ ] Save All Widgets when Saving Notebook
----------------
Download Widget State
Export Widgets to HTML
```
The two Save options make sure that at most one is selected between them. The setting is saved with the notebook. Saving *all* widgets saves all widget state with the notebook (including possibly widgets from previous kernels or state that was loaded from the notebook). 

Toggling this option would *not* also save the notebook immediately, but would just set up things so that when the notebook *is* saved, the state is saved.

ricklupton:
Hi Jason, thanks for keeping on with this!

I like the idea of having a toggle to save widgets along with the notebook, but I'm not sure about the live/all distinction.

Does "live" mean "model connected to a kernel" or "model connected to a visible view"? I'm assuming the first in what I write below.

In my case I think I would have to frequently toggle between the two.

1. I open a notebook with save widgets to fix a typo in a markdown cell. I don't want to rerun everything so I have to remember to switch the toggle to "save all", as there are no live widgets, otherwise I'll accidentally end up with a notebook without widgets -- I guess even if it is autosaved.

2. I have a new version of the widget that I want to update the notebook to use. I have to either "clear all" before starting (which is fine) or remember to switch the toggle to "save live", otherwise I'll end up with outdated state in the notebook.

So in this case it would be best to always leave it as "save all". But maybe there are other uses when you do want to "save live" every time you open the notebook?

jasongrout:
> So in this case it would be best to always leave it as "save all". But maybe there are other uses when you do want to "save live" every time you open the notebook?

I think if your typical workflow is to open a notebook from scratch, run the cells, and save, the "save live" version is what you would usually want - it erases the previous state, and you don't have to remember to run "clear all".

vidartf:
> I open a notebook with save widgets to fix a typo in a markdown cell. [...]

When opening a notebook with stored widgets, it recreates the widgets (i.e. opens them in the kernel), so these will be live. One could possibly imagine a race condition here, where the widgets are slow to re-open, and save request happens quickly. We would have to make sure that this was handled correctly (wait for promises to resolve, then save?) or warned about.

jasongrout:
It sounds like we need to continue the discussion here, so bumping to 7.3.

thomasaarholt:
Since it hasn't been explicitly made clear here, I'd like to add the desire for a default preference setting for whether or not to save the workbooks. 

Motivation: I generate a new collection of workbooks with each simulation experiment I do, and it's very useful to be able to look up the the results of analysis within the workbook without having to rerun the code. Currently I am often forgetting to press the save widget state menu item.

SylvainCorlay:
@jasongrout do you want to get this in for 7.5?

jasongrout:
> @jasongrout do you want to get this in for 7.5?

No.

jasongrout:
Since this hasn't been worked on in 2 years, I'm closing this. If someone wants to push it forward, please add a note here and open a new PR.

PR #2014, opened 2018-03-23
Change float log slider default format to .3g

jasongrout:
After playing with the float slider for a bit, I realized it is very easy to get very small or large orders of magnitude (duh!). These formats are much more readable using .3g rather than .2f as the default format.

Also, I added it to the widget list doc page.

CC @sebasguts - what do you think about changing the default format? For example, `FloatLogSlider(min=-10, max=10, readout_format='.2f')` has a huge number of values that are either difficult to understand (at the max) or just plain not visible (at the min). Changing the format to .3f makes both cases much easier to understand.

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -40,7 +40,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


sebasguts:
I like this a lot. Specially with `orientation=vertical`, using `.2f` does not completely display the values in that setting, whereas `.3g` does.

jasongrout:
Great, thanks!

Issue #2006, opened 2018-03-20
Installing the ipywidgets for jupyterlab breaks


AbdealiJK:
We are using `ipywidgets` 7.0.0 in our systems and are using `jupyter-widgets/jupyterlab-manager` v0.27.0 with `jupyterlab` 0.27.0

It looks like there was a `jupyter-widgets/base` v1.1.8 which is incompatible with these but the previous `jupyer-widgets/jupyterlab-manager` is using ^1.0.1 implying any system which is trying to use `jupyter-widgets/jupyterlab-manager` will now build with `jupyter-widgets/base` v1.1.8 and will be broken.

Seems to us like the `jupyter-widgets/base` which seems to be backward incompatible has been released as a 1.x series and braking other packages - as this is backward incompatible, it should have been a 2.x series ?

jasongrout:
Do you happen to know what the incompatibility was?

AbdealiJK:
Some issue with the model. Not sure on what specifically it was. 

https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/base/src/manager-base.ts#L347

"Could not instantiate widget" is the error I get. 

Note that if I update ipywidgets to 7.1.2 - it gets fixed. 

The combination that worked a few months ago - ipywidgets==7.0.0 and jupyter-widgets/jupyterlab-manager@0.27.0 doesn't seem to work now

jasongrout:
Was there another error shown? It looks like it prints out that, and then throws the real error.

jasongrout:
I can't even get jupyterlab 0.27 running - apparently it depends on the latest jupyterlab_launcher, which is maybe incompatible?

AbdealiJK:
Hm, I'm not sure on this. We have a fairly old installation of jupyterlab from last September

 @rsinha25 - any comments?

jasongrout:
(apparently it depends on jupyterlab_launcher 0.4.0 - explicitly installing that makes jlab 0.27.0 work).

jasongrout:
I just duplicated the error. I see:

> Module @jupyter-widgets/controls, semver range 1.0.0 is not registered as a widget module

This comes because the controls package was upgraded to be protocol version 1.1.0, but the python side still is looking for 1.0.0 (see https://github.com/jupyter-widgets/ipywidgets/blob/b41d4dc189e6f5d21a20863384bfc879273326b6/packages/controls/src/version.ts#L9). There's a couple of ways we could solve this:

[...]

PR #2017, opened 2018-03-23
Bump controls version to 1.2.0

jasongrout:
We do this now so we can test the backwards compatibility before the release.

### DIFF
--- a/ipywidgets/_version.py
+++ b/ipywidgets/_version.py
@@ -11,7 +11,7 @@ __version__ = '%s.%s.%s%s'%(version_info[0], version_info[1], version_info[2],
 __protocol_version__ = '2.0.0'
 __jupyter_widgets_base_version__ = '1.0.0'
 __jupyter_widgets_output_version__ = '1.0.0'
-__jupyter_widgets_controls_version__ = '1.1.0'
+__jupyter_widgets_controls_version__ = '1.2.0'
 
 # A compatible @jupyter-widgets/html-manager npm package semver range
 __html_manager_version__ = '^0.11.0'
--- a/packages/controls/src/version.ts
+++ b/packages/controls/src/version.ts
@@ -6,4 +6,4 @@
  * implements.
  */
 export
-const JUPYTER_CONTROLS_VERSION = '1.1.0';
+const JUPYTER_CONTROLS_VERSION = '1.2.0';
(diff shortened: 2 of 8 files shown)


jasongrout:
I just changed the logic of the JupyterLab manager - if a specific module *version* is requested (not a semver range), then it defaults to the `^` semver range to get a compatible module.

This would fix the problem we see in #2006, for example. If someone wants to insist on a specific version, they should use the range specifier `=1.2.3`.

@SylvainCorlay - thoughts? This does change the behavior for plain versions, but I think it makes the behavior more intuitive and better addresses the various places we need version numbers. I think it also probably makes sense to make the same change in the html manager?

jasongrout:
ping @SylvainCorlay

vidartf:
For future reference it would be nice if this had some tests outlining basic expected behavior. A side effect of this is that it is impossible for packages to lock one version to another (or that would require a one-element range).

jasongrout:
> A side effect of this is that it is impossible for packages to lock one version to another (or that would require a one-element range).

Yes, it would require a semver range `=1.2.3`

jasongrout:
(If this is too big of a change for a minor release, we can probably hold off on something like this until 8.0)

jasongrout:
> (If this is too big of a change for a minor release, we can probably hold off on something like this until 8.0)

Since this *is* technically a backwards-incompatible change in general, I special-cased the changes to just the Jupyter-widgets/[base,controls] packages. We can take up the issue of making the more general change for 8.0 later.

jasongrout:
> For future reference it would be nice if this had some tests outlining basic expected behavior. A side effect of this is that it is impossible for packages to lock one version to another (or that would require a one-element range).

The last commits roll back the behavior for general packages.

nscozzaro:
I'm trying to get ipywidgets to work in my JupyterLab extension but I'm getting an error related to this, which I don't understand: "Error: Module @jupyter-widgets/base, semver range ^1.2.0 is not registered as a widget module", and many other similar errors. 

What I did was to copy all of the code from the jupyterlab-manager extension into my extension, changed the INotebookTracker to my modified INotebookTracker, and now I'm getting this error. Would very much appreciate any guidance as to how I might solve this problem. Thanks


image

image

PR #2019, opened 2018-03-26
Added utf-8 header to fix issue with unicode character

fvinas:
Hi,

Added missing UTF-8 encoding header to `widget.py` since unicode character `'…'` was raising an issue.

FV


jasongrout:
Thanks for catching this! Do you think it would be better to just use the unicode escape sequence, rather than the actual unicode character?

jasongrout:
Thanks!

Issue #1935, opened 2018-01-26
Centralise documentation for the output widget


pbugnion:
Currently, documentation for the output widget is scattered across:
 - [async widgets](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Asynchronous.html#Interacting-with-output-widgets-from-background-threads)
 - [widget list](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20List.html#Output)

With PR #1934 , we need to extend the docs for the output widget to expose its use as a decorator. 

Given the output widget's power, we think it makes sense to bring docs on the output widget to the top level. I think it would be useful to structure the notebook as:

 - introduction (what's currently in the widget list)
 - displaying rich output (e.g., with an example from matplotlib?)
 - async widgets
 - using the `@capture` decorator, including as a way to debug widget callbacks.

Happy to tackle this today / this weekend, unless someone else is particularly keen to.

---

Jason deserves the credit for the idea behind this issue.

jasongrout:
Definitely +1. I started an output widget notebook a while ago, but I'm not sure I pushed it anywhere.

jasongrout:
Here is the output notebook I started working on a while ago. Feel free to use it as a start, or feel free to discard it and do what you think is best: https://gist.github.com/anonymous/eb176705a103517179b4f57b6201df1d

jasongrout:
#1936 also has some interesting material that could be incorporated in an output notebook.

jasongrout:
Fixed by #2020. Thanks again!

PR #2020, opened 2018-03-26
Centralize output widget docs

pbugnion:
Addresses issue #1935.

The output widget is powerful since it allows leveraging Jupyter's rich display mechanism. Documentation and recipes for the output widget were scattered across many places in the documentation. This PR tries to centralise mentions of the output widget, and add other contributions previously found in issues and gists.

In particular, we create a new example notebook with:
 - What was previously under the output widget in [widget list](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Output)
 - The [section](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Asynchronous.html#Interacting-with-output-widgets-from-background-threads) on updating output widgets asynchronously
 - This [recipe](https://github.com/jupyter-widgets/ipywidgets/issues/1936) for creating a logging handler that writes to the output widget
 - This [gist](https://gist.github.com/anonymous/eb176705a103517179b4f57b6201df1d) with information on clearing the widget output, appending to the widget directly and controlling the output of `interact`.

----
Left to do:

- [x] it looks like there is an extra left margin in your screenshot. Is the output widget not deleting that left margin? (this seems to predate this PR -- I've opened issue #2025 to fix it).
- [x] The background thread issue is a bit more complicated - threaded output in the ipython kernel isn't cleanly supported, since it's not just scoped to a thread, but also scoped to the execution in the main thread. In other words, if you start a background thread, the background thread output goes into whatever output the main thread happens to be sending things to. What the output widget does is offer a way to redirect that background threaded output to somewhere we choose. We can't rely on the display() mechanism for this since it suffers the problem above. To get around this, we append directly to the output widget, bypassing the ipython display mechanism.
- [x] in debugging callbacks, "An effective tool for debugging this" - it's not clear what "this" refers to - perhaps a small bit of explanation, or at least just replacing "this"?
- [x] Can we add a sentence explaining the `wait=True` argument to the capture decorator?
- [x] Perhaps a word should be added about #1811 - that append_display... doesn't work with widgets for right now, instead explicitly use display().
----

This is what the output looks like:

demo3


jasongrout:
Wow, nice! Very nice!!!!

A couple of things:

1. it looks like there is an extra left margin in your screenshot. Is the output widget not deleting that left margin?
2. The background thread issue is a bit more complicated - threaded output in the ipython kernel isn't cleanly supported, since it's not just scoped to a thread, but also scoped to the execution in the main thread. In other words, if you start a background thread, the background thread output goes into whatever output the main thread happens to be sending things to. What the output widget does is offer a way to redirect that background threaded output to somewhere we choose. We can't rely on the display() mechanism for this since it suffers the problem above. To get around this, we append directly to the output widget, bypassing the ipython display mechanism.
3. in debugging callbacks, "An effective tool for debugging this" - it's not clear what "this" refers to - perhaps a small bit of explanation, or at least just replacing "this"?
4. Can we add a sentence explaining the wait=True argument to the capture decorator?
5. Perhaps a word should be added about https://github.com/jupyter-widgets/ipywidgets/issues/1811 - that append_display... doesn't work with widgets for right now, instead explicitly use display().

pbugnion:
Thanks for the feedback! I'll look into it over the next few days.

maartenbreddels:
great work pascal!
side questions: how do you make a screenshot like the (except on a crazy big monitor in landscape)

jasongrout:
Firefox and Chrome both have facilities for making screenshots of an entire page: https://developer.mozilla.org/en-US/docs/Tools/Taking_screenshots

pbugnion:
Thanks!

Being able to take whole page screenshots is super useful for docs -- I learnt this from a work colleage. I use a Chrome extension called [fireshot](https://chrome.google.com/webstore/detail/take-webpage-screenshots/mcbpblocgmgfnpjjppndjkmgjaogfceg?hl=en), but I'll definitely look into the links Jason posted.

pbugnion:
For the extra left margin, I've opened issue #2025 . The error seems to predate this PR: it's already present on [docs built off master](http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Output).

I'm happy to tackle that next, but given that it (presumably) isn't a regression introduced by this PR, and given that merge conflicts in the docs notebooks can be fairly painful, I'd like to do it off a separate branch (so we can merge this as soon as possible).

That aside, I think this PR is ready for another review.

jasongrout:
Very nice! Thanks!

pbugnion:
Great, thanks a lot for the help!

PR #2021, opened 2018-03-27
Use _repr_mimebundle_ and require IPython 6.1 or later.

### LINKED ISSUE #1811: output widget append_display_data does not display widgets

jasongrout:
Fixes #1811 

This has the problem (noted in the code) that the display callback now happens *before* the display message is sent. Before this, the display callback happened *after* the display message was sent.

CC @SylvainCorlay - what do you think about this? I'm thinking that this backwards compatibility difference would make this actually an 8.0 change, and may cause other problems.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -13,7 +13,6 @@ from ipykernel.comm import Comm
 from traitlets import (
     HasTraits, Unicode, Dict, Instance, List, Int, Set, Bytes, observe, default, Container,
     Undefined)
-from IPython.display import display
 from json import loads as jsonloads, dumps as jsondumps
 
 from base64 import standard_b64encode
@@ -381,7 +380,6 @@ class Widget(LoggingHasTraits):
     _property_lock = Dict()
     _holding_sync = False
     _states_to_send = Set()
-    _display_callbacks = Instance(CallbackDispatcher, ())
     _msg_callbacks = Instance(CallbackDispatcher, ())
 
     #-------------------------------------------------------------------------
@@ -449,7 +447,7 @@ class Widget(LoggingHasTraits):
             Widget.widgets.pop(self.model_id, None)
             self.comm.close()
             self.comm = None
-            self._ipython_display_ = None
+            self._repr_mimebundle_ = None
 
     def send_state(self, key=None):
         """Sends the widget state, or a piece of it, to the front-end, if it exists.
@@ -548,21 +546,6 @@ class Widget(LoggingHasTraits):
         remove: bool
             True if the callback should be unregistered."""
         self._msg_callbacks.register_callback(callback, remove=remove)
-
-    def on_displayed(self, callback, remove=False):
-        """(Un)Register a widget displayed callback.
-
-        Parameters
-        ----------
-        callback: method handler
-            Must have a signature of::
-
-                callback(widget, **kwargs)
-
-            kwargs from display are passed through without modification.
-        remove: bool
-            True if the callback should be unregistered."""
-        self._display_callbacks.register_callback(callback, remove=remove)
 
     def add_traits(self, **traits):
         """Dynamically add trait attributes to the Widget."""
@@ -671,10 +654,6 @@ class Widget(LoggingHasTraits):
         """Called when a custom msg is received."""
         self._msg_callbacks(self, content, buffers)
 
-    def _handle_displayed(self, **kwargs):
-        """Called when a view has been displayed for this widget instance"""
-        self._display_callbacks(self, **kwargs)
-
     @staticmethod
     def _trait_to_json(x, self):
         """Convert a trait value to json."""
@@ -685,9 +664,8 @@ class Widget(LoggingHasTraits):
         """Convert json values to objects."""
         return x
 
-    def _ipython_display_(self, **kwargs):
-        """Called when `IPython.display.display` is called on the widget."""
-
+    def _repr_mimebundle_(self, **kwargs):
+        """Called when `IPython.display.display` is called."""
         plaintext = repr(self)
         if len(plaintext) > 110:
             plaintext = plaintext[:110] + '…'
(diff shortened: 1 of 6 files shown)


jasongrout:
Resetting to major release milestone until we figure out the display callback issue noted above.

SylvainCorlay:
@jasongrout I think that it is very unlikely that this order has any consequences.

I can see two cases where this would be an issue:

 - when someone overrides Jupyter's display hook to do more than sending the message, and they also rely on the display callbacks, and expect them to be called synchronously after the display hook is called.
 - in cases where the display callbacks would send other messages to the frontend (such as instantiating other widgets) and rely on the front end processing messages in order.

Both situations seem very unlikely. I would be 👍 for merging this in the 7.x series.

jasongrout:
I think a relevant question is: what are the on_displayed python-side hooks being used for? Does it make sense to get rid of the concept altogether? It's a bit of a weird concept when you have possibly multiple views of a widget.

CC also @maartenbreddels, who wanted the view count attribute, which is somewhat related.

ipywidgets/widgets/widget.py
```diff
@@ -717,9 +722,20 @@ def _ipython_display_(self, **kwargs):
...
 
-            self._handle_displayed(**kwargs)
+    def _ipython_display_(self, **kwargs):
+        """Called when `IPython.display.display` is called on a widget.
+
+        Note: if we are in IPython 6.1 or later, we return NotImplemented so
+        that _repr_mimebundle_ is used directly.
+        """
+        if ipython_version_info >= (6, 1):
+            raise NotImplementedError
```
maartenbreddels: I guess this should be `raise NotImplemented`?

maartenbreddels:
I doubt the order matters, although you can add it to the ioloop to be executed later. But indeed, _view_count is more useful.

It does make it easier to use, and in order not to break too much, I'd say, let us break the order changing, but keep the event.

SylvainCorlay:
> > in cases where the display callbacks would send other messages to the frontend (such as instantiating other widgets) and rely on the front end processing messages in order.
> 
> I thought that was one of the primary uses of the display callbacks.

Not specifically to create widgets though? Do you have an example?

ipywidgets/widgets/widget.py
jasongrout: No, `NotImplemented` is not an error class, so raising that will throw a `TypeError`.

jasongrout: (compare methods *return* `NotImplemented`, while subclasses with undefined methods *raise* `NotImplementedError`: https://docs.python.org/3.6/library/exceptions.html#NotImplementedError, https://docs.python.org/3.6/library/constants.html#NotImplemented)

jasongrout:
> Not specifically to create widgets though? Do you have an example?

I was searching for where `on_displayed` is used, and I couldn't find very many examples anywhere I looked on GitHub. It's used once in `interact` to run the function initially, for example.

Hence my question: where do people use this after all?

ipywidgets/widgets/widget.py
```diff
@@ -717,9 +722,20 @@ def _ipython_display_(self, **kwargs):
                     'model_id': self._model_id
                 }
             }
-            display(data, raw=True)
+            return data
 
-            self._handle_displayed(**kwargs)
+    def _ipython_display_(self, **kwargs):
```
SylvainCorlay: @jasongrout the definition of `_ipython_display_` could be behind a `if` statement.

```python
    if ipython_version_info < (6,  1):
        def _ipython_display_(self, **kwargs):
        """Called when `IPython.display.display` is called on a widget.
        Note: only defined for IPython <6.1.
        """
        data = self._repr_mimebundle_(**kwargs)
        if data:
            display(data, raw=True)
```

Madhu94:
Hi, any chance this version check can be made for IPython 5.4 and up? The ``_repr_mimebundle`` definition was backported to 5.x branch - https://github.com/ipython/ipython/pull/10544

ipywidgets/widgets/widget.py
flying-sheep: or do what @SylvainCorlay said and just put the definition into an if statement.

flying-sheep:
that means the check would be the following?

```py
if (5,4) < ipython_version_info or (6,  0) <= ipython_version_info < (6,  1):
    def _ipython_display_(self, **kwargs):
         ...
```

jasongrout:
Perhaps we just make a new major release of ipywidgets that cleans this up and updates our dependency on ipython. IPython 6.1 was released in May 2017 - it's been nearly two years, so I think it's okay to update the dependency.

jasongrout:
Now this is rebased on #2654. Again, I looked in a number of packages, and no one seemed to be using the on_displayed python callback (either for something they needed, or even correctly propagating it to children), so I just removed it. ipyleaflet seemed to be the only one to propagate it to children, but no children were using it to do anything.

SylvainCorlay:
> ipyleaflet seemed to be the only one to propagate it to children, but no children were using it to do anything.

this is an artefact of an old version of ipywidgets when this was really used for all nested widgets.

SylvainCorlay:
There seems to be commits unrelated to this in the PR (update to lumino etc).

jasongrout:
> There seems to be commits unrelated to this in the PR (update to lumino etc).

It is rebased on #2654, so it includes those commits (that should be merged first)

jasongrout:
Never mind, I rebased instead on master so it is easier and less controversial to review/merge.

jasongrout:
Unrelated failure in the new arm64 test run. I think if it fails a few more times, we should remove the arm64 test runs.

jasongrout:
closed/reopened to kick the tests. I couldn't find a way to restart the tests using the travis ui.

PR #2023, opened 2018-03-28
When a view is removed, hide the container and delete the model id

### LINKED ISSUE #1845: Close a widget but the associate '<div class="output_area">' still exists
Since 7.0 widgets in notebook are rendered in ` <div class="output_area"> `,
but when I close a widget, the associate ` <div class="output_area"> ` still exists.

```
from ipywidgets import Layout, Button, Box
items = [Button(description='example') ]
box1 = Box(children=items)
box2 = Box(children=items)
box3= Box(children=items)
display(box1)
display(box2)
display(box3)
box2.close()
```
output:
div_output

 How could I remove this part ?

jasongrout:
We modify the actual data in the view mimetype to change the model_id to be the empty string. This is valid according to the view schema. As a new convention, an empty string for the model id indicates that the view was explicitly removed and should be hidden.

Fixes #1845

This actually doesn’t quite remove all of the empty space in JuptyerLab because there is an extra nesting of containers for rendered outputs, but it does make it better, and does seem to remove space in the separate (simplified) output area views.

CC @SylvainCorlay

### DIFF
--- a/packages/jupyterlab-manager/src/renderer.ts
+++ b/packages/jupyterlab-manager/src/renderer.ts
@@ -30,13 +30,27 @@ class WidgetRenderer extends Panel implements IRenderMime.IRenderer, IDisposable
 
   async renderModel(model: IRenderMime.IMimeModel) {
     const source: any = model.data[this.mimeType];
+
+    // If there is no model id, the view was removed, so hide the node.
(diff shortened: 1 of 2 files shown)


marscher:
Thanks for this workaround!

PR #2026, opened 2018-03-28
Better way to use require fallback?

vidartf:
See discussion in https://github.com/jovyan/pythreejs/issues/171 .

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -35,7 +35,7 @@ function moduleNameToCDNUrl(moduleName: string, moduleVersion: string) {
         fileName = moduleName.substr(index+1);
         packageName = moduleName.substr(0, index);
     }
-    return `https://unpkg.com/${packageName}@${moduleVersion}/dist/${fileName}.js`;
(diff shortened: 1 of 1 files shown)


jasongrout:
@maartenbreddels - can you review this one, as you were involved in the issue over in pythreejs?

jasongrout:
This can also go into a patch release of html manager, so we don't have to rush it for 7.2.

maartenbreddels:
Let me check this one, I think the idea is solid.

maartenbreddels:
Ok, the issue is that require already 'knows' 'jupyter-threejs' failed, since it requested it before doing the CDN request (it queried for ./jupyter-threejs.js), and refuses to do so again. This can be fixed using `require.undef('jupyter-threejs')`, say in requireLoader, after the catch.

maartenbreddels:
Didn't work, you need to take off the `.js` part, sth like this:
```
            var url = moduleNameToCDNUrl(moduleName, moduleVersion)
            conf.paths[moduleName] = url.substring(0, url.length-3);
```

vidartf:
Since the function is not exported, I just removed the `.js` part from the return value.

vidartf:
This works for me locally now.

maartenbreddels:
yes works here too.

PR #2028, opened 2018-03-29
Get rid of output area prompt

### LINKED ISSUE #2025: Remove large left margin in output widget in docs
In the ['latest' docs](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Output), the output widget has an ugly left margin. This seems to be due to a non-zero flex-basis in the prompt area of the output area used for the output widget.

screen shot 2018-03-28 at 15 13 22

Happy to tackle this in the coming days, unless someone gets to it first.

jasongrout:
- [x] Test to make sure it fixes #2025 
- [x] Check to see if the jlab manager needs a similar fix

### DIFF
--- a/packages/html-manager/css/output.css
+++ b/packages/html-manager/css/output.css
@@ -3,13 +3,6 @@
 | Distributed under the terms of the Modified BSD License.
 |----------------------------------------------------------------------------*/
 
-.jp-OutputArea-child {
-    margin-top: 1em;
-    margin-bottom: 1em;
-}
-
-.jp-RenderedText pre {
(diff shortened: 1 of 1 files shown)


jasongrout:
We don't need to worry about jlab, since we have this bit of CSS in jlab: https://github.com/jupyterlab/jupyterlab/blob/d6b859ad13c3cd719eacbdc7020c5c5fed48ddc8/packages/outputarea/style/index.css#L124

pbugnion:
Ah nice!

PR #2037, opened 2018-03-29
Update all dependencies with ‘yarn upgrade’

jasongrout:
Fixes #20135

This fixes the web3 example by upgrading the outputarea and dependencies to include https://github.com/jupyterlab/jupyterlab/commit/1096c88d735f8320bb424a2648c34c77c1742257


jasongrout:
See also https://github.com/jupyterlab/jupyterlab/issues/4279, which tracks the larger issue of including the output area leading to lots of jlab-specific page config code.

PR #2043, opened 2018-04-03
Ensure we fully resolve lab URLs

vidartf:
Fixes `resolveUrl` on lab manager to give a download URL instead of a server path. The base implementation does not give a server path, so neither should the lab one. This will lead to a disconnect in the widget API name vs that used in jupyterlab, but I think that is acceptable.

### DIFF
--- a/packages/jupyterlab-manager/src/manager.ts
+++ b/packages/jupyterlab-manager/src/manager.ts
@@ -164,7 +164,9 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
    * Resolve a URL relative to the current notebook location.
    */
   resolveUrl(url: string): Promise<string> {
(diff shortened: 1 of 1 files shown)


vidartf:
Xref #1993 .

vidartf:
See also jupyterlab/jupyterlab#4265.

ian-r-rose:
This looks good to me, though I think that the difference in meaning between JupyterLab and ipywidgets with regard to `resolveUrl` is unfortunate.

jasongrout:
As in you think we should probably have called the widget manager function `getDownloadURL`, or perhaps something different?

ian-r-rose:
Just for consistency, yeah. If memory serves, we used to have a single function to do both, and then we broke it up into two functions in `@jupyterlab/services`.

PR #2047, opened 2018-04-12
Updates for JupyterLab 0.32.

jasongrout:
This involves updating `@jupyterlab/services` to 2.0, which will force us to have a major version bump in our base package.


jasongrout:
CC @athorton, @vidartf

jasongrout:
CC also @SylvainCorlay.

@jupyterlab/services had a major version bump to 2.0.0. The change there practically had no effect on our code (the one change we made in our shim is actually in code we don't use anymore). So we could make the argument that for our users, nothing changed.

However, If someone downstream had both a dependency on services `^1`, and a dependency on base `^1`, now there is a conflict (because the newest version of base actually depends on services 2). For that reason, I plan to make the following bumps:

`@jupyter-widgets/base` from 1.1.9 to 2.0.0
`@jupyter-widgets/controls` from 1.2.0 to 2.0.0
`@jupyter-widgets/html-manager` from 0.12.0 to 0.13.0
`@jupyter-widgets/html-manager` from 0.34.0 to 0.35.0
`@jupyter-widgets/output` from 1.0.16 to 2.0.0

widgetsnbextension python package from 3.2.0 to 3.3.0 (since this is not a js library people depend on)

ipywidgets python package from 7.2.0 to 7.2.1 for a single change, a bump in the html manager version for the embedding code: https://github.com/jupyter-widgets/ipywidgets/blob/master/ipywidgets/_version.py#L17

Note that the actual widget protocol versions stay *exactly* the same. This now leads to a wider divergence between the widget "module_version" protocol number, and the actual npm package version.

jasongrout:
I added @SylvainCorlay and @vidartf  as reviewers, but if anyone else wants to review/chime in, please feel free to do so!

jasongrout:
just thinking through this, this means that a 3rd party widget, like bqplot, must now bump the js version and dependency to handle the base package version bump, which means that bqplot must now release a new version for python (since that includes the compiled js).

Actually, looking at it more carefully, it looks like we are only using services for its typings. It's possible we could move services to devdependencies and just use it for its typings. Any downstream package from us would also need to include the right services typings.

vidartf:
> It's possible we could move services to devdependencies and just use it for its typings.

That sounds like the right solution to me. Is it possible to reexport the typings as well? I.e.:

```ts
import {
  KernelMessage
} from '@jupyterlab/services';

export{
  KernelMessage
} from '@jupyterlab/services';
```

vidartf:
Note: I pushed a cleanup commit on some linting.

vidartf:
> Is it possible to reexport the typings as well?

Actually, that would only be relevant for interfaces, not classes etc.

jasongrout:
> This does technically violate semver, in that we are deleting an exported interface

I pushed some changes that, instead of *deleting* the comm manager interface, just modify the shim new_comm return value to be a promise. I think we can do a minor version bump in the base package. This technically breaks semver, but in practice I don't think it will cause problems. If anyone is using new_comm, they will need to either update their code, or pin to the base package patch version. I think the benefits of not requiring all of the third party widgets to update is worth it.

jasongrout:
Actually, if people really aren't using the shim comm manager, we should probably just do a patch version bump.

So, how about this plan:

`@jupyter-widgets/base` from 1.1.9 to 1.1.10
`@jupyter-widgets/controls` from 1.2.0 to 1.2.1
`@jupyter-widgets/html-manager` from 0.12.0 to 0.13.0 (upgrading services to 2.0)
`@jupyter-widgets/jupyterlab-manager` from 0.34.0 to 0.35.0 (upgrading services to 2.0)
`@jupyter-widgets/output` from 1.0.16 to 1.0.17

widgetsnbextension from 3.2.0 to 3.2.1
ipywidgets from 7.2.0 to 7.2.1

@vidartf? @SylvainCorlay?

Issue #1864, opened 2017-12-08
Balloons for widgets


RDeRenzi:
I think that ipywidget is great, therefore I venture with a suggestion.
Presently a long widget description with not enough space to displays produces dots, and a balloon when hovering over the text. The balloon is nice, the dots not so much.

In largish data analysis GUIs, very short help (i.e. a slightly longer description) would nicely fit, in my opinion, in a separate trait, that behaves like the balloon: the description is shown statically, the balloon is shown when hovering over the former.

Or in two separate parameters of the description, short and [long], the latter behaving like a balloon.

I hope you may consider it as a future feature, if others agree with the purpose. 

Thank you so much 
Roberto

P.S the same applies to other widget/trait pairs that have the same functionality, e.g. ToggleButton/value

jasongrout:
Many traits have a `description` attribute - are you proposing an additional `description_long` (or some similar name) which would populate the balloon tooltip?

jasongrout:
> In largish data analysis GUIs, very short help (i.e. a slightly longer description) would nicely fit, in my opinion, in a separate trait, that behaves like the balloon: the description is shown statically, the balloon is shown when hovering over the former.

Could you give some example code illustrating your idea?

RDeRenzi:
Dear jasongrout, 
I hope this screenshot helps to understand what I mean. Not the best worked out example, I admit.

![fit_tab](https://user-images.githubusercontent.com/25430702/33769567-fbb13ba4-dc2a-11e7-8976-baaf1541b7fb.png "mujpy screenshot, from https://github.com/RDeRenzi/mujpy")

(btw, Markdown in the above figure does what I need)
The example in the screenshot is obtained exploiting a loooong description. With my proposal one could avoid the dots that cram the static GUI.

Regarding your previous question: I do not really know the ipywidgets innards. Syntactically it could be either an independent trait, e.g.  balloon or an optional separate parameter of description (like in the Markdown syntax, I suppose), performing the same functionality. I am not sure I have answered  your questions.

RDeRenzi:
Perhaps you wanted this. Present code in the screenshot:

        self.plot_range = Text(description='plot range\nstart,stop\n[,pack]\n[last,pack]',value='0,500',
                               layout=Layout(width='22%'),continuous_update=False)
        self.plot_range.style.description_width='36%'
        self.plot_range.observe(on_range,'value')

Proposed code

        self.plot_range = Text(description='plot range',
[...]

PR #2070, opened 2018-05-10
Add tooltip attribute to description widgets #1864

piyushrungta25:
for #1864

### DIFF
--- a/packages/controls/src/widget_description.ts
+++ b/packages/controls/src/widget_description.ts
@@ -43,6 +43,7 @@ class DescriptionModel extends DOMWidgetModel {
             _view_module_version: JUPYTER_CONTROLS_VERSION,
             _model_module_version: JUPYTER_CONTROLS_VERSION,
             description: '',
+            tooltip: '',
         };
     }
 }
@@ -56,6 +57,7 @@ class DescriptionView extends DOMWidgetView {
         this.label.style.display = 'none';
 
         this.listenTo(this.model, 'change:description', this.updateDescription);
+        this.listenTo(this.model, 'change:tooltip', this.updateDescription);
         this.updateDescription();
     }
 
@@ -65,6 +67,11 @@ class DescriptionView extends DOMWidgetView {
 
     updateDescription() {
         let description = this.model.get('description');
+        let tooltip = this.model.get('tooltip');
+        if (tooltip.length === 0) {
+            tooltip = description;
+        }
+
         if (description.length === 0) {
             this.label.style.display = 'none';
         } else {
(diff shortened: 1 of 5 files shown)


packages/controls/src/widget_description.ts
```diff
@@ -65,14 +67,19 @@ class DescriptionView extends DOMWidgetView {
 
     updateDescription() {
         let description = this.model.get('description');
+        let tooltip = this.model.get('tooltip');
+        if (tooltip.length === 0) {
+            tooltip = description;
```
jasongrout: How would we disable the tooltip if we didn't want it to show up?

jasongrout:
Thanks! One design comment about disabling the tooltip if we'd like. Setting the tooltip to `''` seems natural to disable it - perhaps None should be the default, which means take it from the description?

I agree that this change is strictly better than what we have (we can't turn off tooltips right now either).

packages/controls/src/widget_description.ts
piyushrungta25: As far as I understand, it couldn't be disabled completely previously also since it always defaults to the description. Not specifying the tooltip or setting it to an empty string will result in closest behavior.

jasongrout:
Also, can we call it `description_tooltip`? An arbitrary control might have other places where tooltips make sense, so scoping the name to where it is used makes sense, even if it is a bit long. We do have precedence for a name like that, in that the style has a `description_width` attribute.

jasongrout:
Does that make sense to you as a user? It's a bit awkward that we use "None" as a default value, but it is a convention. The way I see it, if you have a string, set the tooltip to that (including `''`). If you have None, do the default behavior.

maartenbreddels:
I think it makes sense, looks great

(from mobile phone)

piyushrungta25:
That does make sense. I made the discussed changes.

jasongrout:
Thanks. We'll need to update the spec to have this trait, then I think we are good to merge. I'll look at this hopefully tomorrow.

jasongrout:
@SylvainCorlay - what are your thoughts about the model change here? In particular, the attribute can be either null or a string - does that make things difficult for xwidgets?

SylvainCorlay:
Hey @jasongrout thanks for the heads up.

The model changes are fine with me. We handle optional widget attributes very well in xwidgets. 

For example, in the layout widget, we do:
```
XPROPERTY(xtl::xoptional<std::string>, derived_type, align_items, {}, XEITHER_OPTIONAL("flex-start", "flex-end", "center", "baseline", "stretch", "inherit", "inital", "unset"));
```

jasongrout:
Great, then I'm okay with merging this for the beta.

jasongrout:
Thanks @piyushrungta25 for pushing this through!

PR #2090, opened 2018-05-30
fix: only handle exception when in IPython

maartenbreddels:
When doing unittests ip is None and will give another exception, this avoids it.


jasongrout:
Thanks!

PR #2094, opened 2018-06-01
[WIP] Change how buffers serialize

vidartf:
CC @SylvainCorlay @jasongrout @maartenbreddels

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -11,6 +11,7 @@ from contextlib import contextmanager
 import collections
 import sys
 import numbers
+import inspect
 
 from IPython.core.getipython import get_ipython
 from ipykernel.comm import Comm
@@ -285,9 +286,7 @@ class Widget(LoggingHasTraits):
                                                state['_view_module_version'],
                                                state['_view_name'])
         widget = widget_class(comm=comm)
-        if 'buffer_paths' in data:
-            _put_buffers(state, data['buffer_paths'], msg['buffers'])
-        widget.set_state(state)
+        widget.set_state(state, msg['buffers'])
 
     @staticmethod
     def get_manager_state(drop_defaults=False, widgets=None):
@@ -311,13 +310,8 @@ class Widget(LoggingHasTraits):
             'model_module': self._model_module,
             'model_module_version': self._model_module_version
         }
-        model_state, buffer_paths, buffers = _remove_buffers(self.get_state(drop_defaults=drop_defaults))
+        model_state, buffers = self.get_state(drop_defaults=drop_defaults)
         state['state'] = model_state
-        if len(buffers) > 0:
-            state['buffers'] = [{'encoding': 'base64',
-                                 'path': p,
-                                 'data': standard_b64encode(d).decode('ascii')}
-                                for p, d in zip(buffer_paths, buffers)]
         return state
 
     def get_view_spec(self):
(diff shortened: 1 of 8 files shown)


ipywidgets/widgets/widget.py
```diff
@@ -638,15 +596,17 @@ def _should_send_property(self, key, value):
...
+                serializer_value = { key: to_json(value, self)}
+            split_lock = { key: self._property_lock[key]}
             # A roundtrip conversion through json in the comparison takes care of
             # idiosyncracies of how python data structures map to json, for example
             # tuples get converted to lists.
-            if (jsonloads(jsondumps(split_value[0])) == split_lock[0]
-                and split_value[1] == split_lock[1]
-                and _buffer_list_equal(split_value[2], split_lock[2])):
+            if (jsonloads(jsondumps(serializer_value)) == split_lock
+                and _buffer_list_equal(buffers, self._buffer_lock)):
```
vidartf: @SylvainCorlay This solution here is broken, since the buffers are not checked *per key*. There needs to be a way to check the buffers "owned" by a specific key/trait. E.g. by having a mapping (key -> buffer indices).

maartenbreddels:
What is the reason for this change?

vidartf:
I think @SylvainCorlay har some problems with reimplementing the recursive serializers for C efficiently, so he suggested this simpler behavior. I suggested I could sprint it on a workshop to use as a base for discussion etc.

maartenbreddels:
👍it looks good, afaik tell, but for backwards compatibility we may want to keep the old behaviour? Or is the plan a breaking change for a major release?

SylvainCorlay:
The main problem with the current approach to the serialization of binary buffers is that

1. it involves an extra global processing of the patches in both ways, which may be very costly especially for serialization.

When deserializing, Python's JSON module first generates nested lists and dicts, and then we proceed with the insertion of the buffers.

When serializing, user-specified serializers return nested lists and dicts, (and form something that is almost JSON-serializable) and leave some binary types in it. **Then `_separate_buffers` is recursively called on all list and dicts.** which is very costly, especially when the state includes long sequences.

2. Serialization and deserialization cannot be done without looking at the entire patch.

The whole thing assumes that there exists *somewhere* an intermediate representation that is a "JSON plus binary buffers*, which may not be the case on the kernel side. In our C++ implementation, a JSON object is only JSON, and cannot contain other objects.

SylvainCorlay:
The solution that we have adopted in the `xwidgets` implementation is that

 - we removed the `buffer_paths` part of the message.
 - instead, buffers are replaced by a string indicating the index of the buffer in the sequence.

Since the wire format has not changed, we implemented a shim between that and the current version of the protocol which I would like to remove when this gets in.

SylvainCorlay:
> Or is the plan a breaking change for a major release?

I think that this is a breaking change for a major release.

jasongrout:
> instead, buffers are replaced by a string indicating the index of the buffer in the sequence.

What's the chance of a conflict between a random string in the data structure and your special buffer pointer string?

SylvainCorlay:
> > instead, buffers are replaced by a string indicating the index of the buffer in the sequence.
> 
> What's the chance of a conflict between a random string in the data structure and your special buffer pointer string? (i.e., is your special buffer pointer sufficiently distinct that there is little chance of interpreting some other way, or having some other random string accidentally be considered as a buffer index?)

None. The string could even simply contain the integer value, or be an integer. The `@buffer_reference@` prefix is only meant for human readability of messages.

The fact that it corresponds to a buffer is part of the schema for the said widget. The idea is that there is no  "generic" pass replacing magic strings.

jasongrout:
Now is the time to take a look at this again for ipywidgets 8.0, if we want to change things.

vidartf:
Ping @SylvainCorlay which initially requested it.

Issue #2052, opened 2018-04-16
Add a bit of padding to Dropdown to make room for the arrow


jasongrout:
Right now, if the content goes to the edge of the dropdown, the arrow and the content are on top of each other:

```python
from ipywidgets import *

Dropdown(options=['very very very very very very very very very very very very very very long option', 'another option'])
```
This gives:
<img width="328" alt="screen shot 2018-04-16 at 8 01 55 am" src="https://user-images.githubusercontent.com/192614/38817188-b333aee8-414c-11e8-93d5-8122f2d210fa.png">

However, adding `padding-right: 18px` in the inner select: https://github.com/jupyter-widgets/ipywidgets/blob/e1af98674e46c93c1fffe1e42a20adb2a62dad1e/packages/controls/css/widgets-base.css#L659

gives:
<img width="322" alt="screen shot 2018-04-16 at 8 03 18 am" src="https://user-images.githubusercontent.com/192614/38817191-b623dbe6-414c-11e8-9f69-f5c694ea2ba6.png">

jasongrout:
(perhaps 20px of right padding looks better? Probably whatever it takes to center that svg from the content edge and the border edge on the other side...

tjann:
Hi, I am interested in getting into open source / Jupyter with this issue. I thought it looked pretty simple but my changes aren't showing up. Could this be related to https://github.com/jupyter-widgets/ipywidgets/commit/f68bfbc566f3a78a8f3350b438db8ed523ce3642 by any chance?

jasongrout:
Thanks!

It's most likely not related to that commit because my comment is way after that commit removed the code. Perhaps it's the browser caching your code so it's not showing up? Can you open the browser developer tools and verify that your CSS is showing up on the page?

tjann:
Ah, so glad I asked! Indeed my source code/CSS wasn't showing up, so I switched from jupyter lab to jupyter notebook and it worked. Makes sense in hindsight since I only have the developer release of the notebook and not the lab. I have some questions that might help others:

I started wondering why we needed the latest jupyter lab to do dev work on widgets: 
https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/dev_install.md

But before asking, I wanted to check, and found another guide for developing ipywidgets which doesn't mention jupyter lab but seems to be a bit outdated: 
[...]

PR #2101, opened 2018-06-16
added 20px of right padding to the select dropdown styling

tjann:
addressing issue #2052


SylvainCorlay:
THanks @tjann this looks good to me. Merging!

Issue #1942, opened 2018-01-30
GridBox widget


jasongrout:
Now that [CSS grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) is making its way into [browsers](https://caniuse.com/#feat=css-grid), I think it would be great to have (in core widgets) a GridBox widget that places its widgets in a grid using the CSS grid spec.

jasongrout:
Here are some interesting resources from a HN article on CSS grids: https://news.ycombinator.com/item?id=16292094

Madhu94:
I'd like to give this a shot.

Madhu94:
Hi, I had a question - 
Should GridBox inherit from Box? 

There is a lot of shared code that I could reuse. But Box was meant to be a flexbox implementation. I could (1) create an abstract class, say "CSSLayout" (name TBD, but users shouldn't touch this) and move the common functionality into that. Box and GridBox would both subclass this and set ``display`` to either ``grid`` or ``flex``. 

(Note - I was able to get this to work even otherwise though, just by making GridBox inherit from Box and changing display in the child's constructor )

jasongrout:
I think Gridbox should not inherit from Box. The big thing that Box gives you is a children attribute, but that's probably not useful in a 2d layout to have a single list of children.

Madhu94:
I agree that the BoxModel contains just a children attribute, but the BoxView seems to contain some logic that we might end up re-implementing ?

> that's probably not useful in a 2d layout to have a single list of children.
Seems like a good time to finalize the API - did you have a 2D list (list of lists) in mind, as the input ?
I thought it would be a simple list, and since we will expose the grid spec as part of the ``Layout``, the user can set the ``grid-template-columns`` css property to tweak the layout. I've exposed ``grid``, ``grid_area`` and ``grid_gap`` - following the convention of using only shorthand css properties.

Madhu94:
Example : 
``` 
GridBox([w1, w2, w3, w4, w5, w6], layout=Layout(grid='none / repeat(3, 1fr)'))
``` 

This will create 2 rows of 3 columns each

jasongrout:
> Seems like a good time to finalize the API

Good question. I haven't read too deeply into the grid spec, so I'm not up to speed on what would be a natural api.

Layouts are used as a generic mechanism across all widgets, though. I imagine we would have more specific grid-type properties on the widget itself?

Madhu94:
[...]

PR #2107, opened 2018-06-28
Grid layout

SylvainCorlay:
screen shot 2018-07-05 at 12 18 18 am


Madhu94:
Both commits taken together, my changes are effectively wiped out and might as well not be included ! :(  
The changes bear close resemblance to what the state was when I made this comment https://github.com/jupyter-widgets/ipywidgets/issues/1942#issuecomment-381471396 => Basically GridBox inherited from Box, and all CSS properties (not redundant) were in Layout.

Could you please let me know what were the reasons for proposing it should be a separate class ?
Thanks.

SylvainCorlay:
> Both commits taken together, my changes are effectively wiped out and might as well not be included ! :(
> The changes bear close resemblance to what the state was when I made this comment #1942 (comment) => Basically GridBox inherited from Box, and all CSS properties (not redundant) were in Layout.

Yeah, I am really sorry if you spent too much time on separating it in multiple widgets.

So I was playing with your implementation yesterday and my initial change was to group things using the shorthand properties like elsewhere in ipywidgets. Although I did not go all the way, because there is a catch-all `grid` short hand property that is pretty hard to deal with.

Then we discussed with @jasongrout and went back & forth between the initial design of adding grid properties to layout that I think you proposed and your PR.

In any case, I would be happy to squash my commit into yours especially as you pretty much had this version already earlier.

From what I gather in the original thread, you seem to lean towards a single layout attribute. Also, coming back to your proposal of having a hierarchy of `Layout` types, it seems that at least the item properties should be in the base one, and maybe only the container grid properties should be in the `GridBoxLayout` class.

What do you think? Should we go ahead with something like this or would you like to iterate more upon it in the earlier PR?

Madhu94:
> From what I gather in the original thread, you seem to lean towards a single layout attribute. Also, coming back to your proposal of having a hierarchy of Layout types, it seems that at least the item properties should be in the base one, and maybe only the container grid properties should be in the GridBoxLayout class.    

The current state is similar to how ``Box`` is implemented though.  

> Should we go ahead with something like this or would you like to iterate more upon it in the earlier PR?  

No, please go ahead with your changes, my apologies if the earlier comment seemed a little aggressive.

SylvainCorlay:
>> Should we go ahead with something like this or would you like to iterate more upon it in the earlier PR?
>>
>
> No, please go ahead with your changes, my apologies if the earlier comment seemed a little aggressive.

Not at all!

SylvainCorlay:
screen shot 2018-07-04 at 7 55 13 pm

SylvainCorlay:
@jasongrout this is ready to go!

PR #2111, opened 2018-07-02
Remove value from the Image repr

mwcraig:
Right now `Image.__repr__` includes the value of the image...which bring my browser to a crashing halt for large images.

This is likely not the ideal way to handle this; I'd be happy to change `widget.__repr__` to truncate all large values (over maybe 10k characters?) in the repr instead.

### DIFF
--- a/ipywidgets/widgets/widget_image.py
+++ b/ipywidgets/widgets/widget_image.py
@@ -6,6 +6,7 @@
 Represents an image in the frontend using a widget.
 """
 import mimetypes
+import hashlib
 
 from .widget_core import CoreWidget
 from .domwidget import DOMWidget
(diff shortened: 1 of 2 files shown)


vidartf:
One thing we did e.g. in nbdime and nbval was to say something like `<base64 string, hash=ea628...>`. You could do something similar (but maybe with a cut-off value of only around a 100 characters).

mwcraig:
I modified this to always present the value (truncated along the lines that @vidartf suggested).  A more general solution in `widgets._repr_from_keys` would probably be ideal, but for now this presents a value for `Image` but won't kill your browser.

ipywidgets/widgets/widget_image.py
```diff
@@ -128,3 +129,27 @@ def _guess_format(cls, filename):
             return mtype[len('image/'):]
         except Exception:
             return None
+
+    def __repr__(self):
+        # Truncate the value in the repr, since it will
+        # typically be very, very large.
+
+        def hash_string(s):
```
jasongrout: Let's not define a new function every time we do a repr.

ipywidgets/widgets/widget_image.py
```diff
@@ -128,3 +129,27 @@ def _guess_format(cls, filename):
...
+            return hashlib.md5(s.encode("utf8")).hexdigest()
+
+        class_name = self.__class__.__name__
+
+        # Return value first like a ValueWidget
+        signature = []
+        sig_value = str(self.value)
+        if len(str(self.value)) > 100:
+            sig_value = '<base64, hash={}...>'.format(hash_string(sig_value)[:16])
```
jasongrout: We don't know that it is base64, do we? It might be a very long URL.

I like your previous approach of just truncating. Or perhaps we should just enhance the existing repr keys mechanism to let us return a repr form for any key, rather than just whether to include the key or not.

mwcraig:
This just truncates long `Image.value` (added a couple tests too).

jasongrout:
Looks great to me, and works well when I test it too. Thanks!

PR #2116, opened 2018-07-05
Upgrade to JupyterLab 0.33.0rc1

jasongrout:
Upgrades JupyterLab support to 0.33.0rc1.

TODO:
- [x] Fix examples. In particular, the web-tmpnb example needs the same sort of kernelChanged logic that the jlab manager had.


jasongrout:
I think this works with the latest JupyterLab 0.33.0rc1. Can someone check it out before merging?

jasongrout:
CC @SylvainCorlay @vidartf @pbugnion @maartenbreddels - if one of you can check this, that would be great

jasongrout:
Thanks!

PR #2122, opened 2018-07-06
Update the widget model spec versions.

jasongrout:
Since our next release is a beta release, this spec is not final yet.


jasongrout:
We need to update both the base and controls version because of the grid layout attributes.

PR #2123, opened 2018-07-07
[WIP]:Added Speedup and slowdown functionality to play widget

raviteja-kvns:
Speedup and slowdown functionality has been added to the play widget. 
Implementation is done on the typescript side. No changes on the python side were necessary.

Please review and provide feedback


SylvainCorlay:
Sorry for the super late feedback @raviteja-kvns, I am looking at the backlog, and realized that we did not have a loot at this.

The code looks fine. How about using a multiplicative scale for the speedup instead?

PR #2128, opened 2018-07-10
Add 'transparent' as a color

mwcraig:
I believe `'transparent'` is a valid HTML color (see https://developer.mozilla.org/en-US/docs/Web/CSS/color_value for documentation of transparent as a color). It would be handy in bqplot for making unfilled markers.


SylvainCorlay:
Fastest merge in the west :)

SylvainCorlay:
For the record:

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

*The transparent keyword represents a fully transparent color. This makes the background behind the colored item completely visible. Technically, transparent is a shortcut for rgba(0,0,0,0).*

SylvainCorlay:
The failure on Python 3.3 is unrelated to that change and is due to Python 3.3 now being deprecated.

I am merging an opening a PR removing it from the build matrix.

PR #2129, opened 2018-07-10
Remove Python 3.3 from the build matrix

SylvainCorlay:


jasongrout:
For everyone else, python 3.3 EOL was September 2017.

PR #2130, opened 2018-07-11
Add DeprecationWarning when using Mapping for options

### LINKED ISSUE #1958: Drop support for mutable types as dropdown options
I know this has been reviewed before, but I humbly request it be reviewed again.. :-)

I don't see any upside to supporting dictionaries as an options list (you can just use `widget.options = dict.items()` if you want the same behaviour, which more clearly separates ownership semantics), and it is a source of many pitfalls.  I suspect that many people use dictionaries simply because it's more prominently featured in the docs than the list-of-tuples approach.

I realize this is a backwards-incompatible change, but I think moving forward it would reduce community support burden (and frustrated user time).  Certainly this could be done in a phased manner with `DeprecationWarning` (making sure users can actually see it, of course).

jtpio:
Fixes #1958.

For Selection Widgets, add a `DeprecationWarning` whenever a `Mapping` (dictionaries) is used as options.

I'll do a second pass on the documentation to remove all mentions to dictionaries as possible values for `options`.

### TODO

- [x] Fix Python 2 test
- [x] Second pass on the documentation


jasongrout:
Thank you!

Can you remove the `model_id` changes in the PR for the notebook? That's churn that breaks the saved widget state in the notebook. We'll run all the notebooks before release and commit one big update with the updated widget state.

(to be clear, please keep the source text changes in the notebook, and the one place where we have printed output that would change).

jtpio:
Sure thing! And it also makes the diff cleaner.

jasongrout:
Adding to 7.3 - let's see if we can get this in...

jasongrout:
Looks like there is a test failure. We'll try to get this in first thing for 7.4.

jtpio:
The failure is related to the new `TestDropdown` test and seems to be only happening for Python 2.7.

Not sure what it is yet as the tests pass locally with Python 2.7.14

jtpio:
Tests are now green after the last commit.

There seems to be a different behavior between Python 2.7 and Python 3 when it comes to testing warnings (even though the documentation looks identical between the two versions: https://docs.python.org/3.7/library/warnings.html#testing-warnings).

Python 2 requires to manually clear the warning registry if we want to be able to catch them:

```python
import inspect
module = inspect.getmodule(Dropdown)
getattr(module, '__warningregistry__', {}).clear()
```

This has also been mentioned in other project testing for warning being triggered, for example with Matplotlib: https://github.com/matplotlib/matplotlib/commit/0017d3e19f4619cb9a79f7432b17b02c4ebd315b

SylvainCorlay:
@jtpio these seems to be a conflict. Could you please rebase your PR on master?

jasongrout:
@SylvainCorlay - We should try to get this in for 7.4 as well.

jtpio:
`test_widget_selection.py` was added in both `master` and this branch.

Should be good now :+1:

jasongrout:
That's an awesome problem, that we're tripping over ourselves to add more testing :).

PR #2146, opened 2018-07-24
Bug: cannot deserialize Image.value

maartenbreddels:
If the `Image.value` value comes off the wire, it will/can be a memoryview, which will make deserializing (quietly fail).
We need a custom deserializer for this.


maartenbreddels:
Seems fixed by the 2nd commit.

PR #2149, opened 2018-07-27
Update to released jlab 0.33 packages

jasongrout:
Also change the jlab manager to target es2015 like jlab does now.


mprogram:
@jasongrout, for this update to jlab 0.33 packages may I ask you, out of my curiosity, why `@jupyterlab/services` are being updated to `^3.0.0` in `examples`' and `packages`' `package.json`s, but not in `widgetsnbextension/package.json`?

jasongrout:
> @jasongrout, for this update to jlab 0.33 packages may I ask you, out of my curiosity, why @jupyterlab/services are being updated to ^3.0.0 in examples' and packages' package.jsons, but not in widgetsnbextension/package.json?

Great question. services (and all of the jlab packages) now compile to es2015 instead of es5, but widgets in the classic notebook still supports browsers that don't support es2015. Lucky for us, though, we don't use any new advanced features of services, so actually in the classic notebook, services 1 or 2 or 3 is just fine for us. However, for JupyterLab, we need to keep up to the latest release.

For widgets 8, we'll probably drop support for older browsers, and then will upgrade to services 3 across the board.

jasongrout:
> …and I'm getting warnings like "workspace-aggregator-14e3c834-0b61-4d2c-b25d-064eaf7f0a6d > @jupyter-widgets/controls > postcss-cssnext@3.1.0" has unmet peer dependency "caniuse-lite@^1.0.30000697" and further. It seems that postcss-cssnext has been deprecated in favor of postcss-preset-env as posted on Deprecating cssnext

Thanks for looking this up! Sounds like we should upgrade.

Our use of postcss is again basically just to support older browsers in the classic notebook.

jasongrout:
> Thanks for looking this up! Sounds like we should upgrade.

Can you make a new issue? If you can, a pull request would be even better!

mprogram:
Thank you for your great answer. As a follow-up I've just opened the issue (#2151).

mprogram:
One more question: 
`#packages/jupyterlab-manager/src/tsconfig.json`

`-    "lib": ["dom", "es5", "es2015.promise"],`
`+    "esModuleInterop": true,`
`+    "lib": ["dom", "es2015"],`

`-    "target": "ES5",`
`￼+     "target": "es2015",`

…and I've seen a suggestion on gitter on moving to the `ES6` target in a particular extension(s). Is there an implied policy coming from `jupyterlab`, which target should be set in `tsconfig.json` of those?

jasongrout:
No, no implied policy.

jasongrout:
By the way, es2015 is the same as es6. The official name of the standard is es2015. Is that what you were asking?

There is no implied policy between targeting es5 and es2015.

mprogram:
Thank you for explaining. It was, indeed, the meaning of my question.

PR #2158, opened 2018-07-30
Update jupyterlab-manager version compatibility

jtpio:
So JupyterLab 0.33.x matches with `@jupyter-widgets/jupyterlab-manager@0.36`.


jasongrout:
Thanks!

PR #2162, opened 2018-08-02
Implement Video and Audio widgets

martinRenou:
Implement a Video widget, at first I wanted to synchronize the progress of the video between views, but it turned out that it's quite difficult to do it. There is a `timeupdate` event on HTML video elements, but it's not triggered often enough to synchronize the views.

Edit: Also implement an Audio widget. `Image`, `Video` and `Audio` widgets have a common base class `_Media`


martinRenou:
@maartenbreddels

SylvainCorlay:
Lgtm. For the tests to pass you will need to update the spec file as per the test error message.

martinRenou:
I also plan to make an `Audio` widget and it might be better to implement a base class named `Media` for widgets like `Image`, `Video` and `Audio`. The code will be quite the same for those three widgets.

SylvainCorlay:
This looks good to merge.

Leaving it open for a bit so that jason or other people can comment.

maartenbreddels:
Looking good!
I guess the Image unittests covers basically everything?
The only open issue I have is that a video can have multiple urls, for instance, if it does not support webm, it can take a mp4 file. Should we support this?

SylvainCorlay:
Using multiple `source` tags would require providing different data for each case. That would completely modify the model to be a lot more complex (like a dict: `format: data`)... Not sure that it would be good for usability.

martinRenou:
Also @maartenbreddels we may want to implement something like `Video.from_download("bla.mp4")` ?

SylvainCorlay:
After the in-person discussion with @maartenbreddels and @martinRenou we decided to not add the possibility to have extra sources for the video widgets.

I think that eventually, we will add a `sources` attribute to `media` in ipywidgets 8, that will apply to 

 - video
 - audio
 - and picture, which should maybe underlie the Image widget?

maartenbreddels:
For Python2 and 3 I use this:
```python
try:
    from urllib import urlopen  # py2
except ImportError:
    from urllib.request import urlopen  # py3
value = urlopen(url).read()
```
Except that Python2 doesn't give an exception on a 404, and will happily feed you the 404 page. For Python 3 you get a proper exception. For the convenience of this, I suggest we add a `from_download` method.

SylvainCorlay:
We discussed further on the means to `download` the media 

 - to the kernel
 - to the client

and APIs.

This should probably be another PR.

Merging as per agreement with @jasongrout and @maartenbreddels.

PR #2166, opened 2018-08-06
Fix "options.output is undefined"

martinRenou:


SylvainCorlay:
(Context: calling `create_view` with no options.)

SylvainCorlay:
Thanks!

PR #2170, opened 2018-08-08
Fix #1810

### LINKED ISSUE #1810: Jupyterlab output widget doesn't send appended outputs back to kernel
The output widget should send the appended outputs back to the kernel, but doesn't. It should sync the outputs on both the `add` and `clear` methods: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/jupyterlab-manager/src/output.ts#L78

The classic notebook output widget does do this:

* https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/src/widget_output.js#L43
* https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/src/widget_output.js#L48

The output widget should also listen to syncs from the outputs in the kernel, like the classic notebook output widget does: https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/src/widget_output.js#L50

fcollonval:
Here is a fix for #1810 and https://github.com/jupyterlab/jupyterlab/issues/4319

Thanks @jasongrout  to point out exactly what to do and where to look.

### DIFF
--- a/packages/jupyterlab-manager/src/output.ts
+++ b/packages/jupyterlab-manager/src/output.ts
@@ -40,7 +40,8 @@ export
 class OutputModel extends outputBase.OutputModel {
   defaults() {
     return {...super.defaults(),
-      msg_id: ''
+      msg_id: '',
+      outputs: []
     };
   }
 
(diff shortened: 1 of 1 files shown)


jasongrout:
Thanks! This looks great to me. I *really* appreciate you fixing this.

jasongrout:
This is now in `@jupyter-widgets/jupyterlab-manager` 0.36.2.

Issue #2127, opened 2018-07-10
Display bug in embedded Output widget - spotted in the context of ipysheet


oscar6echo:
When embedding [ipysheets](https://github.com/QuantStack/ipysheet) in a widget Output there maybe a display bug.

In short it happens when you want to display an ipysheet, contained in an ipywidget.Output, in an ipywidgets.Tab. Then if this ipysheet is not in the default tab, it won't be properly displayed (typically only the first column shows and a slider appears on the right) until you click on it.

See the display bug in [this notebook](https://nbviewer.jupyter.org/urls/gist.githubusercontent.com/oscar6echo/2a78d98f4f3ff242ac2dd246e7d580b6/raw/b5d658873d897f8a9b76cacc1b11dd90d2f3c4cd/display_bug_widget_tab_output_ipysheet.ipynb).

I understand from @maartenbreddels  that there is a messaging issue in the encapsulation of widgets that results in the Output widget not being 'refreshed' often enough or something along these lines.

This display bug can be circumvented using Box instead of Ouput, or no wrapper widget at all, as shown in the notebook. So it is not critical in my view. However that limitation should be known, and may lead to more annoying problems in other contexts.

jasongrout:
@maartenbreddels - is this the thing you were talking about the other day, about widgets inside an output widget not getting resize messages or something?

oscar6echo:
Solved by #2172 for JupyterLab.  
Unchanged for Jupyter classic notebook.

PR #2172, opened 2018-08-10
The widget renderer should call its parent’s dispose method.

### LINKED ISSUE #2171: Remove method for a WidgetView not called when clearing outputs
When clearing a view using "Clear Outputs" from the context menu (or re-executing the notebook cell), the user defined `remove` method from a `DOMWidgetView` is never called.

This seems to happen only in JupyterLab (and also when using an `Output` widget). As far as I remember it used to work fine when creating widgets on the fly in the classic notebook (from [Building a Custom Widget](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Custom.html)).

A simple way to reproduce this is to start from the [Jupyter Widget Hello](https://github.com/vidartf/jupyter-widget-hello) and add the following to the view:

```typescript
export
class ExampleView extends DOMWidgetView {

  interval : number

  initialize() {
    console.log('initialize');
  }

  render() {
    super.render();
    console.log('render');
    this.interval = window.setInterval(() => {
      console.log(`Interval: ${this.model.get('value')}`);
    }, 1000);
    console.log(`create interval ${this.interval}`);
    this.el.textContent = this.model.get('value');
  }

  remove() {
    console.log('remove')
    window.clearInterval(this.interval);
    console.log(`cleared interval ${this.interval}`);
    super.remove();
  }
}
```

Since the `remove` method is not called, the interval keeps running in the background and new ones are added when re-executing the cells. This can be problematic when the view allocates more memory (sprites, images and so on) and needs to do some cleanup before being removed.

After digging a little bit in the code, the further I could get down the chain is in the `dispose` method of the `WidgtetRenderer`, which doesn't seem to hold information related to the current `viewId`.

https://github.com/jupyter-widgets/ipywidgets/blob/537b171d5faec7883a38098dcf5c7a952499cb23/packages/jupyterlab-manager/src/renderer.ts#L77-L82

The view is however properly disposed with the `remove` method being triggered when closing the widget with `widget.close()`:

https://github.com/jupyter-widgets/ipywidgets/blob/537b171d5faec7883a38098dcf5c7a952499cb23/packages/base/src/manager-base.ts#L180

Is there a way to know which views are attached to an output area, so they could be removed whenever the output area is cleared?

jasongrout:
As it was, any child widget was never getting disposed. Now, the widget renderer parent (a panel) will dispose its children (i.e., any widget that is rendered). This also should ensure that all widget remove methods are called.

Fixes #2171

### DIFF
--- a/packages/jupyterlab-manager/src/renderer.ts
+++ b/packages/jupyterlab-manager/src/renderer.ts
@@ -78,6 +78,7 @@ class WidgetRenderer extends Panel implements IRenderMime.IRenderer, IDisposable
     if (this.isDisposed) {
       return;
     }
+    super.dispose();
     this._manager = null;
   }
 


jasongrout:
CC @maartenbreddels - does this take care of the issue you mentioned the other day about widgets inside of an output widget not getting relevant Phosphor lifecycle events?

jasongrout:
This is now in `@jupyter-widgets/jupyterlab-manager` 0.36.2.

maartenbreddels:
If you have time, can you check if this solved putting a ipsheet in a tabbed output inside of an Output widget @oscar6echo ?

jasongrout:
@maartenbreddels - is this issue the one you meant to post to?

jasongrout:
Ah, never mind. I think I understand your comment now.

oscar6echo:
as asked by @maartenbreddels I
+ updated ipywidgets (to 7.4.0), jupyterlab (0.34.0), @jupyter-widgets/jupyterlab-manager (0.37.2) 
+ reran the [sample notebook](https://nbviewer.jupyter.org/urls/gist.githubusercontent.com/oscar6echo/2a78d98f4f3ff242ac2dd246e7d580b6/raw/b5d658873d897f8a9b76cacc1b11dd90d2f3c4cd/display_bug_widget_tab_output_ipysheet.ipynb) in #2127 in JupyterLab

and the problem disappeared.  
(Obviously it still does not work with the Jupyter classic notebook - as it is unrelated to Phosphor)  

Remark (not directly connected to the issue, I recognize):
But JupyterLab is still lacking in terms of ipywidgets: It does not save the ipywidgets state automatically and there is no way to do it manually, nor programmatically (I think). Hopefully ipywidgets will gain Jupyter first calls citizen status soon, as they are key in Jupyter workflows in my company.

PR #2173, opened 2018-08-11
Add jupyter-widgets CSS class to error messages

### LINKED ISSUE #2076: Error messages with JupyterLab Dark Theme
Some of the error messages are difficult to read when using JupyterLab with the Dark Theme:

image

Equivalent for the Light Theme:

image

One fix for that is to add the `jupyter-widgets` css class to the node everytime `textContent` is set manually:

```javascript
this.addClass('jupyter-widgets');
```

For example right after this line:

https://github.com/jupyter-widgets/ipywidgets/blob/4aceb8670a64aff89c528d5d00205f1428c68d16/packages/jupyterlab-manager/src/renderer.ts#L59

This is also the case when using the Hello World widget from the two cookie cutters:
- https://github.com/jupyter-widgets/widget-cookiecutter
- https://github.com/jupyter-widgets/widget-ts-cookiecutter

Should this css class be added when needed? (including ipywidgets core error messages)
Another way would be to default the color of `jp-OutputArea-output` to `var(--jp-content-font-color1)`, directly in JupyterLab?

jtpio:
Fixes #2076.

Went for the simpler solution for now, as the most important is to better see these error messages when using the Dark Theme in JupyterLab.

### Before

image

### After

image

### DIFF
--- a/packages/html-manager/src/output_renderers.ts
+++ b/packages/html-manager/src/output_renderers.ts
@@ -28,9 +28,11 @@ export class WidgetRenderer extends Widget implements IRenderMime.IRenderer {
                 console.log('Error displaying widget');
                 console.log(err);
                 this.node.textContent = 'Error displaying widget';
+                this.addClass('jupyter-widgets');
             }
         } else {
             this.node.textContent = 'Error creating widget: could not find model';
+            this.addClass('jupyter-widgets');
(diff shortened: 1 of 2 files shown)


jasongrout:
Thanks!

PR #2182, opened 2018-08-25
Add support for specifying custom icons in ToggleButtons

astrofrog:
This adds support for specifying custom icons via the data URI mechanism in ToggleButtons. Here is a preview:

screen shot 2018-08-25 at 16 29 33

A few open questions:

* One thing I wasn't quite sure about is how the icon size should be handled, and how to make sure the icons are properly centered (in the example above there is a slight vertical offset for the two custom icons) - I'm not familiar enough with the notebook CSS.

* What is the mechanism for adding tests? I can't see any tests currently for ToggleButtons currently?

* I can add the same mechanism for ToggleButton once things are polished with ToggleButtons.

_Impremented during the JupyterCon 2018 sprint_

### DIFF
--- a/ipywidgets/widgets/widget_button.py
+++ b/ipywidgets/widgets/widget_button.py
@@ -63,16 +63,6 @@ class Button(DOMWidget, CoreWidget):
         self._click_handlers = CallbackDispatcher()
         self.on_msg(self._handle_button_msg)
 
-    # @validate('icon')
-    # def _validate_icon(self, proposal):
-    #     """Strip 'fa-' if necessary'"""
-    #     value = proposal['value']
-    #     if value.startswith('fa-'):
-    #         warnings.warn("icons names no longer start with 'fa-', "
-    #         "just use the class name itself (for example, 'check' instead of 'fa-check')", DeprecationWarning)
-    #         value = value[3:]
-    #     return value
-
     def on_click(self, callback, remove=False):
         """Register a callback to execute when the button is clicked.
 
--- a/ipywidgets/widgets/widget_button.py
+++ b/ipywidgets/widgets/widget_button.py
@@ -11,7 +11,8 @@ from .domwidget import DOMWidget
 from .widget import CallbackDispatcher, register, widget_serialization
 from .widget_core import CoreWidget
 from .widget_style import Style
-from .trait_types import Color, InstanceDict
+from .widget_media import Icon
+from .trait_types import Color, InstanceDict, InstanceString
 
 from traitlets import Unicode, Bool, CaselessStrEnum, Instance, validate, default
 import warnings
@@ -49,7 +50,7 @@ class Button(DOMWidget, CoreWidget):
     description = Unicode(help="Button label.").tag(sync=True)
     tooltip = Unicode(help="Tooltip caption of the button.").tag(sync=True)
     disabled = Bool(False, help="Enable or disable user changes.").tag(sync=True)
-    icon = Unicode('', help="Font-awesome icon name, without the 'fa-' prefix.").tag(sync=True)
+    icon = InstanceString(Icon, Icon.fontawesome, default_value=None, allow_none=True, help= "Optional button icon.").tag(sync=True, **widget_serialization)
 
     button_style = CaselessStrEnum(
         values=['primary', 'success', 'info', 'warning', 'danger', ''], default_value='',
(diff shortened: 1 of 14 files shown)


jasongrout:
Thanks!

I've been thinking more about this, and particularly thinking about consistency with the Image widget. What do you think of having a new attribute, `icon_format`, that can be a mimetype, or the string 'url', or the string 'fontawesome' (defaulting to 'fontawesome'). The icon attribute itself would be changed to binary, and the format string would be used to determine how to interpret the icon attribute.

This would be a backwards incompatible change, and might be overkill, considering icons usually are fairly small, so it's not a big deal to transmit them as data URIs rather than in binary. It does make it slightly easier to give an image, as in you don't have to construct a data uri for the image (you can just read in the binary image and set the attribute to the data). Thoughts? How easy is it to construct data URIs in python?

maartenbreddels:
What about making icon a union of string and image? String means
fontawesome otherwise an image

(from mobile phone)

jasongrout:
Do you mean the icon attribute would actually be a reference to an image widget?

maartenbreddels:
Indeed, so all logic of reading images etc on the kernel side is in that
class.

(from mobile phone)

jasongrout:
I don't think we have precedent for having an attribute that is the union of a widget reference or a string. @SylvainCorlay - do you have thoughts about having a union attribute of either a widget reference or a string? In the serialization, this really means that if the string starts with the magic value, it's considered a widget reference, otherwise it's a fontawesome name.

SylvainCorlay:
The icon trait could be of a special `Icon` type.

Assigning a string would create a font-based icon upon traitlet validation. To get an image, explicitely instantiate an image-based icon.

maartenbreddels:
I like it, so icon could be similar to Image, except it would create a li
dom element for format=='fontawesome'? And button would create a view, and
would attach that Dom element?

(from mobile phone)

SylvainCorlay:
Yes, probably. We also need to do some work on line height for the containing element for centering.

maartenbreddels:
I've turned all icon/icons into Icon widgets/List of icon widgets. Icon is basically Image + fontawesome support. 
See this demo for features, setting a custom .svg image as icon.
icon

maartenbreddels:
I've started supporting non-unique labels with this PR, but it would make sense to do this with all of the _Selection widget. Shall I leave this in this PR and once we agree do another PR to support it in the others as well?

ipywidgets/widgets/widget_button.py
```diff
@@ -62,15 +63,15 @@ def __init__(self, **kwargs):
...
-        return value
+    # @validate('icon')
+    # def _validate_icon(self, proposal):
+    #     """Strip 'fa-' if necessary'"""
+    #     value = proposal['value']
+    #     if value.startswith('fa-'):
+    #         warnings.warn("icons names no longer start with 'fa-', "
+    #         "just use the class name itself (for example, 'check' instead of 'fa-check')", DeprecationWarning)
+    #         value = value[3:]
+    #     return value
```
SylvainCorlay: leftover?

ipywidgets/widgets/widget_media.py
```diff
@@ -170,6 +170,45 @@ def __repr__(self):
...
+    method.
+
+    """
+    _view_name = Unicode('IconView').tag(sync=True)
+    _model_name = Unicode('IconModel').tag(sync=True)
+
+    # Define the custom state properties to sync with the front-end
+    format = Unicode('png', help="The format of the icon.").tag(sync=True)
+    width = CUnicode(help="Width of the icon in pixels.").tag(sync=True)
+    height = CUnicode(help="Height of the icon in pixels.").tag(sync=True)
```
SylvainCorlay: maybe not "in pixel". Should we say 'width/height' of the icon DOM element.

packages/controls/src/widget_bool.ts
```diff
@@ -194,16 +202,22 @@ class ToggleButtonView extends DOMWidgetView {
...
             } else {
                 this.el.textContent = '';
-                if (icon.trim().length) {
-                    let i = document.createElement('i');
-                    this.el.appendChild(i);
-                    i.classList.add('fa');
-                    i.classList.add('fa-' + icon);
+                if (icon) {
+                    this.iconView = <IconView> await this.create_child_view(icon)
+                    if (description.length === 0) {
```
maartenbreddels: since of the async we should first put in a placeholder element.

maartenbreddels:
@SylvainCorlay issues addressed that we discussed, should we merge?

maartenbreddels:
oops, I see I need to follow the same pattern in ToggleButton

maartenbreddels:
Ok, this PR led me to find https://github.com/jupyter-widgets/ipywidgets/pull/2256, we should first solve that and then rebase this on master.

maartenbreddels:
With #2256 in and some cleanups, this is ready for a new review or merge.

maartenbreddels:
I think with https://github.com/mariobuikhuizen/ipyvuetify I personally don't need this anymore. Happy for someone else to pick this up.

PR #2185, opened 2018-08-27
Allow setting custom CDN for loading widgets.

tarzzz:
Details: https://github.com/jupyter-widgets/ipywidgets/issues/1627#issuecomment-415456828

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -1,8 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-declare var __widgets_cdn__:string;
-__widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
+const __widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
 
 import * as libembed from './libembed';
 
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -1,6 +1,9 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+declare var __widgets_cdn__:string;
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks!

Do you know if any other CDN supports semver suffixes in URLs like unpkg?

tarzzz:
[jsdelivr](https://www.jsdelivr.com/) does.

vidartf:
Seems nice! Is there a natural place to add documentation for this?

tarzzz:
@SylvainCorlay Any concerns with merging this PR ..?

packages/html-manager/src/libembed-amd.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+const __widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
```
jasongrout: I think it might be a bit cleaner if we make things more specific, like `__jupyter_widgets_cdn__`, or if we could figure out a nice way to not have a global variable.

tarzzz: Updated the var name to `__jupyter_widgets_cdn__` : )

maartenbreddels:
I think it is ok like this, but indeed, maybe we can do without a global, ala requirejs `data-main`.
```html
<script type="text/javascript" data-cdn-prefix="https://unpkg.com" src="https://unpkg.com/@jupyter-widgets/html-manager@^0.14.0/dist/embed-amd.js"></script>
```
I don't know how you can access this attribute, but requirejs seems to be able to do it.

tarzzz:
@maartenbreddels https://github.com/jupyter-widgets/ipywidgets/pull/2191 looks good to me, but I will let the team members @SylvainCorlay or @jasongrout decide : )

jasongrout:
Closing in favor of #2191. Thanks for pushing this forward, @tarzzz!

@maartenbreddels - it'd be great if you could mention @tarzzz in at least a commit message, or rebase your PR on top of this so @tarzzz gets credit in the repo for the work done here.

PR #2191, opened 2018-08-31
Make the CDN configurable

maartenbreddels:
An alternative to #2185 using a custom data-cdn attribute. Note that this in principle will accept this attribute from any script tag.

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -1,8 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-declare var __widgets_cdn__:string;
-__widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
+const __widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
 
 import * as libembed from './libembed';
 
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -1,6 +1,9 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+declare var __widgets_cdn__:string;
+__widgets_cdn__ = (window as any).__widgets_cdn__ || 'https://unpkg.com';
+
 import * as libembed from './libembed';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/html-manager/src/libembed-amd.ts
```diff
@@ -3,6 +3,14 @@
 
 import * as libembed from './libembed';
 
+let cdn = 'https://unpkg.com/';
+
+// find the data-cdn for any script tag, assuming it is only used for embed-amd.js
+const scripts = document.getElementsByTagName('script');
+Array.prototype.forEach.call(scripts, (script) => {
```
jasongrout: The standard trick here is to just look at the *last* script tag, which should be the one that loaded this script. I'm not sure how doing an async script tag affects this, though.

On the other hand, it looks like `currentScript` is a thing: https://stackoverflow.com/questions/403967/how-may-i-reference-the-script-tag-that-loaded-the-currently-executing-script

currentScript is supported on all the browsers that we care about: https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript

jasongrout: Some comments on that stack overflow post seem to indicate that the `currentScript` sometimes doesn't work, but the last script does? Maybe we implement currentScript and if doesn't work in all cases, we fall back to the last script tag?

maartenbreddels: By the time this code snippet gets executed, document.currentScript is null and all script tags are already all there (I guess it is after document.onload, already deep in require's callstack). So it is more robust to go over all script tags. Maybe avoid a name collision by renaming the attribute to `data-jupyter-widgets-cdn` ?

maartenbreddels: FYI, i'm copying requirejs' behaviour with its data-main attribute, I guess they also have a reason for not trying to be smarter.

maartenbreddels: I have no idea how using webpack+requirejs document.currentScript can ever be used, so I'd go with the `data-jupyter-widgets-cdn`.

tarzzz: @jasongrout Yes, that looks good to me.

tarzzz:
@maartenbreddels Yes, I checked this locally and it works well for me. Thanks for the fix. 

Is there an approximate to when this will be available in a release (after merging)?

maartenbreddels:
I see I still need to update the docs, but jason releases quite often actually, is it urgent?

tarzzz:
It would be nice to get it in a release soon. It's blocking some of other stuff I am working on.

jasongrout:
Looks good!

jasongrout:
> It would be nice to get it in a release soon. It's blocking some of other stuff I am working on.

Since html manager is still pre-1.0, we can release this without waiting for ipywidgets 7.5. I'll look through and see if there are other issues that need to merged, and then release sometime this week.

Thanks again!

tarzzz:
@jasongrout Checking in on when can I expect this to be available in a release.. Thanks.. !! 😸

jasongrout:
> @jasongrout Checking in on when can I expect this to be available in a release.. Thanks.. !! 😸

Done!

PR #2205, opened 2018-09-10
Fix JS remove_buffers check for data views

### LINKED ISSUE #1748: serializers returning an object with buffer key in it get treated as buffers
https://github.com/jupyter-widgets/ipywidgets/blob/24f3312897f2694f10fddd207206af0aae10ecce/packages/base/src/utils.ts#L184 and https://github.com/jupyter-widgets/ipywidgets/blob/24f3312897f2694f10fddd207206af0aae10ecce/packages/base/src/utils.ts#L155 are too generous, they should whitelist specific types etc, I noticed this when I returned 
`{'buffer': ar.data.buffer, dtype:arrayToTypes[ar.data.constructor.name], shape:ar.shape}` from a serializer. On the python side I got passed a memory object, instead of a dict.
Might be worth to take a look at https://github.com/jashkenas/underscore/pull/2694/files#diff-0f36b362a0b81d6f4d4bfd8a7413c75dR62 for solving this.

PS: Btw, a simple workaround is not to name it buffer, but other people may get bitten by this.

vidartf:
Uses `ArrayBuffer.isView(value)` instead of `value.buffer instanceof ArrayBuffer` to check if `value` is a typed array or a `DataView`.

Fixes #1748.

### DIFF
--- a/packages/base/src/utils.ts
+++ b/packages/base/src/utils.ts
@@ -167,12 +167,12 @@ function remove_buffers(state): {state: any, buffers: ArrayBuffer[], buffer_path
             for (let i = 0; i < obj.length; i++) {
                 let value = obj[i];
                 if (value) {
-                    if (value.buffer instanceof ArrayBuffer || value instanceof ArrayBuffer) {
+                    if (value instanceof ArrayBuffer || ArrayBuffer.isView(value)) {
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Indeed, thanks a lot!

PR #2213, opened 2018-09-16
Fix Image display inconsistency between classic and lab

mwcraig:
Display of an image widget is different in classic notebook and lab. This PR fixes that by adding the CSS necessary to make an Image widget display in Lab as it does in notebook. Though one could argue as to which behavior is better, the behavior in classic notebook has been present for a long time and so I think the right fix is to make lab behave the same way.

### Input image

All of the examples below use a 600 x 300 image of a gaussian as input; a copy is below.

gaussian_600_x_300

### plain `Image` widget output differs

The code cell below generates different output in classic than in lab:

```python
i = Image(value=b)
i.width = 100
i.height = 150
i
```

| Classic notebook | Lab  |
| ----------------- | ---- |
| Preserves aspect ratio | Is 100 x 150 |
|             untitled               |      jupyterlab      |

I believe the root cause is this CSS snippet from [outputarea.less in classic notebook](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/less/outputarea.less#L70):

```css
div.output_area img, div.output_area svg {
  max-width: 100%;
  height: auto;
}
```

In this case, `height: auto` overrides `img height='150'` and the aspect ratio is preserved.

### `Image` in a `Box` output differs

This code snippet also generates different results in classic than in lab:

```python
i2 = Image(value=b)
box2 = Box()
i2.width = 500
box2.layout.width = '200px'
box2.children = [i2]
box2
```
| Classic notebook | Lab  |
| ----------------- | ---- |
| Squeezes image | 200px wide chunk displayed, image distorted |
|       untitled       |         jupyterlab       |

Again, I believe the difference is due to the CSS snippet above. In this case, in the classic notebook, the maximum width of the image is set to `100%`, forcing it to fit inside the box. In lab, one can scroll to see the rest of the image because the `Box` has `overflow: visible`.

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -1074,3 +1074,11 @@
     /* Make it possible to have absolutely-positioned elements in the html */
     position: relative;
 }
+
+
+/* Image widget  */
+
+.widget-image {
+    max-width: 100%;
+    height: auto;
+}


mwcraig:
I think this should be closed in favor of #2213

SylvainCorlay:
> I think this should be closed in favor of #2213

Hey @mwcraig thanls for fixing this. Which PR are you refering to here? (You are including a link to the current PR.)

SylvainCorlay:
Ah, found it in #2214

SylvainCorlay:
I am willing to get this in and discuss the more intrusive #2214 after 7.5.

@jasongrout do you want to merge this?

jasongrout:
> I am willing to get this in and discuss the more intrusive #2214 after 7.5.

@mwcraig - how does that plan sound to you?

mwcraig:
> @mwcraig - how does that plan sound to you?

See https://github.com/jupyter-widgets/ipywidgets/pull/2214#issuecomment-486665820, but in a nutshell I think this version needs, at a minimum, something in the documentation explaining the styling of images, and ideally the CSS we are applying is added to a default `Layout` attached to the `Image` widget, so that it is discoverable/more easily settable from Python.

This PR as it is now is still definitely better than the current state of affairs....different behavior out of the box in lab and classic is undesirable.

SylvainCorlay:
So I went ahead and merged this version (tested locally as well) as it is the least intrusive, it fixes the issue, and makes jupyterlab behave consistent with the classic notebook behavior.

Also, I think we should continue discussing the other PR which could potentially be a cleaner solution in the future.

SylvainCorlay:
> And ideally the CSS we are applying is added to a default Layout attached to the Image widget, so that it is discoverable/more easily settable from Python.

Commenting on that one specifically:

All core widgets that have some layout-related CSS applied to their top element have this issue. For example, Button, Slider, etc all have the same height, which is a "natural height" for basic widgets, which cannot be read from the Python API.

Should `max_height` be set through css, this would also prevent `layout.height` to fully take effect unless they also set `layout.max_height`, but theire is not means to learn about that CSS property being applied from the Python API and it will probably never be the case.

Hence, I don't think we should add specific "Image" documentation, as much as finding a more appropriate means to deal with "natural" layout properties (set through CSS), and layout properties set through ipywidgets' APIs.

PR #2214, opened 2018-09-17
Add object_fit to Layout, document how it applies to  Image, and fix Lab/Classic inconsistency

mwcraig:
Unfortunately this ends up encompassing a few things.

1. As an alternative to #2213, explicitly set `Image.layout.max_width` and `Image.layout.height` to the values they had as a result of notebook CSS. While #2213 accomplishes the same thing by changing the CSS applied to an Image widget, it still leaves the user in the dark about layout that is being applied to the element. I prefer this explicit approach.
2. Add the `object_fit` CSS property to `Layout`. `object_fit` is used to specify how replaced content like an image fills a container. See https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit for description and browser compatibility (everything except IE).
3. Add lots of documentation about how to get an image positioned inside a container.


mwcraig:
## One problem this is trying to solve

### Input image

All of the examples below use a 600 x 300 image of a gaussian as input; a copy is below.

gaussian_600_x_300

### plain `Image` widget output differs

The code cell below generates different output in classic than in lab:

```python
i = Image(value=b)
i.width = 100
i.height = 150
i
```

| Classic notebook | Lab  |
| ----------------- | ---- |
| Preserves aspect ratio | Is 100 x 150 |
|             untitled               |      jupyterlab      |

I believe the root cause is this CSS snippet from [outputarea.less in classic notebook](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/less/outputarea.less#L70):

```css
div.output_area img, div.output_area svg {
  max-width: 100%;
  height: auto;
}
```

In this case, `height: auto` overrides `img height='150'` and the aspect ratio is preserved.

### `Image` in a `Box` output differs

This code snippet also generates different results in classic than in lab:

```python
i2 = Image(value=b)
box2 = Box()
i2.width = 500
box2.layout.width = '200px'
box2.children = [i2]
box2
```
| Classic notebook | Lab  |
| ----------------- | ---- |
| Squeezes image | 200px wide chunk displayed, image distorted |
|       untitled       |         jupyterlab       |

Again, I believe the difference is due to the CSS snippet above. In this case, in the classic notebook, the maximum width of the image is set to `100%`, forcing it to fit inside the box. In lab, one can scroll to see the rest of the image because the `Box` has `overflow: visible`.

maartenbreddels:
So `object_fit` would solve the above issues, or do you think something needs to change in jupyter lab?

mwcraig:
Adding `object_fit` allows a straightforward way to dictate scaling in both classic and lab.

Setting `Image.layout.max_width = '100%` and `Image.layout.height = 'auto'` will ensure it same behavior in both classic and lab. 

I don't know that those choices for width/height were a deliberate choice made in ipywidgets; the CSS that gets applied comes from the classic notebook CSS.

I think explicitly setting them in the widget (at least in the 7.x series) fixes the difference between classic and lab and makes more explicit to the user what styling is being applied in the case of classic.

ipywidgets/widgets/widget_media.py
```diff
@@ -159,8 +159,19 @@ class Image(_Media):
...
+                           "for styling the widget.").tag(sync=True)
+
+    def __init__(self, *args, **kwargs):
+        super(Image, self).__init__(*args, **kwargs)
+        # This is done to make explicit a CSS rule that was applied
+        # to all img tags, including the Image widget, in the
+        # classic notebook. Removing it in the future would not
+        # be unreasonable.
+        self.layout.max_width = '100%'
+        self.layout.height = 'auto'
```
vidartf: There will be a conflict here if I supply a layout in args/kwargs. Maybe this should only be set if the layout used is the default one?

mwcraig:
Good catch @vidartf, I addressed that.

mwcraig:
Friendly bump....if this looks good I'll close #2213

mwcraig:
Friendly bump....if this looks good I'll close #2213

SylvainCorlay:
@mwcraig @vidartf I am thinking of including #2213 as the least intrusive change for the 7.5 release, and leave this one open. Thoughts?

mwcraig:
Not surprisingly, I’d prefer this one. If we want something like #2213 instead of this I think it would be helpful to set the styling traits explicitly on the Image in python instead of implicitly in the CSS, and add some documentation about image styling. 

I could add the python bits tomorrow could move some docs over to #2213 on Saturday. 

The main reason I don’t want #2213 as-is is that there is no way to figure out from python the CSS style that has been applied to the image, which makes it really confusing to work out why images display the way they do in-browser.

vidartf:
I have no input on this at this time. My previous comment was only on the consistency of the code. I have not had a look on the other bits, so cannot speak for/against.

SylvainCorlay:
If we add `object-fit`, it probably makes sense to add `object-position` as well?

mwcraig:
Yeah, it probably should. Don't remember why I didn't include it originally. I'll rebase and add `object-position` too.

mwcraig:
`object-position` added, but want to work on the documentation a bit more. Should be ready for a review later today.

mwcraig:
This is ready for a look, especially at the documentation piece of this. I may have gone a little overboard there...

ipywidgets/widgets/widget_media.py
```diff
@@ -150,8 +150,21 @@ class Image(_Media):
...
+    def __init__(self, *args, **kwargs):
+        super(Image, self).__init__(*args, **kwargs)
+        # This is done to make explicit a CSS rule that was applied
+        # to all img tags, including the Image widget, in the
+        # classic notebook. Removing it in the future would not
+        # be unreasonable.
+        if self.layout.max_width is None:
+            self.layout.max_width = '100%'
+        if self.layout.height is None:
+            self.layout.height = 'auto'
```
SylvainCorlay: Why is this required?

SylvainCorlay:
@mwcraig this looks good to me. I would like to get this in for 7.5.

I just left a minor comment inline. Let me know what you think.

mwcraig:
It isn't necessary since #2213 was merged. I can remove tonight (more likely saturday).

mwcraig:
@SylvainCorlay -- updated as requested

SylvainCorlay:
Yeah!

PR #2219, opened 2018-09-21
Use pytest instead of nose for tests

mwcraig:
The main reason for making this switch is that nose is no longer being developed, though I also find the pytest style of writing tests a little cleaner.

### DIFF
--- a/docs/source/dev_testing.md
+++ b/docs/source/dev_testing.md
@@ -3,7 +3,7 @@ Testing
 
 To run the Python tests:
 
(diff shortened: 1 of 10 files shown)


jasongrout:
Thank you!!

jasongrout:
> Would you mind getting rid of those too?

I had a few extra minutes, so fixed those mentions of nose. I think we're good to go if tests pass. Thanks!

jasongrout:
Thanks again!

mwcraig:
My pleasure, thanks for the cleanup!

PR #2222, opened 2018-09-21
Respect allow_none for Color, add FullColor

vidartf:
- Respects `allow_none` for `Color` trait.
- Ensure `Color.validate` raises a `TraitError` if value is not a string type, instead of `AttributeError`. E.g. `Union.validate` relies on catching `TraitError`.
- Adds trait type `FullColor` that uses the expanded HTML color definitions.


vidartf:
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

vidartf:
Maybe a better name than "FullColor" exists. What about `CSSColor` ?

SylvainCorlay:
I like the validation of rgba color.

How about making it part of the base color class?

jasongrout:
Is there a reason not to just expand the base color class with no option to restrict it to the current definition? Why not just adopt the CSS color spec completely in the base color class?

vidartf:
If you figure this as backwards compatible enough to not wait for a major release, I'd be happy to add the changes to Color!

jasongrout:
I think if it is strictly expanding the allowed values, it's okay to put in a minor release.

vidartf:
Now expands the Color trait!

maartenbreddels:
awesome work, looking good!

SylvainCorlay:
As per discussion with @jasongrout we wanting on @jasongrout to make a patch release for jupyterlab 0.35 before merging this.

vidartf:
@jasongrout I've lost track here. Is the relevant patch release out?

jasongrout:
There are some patch releases out. Can you remind me why we were waiting for a jlab release for this?

vidartf:
> Can you remind me why we were waiting for a jlab release for this?

Not really. @SylvainCorlay / @maartenbreddels do either of you remember?

SylvainCorlay:
Sounds good to me!

vidartf:
TODO: Include HSL values in tests.

SylvainCorlay:
@vidartf I am fine with merging this now (for the 7.5 minor release).

Is there anything that you think we should add?

vidartf:
@SylvainCorlay I added some simple HSL(A) tests (the original ones I wrote got lost in the previous refactor), so now I am happy.

SylvainCorlay:
:tada:

SylvainCorlay:
Thanks @vidartf this is going to be super useful!

PR #2232, opened 2018-10-01
WIP: Menu widget based on phosphorjs

maartenbreddels:
(requires #2182 )
Adds Menu option to ipywidgets (still working out the details, very preliminary).
The plan is for a Button to have a Menu trait, like icon, so it can have a popup menu. We could also think of Menubars etc later on, or context menus.

menus

### DIFF
--- a/application.ipynb
+++ b/application.ipynb
@@ -1,0 +1,318 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Simple example"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ipywidgets as widgets"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "menu_maximize   = widgets.Menu(description='Fullscreen', icon='arrows-alt', command='fullscreen')\n",
+    "menu_fullscreen = widgets.Menu(description='Maximize', icon='window-maximize', command='maximize')\n",
+    "menu_restore = widgets.Menu(description='Restore', icon='window-restore', command='restore')\n",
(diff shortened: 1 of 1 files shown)


maartenbreddels:
The design is pretty similar to Qt, actually, I added the Action widget, just becase otherwise Button.default_action would be called Button.default_menu, but I think we could do without the Action widget.
I think we should discuss this at the next telecon.
A preview:

screen shot 2018-10-05 at 19 36 27

# TODOS
 - [ ] when delay=None, the caret should be its own button, like bootstrap's split button. This is also how Qt behaves.
 - [ ] Resolve circular dependency for context menu both in ipywidgets and jupyter-widget/base+control
 - [ ] Have a menu bar
 - [ ] Think about an Application Widget, that will have a menu, status bar, toolbar with fullscreen button?

maartenbreddels:
I've added an Application widget, which includes a menubar and a toolbar. The menubar is actually just a Menu widget, since otherwise a MenuBar widget would be a meaningless Container class. The toolbar is actually just a List of DOMWidgets, but added do a div with a special class to allow some custom styling through css.
The way that evens are delivered through from view to view is using the DOM Event mechanism. All toolbar and menu dom elements can produce a custom event, which bubbles up till it reaches the Application view, and gets send to the central_widget_view, or a default handler is called.

It's all rough, but I think it shows where I want to go, and allows us to create really nice UI's for bqplot and ipyvolume (using only jslink), as this demo hopefully shows:

ipywidgets-application

Discussion material for our next telecon I'd say!

PS: Functionality like this would make this PR https://github.com/bloomberg/bqplot/pull/724 obsolete.

martinRenou:
It looks really good! Do you use jupyter-lab CSS variables in order to support dark theme?

maartenbreddels:
No, the CSS is a quick copy paste + edit from http://phosphorjs.github.io/examples/menus/, some pointers/help on how to proceed with the css would be appreciated

martinRenou:
Sure! I can help. Do I have commit rights on your branch?

maartenbreddels:
Yes, you should have.

PR #2238, opened 2018-10-06
JupyterLab 0.35 upgrade

jasongrout:
This only does a very minimal upgrade to JupyterLab 0.35.

Later on, we should upgrade Typescript, etc.


jasongrout:
CC @athornton

PR #2244, opened 2018-10-13
feat(packages): IntText now ignores decimal portion of numbers

### LINKED ISSUE #1999: IntText Displays Invalid Values
If I enter a decimal value into an InText, it will show me exactly what I typed in, but the value will be cast when I go to retrieve it. Instead, it should block typing in a '.' or verify the number and change it to the cast value when on unfocus.

image

Widget Version: 7.0.3

realslimshanky:
Closes #1999


SylvainCorlay:
Closing as superseeded by #2354.

PR #2265, opened 2018-11-15
Allows lab manager to load state from kernel/notebook

vidartf:
When connecting to a new kernel in lab, this code now:
- Loads all live widgets from the kernel.
- Loads all stored widgets from the notebook (where the comm id does not overlap with a live one).

This attempts to mirror the logic in the widgetsnbextension code, and mostly the code is a hard copy (not DRY as such, and maintainers should note the possible need to keep these in sync in the future).


vidartf:
One issue for a TODO:
- [ ] When restarting kernel, widgets are not set as disconnected.

jasongrout:
> Note, this resolves a TODO on `get_model` that might not be fully backwards compatible.

Unfortunately, this makes it backwards incompatible, which technically means we should bump the base package up a major version number. It feels like we should have a bigger carrot for a major version bump in the base package.

How about we separate these two changes, so we can merge the jlab manager changes in immediately with essentially no change in base, and we take up fixing the get_model function in the next base major release?

(I'm okay with the typings enhancements you did in base, as well as the function you added, so I imagine this would be a minor base version bump, just not a major one yet.)

jasongrout:
> Unfortunately, this makes it backwards incompatible, which technically means we should bump the base package up a major version number.

The reason I think we should have a bigger reason to bump a major version is that a major bump of base requires everyone to release their widgets again (presuming they use the standard `^1` semver for base).

Maybe this conflation points us to having the widget manager base as a separate package from the widgets base package in some future version.

jasongrout:
@vidartf - FYI, I'm working on further enhancements to this PR...

vidartf:
@jasongrout OK, give me a ping when you are satisfied, and I'll have a look.

jasongrout:
I rebased on master to pull in changes needed to work with the current jlab prerelease.

jasongrout:
@vidartf - I think this is good to look at now. I also rolled in #2326, which gives us:

* move to using the tracker instead of the jlab nb widget
* widgets work across new views, since the widget manager is tied to the context now.
* widget renderers wait until a widget manager is initialized before rendering. We have to do this since a widget manager won't be assigned until *after* the initial rendering of the document.

jasongrout:
I added automatic saving of live widget state to the notebook metadata when a preference is set.

jasongrout:
I also added a settings menu item to toggle the widget save setting.

jasongrout:
If the widget save setting is 'false', should we clear the existing widget state in the notebook, or just leave it alone? Or perhaps we should clear it from the model after we load it?

Also, does it seem reasonable to only save live widgets? This may be a good first attempt. This may mean that opening a notebook with saved state, running just a few cells, and saving means that other cells with saved widgets then stop working. Perhaps it's better to somehow save live widgets plus any widgets referenced from a view (hard)? Or perhaps any widgets that have been displayed at least once (again, probably hard)? Or perhaps we just have an option to save *all* widget state, including unseen and unused widgets (easy, but confusing, I think)?

jasongrout:
And another possible TODO: right now, if we can't find a widget model, we just leave a "Loading widget..." message. Perhaps we should start a timeout when we do that, so that if the model doesn't load after a while (5 seconds?), we just explain there is an error, rather than having the user keep waiting for the widget to load.

jasongrout:
> And another possible TODO: right now, if we can't find a widget model, we just leave a "Loading widget..." message. Perhaps we should start a timeout when we do that, so that if the model doesn't load after a while (5 seconds?), we just explain there is an error, rather than having the user keep waiting for the widget to load.

I did this in the latest commit.

jasongrout:
Wrapping this up until we can discuss with @vidartf (or whoever else wants to review it). This I think requires a minor version bump in base and a major version bump in jlab manager.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -111,31 +153,111 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+
+  async _loadFromKernel(): Promise<void> {
+    if (!this.context.session.kernel) {
+      return;
+    }
+    const comm_ids = await this._get_comm_info();
+
+    // For each comm id, create the comm, get the state, and create a widget.
+    const widgets_info = await Promise.all(Object.keys(comm_ids).map(async (comm_id) => {
+      const comm = await this._create_comm(this.comm_target_name, comm_id);
```
vidartf: Note: This changes the sync logic (does not await all `_create_comm` first), but I do not think this is significant.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -213,14 +332,114 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+        }
+      }
+      return serialize_state(models, options);
+  }
+
+  /**
+   * Set the dirty state of the notebook model if applicable.
+   *
+   * TODO: should also set dirty when auto-save state is turned on (easy), and when any
+   * model changes any data (likely hard right now).
```
vidartf: Note to self: Just before merging, if this TODO note is still here, it should be replaced by an issue.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -213,14 +332,114 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+   * Get a model
+   *
+   * #### Notes
+   * Unlike super.get_model(), this implementation always returns a promise and
+   * never returns undefined. The promise will reject if the model is not found.
+   */
+  async get_model(model_id: string): Promise<WidgetModel> {
+    const modelPromise = super.get_model(model_id);
+    if (modelPromise === undefined) {
+      throw new Error('widget model not found');
```
vidartf: @jasongrout I initially [added some code here](https://github.com/jupyter-widgets/ipywidgets/pull/2265/commits/6078163210f4c579fb69dc1fa8226657dbdfa913#diff-d4bf928d4fe8ff74753016095895f635R309) where if widgets were in the process of being restored from live / state, this would await the completion of the restore. Was this not needed?

vidartf:
> This may mean that opening a notebook with saved state, running just a few cells, and saving means that other cells with saved widgets then stop working.

Not really, as widgets in the state will be restored (will become live). Your scenario will only happen if you are working super fast (and/or the restore is super slow). Still, it could leave to some weird behavior.

packages/jupyterlab-manager/src/renderer.ts
```diff
@@ -16,51 +20,84 @@ import {
...
+      // Let's be optimistic, and hope the widget state will come later.
+      this.node.textContent = 'Loading widget...';
       this.addClass('jupyter-widgets');
-      return Promise.resolve();
+
+      // If we haven't rerendered in 5 seconds with the right model, display an
+      // error message to the user.
+      setTimeout(() => {
+        if (this._rerenderMimeModel) {
+          this.node.textContent = 'Error displaying widget: model not found';
```
vidartf: Maybe we should tweak this error message to indicate that, while no actual error has occurred, getting the widget state is taking a lot longer than expected.

packages/jupyterlab-manager/src/plugin.ts
```diff
@@ -1,101 +1,189 @@
...
+  if (!wManager) {
+    wManager = new WidgetManager(context, rendermime, SETTINGS);
+    WIDGET_REGISTRY.forEach(data => wManager.register(data));
+    Private.widgetManagerProperty.set(context, wManager);
+  }
 
+  // For any widgets that have already been rendered with the placeholder, set
+  // the renderer. This iteration structure is closely tied to the structure of
+  // the notebook widget.
+  nb.widgets.forEach(cell => {
```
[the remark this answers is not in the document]
jasongrout: I just pushed a commit that (a) refactors the awkward logic into a self-contained iterator, and (b) attempts to also iterate through linked views, using even *more* awkward logic that relies on specific jlab details. It's hard for me to test, though, since linked output views aren't restored in the latest jlab release.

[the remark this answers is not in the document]
jasongrout: (note that both of these awkward iterators are just an interim solution until there might be a more supported clean way of iterating through renderers in jlab itself)

packages/jupyterlab-manager/src/manager.ts
jasongrout: right, instead of doing things in batches, it does each as soon as it can. I think it's more streamlined. Note that I did put in a sync barrier so that hydrating widgets works well even when there are links between them.

packages/jupyterlab-manager/src/manager.ts
jasongrout: I simplified this function, and instead made the renderer responsible for rerendering when a restore happened: https://github.com/jupyter-widgets/ipywidgets/pull/2265/files#diff-cc0d118ac726eae45225b560ae664079R42

One of the reasons to do this was to make the restore procedure simpler, since the restore itself can use get_model to rehydrate references to other models.

packages/jupyterlab-manager/src/renderer.ts
jasongrout: I pushed a commit that tries to make the error messages more intelligent, based on whether the manager has restored its status or not.

jasongrout:
@vidartf - I think I addressed the issues you raised above. Thoughts now?

jasongrout:
> > This may mean that opening a notebook with saved state, running just a few cells, and saving means that other cells with saved widgets then stop working.
> 
> Not really, as widgets in the state will be restored (will become live). Your scenario will only happen if you are working super fast (and/or the restore is super slow). Still, it could leave to some weird behavior.

I was talking about dead widgets that only live in the notebook metadata, not restored widgets from the kernel.

packages/jupyterlab-manager/src/manager.ts
vidartf: After talking directly, we determined that the only dangling references for `get_model` would be from trying to render yet-unrestored models. The logic I added was therefore not needed (and introduced a possible deadlock). The current logic deals with such dangling references in the view.

vidartf:
> Not really, as widgets in the state will be restored (will become live).

I was mistaken about this, widgets in the notebook without a kernel equivalent, will not become live (they won't get a comm).

packages/base/src/utils.ts
```diff
@@ -84,14 +84,14 @@ class WrappedError extends Error {
  * Returns a single Promise.
  */
 export
-function resolvePromisesDict(d): Promise<any> {
+function resolvePromisesDict<V>(d: {[keys: string]: PromiseLike<V>}): Promise<{[keys: string]: V}> {
```
vidartf: I'll add a type declaration for this, as the inline typing is distracting.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -111,31 +153,112 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
     }
   }
 
+  _handleKernelStatusChange(args: Kernel.Status) {
+    switch (args) {
+    case 'connected':
+      // TODO: should we clear away any old widgets before restoring?
```
vidartf: Did we answer this? If so, we should change this comment to explain what we chose (and why).

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -78,7 +98,7 @@ class BackboneViewWrapper extends Widget {
  */
 export
 class WidgetManager extends ManagerBase<Widget> implements IDisposable {
-  constructor(context: DocumentRegistry.IContext<DocumentRegistry.IModel>, rendermime: RenderMimeRegistry) {
+  constructor(context: DocumentRegistry.IContext<INotebookModel>, rendermime: RenderMimeRegistry, settings: {saveState: boolean}) {
```
vidartf: I'll add a type for `{saveState: boolean}` as well.

packages/jupyterlab-manager/src/manager.ts
vidartf: Should we also reset `this._restoredStatus` here?

packages/jupyterlab-manager/src/plugin.ts
```diff
@@ -1,101 +1,236 @@
...
+  let linkedViews = filter(
+    app.shell.widgets(),
+    w => w.id.startsWith('LinkedOutputView-') && (w as any).path === path
+  );
+  for (let view of toArray(linkedViews)) {
+    for (let outputs of toArray(view.children())) {
+      for (let output of toArray(outputs.children())) {
+        if (output instanceof WidgetRenderer) {
+          yield output;
+        }
```
vidartf: @jasongrout Just for posterity: Would you mind detailing what the plan is for replacing these rather hackish iterators?

jasongrout: No specific plan yet. Basically, it would be nice to have a way in JLab to iterate through all renderers associated with a document. Before adding that to the JLab API, though, I'd love to have one more use case for this to double check that we have the right level of generality. I'm also thinking that we might find a more elegant way to do this in the future.

packages/jupyterlab-manager/src/manager.ts
jasongrout: Is this the only thing left to resolve here?

vidartf: I think so. I.e. both the TODO, and reviewing if `this._restoredStatus` should ever be reset (and if so, when and where).

jasongrout: I went ahead and reset the restored status in disconnect. I think it makes most sense to reset it there, since a reconnect will then restore widgets again.

vidartf: Was this actually done? Since you force pushed, the commit log doesn't show any partial diffs for this, so its super hard to track what was changed.

jasongrout: Sorry, and in general this PR was already hard to track being so long and involved. Are you asking if the TODO was fixed, or if the restored status was reset?

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -213,14 +333,131 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+
+
+  /**
+   * Close all widgets and empty the widget state.
+   * @return Promise that resolves when the widget state is cleared.
+   */
+  async clear_state(): Promise<void> {
+    await super.clear_state();
+    this._modelsSync = new Map();
+    this.setDirty();
```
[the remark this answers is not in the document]
jasongrout: Should we clear away old widgets before restoring? Actually, since we're only storing live widget models right, I think it doesn't hurt to leave old widget models in - we won't get file bloating from disconnected widgets.

jasongrout:
To answer the conversation above: should we clear away old widgets before restoring? Actually, since we're only storing live widget models right, I think it doesn't hurt to leave old widget models in - we won't get file bloating from disconnected widgets.

PR #2268, opened 2018-11-18
Code to implement issue 1936

sthapa:
PR to implement a convenience function for https://github.com/jupyter-widgets/ipywidgets/issues/1936

### DIFF
--- a/ipywidgets/widgets/widget_output.py
+++ b/ipywidgets/widgets/widget_output.py
@@ -140,6 +140,43 @@ class Output(DOMWidget):
         """Append text to the stderr stream."""
         self._append_stream_output(text, stream_name='stderr')
 
+    def register_logger(self, logger, *args, **kwargs):
+        """Registers a handler to given logger to send output to output widget"""
+
+        import logging
+
+        class WidgetLogger(logging.Handler):
+            """ Class to implement a logging interface that outputs to the
+             Output widget"""
+
+            # have a class member to store the existing logger
+            logger_instance = logging.getLogger("__name__")
+
+            def __init__(self, output_widget):
+                # Initialize the Handler
+                logging.Handler.__init__(self, *args)
+
+                # save outer_instance
+                self.output = output_widget
+
+                # optional take format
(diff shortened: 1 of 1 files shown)


marscher:
This seems very useful, why hasn't was this never merged?

ipywidgets/widgets/widget_output.py
```diff
@@ -140,6 +140,43 @@ def append_stderr(self, text):
...
+                self.logger_instance.addHandler(self)
+
+            def emit(self, record):
+                """ Overload of logging.Handler method """
+
+                record = self.format(record)
+                self.output.outputs = ({'name': 'stdout',
+                                                'output_type': 'stream',
+                                                'text': (record + '\n')},) + self.output.outputs
+        logger.addHandler(WidgetLogger(self))
```
jgunstone, line 178: i think this needs to be:

`logger.addHandler(WidgetLogger(self, *args, **kwargs))`

jgunstone (commented):
just a quick not that I think you need to pass the *args and **kwargs when registering the logger. 
this allows the "format" of the handler to be set

ipywidgets/widgets/widget_output.py
```diff
@@ -140,6 +140,43 @@ def append_stderr(self, text):
...
+                # make the logger send data to this class
+                self.logger_instance.addHandler(self)
+
+            def emit(self, record):
+                """ Overload of logging.Handler method """
+
+                record = self.format(record)
+                self.output.outputs = ({'name': 'stdout',
+                                                'output_type': 'stream',
+                                                'text': (record + '\n')},) + self.output.outputs
```
marscher, line 177: So there is no append for the output, right? For huge log outputs, this can crash both browser and more horrible Jupyter lab 4. So it'd be a nice improvement, if we could append to outputs, right?

sthapa: If there's interest in merging this and this approach looks reasonable, I can make the suggested changes and resolve the conflicts with the current code.

PR #2271, opened 2018-11-24
Link JL extensions in dev-install script

jtpio:
Looks like the `base`, `controls` and `output` packages need to be linked, so the changes can be picked up automatically when developing in JupyterLab.

### DIFF
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -52,5 +52,8 @@ echo -n "ipywidgets"
 pip install -v -e .
 
 if test "$skip_jupyter_lab" != yes; then
+    jupyter labextension link ./packages/base
+    jupyter labextension link ./packages/controls
+    jupyter labextension link ./packages/output
     jupyter labextension install ./packages/jupyterlab-manager
 fi


SylvainCorlay:
That is correct. This issues has bitten me a few times!

maartenbreddels:
Nice, thanks!

maartenbreddels:
I think link will rebuild right, maybe it should be:
```
if test "$skip_jupyter_lab" != yes; then
    jupyter labextension link ./packages/base --no-build
    jupyter labextension link ./packages/controls --no-build
    jupyter labextension link ./packages/output --no-build
    jupyter labextension install ./packages/jupyterlab-manager
fi
```

jtpio:
Right, `--no-build` will defer the build until the next `jupyter lab build`

jasongrout:
Does this work? These days, link is redundant with install (install on a local file just links). And both of those are just intended to work with actual JLab extensions.

jasongrout:
(I can imagine it might work, but there might be an assumption somewhere that the package is a jlab extension.)

vidartf:
> And both of those are just intended to work with actual JLab extensions.

I don't think this is correct. Link is explicitly meant to work with non-lab packages. If you use `install` with an npm package that does not have the `jupyterlab` key in `package.json`, it will tell you that it does a link instead. I would say that `link` is made for exactly these kind of situations.

PR #2272, opened 2018-11-26
Add kernel.py file containing kernel-related routines

martinRenou:
We need those changes for `xeus_python`. We implemented widgets support, but it enforces us to mock IPython an ipykernel at different places: `IPython.display.display`, `IPython.display.clear_output`, `IPython.get_ipython`, `ipykernel.comm.Comm`. See https://github.com/QuantStack/xeus-python/blob/master/src/xinterpreter.cpp#L41.
Those changes help by putting all those imports in one place: `kernel.py`. So that we just need to mock `ipywidgets.kernel` in `xeus_python`.

### DIFF
--- a/ipywidgets/__init__.py
+++ b/ipywidgets/__init__.py
@@ -31,9 +31,6 @@ def register_comm_target():
 
 # deprecated alias
 handle_kernel = register_comm_target
+_handle_ipython = register_comm_target
 
-def _handle_ipython():
-    """Register with the comm target at import if running in IPython"""
-    register_comm_target()
-
-_handle_ipython()
+register_comm_target()
--- a/ipywidgets/__init__.py
+++ b/ipywidgets/__init__.py
@@ -20,33 +20,20 @@ accessible as a `value` attribute.
 
 import os
 
-from IPython import get_ipython
 from ._version import version_info, __version__, __protocol_version__, __jupyter_widgets_controls_version__, __jupyter_widgets_base_version__
 from .widgets import *
 from traitlets import link, dlink
 
 
-def load_ipython_extension(ip):
-    """Set up IPython to work with widgets"""
-    if not hasattr(ip, 'kernel'):
-        return
-    register_comm_target(ip.kernel)
-
-
-def register_comm_target(kernel=None):
+def register_comm_target():
     """Register the jupyter.widget comm target"""
-    if kernel is None:
-        kernel = get_ipython().kernel
-    kernel.comm_manager.register_target('jupyter.widget', Widget.handle_comm_opened)
+    register_target('jupyter.widget', Widget.handle_comm_opened)
 
 # deprecated alias
 handle_kernel = register_comm_target
 
 def _handle_ipython():
     """Register with the comm target at import if running in IPython"""
-    ip = get_ipython()
-    if ip is None:
-        return
-    load_ipython_extension(ip)
+    register_comm_target()
 
 _handle_ipython()
--- a/ipywidgets/widgets/kernel.py
+++ b/ipywidgets/widgets/kernel.py
@@ -21,4 +21,4 @@ if (ipython is not None and hasattr(ipython, "kernel")):
 else:
     register_target = mock_register
 
-get_kernel = get_ipython
+get_ipython = get_ipython
--- a/ipywidgets/widgets/kernel.py
+++ b/ipywidgets/widgets/kernel.py
@@ -1,0 +1,19 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+"""Utils functions for retrieving IPython kernel and display logic."""
+
+from ipykernel.comm import Comm
+from IPython import get_ipython
+from IPython.display import display, clear_output
+
+def mock_register(*args, **kwargs):
+    pass
+
+ipython = get_ipython()
+if (ipython is not None and hasattr(ipython, "kernel")):
+    register_target = get_ipython().kernel.comm_manager.register_target
+else:
+    register_target = mock_register
+
+get_kernel = get_ipython
--- a/ipywidgets/widgets/kernel.py
+++ b/ipywidgets/widgets/kernel.py
@@ -1,7 +1,12 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-"""Utils functions for retrieving IPython kernel and display logic."""
+"""
+Utils functions for retrieving IPython kernel and display logic.
+This is helpful for other Python kernels like xeus-python, so that they can
+mock `ipywidgets.kernel` module functions instead of mocking IPython and
+ipykernel
+"""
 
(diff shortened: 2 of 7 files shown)


ipywidgets/widgets/widget.py
```diff
@@ -388,7 +386,7 @@ def get_view_spec(self):
 
     _view_count = Int(None, allow_none=True,
         help="EXPERIMENTAL: The number of views of the model displayed in the frontend. This attribute is experimental and may change or be removed in the future. None signifies that views will not be tracked. Set this to 0 to start tracking view creation/deletion.").tag(sync=True)
-    comm = Instance('ipykernel.comm.Comm', allow_none=True)
+    comm = Instance(Comm, allow_none=True)
```
SylvainCorlay: We should check that there was not a reason for using the string version.

In that case, it would be `ipywidgets.kernel.Comm`

ipywidgets/__init__.py
```diff
@@ -20,33 +20,20 @@
 
 import os
 
-from IPython import get_ipython
 from ._version import version_info, __version__, __protocol_version__, __jupyter_widgets_controls_version__, __jupyter_widgets_base_version__
 from .widgets import *
 from traitlets import link, dlink
 
 
-def load_ipython_extension(ip):
```
martinRenou: I'm not sure about the changes on the `__init__.py`, I didn't know what to keep, what to change, and what to remove.

ipywidgets/__init__.py
```diff
@@ -20,33 +20,20 @@
...
-    register_comm_target(ip.kernel)
-
-
-def register_comm_target(kernel=None):
+def register_comm_target():
     """Register the jupyter.widget comm target"""
-    if kernel is None:
-        kernel = get_ipython().kernel
-    kernel.comm_manager.register_target('jupyter.widget', Widget.handle_comm_opened)
+    register_target('jupyter.widget', Widget.handle_comm_opened)
```
maartenbreddels: Is this not imported, or am I missing something? If an error, please add a unittest to expose it.

ipywidgets/__init__.py
```diff
@@ -20,33 +20,20 @@
...
 # deprecated alias
 handle_kernel = register_comm_target
 
 def _handle_ipython():
     """Register with the comm target at import if running in IPython"""
-    ip = get_ipython()
-    if ip is None:
-        return
-    load_ipython_extension(ip)
+    register_comm_target()
```
maartenbreddels: Can't we skip this function? Just have register_comm_target and invoke that at import? Much simpler? If for compatibilty reasons, you could do: `_handle_ipython = register_comm_target`

ipywidgets/__init__.py
```diff
@@ -20,33 +20,20 @@
...
 
 _handle_ipython()
```
maartenbreddels: register_comm_target() instead?

ipywidgets/widgets/kernel.py
```diff
@@ -0,0 +1,19 @@
...
+def mock_register(*args, **kwargs):
+    pass
+
+ipython = get_ipython()
+if (ipython is not None and hasattr(ipython, "kernel")):
+    register_target = get_ipython().kernel.comm_manager.register_target
+else:
+    register_target = mock_register
+
+get_kernel = get_ipython
```
maartenbreddels: I think this is confusing, I'd expect: `get_kernel = lambda: get_ipython().kernel` If not, please add a comment explaining why

ipywidgets/widgets/widget.py
```diff
@@ -184,7 +182,7 @@ def __call__(self, *args, **kwargs):
             try:
                 local_value = callback(*args, **kwargs)
             except Exception as e:
-                ip = get_ipython()
+                ip = get_kernel()
```
maartenbreddels: Either `kernel = get_kernel` or `ip = get_ipython`

ipywidgets/widgets/widget.py
maartenbreddels: Could you explain why ipywidgets.kernel.Comm?

maartenbreddels (commented):
I like the title and comments and the description of the PR, and I'd really like to understand the issue, and the solution. Maybe a video call can help, or some more context. Happy to help with the PR though!

ipywidgets/widgets/widget.py
martinRenou: Because I moved the import of `Comm` in `ipywidgets.kernel`. I can change it to `'ipywidgets.kernel.Comm'` like Sylvain suggested, but `Comm` is imported in this file so it doesn't make a big difference.

ipywidgets/__init__.py
martinRenou: It is imported, through `from .widgets import *`

ipywidgets/__init__.py
martinRenou: Yes I can do that :)

ipywidgets/__init__.py
martinRenou: Yes

martinRenou:
We need those changes for `xeus_python`. We implemented widgets support, but it enforces us to mock IPython an ipykernel at different places: `IPython.display.display`, `IPython.display.clear_output`, `IPython.get_ipython`, `ipykernel.comm.Comm`. See https://github.com/QuantStack/xeus-python/blob/master/src/xinterpreter.cpp#L41.
Those changes help by putting all those imports in one place: `kernel.py`. So that we just need to mock `ipywidgets.kernel` in `xeus_python`.

Sorry that it wasn't clearer at first.

maartenbreddels:
> Sorry that it wasn't clearer at first.

No worries. Starting to understand it better.

I didn't understand it was for xeus_python, now it starts to make more sense. I do think the get_ipython and get_kernel is a bit confusing. If there is a good reason to do I think some comments may help.

martinRenou:
The thing is that `get_ipython()` is used at different places in the code, and not only to retrieve `get_ipython().kernel`, so your solution `get_kernel = lambda: get_ipython().kernel` does not really work. Also I changed the name because I would like, in `xeus_python`, to mock `ipywidgets.kernel.get_kernel` and not `ipywidgets.kernel.get_ipython`. Because at the end it's not `ipython` anymore

ipywidgets/__init__.py
martinRenou: After in-person discussion with Sylvain, this doesn't seem to be useful. Maybe @jasongrout you have an idea? Do you think this can be removed? Or stay as an alias for `register_comm_target` for backward compatibility?

martinRenou:
I think we can close this PR. It is a bit outdated. And IMO we should not mock ipywidgets to have it working in xeus-python, as in ipywidgets should not be a special case. I made a PR using IPython in xeus-python: https://github.com/jupyter-xeus/xeus-python/pull/400, which should fix most of these issues. 

We still need to mock `ipykernel.comm` though, as it is used by multiple libraries for using comms. We should find a nicer way to do this, but for now, mocking is our only solution.

PR #2273, opened 2018-11-26
Export require loader

SylvainCorlay:
The requireLoader which is internal to the HTML manager is duplicated in a number of downstream packages

 - [voila](https://github.com/QuantStack/voila/blob/master/js/loader.js#L40)
 - [thebelab](https://github.com/minrk/thebelab/blob/master/src/loader.js#L33)
 - [ipywidgets_server](https://github.com/pbugnion/ipywidgets_server/blob/master/js/loader.js#L14)

In order to prevent this duplication, I am exporting the function, and we will be able to remove that code from all of these projects.

### DIFF
--- a/packages/html-manager/package-lock.json
+++ b/packages/html-manager/package-lock.json
@@ -1,0 +1,10082 @@
+{
+  "name": "@jupyter-widgets/html-manager",
+  "version": "0.15.0",
+  "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Same for https://github.com/maartenbreddels/flask-ipywidgets/, looking good!

PR #2278, opened 2018-11-30
WIP Fix regexp escape issue in embed_minimal_html & Python 3.7 CI

rth:
Aims to fix the regexp escape issue in `embed_minimal_html`

Closes https://github.com/jupyter-widgets/ipywidgets/issues/2277

Also adds Python 3.7 CI as it doesn't appear to be tested.

### DIFF
--- a/ipywidgets/embed.py
+++ b/ipywidgets/embed.py
@@ -238,7 +238,7 @@ def escape_script(s):
     We only replace these three cases so that most html or other content
     involving `<` is readable.
     """
-    return script_escape_re.sub(r'\u003c\1', s)
+    return script_escape_re.sub(u'\u003c\1', s)
 
 @doc_subst(_doc_snippets)
 def embed_snippet(views,
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
     - 3.4
     - 3.5
     - 3.6
+    - 3.7
 addons:
   apt_packages:
     - pandoc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 # http://travis-ci.org/#!/jupyter-widgets/ipywidgets
 language: python
+dist: xenial
 python:
     - 2.7
     - 3.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
     - 3.4
     - 3.5
     - 3.6
-    - 3.7
 addons:
   apt_packages:
     - pandoc
@@ -37,5 +36,8 @@ matrix:
           env: GROUP=doc
         - python: 3.6
           env: GROUP=js BROWSER=firefox TRAVIS_NODE_VERSION=6.11.0
+        - python: 3.7
+          dist: xenial
+          sudo: true
 after_success:
     - coveralls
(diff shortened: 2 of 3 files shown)


rth:
Added Python 3.7 CI.

But I must be missing something between,
```
Python 3.7.1 | packaged by conda-forge | (default, Nov 13 2018, 18:15:35) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> u'\u003c\1'
'<\x01'
>>> '\u003c\1'
'<\x01'
>>> r'\u003c\1'
'\\u003c\\1'
```
and which one is expected to be used here (while keeping a Python 2 compatibility).

You can see the failing test with the original code in the first commit.

pbugnion:
Thanks for raising this, and sorry it took a while for one of us to have a look. 

## Problems with your solution

I'm not sure your implementation does quite what we need here. It's useful to refer to the issue where this was originally added to understand what's going on:

https://github.com/jupyter-widgets/ipywidgets/pull/2189

Basically, we want to replace `<` with the equivalent *bytes* `\u03c` (which happen to correspond to the unicode code point for `<`). We definitely don't want to replace the raw string by a unicode string literal. While it makes the test pass, it reverts the benefits of the original PR.

I have been using the following minimal example:

```py
import ipywidgets as widgets
from ipywidgets.embed import embed_minimal_html

t = widgets.Text("hello</script>")

embed_minimal_html("out.html", views=[t])
```

Without the escaping, when you view the html you just see a bunch of `}` that correspond to the end of the `application/vnd.jupyter.widget-state+json` script tag.

## Moving forwards

I can't quite convince myself that this isn't a Python bug. For instance, consider the following minimal reproduction (I've deliberately not used `\u` to avoid unicode-based confusion, since I don't think the issue has anything to do with unicode).

```
import re
re.sub("<", r"\h", "<")
```

In Python 3.6, this correctly outputs `"\\h"` (or `r"\h"`). In Python 3.7, this throws the same exception about an escape character.

@jasongrout You know much more about this than I do. Am I missing something?

antlarr-suse:
Just for the records, e253d3d77bc52965eecc3ae77e56231f9e360c49 by itself breaks the existing tests and thus, means that it changes the current behaviour with existing code, while using

```python
script_escape_re.sub(lambda match: r"\u003c" + match.group(1), s)
```

works fine with the current tests (as does #2316).

ipywidgets/embed.py
```diff
@@ -238,7 +238,7 @@ def escape_script(s):
     We only replace these three cases so that most html or other content
     involving `<` is readable.
     """
-    return script_escape_re.sub(r'\u003c\1', s)
+    return script_escape_re.sub(u'\u003c\1', s)
```
vidartf: I assume the second backslash here needs to be escaped? I.e. 
```suggestion
    return script_escape_re.sub(u'\u003c\\1', s)
```
This ensures that the `.sub()` method sees it as a regexp backreference (`\1`). In non-raw strings, `\1` means "byte expressed in octal (1 here)".

vidartf: Never mind this, after re-reading the initial PR, this is the wrong solution.

vidartf:
> Basically, we want to replace `<` with the equivalent _bytes_ `\u03c` (which happen to correspond to the unicode code point for `<`).

Referring to the original PR, we don't actually want to replace it with the *bytes*, but rather the *literal string* `\u003c`, as that is the *JSON escape* for that unicode character.

vidartf:
Would it not simply be enough to replace `r'\u003c\1'` with `'\\u003c\\1'`?

jasongrout:
Superseded by #2331. Thanks!

PR #2281, opened 2018-12-04
Improved dev experience, add watch commands

maartenbreddels:
My workflow is as follows:
from root I have two terminals open:
```
$ (cd packages/base; npm run watch)
$ (cd widgetsnbextension; npm run watch)
```
A change in base is propagated in about ~1 second.

### DIFF
--- a/packages/jupyterlab-manager/package.json
+++ b/packages/jupyterlab-manager/package.json
@@ -31,7 +31,8 @@
     "clean": "rimraf docs && rimraf lib",
     "docs": "typedoc --mode file --module commonjs --excludeNotExported --target es5 --moduleResolution node --out docs/ src",
     "lint": "tslint --project tslint.json --format stylish",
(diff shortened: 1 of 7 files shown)


packages/jupyterlab-manager/package.json
```diff
@@ -31,7 +31,8 @@
     "clean": "rimraf docs && rimraf lib",
     "docs": "typedoc --mode file --module commonjs --excludeNotExported --target es5 --moduleResolution node --out docs/ src",
     "lint": "tslint --project tslint.json --format stylish",
-    "prepublish": "npm run clean && npm run build"
+    "prepublish": "npm run clean && npm run build",
+    "watch": "tsc -w --project src"
```
vidartf: Do you need to run this when using `jupyter lab --watch`, or will it get in the way? I.e. do you end up with two processes trying to write to the same files?

PR #2283, opened 2018-12-05
Play widget: expose playing and repeat 

### LINKED ISSUE #1897: Play widget will stop programmatically, but won't start
Using a Play() widget, I can stop the play with:

```python
play._playing = False
```
But this doesn't start it:

```python
play._playing = True
```
The Play button is highlighted, but it doesn't start playing.

jtpio:
Fixes #1897.

Compared to the other widgets, there is no straightforward way to control the `Play` widget programmatically, except for `_playing` attribute as discussed in https://github.com/jupyter-widgets/ipywidgets/issues/1897.

Since the animation logic is happening on the frontend, this change is mostly about triggering the `click` method for the corresponding buttons. 

play_stop_pause_methods

### DIFF
--- a/ipywidgets/widgets/widget_int.py
+++ b/ipywidgets/widgets/widget_int.py
@@ -228,7 +228,7 @@ class Play(_BoundedInt):
     _model_name = Unicode('PlayModel').tag(sync=True)
 
     playing = Bool(help="Whether the control is currently playing.").tag(sync=True)
-    _repeat = Bool(help="Whether the control will repeat in a continous loop.").tag(sync=True)
+    repeat = Bool(help="Whether the control will repeat in a continous loop.").tag(sync=True)
 
     interval = CInt(100, help="The time between two animation steps (ms).").tag(sync=True)
     step = CInt(1, help="Increment step").tag(sync=True)
--- a/ipywidgets/widgets/widget_int.py
+++ b/ipywidgets/widgets/widget_int.py
@@ -224,16 +224,17 @@ class _IntRange(_Int):
 class Play(_BoundedInt):
     """Play/repeat buttons to step through values automatically, and optionally loop.
     """
-    interval = CInt(100, help="The time between two animation steps (ms).").tag(sync=True)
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
I am +1 on this feature.

I am not sure actually calling the `click` method on the button is the way to go.

However, this looks simple enough, so am fine with merging as-is.

@jasongrout do you want to press the green button?

jasongrout:
How about using state changes instead of messages? If we have a play state attribute that indicates whether the widget is playing or not, `pause` becomes setting that attribute to `False`, pressing play is setting that attribute to `True`, and clicking stop is equivalent to setting the play state to `False` and the value attribute to the min value.

In general, I think state changes are simpler to reason about than sending messages, and the bonus here is that we get information into what is happening on the python side.

jasongrout:
We already have `_playing`, as noted in #1897. Perhaps just exposing `playing` and `repeat` as official state, and making them actually work bidirectionally would be great.

vidartf:
The methods could also just set the state of `_playing`, could they not?

jtpio:
As far as I can recall setting the state of `_playing` was not sufficient without significant changes to the frontend animation logic.

But yes it would be good to have both `playing` and `repeat` so it's similar to what the other widgets expose.

jtpio:
Reworking this PR to control the `Play` widget by exposing and setting `playing` and `repeat`:

play-playing-repeat

To trigger a `stop` programmatically, this would be equivalent to do:

```python
play.playing = False
play.value = play.min
```

jtpio:
This should also fix the issue where the play button could be hit multiple times in a row, which would "stack up" the animation.

play-multiple-plays

jtpio:
Another thing to look at in a separate PR would be to drop the pause button, and instead have a toggle play / pause button.

SylvainCorlay:
This looks good to me! Thanks @jtpio.

This needs a rebase before it is merged.

jtpio:
Done.

PR #2303, opened 2019-01-29
Fix DeprecationWarning in Python 3.7

Juanlu001:
Fix `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`.


Juanlu001:
Failures are related to pip not upgrading dependencies when it should (or we'd want it to), I think.

vidartf:
I re-ran the failed tests, and the current failure seems significant. `collections.abc` are not present on Python 2.7. I say we should keep this PR open until the point where Python 2 support is dropped. Or, if we want this before, this PR should include a `try`/`except ImportError` clause.

Juanlu001:
Thanks! True, I forgot about Python 2. What's the roadmap to drop it?

vidartf:
I'm not sure if there is a specific plan, but unless we decide to drop it early, I would say the upper bound is when [Python stops supporting it](https://pythonclock.org/). If nothing else, there are many other projects that will stop supporting it at the same time, making it a coordinated push for those still on Python 2 to make the jump.

jasongrout:
> What's the roadmap to drop it?

I'd like to drop python 2 with our next major release. We don't have a roadmap/timeline for that, though.

jasongrout:
@SylvainCorlay, @Juanlu001 - I fixed to work on python 2 and 3.

Juanlu001:
Thanks @jasongrout!

PR #2314, opened 2019-02-12
Expose JupyterLab Widget Manager

SylvainCorlay:
This exposes jupyterlab-manager's `Output` and `Manager` implementation.

This is required by https://github.com/QuantStack/voila/pull/46 and the corresponding PR to thebelab to support the output widget and the `@interact` magic.

### DIFF
--- a/packages/jupyterlab-manager/src/index.ts
+++ b/packages/jupyterlab-manager/src/index.ts
@@ -2,9 +2,23 @@
 // Distributed under the terms of the Modified BSD License.
 
 import WidgetManagerProvider from './plugin';
+import * as output from './output';
 
 export default WidgetManagerProvider;
 
 export {
   INBWidgetExtension
 } from './plugin';
+
+export {
+  WidgetManager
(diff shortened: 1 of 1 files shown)


packages/jupyterlab-manager/src/index.ts
```diff
@@ -2,9 +2,23 @@
...
+  WidgetManager
+} from './manager';
+
+export {
+  WidgetRenderer
+} from './renderer';
+
+export {
+  output
+};
```
jasongrout: Can this just be `export {output} from './output';`, instead of having the `import *` above?

jasongrout: never mind, I see that you're just exporting a namespace.

SylvainCorlay:
@jasongrout I leave it to you to merge?

I would like to get a patch version of the jupyterlab-manager out after that so that we can move forward with the output widget in voila and jupyterlab.

PR #2315, opened 2019-02-13
Enable noImplicitAny compiler check, fix tslint errors.

blois:
This enables the noImplicitAny TS compiler check, with types specified to the closest approximation as I could determine. Also fixed up a handful of tslint warnings and disabled them for the test code to allow `yarn lint` to exit clean.

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -28,20 +28,23 @@ import {
     JUPYTER_WIDGETS_VERSION
 } from './version';
 
+import {
+    KernelMessage
+} from '@jupyterlab/services';
+
 /**
  * Replace model ids with models recursively.
  */
 export
-function unpack_models(value, manager): Promise<any> {
-    let unpacked;
+function unpack_models(value: any, manager: managerBase.ManagerBase<any>): Promise<any> {
     if (Array.isArray(value)) {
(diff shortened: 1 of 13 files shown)


SylvainCorlay:
Thanks for this @blois. Sorry for not responding earlier.

This appears to need a rebase before it is merged.

SylvainCorlay:
By the way @blois we would like to get ipywidgets 7.5 out shortly. Woud be great to have this in!

packages/base/src/widget.ts
```diff
@@ -79,9 +82,9 @@ class WidgetModel extends Backbone.Model {
             _model_name: 'WidgetModel',
             _model_module_version: JUPYTER_WIDGETS_VERSION,
             _view_module: '@jupyter-widgets/base',
-            _view_name: null,
+            _view_name: null as string,
```
vidartf: Maybe better to change the typing of this to `string | null` ? And similarly for others like this?

SylvainCorlay:
Hey @blois I am following up on this. We would like to tag a release of ipywidgets soon and it would be great to get that PR in!

blois:
Apologies for the delay, just updated it to be based on the latest version!

jasongrout:
Thanks!

PR #2326, opened 2019-02-27
Use JupyterLab notebook tracker rather than the document widget extension mechanism

jasongrout:
This explores using the tracker mechanism in JLab instead of the document widget extensions, as explained in https://github.com/jupyterlab/jupyterlab/issues/3974.

It seems to work pretty well!

### DIFF
--- a/packages/jupyterlab-manager/package.json
+++ b/packages/jupyterlab-manager/package.json
@@ -45,14 +45,18 @@
     "@jupyterlab/rendermime": "^1.0.0-alpha.3",
     "@jupyterlab/rendermime-interfaces": "^1.3.0-alpha.3",
     "@jupyterlab/services": "^4.0.0-alpha.3",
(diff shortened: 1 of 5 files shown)


jasongrout:
Currently the widget manager can't insert into the rendermime early enough so that the initial render uses the widget renderer. A couple of ideas, discussed also with @afshin and @ian-r-rose, about resolving this initial render issue:

1. Have a new hook for inserting something into a notebook's rendermime *before* the notebook gets rendered
2. Insert something into the global rendermime that somehow delays rendering until necessary information is available.
3. Insert into the rendermime after the initial render, but then go back through and manually rerender things that contain widgets. This has the side effect of possibly having a flash of text content before replacing it with a widget.

jasongrout:
I merged these changes into #2265, since they were starting to build on each other.

I'll close this since #2265 now supersedes it.

PR #2331, opened 2019-03-01
Fix re escape

### LINKED ISSUE #2277: Regexp escape issue in embed_minimal_html
When running the example https://ipywidgets.readthedocs.io/en/stable/embedding.html#python-interface, I'm getting the following error,

```py
Python 3.7.1 (default, Oct 23 2018, 19:19:42) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from ipywidgets import IntSlider
>>> from ipywidgets.embed import embed_minimal_html
>>> 
>>> slider = IntSlider(value=40)
>>> embed_minimal_html('export.html', views=[slider], title='Widgets export')
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/sre_parse.py", line 1021, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\u'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/site-packages/ipywidgets/embed.py", line 302, in embed_minimal_html
    snippet = embed_snippet(views, **kwargs)
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/site-packages/ipywidgets/embed.py", line 268, in embed_snippet
    for view_spec in data['view_specs']
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/site-packages/ipywidgets/embed.py", line 268, in <genexpr>
    for view_spec in data['view_specs']
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/site-packages/ipywidgets/embed.py", line 241, in escape_script
    return script_escape_re.sub(r'\u003c\1', s)
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/re.py", line 309, in _subx
    template = _compile_repl(template, pattern)
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/re.py", line 300, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "/home/rth/.miniconda3/envs/ipywidget-env/lib/python3.7/sre_parse.py", line 1024, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \u at position 0
```

ipywidgets 7.4.2 was installed in a clean conda environement with Python 3.7.

vidartf:
Fixes #2277.

Supersedes #2278 as that seems to have stagnated for a few months.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ addons:
   apt_packages:
     - pandoc
   firefox: latest
+services:
+  - xvfb
 cache: yarn
 env:
     - GROUP=python
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 # http://travis-ci.org/#!/jupyter-widgets/ipywidgets
 language: python
+dist: xenial
 python:
     - 2.7
     - 3.4
     - 3.5
     - 3.6
+    - 3.7
 addons:
   apt_packages:
     - pandoc
--- a/ipywidgets/embed.py
+++ b/ipywidgets/embed.py
@@ -238,7 +238,7 @@ def escape_script(s):
     We only replace these three cases so that most html or other content
     involving `<` is readable.
     """
-    return script_escape_re.sub(r'\u003c\1', s)
+    return script_escape_re.sub(r'\\u003c\1', s)
 
 @doc_subst(_doc_snippets)
 def embed_snippet(views,
--- a/scripts/travis_script_js.sh
+++ b/scripts/travis_script_js.sh
@@ -1,7 +1,5 @@
 set -ex
 export CHROME_BIN=chromium-browser
-export DISPLAY=:99.0
-sh -e /etc/init.d/xvfb start
 
 yarn run integrity
 


vidartf:
Note: There are three escape mechanisms involved:
- That of JSON. This is where we want `\u003c` to be decoded to bytes.
- That of python strings. This is avoided by using a raw string.
- That of `re`. Previously, any sequence not matching [the special sequences of re](https://docs.python.org/3.7/library/re.html#regular-expression-syntax) would simply be spat back out. With 3.7 this gives an error, so we now need to be explicit that we want a literal backslash.

maartenbreddels:
3.7 is quite common now, and also ipyvolume users get bitten by this https://github.com/maartenbreddels/ipyvolume/issues/245
This seems like a simple fix, but is it tested like in https://github.com/jupyter-widgets/ipywidgets/pull/2278 ?
Would be great to have this merged soon.

jasongrout:
>  Would be great to have this merged soon.

I think you can merge it, right?

jasongrout:
> Note: There are three escape mechanisms involved:

Thanks for the very clear explanation about why the change is needed!

maartenbreddels:
@vidartf could you add the test for this?

maartenbreddels:
As discussed over gitter, the tests already covered this, and confirmed.

maartenbreddels:
Thanks @vidartf !

PR #2338, opened 2019-03-06
FIX: displaying a widget without a view should only have text/plain in the mimebundle

maartenbreddels:
The current _ipython_display_ always adds the `'application/vnd.jupyter.widget-view+json'` mimetype to the mimebundle, that causes nothing to show for non-dom/view widgets:
image

The expected output is:

image

Related PR #2021 
Reported in https://github.com/QuantStack/ipysheet/issues/7


SylvainCorlay:
Discussed in person.

Issue #2328, opened 2019-02-28
tooltip on Text input itself if no description is present?


zerline:
```
import ipywidgets
ipywidgets.__version__
```
'7.4.2'

```
t = ipywidgets.Text('ok')
t.description_tooltip = "my tooltip"
t
```
 -> I see no tooptip on mouse over ..

jasongrout:
Thanks for posting code to illustrate what you are seeing.

The description tooltip is only for the description, which you don't have in the above code. Try this:

```python
t = ipywidgets.Text('ok')
t.description_tooltip = "my tooltip"
t.description="description"
t
```
and hover over 'description'.

zerline:
That's right, thanks for the enlightenment!
Now: would it be possible to have tooltips on the text input itself? Tell me. If you find that unreasonable, then I will subclass it, np.

jasongrout:
> Now: would it be possible to have tooltips on the text input itself? Tell me. If you find that unreasonable, then I will subclass it, np.

I think the browser may make it possible. Whether it's a reasonable ux or not is a UX decision, probably not a browser capability. I won't have time to work on this anytime soon, though, so that may also help you decide what to do.

zerline:
I'll try to do it then.

jasongrout:
Are you saying you'd like to contribute it to ipywidgets, or subclass the widget for your own purposes? If it's for upstream, we would need to have that UX discussion about having tooltips on widget contents (not just descriptions).

zerline:
Yes, that's the idea: contribute if it's useful (subclassing is done already, and works afaik). My purpose is perhaps somewhat particular : a grid of small text inputs with no description.

What about a tooltip that would appear either on the description if any, or on the actual input if no description is present?

jasongrout:
> What about a tooltip that would appear either on the description if any, or on the actual input if no description is present?

That sounds like a reasonable compromise to me. Great idea!

jasongrout:
do you just want to use the existing description_tooltip field then, without introducing another tooltip field?

zerline:
I think that's possible. 
NB: I'm using the HTML attribute named 'title'. It's a rather general attribute. So I was wondering where to put the code. Maybe _String.

jasongrout:
I was imagining it would go around https://github.com/jupyter-widgets/ipywidgets/blob/092c085aaf3b1165a5c8f24a9ce663ae523e7969/packages/controls/src/widget_string.ts#L275

[...]

PR #2342, opened 2019-03-07
Tooltip on Text input itself if no description is present

zerline:
This intends to do the work described on issue #2328.


packages/controls/src/widget_string.ts
```diff
@@ -274,6 +274,10 @@ class TextView extends DescriptionView {
 
         this.textbox = document.createElement('input');
         this.textbox.setAttribute('type', this.inputType);
+        if (this.model.get('description').length === 0) {
```
jasongrout: This check should probably be moved into update_title, and update_title called when the description changes too. That way if someone changes the description text, the appropriate thing is done with the title attribute.

packages/controls/src/widget_string.ts
```diff
@@ -292,6 +296,13 @@ class TextView extends DescriptionView {
         this.textbox.setAttribute('placeholder', value);
     }
 
+    update_title() {
+        let title = this.model.get('description_tooltip');
+        if (title.length > 0) {
+            this.textbox.setAttribute('title', title);
+        }
```
jasongrout: If the title length is now 0, we should probably unset the textbox title attribute.

jasongrout:
To build, you should be able to create a new environment and use the dev_install.sh script in the repo root (if on linux or mac) to install the dev version of widgets.

zerline:
Ok

zerline:
Done.

zerline:
Is this very _non intrusive_ PR to be merged in a near future as is?

SylvainCorlay:
This looks good to me!

Will merge later today if no one has a problem with this.

PR #2351, opened 2019-03-22
FIX: inconsistency between actual notebook output and Output.outputs

### LINKED ISSUE #2350: Output widget + clear_display(wait=True) bug/inconsistency
For both the classical notebook as Jupyter lab the outputs trait is always set to an empty list, even when wait=True. The actualy displaying of the content is left to the notebook or Jupyter lab, so the output is still visible (in no new display call is made), but after a page refresh the output is not visible anymore (since the state of the widget does not contain the content anymore).

I think both Output widgets should not reset the outputs trait, when wait is set to true, but only do it at the next display call.

maartenbreddels:
Fixes #2350
I think this approach of updating the outputs trait will always make the notebook's real output and the outputs trait the same.

### DIFF
--- a/packages/jupyterlab-manager/src/output.ts
+++ b/packages/jupyterlab-manager/src/output.ts
@@ -107,7 +107,7 @@ class OutputModel extends outputBase.OutputModel {
       break;
     case 'clear_output':
       this.clear_output((msg as KernelMessage.IClearOutputMsg).content.wait);
-      this.set('outputs', [], {newMessage: true});
+      this.set('outputs', this._outputs.toJSON(), {newMessage: true});
       this.save_changes();
       break;
     default:
--- a/packages/jupyterlab-manager/src/output.ts
+++ b/packages/jupyterlab-manager/src/output.ts
@@ -102,17 +102,15 @@ class OutputModel extends outputBase.OutputModel {
       let model = msg.content as nbformat.IOutput;
       model.output_type = msgType as nbformat.OutputType;
       this._outputs.add(model);
-      this.set('outputs', this._outputs.toJSON(), {newMessage: true});
-      this.save_changes();
       break;
     case 'clear_output':
       this.clear_output((msg as KernelMessage.IClearOutputMsg).content.wait);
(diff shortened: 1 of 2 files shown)


jasongrout:
This looks right to me. Do you have a concrete example/list of steps showing the difference between status quo and this PR? You mentioned refreshing in the issue?

maartenbreddels:
```python
import ipywidgets as widgets
from IPython.display import clear_output

output2 = widgets.Output()
with output2:
    print("in output2")
    clear_output(wait=True)
display(output2)
```
This will show `in output2`, but the widget state does not have the content anymore. If you refresh the notebook, nothing will be shown. With this PR it still is in the widget state and will be shown after a refresh.

packages/jupyterlab-manager/src/output.ts
```diff
@@ -107,7 +107,7 @@ class OutputModel extends outputBase.OutputModel {
       break;
     case 'clear_output':
       this.clear_output((msg as KernelMessage.IClearOutputMsg).content.wait);
-      this.set('outputs', [], {newMessage: true});
+      this.set('outputs', this._outputs.toJSON(), {newMessage: true});
       this.save_changes();
```
jasongrout: Since the two cases now share the last two lines, you could refactor them down to below the switch statement, i.e., update the internal outputs object, then *always* copy it to the outputs store and save changes.

jasongrout: Never mind - that's not a good idea for kernel iopub messages that aren't in the cases.

jasongrout:
> With this PR it still is in the widget state and will be shown after a refresh.

And, just checking, if you then display one more thing after the refresh, it replaces the output (i.e., the pending clear is still in effect?)

maartenbreddels:
No, the pending clear is not in any state, outside of the run of a single cell AFAIKT. For instance, using display/clear_display(wait=True) (without an Output widget) will only work in a single cell. This is consistent with nbconvert now as well https://github.com/jupyter/nbconvert/pull/969 But I have not looked at the sourcecode of the classical notebook, this is purely observational.

PR #2357, opened 2019-04-04
Set value=null on DatePicker on Invalid Dates when it lost focus

### LINKED ISSUE #1895: DatePicker on Invalid Dates
When a DatePicker has an invalid value it is highlighted in orange as expected. The widget.value though is the last valid value which seems strange since there is a mismatch in the UI when something goes to process this.

Example:
```
import ipywidgets
def print_value(v):
    print(v)
w = ipywidgets.DatePicker()
w.observe(print_value, 'value')
w
```
Change the date from 01/31/2017 and keep pushing up on the first value as that you get 02/31/2017. `w.value` is `2017-01-31` still although this is not what the UI says.

Starting:
image
Ending:
image

 I would have expected it to say None as it does when it was blank when first created.

xmnlab:
Resolves #1895

### DIFF
--- a/packages/controls/src/widget_date.ts
+++ b/packages/controls/src/widget_date.ts
@@ -111,7 +111,7 @@ class DatePickerView extends DescriptionView {
     events(): {[e: string]: string} {
         return {
             'change [type="date"]': '_picker_change',
-            'blur [type="date"]': '_picker_blur'
+            'focusout [type="date"]': '_picker_focusout'
         };
     }
 
@@ -127,8 +127,8 @@ class DatePickerView extends DescriptionView {
         }
     }
 
-    private _picker_blur() {
-        if (!this._datepicker.validity.badInput) {
+    private _picker_focusout() {
+        if (this._datepicker.validity.badInput) {
             this.model.set('value', null);
             this.touch();
         }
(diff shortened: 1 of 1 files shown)


xmnlab:
hey everyone, I am trying to trigger a function when the object lost the focus, but it seems is doesn't work. 

any thoughts?

saulshanabrook:
@xmnlab You could try adding an event listener to the `focusout` event maybe? https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event

xmnlab:
hey @saulshanabrook thanks! I will try that! thanks!

xmnlab:
@saulshanabrook it seems it worked. thanks!

xmnlab:
it is done for review.

xmnlab:
thanks @gnestor !

could it be merged?

gnestor:
Yes, I don't have merge rights but I approve 👍

xmnlab:
thanks @gnestor ! :)

xmnlab:
hey everyone, any feedback about this PR? could it be merged?

gnestor:
@jasongrout Would you remind reviewing?

jasongrout:
Sure, though I may not be able to in the near future. I plan to make a pass through all ipywidgets open PRs before the next ipywidgets release.

SylvainCorlay:
Thanks for looking into this @xmnlab!

xmnlab:
thank you all @SylvainCorlay @gnestor @jasongrout @saulshanabrook ! :)

PR #2372, opened 2019-04-13
Add jlab manager disconnect function

jasongrout:
Fix unfinished items from #2265, including some code that did not get pushed to that PR and following up on some TODO items.

CC @SylvainCorlay, @vidartf


vidartf:
:+1:

PR #2376, opened 2019-04-14
Added a layout widget for 'stacked' layout

astrofrog:
This adds a new layout widget that I've called 'Stacked' for now which behaves like a ``Box`` widget where one can switch between different widgets depending on a ``selection_index``. This is very useful when used in combination with another widget that can be used to select something. Here's the example I added to the docs:

Screen-Recording-2019-04-14-at-22 53 35

Of course this behavior can be emulated by simply manually adding/removing widgets from a ``Box`` widget but having a dedicated widget is much more efficient.

This is inspired by [QStackedLayout](https://doc.qt.io/qt-5/qstackedlayout.html) in Qt5.

Is there any interest in including something like this? If so let me know if you think of a better name! Also, I wasn't sure what the best way to update the docs notebook - for now I just added some cells, but should I re-run it and save the widget state for the whole notebook? (which might make a large diff). Is there a script to automate this?

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -1112,7 +1112,46 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Accordion and Tab use `selected_index`, not value\n",
+    "### Stacked\n",
+    "\n",
+    "The `Stacked` widget can have multiple children widgets as for `Tab` and `Accordion`, but only shows one at a time depending on the value of ``selected_index``:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "button = widgets.Button(description='Click here')\n",
+    "slider = widgets.IntSlider()\n",
+    "stacked = widgets.Stacked([button, slider])\n",
+    "stacked  # will show only the button"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This can be used in combination with another selection-based widget to show different widgets depending on the selection:"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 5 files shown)


maartenbreddels:
I think this is a pretty good idea, I think it fits into quite good into ipywidgets I think.

At the same time, it exposes an issue with ipywidgets I think. I believe that the ipywidget library should have more atomic building blocks such that this can be built from it.

I'm in favor of merging this.

SylvainCorlay:
> I'd like to make a quick and small release supporting the new JLab prerelease. Can we hold off on merging at least for today?
> 
> As for a stack panel - a typical design for having a tab panel is to have a standalone tab bar and then a stack widget that it controls. Perhaps we should refactor our tab panel to use something like this as well.
>

+1

maartenbreddels:
Yes, that sounds like a good approach, I think ipyvuetify will also be like that right @mariobuikhuizen?
The tab bar does not control the visibility of anything; that part you need to implement separately.

mariobuikhuizen:
Vuetify (and thus ipyvuetify) supports both an integrated and a separate tab bar.

jasongrout:
FYI, phosphor has a tab bar, a stack panel, and a tab panel that combines the two.

maartenbreddels:
> a stack panel

I see some Qt influence here as well (great package btw!) So I'm ok with using Qt inspired names.

astrofrog:
So just to check, how do we want to proceed here? I could close this and open a feature request issue to keep track of it. Or we could consider reviewing/merging this and just updating the JS view later?

vidartf:
Merging this for now! Long term, we might want to use the lumino `StackedPanel` class for the view, as that will keep everything in the DOM, simply showing/hiding the different views when changing the selection. For now, I think this is sufficient, and I think that this can be updated in a backwards compatible way.

vidartf:
(will merge in master to fix merge conflicts first)

PR #2394, opened 2019-04-25
Fix number input css

SylvainCorlay:
Before:
Screenshot from 2019-04-26 17-10-55
After:
Screenshot from 2019-04-26 22-03-33


SylvainCorlay:
This is ready to go.

PR #2409, opened 2019-05-09
Remove the overflow exception for the shorthand property

SylvainCorlay:
This PR removes the `overflow-x` and `overflow-y` properties from the `Layout` widget.

- `overflow` is not a CSS shorthand property which can be used to set both `overflow-x` and `overflow-y` (just like e.g. `margin` can be used to set ` margin-[top, right, bottom, left]`). In ipywidgets layout, whenever there is a shorthand property, it is the only one to be exposed to the Python API (see documentation [here](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Styling.html#Shorthand-CSS-properties)).
- However, we were making an exception for the case of `overflow-x` and `overflow-y` because for some time, browsers that introduced these new CSS properties did not allow using the `overflow` as a shorthand property to set both.
- This pull request stops making them an exception because latest versions of [chrome and firefox support it](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#Browser_compatibility). For these browsers, this pull request solves a bug, which is that it is not possible anymore to set `overflow` because it is being overriden by `overflow-x` and `overflow-y`.


SylvainCorlay:
cc @martinRenou

jasongrout:
This is a backwards-incompatible change, so I think it will need to wait for 8.0. Thoughts?

SylvainCorlay:
> This is a backwards-incompatible change, so I think it will need to wait for 8.0. Thoughts?

I disagree. The problem with shorthand properties is that `overflow` and `overflow_[x|y]` override each other making them unusable. Hence I think we need to include that in 7.5.

SylvainCorlay:
It is more of a bug in the current spec, which we introduced knowingly because of the way browsers behaved.

jasongrout:
If people use one or the other, it works fine, right?

SylvainCorlay:
> If people use one or the other, it works fine, right?

Actually not, the layout sets `overflow_[x,y]`  to null which overrides `overflow`.

jasongrout:
> Actually not, the layout sets `overflow_[x,y]` to null which overrides `overflow`.

Ah right. We've had exactly this issue and discussion before. It may have been with a different shortcut attribute. IIRC, we handled it as a special case so that null wouldn't override what was set by the shortcut.

SylvainCorlay:
> Ah right. We've had exactly this issue and discussion before. It may have been with a different shortcut attribute. IIRC, we handled it as a special case so that null wouldn't override what was set by the shortcut.

This does not appear to be in the codebase anymore.

jasongrout:
Maybe we only discussed it, never implementing it. I do remember talking about this.

It looks like from the code that currently, `overflow` *never* works, since it is always overridden with overflow_x/y, right?

Let's come up with some way that if people used `overflow_x/y`, their code still works. For example, if `overflow` is null, use `overflow_x/y`.

SylvainCorlay:
> It looks like from the code that currently, overflow never works, since it is always overridden with overflow_x/y, right?
>

Right, unless you use an older version of chrome, or safari, or IE.

> Let's come up with some way that if people used overflow_x/y, their code still works. For example, if overflow is null, use overflow_x/y in the ts code that actually sets the css.
> 

OK, we can special-case overflow_x/y. Only set it if overflow is not set. Maybe we should include a deprecation of these attributes.

jasongrout:
Sure, deprecation sounds great too.

SylvainCorlay:
> Sure, deprecation sounds great too.

Okay. Do we have a standard way of deprecating stuff?

jasongrout:
I experimented with this, and it looks great. Thanks!

PR #2418, opened 2019-05-23
Drop postcss

SylvainCorlay:
This makes `widgetsnbextension` and `html-manager` use the raw CSS from `@jupyter-widgets/controls`  instead of the version processed with `postcss` and `cssnext`.

- `cssnext` was substituting CSS variables with their values.
- with this change, the CSS variables that are defined to polyfill for JupyterLab in the classic notebook become visible and usable by widgets. This will allow custom widget libraries to handle `lab`, `classic`, and embedded contexts in a more homogeneous way.
- `cssnext` is deprecated.

**Note:** that IE does not support CSS variables, but it is not a supported browser for the classic notebook at the moment.


SylvainCorlay:
@jasongrout

maartenbreddels:
I see a lot of red 👍 (as in lines removed).
To summarize, instead of substituting the variables, we now include them in the controls package? Which means the become visible in the notebook once the controls package is loaded?

SylvainCorlay:
> To summarize, instead of substituting the variables, we now include them in the controls package? Which means the become visible in the notebook once the controls package is loaded?

`widgets.css` includes the variables. (all it does is `@import labvariables.css` and `@import widgtets-base.css`).

SylvainCorlay:
@maartenbreddels @jasongrout how about we press the green button?

jasongrout:
Reviewing this is next on my list, after a short break :)

jasongrout:
Thanks. I'll note that we still build `widgets.built.css` so people can still use it from the `@jupyter-widgets/controls` package if they wish.

PR #2421, opened 2019-05-29
Update for JLab 1.0a5.

jasongrout:
There were a few places where I strategically cast to any to smooth over differences between versions of @jupyterlab/services that different widgets packages are using.

With this and a fresh environment with jlab 1.0a5 installed, `./dev-install.sh` makes ipywidgets work in jlab.


jasongrout:
@SylvainCorlay - I'd love for this to go into 7.5.

Issue #2422, opened 2019-05-30
Failed to build examples:   Parameter 'model' implicitly has an 'any' type.


IlyaKazakevich:
```
src/widget_layout.ts(94,59): error TS7006: Parameter 'model' implicitly has an 'any' type.
src/widget_layout.ts(94,66): error TS7006: Parameter 'value' implicitly has an 'any' type.
```

IlyaKazakevich:
could be fixed with something like
``
this.listenTo(this.model, 'change:' + trait, (model:any, value:any) => {  
``

throwable-one:
PR sent https://github.com/jupyter-widgets/ipywidgets/pull/2423

vidartf:
Closed in https://github.com/jupyter-widgets/ipywidgets/pull/2423

PR #2423, opened 2019-05-30
#2422 provide type explicitly to fix complilation

throwable-one:
Fix for 
https://github.com/jupyter-widgets/ipywidgets/issues/2422


jasongrout:
Thanks!

PR #2425, opened 2019-05-30
Remove unused scrollIfNeeded and StyleModel import

jtpio:
The `scrollIfNeeded` function and the `StyleModel` import don't seem to be used and can be removed.


jasongrout:
> The `scrollIfNeeded` function and the `StyleModel` import don't seem to be used and can be removed.

Indeed! Thanks!

PR #2431, opened 2019-06-03
Add _user_data attribute to the Widget class

martinRenou:
cc. @maartenbreddels @rmorshea 

Add `_user_data` attribute to the Widget class. This allows us, in `voila`, to provide metadata to the widget instance like `mount_id` for putting a given widget to a specific place in the `voila` template.

Multiple things to discuss here:
- the name: is `_user_data` fine or should we use `_metadata` ?
- the serialization: what should we use for the serialization? Do we only allow JSON types in this dict (numbers, strings, lists, dicts..)?
- a util method: should we provide a `tag` method to the `Widget` class? Similar to the `tag` method of `Trait` classes?

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -388,6 +388,8 @@ class Widget(LoggingHasTraits):
 
     _view_count = Int(None, allow_none=True,
         help="EXPERIMENTAL: The number of views of the model displayed in the frontend. This attribute is experimental and may change or be removed in the future. None signifies that views will not be tracked. Set this to 0 to start tracking view creation/deletion.").tag(sync=True)
+    _user_data = Dict({}, help="User defined data.").tag(sync=True)
+
     comm = Instance('ipykernel.comm.Comm', allow_none=True)
 
     keys = List(help="The traits which are synced.")


vidartf:
A few thoughts:
- What's the rationale behind using the `_` prefix for the field name? I don't think that is a good idea if it is meant purely for external use (it is normally meant to signify *internal* use only).
- The `mount_id` is a specific example case, that could possibly be solved by some other means (a `name` field?).
- The more general mechanism for allowing users / lib-authors to tag core widget (or any widget class outside of their control) with generic data is a bigger issue, and has some interesting points:
  - How would the observation pattern be (kernel side / client side)?
  - Are we only allowing JSON primitives (serialization/deserialization)? Buffers? Custom parts?
  - Based on the answers above, is `Dict` really the right type here? Are there other data types that will work better?
  - Do we only want to tag models, or would we want to be able to tag views as well?

SylvainCorlay:
note: we had alternatives means of achieving this without adding user data using the display id.

vidartf:
@SylvainCorlay What were the alternatives? Which alternative did Voilà end up using? Should this issue be closed?

vidartf:
@maartenbreddels Let us know if you still think this is needed/useful!

martinRenou:
I remember it is not needed anymore. I think I remember we had some discussions about using the `display_id` from the IPython display logic instead.

Issue #2392, opened 2019-04-24
Graceful error handling for DatePicker invalid dates


nickpollari:
Whenever you accidentally type in an extra character into the DatePicker widget it automatically registers an exception, see the GIF below. Shouldn't this be handled better like when the user actually clicks outside of the widget space? Or can we introduce a "continuous_update" kind of method here to prevent this error from being raised during user input?

![dp_error](https://user-images.githubusercontent.com/12014615/56685676-81344e00-66a0-11e9-8785-56ded0d41c48.gif)

jasongrout:
> Or can we introduce a "continuous_update" kind of method here to prevent this error from being raised during user input?

I think this makes sense, and like the inttext and floattext widgets, have continuous_update default to false.

jasongrout:
Setting as good first issue, since it probably should mirror the pattern in the IntText/FloatText widgets.

Relevant IntText code is at 

https://github.com/jupyter-widgets/ipywidgets/blob/99b977eabc9f56ca163d228f701a3e2a39f2e8df/ipywidgets/widgets/widget_int.py#L130 

https://github.com/jupyter-widgets/ipywidgets/blob/99b977eabc9f56ca163d228f701a3e2a39f2e8df/packages/controls/src/widget_int.ts#L482  

https://github.com/jupyter-widgets/ipywidgets/blob/99b977eabc9f56ca163d228f701a3e2a39f2e8df/packages/controls/src/widget_int.ts#L574-L588

jasongrout:
Or perhaps we should not have an option to set it back to continuous update, instead just only support non-continuous update. That's probably better. In that case, it may be as easy as figuring out the right event for the change handler to avoid making an update until the user explicitly indicates they are done (focusing out of the widget, pressing enter, etc.) https://github.com/jupyter-widgets/ipywidgets/blob/99b977eabc9f56ca163d228f701a3e2a39f2e8df/packages/controls/src/widget_date.ts#L113

nickpollari:
> Or perhaps we should not have an option to set it back to continuous update, instead just only support non-continuous update. That's probably better. In that case, it may be as easy as figuring out the right event for the change handler to avoid making an update until the user explicitly indicates they are done (focusing out of the widget, pressing enter, etc.)

I like this solution the most. I'm hard pressed to think of an instance where you need the DatePicker to be continuously pushing events as I type in a value.

jasongrout:
[...]

PR #2433, opened 2019-06-04
Fix: better error handling for DatePicker, removing continuous update (issue #2392)

ibdafna:
cc. @jasongrout @nickpollari

I removed the `change [type="date"]` handler as it fired the update function on any state change, causing the issue. 

The logic I implemented instead, for the purpose of determining whether a user is done updating the widget, was to use two handlers: one for focusout, and the other for the enter key code.

### DIFF
--- a/packages/controls/src/widget_date.ts
+++ b/packages/controls/src/widget_date.ts
@@ -110,7 +110,6 @@ class DatePickerView extends DescriptionView {
 
     events(): {[e: string]: string} {
         return {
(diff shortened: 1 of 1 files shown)


jasongrout:
Setting as a minor release since it does involve changing behavior the user sees.

kitchoi:
@ibdafna Just a gentle ping... this looks like a good change but since it has been a while, there are now conflicts blocking merge.

Issue #2434, opened 2019-06-05
Using numpy array as a `value` for FloatRangeSlider?


lheagy:
## Example
Right now, if the `value` keyword is assigned a numpy array (e.g. in the following example), a value error is raised. 
```python
import numpy as np
import ipywidgets as widgets

slider = widgets.FloatRangeSlider(
    value=5*np.r_[-1, 1],
    min=-1,
    max=1
)
```

## Error
The value error is quite opaque to the user: 

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

## Suggested fix
If we replace: 
https://github.com/jupyter-widgets/ipywidgets/blob/825dc5837df6938bfc57abedabd7d617ce1da920/ipywidgets/widgets/widget_float.py#L291

with 
```python
if kwargs.get('value', None) is None:
```
the same logic applies, and a numpy array will be accepted here. 

This will error in traitlets as a `tuple` is expected. At least in this case, a more informative error is provided
```
TraitError: The 'value' trait of a FloatRangeSlider instance must be a tuple, but a value of class 'numpy.ndarray' (i.e. array([-5,  5])) was specified.
```

In my opinion, it would be quite handy to be able to directly provide numpy arrays with two elements and have them be cast to a tuple in the validation step (especially for symmetric bounds, where it is handy to set `5*np.r_[-1, 1]`). What are others thoughts on this?

PR #2435, opened 2019-06-05
FloatRangeSlider `value` logic

lheagy:
Update logic in setting of default `value` field of the FloatRangeSlider so that numpy arrays can be passed as inputs (see #2434). Note that this will still currently error in traitlets as a `tuple` is expected, but at least the error at this point is very clear: A trait error is raised:
```
TraitError: The 'value' trait of a FloatRangeSlider instance must be a tuple, but a value of class 'numpy.ndarray' (i.e. array([-5,  5])) was specified.
```

I would appreciate the ability to simply provide a 2-entry long numpy array in the `value` field. What do folks think of adding something like: 
```python
try:
    proposal_tuple = tuple(proposal['value'])
    proposal['value'] = proposal_tuple
except:
    raise TraitError
```
to the validation of the `FloatRange` value 
https://github.com/jupyter-widgets/ipywidgets/blob/825dc5837df6938bfc57abedabd7d617ce1da920/ipywidgets/widgets/widget_float.py#L276-L281

or would it make more sense to do something at the traitlets level?

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -288,7 +288,7 @@ class _BoundedFloatRange(_FloatRange):
 
     def __init__(self, *args, **kwargs):
         min, max = kwargs.get('min', 0.0), kwargs.get('max', 100.0)
-        if not kwargs.get('value', None):
+        if kwargs.get('value', None) is None:
             kwargs['value'] = (0.75 * min + 0.25 * max,
                                0.25 * min + 0.75 * max)
         super(_BoundedFloatRange, self).__init__(*args, **kwargs)


vidartf:
Currently, the container traits from traitlets have a whitelist of types they can cast from. I don't think we can use that system for adding support for numpy, as that would add a dependency on numpy. Simply casting to tuple should be ok though, since we are checking the type of the sub-elements (i.e. we will avoid a false positive for `tuple('ok')`. So unless someone else can think of some other false positives, I say the `validate` pattern you suggested should be ok.

vidartf:
@jupyter-widgets/jupyter-widgets We should be able to merge this PR as is for the more helpful error message, and leave the other suggestions as future work for further discussion.

SylvainCorlay:
Thanks! This looks good to me.

PR #2436, opened 2019-06-06
implements __copy__ for widgets

### LINKED ISSUE #2352: Duplicating\copying a widget
Hi,
I'm trying to duplicate a widget, and haven't found anything about it and nothing I tried worked so far.
For example:
```
orig_text = widgets.Text(description="work")
dup_text = text.copy()
```
Both copy(), deepcopy(), and trying to manually copy __dict__ didn't work.
I want them to have different ID's, so I can type in each of the text-boxes a different unrelated text, but dup_text should have the not only the same description but every other property of orig_text, while still have a different ID.

Is there any (preferably elegant) way to achieve this?
Thank alot.

DanielAristidou:
implements __copy__ for widgets 

implemented on base widget - and 

Follows the following steps: 

create a new widget from the reference class. 
initialises the widget passing any required arguments. 
reads all relevant attributes and sets them on the new widget

fixes #2352

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -759,3 +759,23 @@ class Widget(LoggingHasTraits):
             for key in keys
         )
         return '%s(%s)' % (class_name, signature)
+
+
+    def __copy__(self):
+
+        cls = self.__class__
+        result = cls.__new__(cls)
+        result.__init__()
+
+        new_state = {key: value for key, value in self.get_state().items() if
+                     not key.startswith('_') and
+                     not key == 'layout' and
+                     not key == 'style'}
+
+        for key, value in new_state.items():
+            setattr(result, key, value)
+
+        result.layout = self.layout
+        result.style = self.style
+
+        return result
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
@DanielAristidou thanks for your PR.

The best practice with GitHub PRs is to not open them from your master branch, but from a feature branch, otherwise, you may have trouble updating your master branch to match master.

Regarding the content, I am adding some comments inline.

ipywidgets/widgets/widget.py
```diff
@@ -759,3 +759,23 @@ def _gen_repr_from_keys(self, keys):
...
+    def __copy__(self):
+
+        cls = self.__class__
+        result = cls.__new__(cls)
+        result.__init__()
+
+        new_state = {key: value for key, value in self.get_state().items() if
+                     not key.startswith('_') and
+                     not key == 'layout' and
+                     not key == 'style'}
```
SylvainCorlay: Maybe the state could be directly passed to `__init__` as kwargs.

Also, rather than overriding the method for each widget type that may have subwidget, a good way to exclude things may be to exclude attributes that are of type Widget?

A list of synchronized keys is available in the `keys` attribute.

Finally, I don't quite know if there may be other state in the object that is no synchronized with the front-end, and would need to be included in a copy operator...

vidartf:
Note that the `_repr_keys` might be relevant to this PR

tiagorangel2011 (approved):

PR #2441, opened 2019-06-13
cast 'value' in range sliders to a tuple

lheagy:
Following up from the conversation in #2435, this would allow numpy arrays (or lists) to be past as the `value` to an `IntRangeSlider` or a `FloatRangeSlider` by trying to cast the `value` input to a `tuple` when the slider is instantiated. This can be really handy when you want to specify a symmetric value about zero, e.g.

image

See if you think this is an appropriate place to do this. If so, I can also add a test to the pr


SylvainCorlay:
Would CTuple do the job?

vidartf:
`CTuple` does not exists :)

PR #2453, opened 2019-06-22
Add  docstring to GridBox and DatePicker

mwcraig:
When I was working on the tutorial I noticed a couple widgets are still missing docstrings.


mwcraig:
(milestoned to 7.5 but punt if you need to)

PR #2456, opened 2019-06-22
Update to JLab 1.0RC

jasongrout:
This also updates a number of packages, moves the tsconfig files up to the package root directories, and fixes numerous typescript errors.


jasongrout:
Widgets now work in JupyterLab (at least the int slider). They don't work in classic because I've changed Typescript to generate es2017, and the classic widget manager is not using classes.

To fix classic notebook, we should change the classic notebook widget manager to be a proper es6 class (we don't need to move to typescript, a plain js class would work just fine).

jasongrout:
@SylvainCorlay - do you want to take this over? I'm not sure when I'll be able to get back to it.

stonebig:
is this error with jlab-1.0.0rc0 related to this ?
image

jasongrout:
Probably? Widgets has not been updated for rc0 yet.

stonebig:
I just cliked on the python icon that creates normally a notebook

jasongrout:
Make sure you are testing with no jupyterlab extensions installed. None have been updated yet.

stonebig:
hum, it doesn't seems to solve the issue. I'm not expert, I'll wait a bit.

jasongrout:
I'd suggest trying a clean install into a new environment, to make sure there aren't old bits of code hanging around.

jasongrout:
I haven't had time to update the examples to use the new typescript base config. Someone can do that on another PR, or someone can take over this PR and finish that.

SylvainCorlay:
Thanks @jasongrout. I can look into it and iterate

jasongrout:
Thanks. This is a good time to hand off to you :).

I *think* everything works. jlab and the classic notebook work, tests pass in base, controls, and html manager, and examples build.

SylvainCorlay:
Ok I can take this from here. Thanks @jasongrout

jasongrout:
Looks like our one test in the examples is failing. Looks like possibly an issue with the karma/karma-webpack plugin (maybe because of the upgrade, and something should be different now?). I'll let you take it from here, though.

jasongrout:
Okay, I think I checked that each of the examples works now. I deleted the tmpnb example since tmpnb is no more.

jasongrout:
FYI, this will require a major bump of each js npm package (but not a major bump of ipywidgets python package).

jasongrout:
Well, technically it should probably be a major bump of base, since it does change the compilation target to es2017, and uses ts 3.5 now, and I've bumped the dependencies to services to *just* be 4.0rc0, instead of previous versions as well. HOWEVER, if all you were doing was importing the token so you could register a custom widget, likely no change needed. So it seems like a bit of unnecessary work to make 3rd party libraries release a new version just to have @jupyter-widgets/base 2.0. However, it does put us on better footing moving forward.

Another alternative would be to try to shoehorn the changes in base to still support old versions of @jupyterlab/services and not upgrade the ts targets, etc. That would enable current widget libraries to keep working without updates.

Thoughts? @SylvainCorlay?

SylvainCorlay:
> Well, technically it should probably be a major bump of base, since it does change the compilation target to es2017, and uses ts 3.5 now, and I've bumped the dependencies to services to just be 4.0rc0, instead of previous versions as well. 

I would be fine with a major bump of `base` (which would probably imply a major bump of the widgetsnbextension python package, but I don't think that it matters so much).

> Another alternative would be to try to shoehorn the changes in base to still support old versions of @jupyterlab/services and not upgrade the ts targets, etc. That would enable current widget libraries to keep working without updates.

Nah, regardless that feels a bit overkill.

jasongrout:
> I would be fine with a major bump of `base` (which would probably imply a major bump of the widgetsnbextension python package, but I don't think that it matters so much).

The biggest ramification of this is that every third-party extension will need to bump and republish their js versions (and hence their python versions for classic extensions) since they all rely on base directly or indirectly, probably with semver range `^1`.

SylvainCorlay:
> The biggest ramification of this is that every third-party extension will need to bump and republish their js versions (and hence their python versions for classic extensions) since they all rely on base directly or indirectly, probably with semver range ^1.

Do you *have* to bump major when bumping a dependency major version?

jasongrout:
I didn't say they would have to bump major, though that would be good practice if they are depended on. The key is that they will need to publish a new version.

SylvainCorlay:
If we don't bump major, what do we know already depends on the content of `lib` and may break?

jasongrout:
Probably anyone depending on base for doing a new widget manager would break (like the examples in our repo broke, for instance). Possibly someone building a third-party extension that doesn't understand js classes (since we have a new ts target). Also someone using `@jupyterlab/services` less than 4.0RC directly would break.

SylvainCorlay:
ok, so voila breaks.

jasongrout:
On the other hand, I think things like ipyvolume, pythreejs, bqplot, etc. probably would all work just fine and require no code changes beyond updating their dependencies (i.e., if we published base as a patch release, these libraries *probably* would work just fine)

jasongrout:
> ok, so voila breaks.

Probably. Perhaps that is also a good test case to see if it really does break.

jasongrout:
Regardless, I think it's important to have this PR in a published version of ipywidgets before the jlab release later this week.

vidartf:
Having base change its compilation target from ES5 to ES2017 is definitely a breaking change that should require a major version bump. I don't see a safe way around it, given that we have a very limited overview of how people consume it (semver is the only real guarantee we give).

jasongrout:
> Having base change its compilation target from ES5 to ES2017 is definitely a breaking change that should require a major version bump.

Note that having services change its compilation target to es2017 may be enough for us to force the major bump here as well.

PR #2462, opened 2019-06-27
Dryrun prerelease

SylvainCorlay:


packages/controls/src/version.ts
```diff
@@ -6,4 +6,4 @@
  * implements.
  */
 export
-const JUPYTER_CONTROLS_VERSION = '1.4.0';
+const JUPYTER_CONTROLS_VERSION = require('../package.json').version;
```
SylvainCorlay: @vidartf I am reading that value from `package.json`. It appears I cannot use the `import` syntax because of the ts root dir though.

jasongrout: No, this is not the package.json version number. It's the controls *spec* version number, which may be different from the package.json version number. For example, you'll notice that these are already different. package.json has version 1.4.5, while this is 1.4.0. This allows us to bump the package version without bumping the attribute spec version.

jasongrout: Our guarantee is that a bump in the spec version will bump the package version, but if something changes in the package that does not change the spec, the package version can bump without changing the spec version.

packages/schema/jupyterwidgetmodels.v7-5.md
```diff
@@ -0,0 +1,1054 @@
+# Model State
+
+This is a description of the model state for each widget in the core Jupyter widgets library. The model ID of a widget is the id of the comm object the widget is using. A reference to a widget is serialized to JSON as a string of the form `"IPY_MODEL_<MODEL_ID>"`, where `<MODEL_ID>` is the model ID of a previously created widget of the specified type.
+
+This model specification is for ipywidgets 7.4.*, @jupyter-widgets/base 1.1.*, and @jupyter-widgets/controls 1.4.*.
```
jasongrout: This should be updated by updating the script that generates this text.

packages/controls/src/version.ts
jasongrout: Did the spec change in a backwards incompatible way? If not, then the controls spec should go to 1.5, while the package version should go to 2.0

jasongrout: (that would allow this lab manager to work with ipywidgets 7.4, which asks for controls *spec* `^1.4`...)

jasongrout:
Awesome, thanks for doing this! I'm sure there are issues with the release instructions, so I'm glad you are doing the release.

I put in a couple of notes above about the subtle tricky things involving the various version numbers in the code.

packages/controls/src/version.ts
vidartf: @jasongrout Doesn't this break the managers that use the version string for CDN?

SylvainCorlay: @vidartf this is an excellent point.

PR #2473, opened 2019-07-01
Backward-compatible-control

SylvainCorlay:
Setting the version of `@jupyter-widgets/controls` to the latest released version before running another RC.


jasongrout:
Master is 1.4.5, right? So this should be 1.5 in the release?

SylvainCorlay:
> Master is 1.4.5, right? So this should be 1.5 in the release?

yes, absolutely. Dunno why I was using 1.4.3.

PR #2489, opened 2019-07-09
Reduce count of deprecation warnings

Jamim:
Hello,

Hope you'll find this PR useful.

Fixed warnings:
 - Widget registration using a string name has been deprecated. Widget registration now uses a plain `@register` decorator.
 - Traits should be given as instances, not types (for example, `Int()`, not `Int`). Passing types is deprecated in traitlets 4.1.
 - Deprecated in traitlets 4.1, use the instance .metadata dictionary directly, like x.metadata[key] or x.metadata.get(key, default)
 - Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

These changes also remove some unused imports.

Best regards!

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -8,12 +8,14 @@ in the IPython notebook front-end.
 """
 
 from contextlib import contextmanager
-import collections
-import sys
+try:
+    from collections.abc import Iterable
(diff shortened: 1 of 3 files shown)


vidartf:
Thanks!

Issue #2499, opened 2019-07-14
CSS class 'jupyter-widgets' should not appear node 'jp-OutputArea-output'


qzchenwl:
| package | version |
| ---          | ---        |
| jupyterlab | 1.0.2 |
| jupyterlab-manager | 1.0.1 |

When customizing my own widget, I need to calculate the output area width. If the node 'jp-OutputArea-output' has css attribute 'overflow: visible', I cannot get the correct width.

And there's no means to change css attribute of node 'jp-OutputArea-output', because it's my parent node.

Without 'jupyter-widgets' class, the node 'jp-OutputArea-output' has css attribute 'overflow: auto' which is fine.

![image](https://user-images.githubusercontent.com/146802/61178909-37695d00-a62a-11e9-88d1-44e921798825.png)

jasongrout:
Closed by #2500.

PR #2500, opened 2019-07-14
remove class `jupyter-widgets` from jp-OutputArea-output node

qzchenwl:
Fix issue #2499

### DIFF
--- a/packages/jupyterlab-manager/src/renderer.ts
+++ b/packages/jupyterlab-manager/src/renderer.ts
@@ -48,7 +48,6 @@ class WidgetRenderer extends Panel implements IRenderMime.IRenderer, IDisposable
 
     // Let's be optimistic, and hope the widget state will come later.
     this.node.textContent = 'Loading widget...';
-    this.addClass('jupyter-widgets');
 
     const manager = await this._manager.promise;
     // If there is no model id, the view was removed, so hide the node.


jasongrout:
The `jupyter-widgets` class is meant to be the very base class for jupyter widgets that sets some CSS defaults (such as font color, etc.). However, I notice that it is set a *lot* in the code, at multiple levels. I think we need to be more deliberate about how we are using it (e.g., is it convention that it is applied at the top-level widget? At the root node of each widget? Above the entire widget hierarchy?)

Also, perhaps we could just change the overflow setting. It was set to `visible` in 2016 (https://github.com/jupyter-widgets/ipywidgets/pull/919).

Whatever change we decide on, I think we need to carefully examine lots of examples using widgets to understand the ramifications of this change. We've seen this before with seemingly small changes that had large ramifications, especially in CSS.

SylvainCorlay:
I think that the `jupyter-widgets` CSS class is only meant for the top-level DOM elements.

I would be in favor of this, but I need to double check that everything is fine in the PR.

jasongrout:
I did some initial investigations. Visually, removing this class removes a layer of padding around the entire set of widgets.

I'm a little curious about the css for this class of `overflow: visible`, which is the main point here. In general, it seems odd to let the content flow out of the widget's containing box, so I'm not sure why it is set at all. Tracing back through the history, it seems this `overflow: visible` was added in https://github.com/jupyter-widgets/ipywidgets/pull/796, and it's not clear what problem it was solving.

jasongrout:
> I'm a little curious about the css for this class of `overflow: visible`, which is the main point here. In general, it seems odd to let the content flow out of the widget's containing box, so I'm not sure why it is set at all. Tracing back through the history, it seems this `overflow: visible` was added in #796, and it's not clear what problem it was solving.

But this is a different issue than this one, i.e., `.jupyter-widgets` being applied at the top level.

Let's merge this and address any other problems that come up. Thanks again @qzchenwl!

PR #2504, opened 2019-07-16
Update instructions for jupyterlab-manager 1.0

jtpio:
Update the instructions to install `jupyterlab-manager` for JupyterLab 1.0.

The fact that `1.0` is missing from the list can be a bit confusing and was brought up by [josh](https://discourse.jupyter.org/u/josh/summary) in https://discourse.jupyter.org/t/ipywidgets-for-jupyterlab-1-0/1675

In the long run I think it would make sense to remove all mentions to the pre-releases. Maybe after JupyterLab 1.0 has settled in.

### DIFF
--- a/packages/jupyterlab-manager/README.md
+++ b/packages/jupyterlab-manager/README.md
@@ -1,15 +1,14 @@
 Jupyter Widgets JupyterLab Extension
 ====================================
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thx!

PR #2516, opened 2019-07-24
Fix setting a selection index when having duplicate values

### LINKED ISSUE #2513: RadioButton with options having the same value
```python
import ipywidgets as widgets
# widgets.__version__ == 7.2.1

colors = [["red",1], ["blue",2] ,["green",3], ["yellow",2]]
blob = widgets.RadioButtons(options=colors)
blob
```
For example selecting "yellow", it will automatically change the selected option to "blue". It picks the lowest indexed option with the same value. To actually select "yellow" a second click is needed.

Is this the intended behavior? If not, it would be real neat if options with different labels and the same values was supported.

jasongrout:
Fixes #2513 

Basically, when changing a selection that had duplicate values or labels, we always ended up resetting the index to the first such value or label. Now we check to see if the current index is the correct label/value before we reset the index.

I added a test which fails before the fix and passes after the fix.

### DIFF
--- a/ipywidgets/widgets/tests/test_widget_selection.py
+++ b/ipywidgets/widgets/tests/test_widget_selection.py
@@ -65,3 +65,32 @@ class TestSelection(TestCase):
         assert select.value == 4
         assert select.label == '4'
         assert observations == [0]
+
+    def test_duplicate(self):
+        select = Select(options=['first', 1, 'dup', 'dup'])
+        observations = []
+        def f(change):
+            observations.append(change.new)
+        select.observe(f, 'index')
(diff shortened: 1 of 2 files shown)


jasongrout:
CC @maartenbreddels @bbettss

bbettss:
Awesome, thank you!

jasongrout:
Thanks for pointing out the issue with a very clear reproducible code snippet!

maartenbreddels:
Awesome, LGTM!

jasongrout:
Feel free to merge @maartenbreddels :)

PR #2519, opened 2019-07-26
Add $el to BoxView and ControllerView

### LINKED ISSUE #2497: $el missing when inheriting from BoxView
The BeakerX team is separating out table display as a standalone widget and in the process is upgrading it to work with the latest versions of the Jupyter ecosystem libraries. We have run into issues. @piorek writes:

> I've changed to @jupyter-widgets/base ^1 || ^2 and that part is working fine while using ipywidgets<7.5 but as soon as I try to use it with ipywidgets 7.5 I have a problem.
> In many our views like `EasyFormView` we are using `Bootstrap.View` stuff like having `this.$el` but after switching to ipywidgets#7.5 it's not there at runtime at least not here:
https://github.com/twosigma/beakerx/blob/lukasz/8056_separate_table_display/js/notebook/src/EasyForm.ts#L66
> maybe it's because of
> * where we are using it,
> * some changes in ipywidgets#7.5,
> * some changes in jupyterlab#1.0 itself

cc: @LeeTZ @lukasz @SylvainCorlay @jasongrout 
cf. https://github.com/twosigma/beakerx/pull/8076

jasongrout:
Fixes #2497

### DIFF
--- a/packages/controls/src/widget_box.ts
+++ b/packages/controls/src/widget_box.ts
@@ -26,6 +26,7 @@ import {
 } from '@phosphor/widgets';
 
 import * as _ from 'underscore';
+import $ from 'jquery';
 
 export
 class BoxModel extends CoreDOMWidgetModel {
@@ -79,6 +80,7 @@ class BoxView extends DOMWidgetView {
         }
 
         this.el = this.pWidget.node;
+        this.$el = $(this.pWidget.node);
     }
 
     initialize(parameters: any) {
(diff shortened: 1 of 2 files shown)


jasongrout:
CC @maartenbreddels, @SylvainCorlay for a sanity check.

afshin:
Thank you!

jasongrout:
This is released in ipywidgets 7.5.1

PR #2521, opened 2019-07-29
Execute docs when building always

jasongrout:
Currently we manually execute notebooks to store the widget state in them for documentation. This manual process is tedious and error-prone, and requires us frequently run various documentation notebooks and commit results in order to keep things up to date.

I just tested, and always executing notebooks seems to work well with storing widget state. I also stripped all output of all the example notebooks to clear things out of the repo.

I think going forward we should make sure the documentation examples do not include output, and we execute on documentation build.

Thoughts?

CC @SylvainCorlay, who worked a lot on documentation building and widget state storage.


maartenbreddels:
Yes good idea, I think we should also do it during CI.

> I just tested, and always executing notebooks seems to work well with storing widget state.

Good to hear that!

mwcraig:
I’m 💯 for this!

SylvainCorlay:
YES!

PR #1734, opened 2017-10-02
Typings improvement

vidartf:
Some progress on #1672, now stranded a bit. See also #1733.

PR #2528, opened 2019-08-07
Make more of lab manager dependencies optional

vidartf:
It would make the lab manager more reusable if it had fewer strict requirements. I kept the rendermime dependency strict, as I see no way to use it without this.

Other things that I added since they were related (see individual commits for split):
- Cleaned up use of types.
- Fixed an assign in `if` clause (should have been compare `===`).
- Added strict null checks to lab manager package.

If any of these auxiliary points are contentious at all, just say the word and I will split up the PR.

### DIFF
--- a/packages/jupyterlab-manager/src/plugin.ts
+++ b/packages/jupyterlab-manager/src/plugin.ts
@@ -157,8 +157,8 @@ export function registerWidgetManager(
  */
 const plugin: JupyterFrontEndPlugin<base.IJupyterWidgetRegistry> = {
   id: '@jupyter-widgets/jupyterlab-manager:plugin',
-  requires: [INotebookTracker, IRenderMimeRegistry, ISettingRegistry],
-  optional: [IMainMenu, ILoggerRegistry],
+  requires: [IRenderMimeRegistry],
+  optional: [INotebookTracker, ISettingRegistry, IMainMenu, ILoggerRegistry],
   provides: base.IJupyterWidgetRegistry,
   activate: activateWidgetExtension,
   autoStart: true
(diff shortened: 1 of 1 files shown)


packages/jupyterlab-manager/src/plugin.ts
```diff
@@ -168,17 +168,37 @@ function updateSettings(settings: ISettingRegistry.ISettings) {
...
-  });
+  if (settingRegistry !== null) {
+    settingRegistry.load(plugin.id).then((settings: ISettingRegistry.ISettings) => {
+      settings.changed.connect(updateSettings);
+      updateSettings(settings);
+    }).catch((reason: Error) => {
+      console.error(reason.message);
+    });
+
+    // Add a command for creating a new Markdown file.
```
jasongrout: This comment apparently was a copy-paste error originally.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -266,8 +266,8 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
    * Get the currently-registered comms.
    */
   async _get_comm_info(): Promise<any> {
-    const reply = await this._context.session.kernel.requestCommInfo({target: this.comm_target_name});
-    if (reply.content.status = 'ok') {
+    const reply = await this._context!.session.kernel!.requestCommInfo({target: this.comm_target_name});
+    if (reply.content.status === 'ok') {
```
jasongrout: Thanks for catching this!

jasongrout:
At first glance, this looks great!

vidartf:
Let's wait on a rebase after #1734 to make this diff smaller...

vidartf:
I cleaned up this PR a lot, so it should now be a lot quicker to review!

jasongrout:
Thanks!

PR #2531, opened 2019-08-09
Load typescript sources directly into Karma using the webpack plugin.

blois:
Drop the explicit tsc & webpack build steps for src and test code and instead integrate the tsc compilation into webpack and integrate webpack into karma.

This only changes the tests in packages/base- looking for feedback on this first.

This has the benefits:
 - Source maps work in the browser for all source & test files.
 - Automatic compilation when modifying TS files- tests can automatically execute or a simple browser refresh will execute updated tests.

If you execute:

```shell
yarn run test:unit:default:debug
```

Then you can open a local browser to Karma and just refresh the page after any file save to execute the tests with the updated changes.

The webpack coverage config appears to not work before this change and this change doesn't improve the situation.

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -18,16 +18,16 @@
   "scripts": {
     "build": "npm run build:src",
     "build:src": "tsc",
-    "build:test": "tsc --project test && webpack --config test/webpack.conf.js",
     "clean": "npm run clean:src",
     "clean:src": "rimraf lib && rimraf tsconfig.tsbuildinfo",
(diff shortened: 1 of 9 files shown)


jasongrout:
@blois - sorry for the delay in looking at this. We had a sprint this last week in which we split apart base into a base widgets module and a base manager module. If you can rebase this (if it still makes sense to do so), I'd love to review and merge it. It looks like the benefits are great!

Another possible option is to switch to jest for our testing, but that may be a bigger change to examine later.

PR #2532, opened 2019-08-13
Labmanager refactor

vidartf:
This PR builds on top of #2528.

**What does this do?** 
Extract parts of the widget manager into an abstract base class, and then export a new widget manager that does not rely on the document context. Instead, it takes a kernel as an argument to the constructor.

**What is the purpose?**
The goal of this PR is to make it easier for others to reuse widgets in a setting outside of Jupyterlab, (but while still using Jupyterlab components as libraries).

**Does this really belong in this package?**
Ideally, the base and kernel-based manager could live in a new, separate package, but I don't want to increase the maintenance burden of the repo/releases any more, so I think it would probably be better to later on optimize the current package for tree-shaking.

### DIFF
--- a/packages/jupyterlab-manager/src/manager.ts
+++ b/packages/jupyterlab-manager/src/manager.ts
@@ -89,58 +89,11 @@ export class BackboneViewWrapper extends Widget {
 /**
  * A widget manager that returns Lumino widgets.
  */
-export class WidgetManager extends ManagerBase<Widget> implements IDisposable {
-  constructor(
-    context: DocumentRegistry.IContext<INotebookModel>,
-    rendermime: IRenderMimeRegistry,
-    settings: WidgetManager.Settings
-  ) {
+export abstract class LabWidgetManager extends ManagerBase<Widget>
+  implements IDisposable {
+  constructor(rendermime: IRenderMimeRegistry) {
     super();
-    this._context = context;
     this._rendermime = rendermime;
-
-    // Set _handleCommOpen so `this` is captured.
-    this._handleCommOpen = async (comm, msg): Promise<void> => {
-      const oldComm = new shims.services.Comm(comm);
-      await this.handle_comm_open(oldComm, msg);
-    };
-
-    context.sessionContext.kernelChanged.connect((sender, args) => {
-      this._handleKernelChanged(args);
-    });
-
-    context.sessionContext.statusChanged.connect((sender, args) => {
-      this._handleKernelStatusChange(args);
-    });
-    context.sessionContext.connectionStatusChanged.connect((sender, args) => {
-      this._handleKernelConnectionStatusChange(args);
-    });
-
-    if (context.sessionContext.session?.kernel) {
-      this._handleKernelChanged({
-        name: 'kernel',
-        oldValue: null,
-        newValue: context.sessionContext.session?.kernel
-      });
-    }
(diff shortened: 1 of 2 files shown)


vidartf:
For reference, here is the diff of this PR onto #2528: https://github.com/vidartf/ipywidgets/compare/lab-optionals...vidartf:labmanager-refactor

jasongrout:
Needs a rebase (and at least some of this is incorporated in other PRs like #2707

vidartf:
Rebased, but will need a careful review, as I'm not sure I caught all the nuances with the new services API.

jasongrout:
@vidartf, I took a look at this today. I pushed a few commits cleaning up a few things. If those look good to you, then I think we can merge this.

jasongrout:
I also fixed a race condition that has been bothering us since updating to JLab 2.0. CC @wolfv.

jasongrout:
@vidartf, do you want to take a look at this again?

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -504,10 +505,14 @@ export class KernelWidgetManager extends LabWidgetManager {
...
-    this._restoredStatus = true;
-    this._initialRestoredStatus = true;
-    this._restored.emit();
+    try {
+      await this._loadFromKernel();
+      this._restoredStatus = true;
+      this._initialRestoredStatus = true;
+      this._restored.emit();
+    } catch (err) {
+      // Do nothing
```
vidartf: This shouldn't really happen for the `KernelWidgetManager`, but I agree with being defensive.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -614,7 +616,12 @@ export class WidgetManager extends LabWidgetManager {
   ): Promise<void> {
     try {
       if (loadKernel) {
-        await this._loadFromKernel();
+        try {
+          this._kernelRestoreInProgress = true;
+          await this._loadFromKernel();
+        } finally {
+          this._kernelRestoreInProgress = false;
+        }
```
vidartf: Would it make more sense to put the setting/unsetting of this flag inside the method?

jasongrout:
> Adding off-topic commits is a little annoying, but I see why it is expedient. Lets merge and iterate on it (comments for consideration for a follow-up).

I assume you are talking about the last two commits that fix a serious bug. If I would have made them in a different PR, I would have had to redo them anyway, so it only made sense to make them in the context of this PR. Thanks for the seeing the expediency.

packages/jupyterlab-manager/src/manager.ts
jasongrout: Do you mean there won't be an error in loading from kernel in the widget manager? We aren't throwing one explicitly, but of course there could be one, and that would leave us in a weird state for the flag, so I think it's important to be defensive.

Probably even better semantically to have it in a finally statement, though.

packages/jupyterlab-manager/src/manager.ts
jasongrout: Can you be more explicit about what you have in mind?

packages/jupyterlab-manager/src/manager.ts
jasongrout: Or maybe even better to move the flag setting into _loadFromKernel...

PR #2534, opened 2019-08-20
feat: produce embed bundle without lab theme variables

maartenbreddels:
This PR moves the HTML manager a bit closer to voila works, making it possible to import the js module from a CDN, and have a custom way of loading the widget state etc.

On top of that, by default the `labvariables.css` were included, which would override any theming done (e.g. jupyter lab's black theme). To still support the default (light) theme (in embed.js and embed-amd.js) the css is move to those modules, while the `widget-base.css` is kept.


maartenbreddels:
Ok, simplified this PR a bit, since the original goal was to have theming work in nbconvert, like voila. This works now with nbconvert using this branch:
image

maartenbreddels:
I think this is replaced by #3342

PR #2536, opened 2019-08-21
FIX: TypeError: NoneType in xeus-python

hainm:
Please see below error:

image

Original issue is here: https://github.com/QuantStack/xeus-python/issues/124#issuecomment-523533272


hainm:
may be this is something should be fixed in xeus-python.

Issue #2511, opened 2019-07-22
The usage of custom widgets with @interact is unnecessarily restricted and incorrectly documented


bermani:
This issue is directly related to #2261 but I thought I would make a new issue to show all of my findings. Essentially, interact has unnecessary restrictions on the widgets that can be used, these restrictions are conflicting with the behavior specified in the documentation.

## Current Documentation

According to the [docs on interact](https://ipywidgets.readthedocs.io/en/stable/examples/Using%20Interact.html#Widget-abbreviations), 

> This examples clarifies how interact processes its keyword arguments:
>
>1. If the keyword argument is a Widget instance with a value attribute, that widget is used. Any widget with a value attribute can be used, even custom ones.
>
>2. Otherwise, the value is treated as a widget abbreviation that is converted to a widget before it is used.

We should be able to pass any Widget instance with a value attribute, however, this is not the case.

### 1. Interact errors on any widget that's not a ValueWidget
https://github.com/jupyter-widgets/ipywidgets/blob/fcadca8fbdc18ef1d8dc5694e8e6e759af6c7749/ipywidgets/widgets/interaction.py#L297-L302

The call to `widget_from_abbrev` on line 297 is intended to convert any widgets to ValueWidgets; the docstring for `widget_from_abbrev` states that it should do this:
https://github.com/jupyter-widgets/ipywidgets/blob/fcadca8fbdc18ef1d8dc5694e8e6e759af6c7749/ipywidgets/widgets/interaction.py#L310-L311

However, when examining this function and the functions it calls, there is no section that attempts to build a ValueWidget instance from a passed widget, it only either accepts a ValueWidget instance or tries to make one from abbreviations.

https://github.com/jupyter-widgets/ipywidgets/blob/fcadca8fbdc18ef1d8dc5694e8e6e759af6c7749/ipywidgets/widgets/interaction.py#L342-L343
The function returns none, and thus the ValueError on line 300 is raised. This is a confusing and incorrect error message when passing a Widget that isn't a ValueWidget, to see "YourWidget cannot be transformed into a widget". 

### 2. Interact errors on any widget that doesn't have a description attribute
https://github.com/jupyter-widgets/ipywidgets/blob/fcadca8fbdc18ef1d8dc5694e8e6e759af6c7749/ipywidgets/widgets/interaction.py#L303-L304
[...]

PR #2544, opened 2019-08-27
Fix documentation and add more helpful error message for interact usage

bermani:
@SylvainCorlay suggested in #2511 that the current design for `interact` should be that only ValueWidgets can be passed as keyword arguments. This PR edits the documentation to reflect this and changes the error message that displays when a non-ValueWidget Widget is passed to let the user know that a ValueWidget is expected. ValueWidgets are currently not mentioned in the documentation nor any accessible error messages which gave me a big headache when trying to start a custom widget project.

### DIFF
--- a/docs/source/examples/Using Interact.ipynb
+++ b/docs/source/examples/Using Interact.ipynb
@@ -197,12 +197,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This example clarifies how `interact` processes its keyword arguments:\n",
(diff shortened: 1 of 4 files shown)


pbugnion:
@bermani This looks great! Thanks very much for submitting it, and apologies for the delay in replying.

I've pushed a commit to re-order things slightly in the documentation.

pbugnion:
Thanks very much for this, @bermani !

PR #2556, opened 2019-09-11
Make unpack_models compatible with ISerializers

vidartf:
Previously, `unpack_models` could not be typed it as `ISerializers['deserialize']` when strict null checks are on.


SylvainCorlay:
Hold until https://github.com/jupyter-widgets/ipywidgets/pull/1734 is merged

PR #2570, opened 2019-09-30
Reloading kernel state race issue

jasongrout:
If there is a slow kernel startup, we might have some time in between the jlab widget manager’s constructor and the kernel’s connect signal. This means we may restore from the kernel on widget manager construction, then execute some cells, then restore again when the kernel signals it is connected. We fix two problems in this situation.

1. Since we might have a cell execution before the kernel connect signal, there may be some widgets already created. In that case, we want to avoid recreating those widgets. In particular, we should not take over their comm objects and replace their on_msg handler.

2. Previously, we were sending a comm message to request state and just listening for the next message back on the comm that was an update message. However, if the comm was already active, the next message back *might* be a normal update message, rather than a full state update, which would mess up our logic. We fix this by listening for an update comm message with a parent message header matching our sent message.

Finally, we fix one more issue - at startup, we want to restore widgets from the notebook and the kernel, but on a kernel connect signal, we only want to restore widgets from the kernel.

This still does not handle the case where we disconnect from the kernel, miss some update messages, and then reconnect. In order to handle that case, it would be good to have a way to ask an existing widget to update its own state with a full refresh from the kernel.

### DIFF
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -52,8 +52,8 @@ echo -n "ipywidgets"
 pip install -v -e .
 
 if test "$skip_jupyter_lab" != yes; then
-    jupyter labextension link ./packages/base
-    jupyter labextension link ./packages/controls
-    jupyter labextension link ./packages/output
-    jupyter labextension install ./packages/jupyterlab-manager
+    jupyter labextension link ./packages/base --no-build
+    jupyter labextension link ./packages/controls --no-build
+    jupyter labextension link ./packages/output --no-build
+    jupyter labextension install ./packages/jupyterlab-manager --minimize=False
 fi
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -55,5 +55,5 @@ if test "$skip_jupyter_lab" != yes; then
     jupyter labextension link ./packages/base --no-build
     jupyter labextension link ./packages/controls --no-build
     jupyter labextension link ./packages/output --no-build
-    jupyter labextension install ./packages/jupyterlab-manager
+    jupyter labextension install ./packages/jupyterlab-manager --minimize=False
 fi
--- a/dev-install.sh
+++ b/dev-install.sh
@@ -55,5 +55,5 @@ if test "$skip_jupyter_lab" != yes; then
     jupyter labextension link ./packages/base --no-build
     jupyter labextension link ./packages/controls --no-build
     jupyter labextension link ./packages/output --no-build
-    jupyter labextension install ./packages/jupyterlab-manager --minimize=False
+    jupyter labextension install ./packages/jupyterlab-manager
 fi
--- a/packages/jupyterlab-manager/src/manager.ts
+++ b/packages/jupyterlab-manager/src/manager.ts
@@ -267,7 +267,7 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
    */
   async _get_comm_info(): Promise<any> {
     const reply = await this._context.session.kernel.requestCommInfo({target: this.comm_target_name});
-    if (reply.content.status = 'ok') {
+    if (reply.content.status === 'ok') {
         return (reply.content as any).comms;
     } else {
         return {};
--- a/packages/jupyterlab-manager/src/manager.ts
+++ b/packages/jupyterlab-manager/src/manager.ts
@@ -198,6 +198,9 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
       return;
     }
     await this.context.session.ready;
+    if ((this.context.session.kernel as any).handleComms === false) {
+      return;
+    }
     const comm_ids = await this._get_comm_info();
 
     // For each comm id that we do not know about, create the comm, and request the state.
--- a/packages/jupyterlab-manager/src/manager.ts
+++ b/packages/jupyterlab-manager/src/manager.ts
@@ -14,6 +14,10 @@ import {
 } from '@phosphor/disposable';
 
 import {
+  PromiseDelegate
+} from '@phosphor/coreutils';
+
+import {
   Widget
 } from '@phosphor/widgets';
 
@@ -156,7 +160,8 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   _handleKernelStatusChange(args: Kernel.Status) {
     switch (args) {
     case 'connected':
-      this.restoreWidgets(this._context.model);
+      // We only want to restore widgets from the kernel, not ones saved in the notebook.
+      this.restoreWidgets(this._context.model, {loadKernel: true, loadNotebook: false});
       break;
     case 'restarting':
       this.disconnect();
@@ -168,9 +173,13 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   /**
    * Restore widgets from kernel and saved state.
    */
-  async restoreWidgets(notebook: INotebookModel): Promise<void> {
-    await this._loadFromKernel();
-    await this._loadFromNotebook(notebook);
+  async restoreWidgets(notebook: INotebookModel, {loadKernel, loadNotebook}: {loadKernel: boolean, loadNotebook: boolean} = {loadKernel: true, loadNotebook: true}): Promise<void> {
+    if (loadKernel) {
+      await this._loadFromKernel();
+    }
+    if (loadNotebook) {
+      await this._loadFromNotebook(notebook);
+    }
     this._restoredStatus = true;
     this._restored.emit();
   }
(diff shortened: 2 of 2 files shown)


jasongrout:
One thing I notice with these changes. Before, in JupyterLab, if you created and displayed a widget:

```python
import ipywidgets as widgets
colors = [["red",1], ["blue",2] ,["green",3], ["yellow",2]]
blob = widgets.RadioButtons(options=colors)
blob
```
then closed the notebook and reopened it, the cell output would just say something like "Loading widgets..." and not complete. Now, the widget properly appears with the current state from the kernel.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +198,51 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+        return;
+      } catch (e) {
+        // If we have the widget model not found error, then we can create the
+        // widget. Otherwise, rethrow the error.
+        if (e.message !== 'widget model not found') {
+          throw e;
+        }
+        const comm = await this._create_comm(this.comm_target_name, comm_id);
+
+        let msgId: string;
```
bbgit-mn: maybe use snake case to be consistent?

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +198,51 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+        let p = new PromiseDelegate<Private.ICommUpdateData>();
```
bbgit-mn: `p` is not very descriptive :)
Also, does not seem you are reassigning the variable should you use `const` instead of `let`?

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -267,7 +301,7 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
    */
   async _get_comm_info(): Promise<any> {
     const reply = await this._context.session.kernel.requestCommInfo({target: this.comm_target_name});
-    if (reply.content.status = 'ok') {
+    if (reply.content.status === 'ok') {
```
bbgit-mn: 👍

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +198,51 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
-              comm: comm,
-              msg: msg
+      try {
+        await this.get_model(comm_id);
+        // If we successfully get the model, do no more.
+        return;
+      } catch (e) {
+        // If we have the widget model not found error, then we can create the
+        // widget. Otherwise, rethrow the error.
+        if (e.message !== 'widget model not found') {
```
bbgit-mn: is it a good idea to rely on the error message?

dev-install.sh
```diff
@@ -52,8 +52,8 @@ echo -n "ipywidgets"
...
 
 if test "$skip_jupyter_lab" != yes; then
-    jupyter labextension link ./packages/base
-    jupyter labextension link ./packages/controls
-    jupyter labextension link ./packages/output
-    jupyter labextension install ./packages/jupyterlab-manager
+    jupyter labextension link ./packages/base --no-build
+    jupyter labextension link ./packages/controls --no-build
+    jupyter labextension link ./packages/output --no-build
+    jupyter labextension install ./packages/jupyterlab-manager --minimize=False
```
bbgit-mn: Can you elaborate what those options will do?

packages/jupyterlab-manager/src/manager.ts
jasongrout: Unfortunately, it's all we can do and still be backwards compatible. Since we don't throw a specific error class, just the generic error class with a specific message, that's all we have available to check.

dev-install.sh
jasongrout: The first three `--no-build` prevent us rebuilding JupyterLab three unnecessary times. The last option builds JupyterLab, but with js minimization off for easier debugging.

packages/jupyterlab-manager/src/manager.ts
jasongrout: Good point. Fix coming right up.

packages/jupyterlab-manager/src/manager.ts
jasongrout: Yes, thank you!

jasongrout:
@bbgit-mn - I think I addressed your review comments.

dev-install.sh
bbgit-mn: Thank you for clarification!

packages/jupyterlab-manager/src/manager.ts
bbgit-mn: Sounds reasonable

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +198,51 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+          throw e;
+        }
+        const comm = await this._create_comm(this.comm_target_name, comm_id);
+
+        let msg_id: string;
+        let widget_info = new PromiseDelegate<Private.ICommUpdateData>();
```
bbgit-mn: forgot to change to `const`?

jasongrout: Oh, right. Coming up.

jasongrout: done in 8303054

packages/jupyterlab-manager/src/manager.ts
bbgit-mn: maybe add a comment at: https://github.com/jupyter-widgets/ipywidgets/blob/830305493969d7e2ded56a219b9318dfe55c9102/packages/jupyterlab-manager/src/manager.ts#L414
in case someone decides to refactor the code in the future?

bbgit-mn:
Looks like the CI checks are failing do to connectivity issue:
```Could not connect to apt.cache.travis-ci.com:80 (34.96.81.152), connection timed out```

**Comment by vidartf** [comment]
To answer some of my own points:
> - If the kernel connection is lost, then regained, [..] We could also say that "any changes made in the client while disconnected will be overwritten by the state from the kernel on reconnect",

Yes, as this PR does: Only restore from kernel. That part is good.

- If the kernel changes to a fresh kernel (e.g.kernel restart), what should happen? [...]

As now, leave state in memory as is ("disconnected" widget shown if outputs not cleared). If all outputs are cleared, do we currently close those widgets? I have some doubts...

- If the kernel changes to an already running kernel (not fresh), what should happen?

Same as reconnect I think (leave in-memory widgets as is, restore from kernel). As I think a kernel change always generate a status change to `connected` (?), I think this is ok as it is.

**Comment by jasongrout** [comment]
> If so, won't a more minimal changeset be to simply not connect to the "statusChanged" signal before the initial `restored` signal is emitted?

Good idea - I just put in a check for that too, to prevent the common double-restore we have now.

I think it's still good to leave in the logic in this PR - it handles any other connect-triggered restore as well.

**Comment by jasongrout** [comment]
@vidartf - does this look good to you now? I'd like to cut a patch release with it to fix some bugs that were reported with slower kernels.

File: packages/jupyterlab-manager/src/manager.ts [hunk 7]
```diff
@@ -189,29 +201,53 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+            && msg.content.data.method === 'update') {
+            let data = (msg.content.data as any);
+            let buffer_paths = data.buffer_paths || [];
+            // Make sure the buffers are DataViews
+            let buffers = (msg.buffers || []).map(b => {
+                if (b instanceof DataView) {
+                    return b;
+                } else {
+                    return new DataView(b instanceof ArrayBuffer ? b : b.buffer);
+                }
```
vidartf: The reason this bit (`DataView` check) was missing previously was basically that the code wasn't DRY. Ideally this could be extracted into a util function (maybe into `put_buffers` itself, or a new function in utils that wraps that?). I think this code is also currently wrong in `widgetsnbextension`, which was the original inspiration for this code.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +201,53 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
...
+        // widget. Otherwise, rethrow the error. We have to check the error
+        // message text explicitly because the get_model function in this
+        // class throws a generic error with this specific text.
+        if (e.message !== 'widget model not found') {
+          throw e;
+        }
+        const comm = await this._create_comm(this.comm_target_name, comm_id);
+
+        let msg_id: string;
+        const widget_info = new PromiseDelegate<Private.ICommUpdateData>();
```
vidartf: It's a little confusing that this is called `widget_info` vs the outside `widgets_info` (I wasted a few cycles tracing this).

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -156,7 +160,11 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   _handleKernelStatusChange(args: Kernel.Status) {
     switch (args) {
     case 'connected':
-      this.restoreWidgets(this._context.model);
+      // Only restore if our initial restore at construction is finished
```
vidartf: I think this logic is faulty:

`_restoredStatus` is also set to `false` during a disconnect/reconnect cycle, and we only want this prevention during the initial construction (as this comment says). I think the right was to do this is to not connect directly in the ctor, but rather only connect after the inital restored signal is sent. Actually, ideally the initial restore would have been a promise so that we also connect if the initial restore throws an exception.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -168,9 +176,13 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
   /**
    * Restore widgets from kernel and saved state.
    */
-  async restoreWidgets(notebook: INotebookModel): Promise<void> {
-    await this._loadFromKernel();
-    await this._loadFromNotebook(notebook);
+  async restoreWidgets(notebook: INotebookModel, {loadKernel, loadNotebook} = {loadKernel: true, loadNotebook: true}): Promise<void> {
```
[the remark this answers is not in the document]
jasongrout: So something like `options = {loadKernel: true, loadNotebook: true}`? I don't think that's significantly easier to read, but maybe that's because I've been using unpacking more and more and sort of like it.

packages/jupyterlab-manager/src/manager.ts
```diff
@@ -189,29 +201,53 @@ class WidgetManager extends ManagerBase<Widget> implements IDisposable {
       return;
     }
     await this.context.session.ready;
+    if ((this.context.session.kernel as any).handleComms === false) {
```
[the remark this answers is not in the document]
jasongrout: Good call - it's because I didn't update the services dependency to 4.1.0 where that parameter was introduced.

packages/jupyterlab-manager/src/manager.ts
jasongrout: widget_info is a single thing in widgets_info list - so it makes sense to me. However, I agree it could be easy to miss the `s` in the middle of the word when reading the code (as opposed to an s at the end).

packages/jupyterlab-manager/src/manager.ts
jasongrout: Can we make this refactoring a separate ticket? I'd like to get the bug fixed in this PR out ASAP.

packages/jupyterlab-manager/src/manager.ts
jasongrout: We need to connect right away, since there are other things happening such as calling the parent disconnect method. I've added a new variable to keep track of the initial restore, though.

jasongrout:
@vidartf - I think I've addressed all of your comments above.

packages/jupyterlab-manager/src/manager.ts
vidartf: Sure, maybe leave a TODO comment? 😕

dev-install.sh
jasongrout: @vidartf clarified that it *says* it is building a minimized build, but it's actually not.

packages/jupyterlab-manager/src/manager.ts
jasongrout: I'll open an issue.

vidartf:
Just pick up those three lints (`--minimize`, `s`, and the issue for the DRY bit), and then you can self-merge 👍

packages/jupyterlab-manager/src/manager.ts
jasongrout: https://github.com/jupyter-widgets/ipywidgets/issues/2588

PR #2578, opened 2019-10-07
DOC: Use a more modern way to activate recommonmark

mgeier:
I don't know if anybody else had those problems, but on my local installation the Markdown files were not parsed correctly.

Anyway, the recommended way to use Markdown source files has changed, see https://www.sphinx-doc.org/en/master/usage/markdown.html.

FYI: Explicitly specifying `'.ipynb'` as source suffix is not necessary, `nbsphinx` takes care of that automatically. If somebody wants to insist on mentioning it, this would be the way to do it:

```python
source_suffix = {
    '.rst': 'restructuredtext',
    '.md': 'markdown',
    '.ipynb': 'jupyter_notebook',
}
```

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,7 +5,7 @@ numpy
 jupyter_client
 jupyter_sphinx
 nbsphinx>=0.2.13
-recommonmark==0.4.0
+recommonmark
 sphinx>=1.4.6
 sphinx_rtd_theme
 matplotlib
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -5,7 +5,6 @@
 import os
 import subprocess
 import sys
-import recommonmark.parser
 
 
 # -- path -------------------------------------------------------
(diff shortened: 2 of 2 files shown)


mgeier:
I've removed the version restriction to `recommonmark==0.4.0` in b77de5a08c93097f9b1f2684faddb0584556f224.

I don't know why this was originally added, but I assume it is not necessary anymore, because `docs/environment.yml` (which is used on RTD) doesn't have this restriction either.

jasongrout:
> I don't know why this was originally added, but I assume it is not necessary anymore, because `docs/environment.yml` (which is used on RTD) doesn't have this restriction either.

The reason is in the commit message for the commit adding that line: https://github.com/jupyter-widgets/ipywidgets/commit/bb8b308b61191d0f078cc1d813008f144049d6a0

If it works now, great!

(FYI, I found that by viewing the file in master at https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/requirements.txt, then clicking on Blame to see the commit that added that line, then clicking on the commit to see the message)

jasongrout:
Thanks!

PR #2590, opened 2019-10-20
Drop notebook dependency from widgetsnbextension

SylvainCorlay:
While we may not drop the `ipywidgets -> widgetsnbextension` dependency in a backward-compatible fashion, a good way to enable the installation of ipykernel in an environment without pulling the notebook server is to drop the `widgetsnbextension -> notebook` dependency.

Could that land in a 7.x release?

### DIFF
--- a/widgetsnbextension/setup.py
+++ b/widgetsnbextension/setup.py
@@ -214,13 +214,5 @@ if 'setuptools' in sys.modules:
     from setuptools.command.develop import develop
     setup_args['cmdclass']['develop'] = js_prerelease(develop, strict=True)
 
-setuptools_args = {}
-install_requires = setuptools_args['install_requires'] = [
-    'notebook>=4.4.1',
-]
-
-if 'setuptools' in sys.modules:
-    setup_args.update(setuptools_args)
-
 if __name__ == '__main__':
     setup(**setup_args)


SylvainCorlay:
cc @jasongrout

jasongrout:
> Could that land in a 7.x release?

I think that would be too much of a backwards-incompatible change in 7.x. But good news is that we plan on an 8.0 soon (January?)

I think dropping ipywidgets -> widgetsnbextension makes more sense that widgetsnbextension->notebook. Or perhaps even better we have a new ipywidgets-base package that is just the python part, and ipywidgets depends on that as well as widgetsnbextension and the jlab extension?

SylvainCorlay:
The thing is that dropping `widgetsnbextension -> notebook` appears to be backward compatible. You still build the static assets and place them in the appropriate configuration directories.

I find it very unlikely that anyone would rely on installing notebook as a byproduct of installing widgetsnbextension and the fix would be trivial to make...

Eventually, we could drop ipywidgets -> widgetsnbextension in a backward incompatible release.

vidartf:
>I find it very unlikely that anyone would rely on installing notebook as a byproduct of installing widgetsnbextension

I can imagine many scenarios for this to be true. 🐱 

> and the fix would be trivial to make...

That is true of many backwards incompatible changes though...

SylvainCorlay:
At the moment there is not means to install ipywidgets without the whole notebook server (without going crazy with no-deps).

Dropping the `widgetsnbextension -> notebook`  dependency seems like a reasonable way to achieve this.

jasongrout:
I can't imagine a scenario where you want to install widgetsnbextension without the notebook. I can imagine lots of scenarios where you want to install ipywidgets with widgetsnbextension. To me, that points to breaking the dependency ipywidgets->widgetsnbextension (which would not be backwards-compatible) or creating a new ipywidgets-base package for just the python code (which would be backwards compatible).

SylvainCorlay:
> I can't imagine a scenario where you want to install widgetsnbextension without the notebook. 

Doing `ipywidgets -----> widgetsnbextension --X--> notebook` makes the installation of ipywidgets only place the classic notebook bundle in your prefix with no other side effect. It is only picked up by the notebook if the latter is installed. This is why it looks like a good way to prevent pulling the notebook server without really breaking backward compatibility.

People will be more confused to have to manually install something like widgetsnbextension  (which they don't know what it is) while having to install both ipywidgets and notebook will makes sense to most.

jasongrout:
To me, a more compelling argument going forward, with jupyter server maturing, is that people may not want to install the notebook, and install jlab (or voila, etc.) without the notebook.

I still feel strongly that this is backwards-incompatible and shouldn't be in 7.x, but I think now is the right time to hash this out for the 8.0 coming soon.

SylvainCorlay:
:unamused:

jasongrout:
> 😒

Help us get 8.0 ready and out! :)

SylvainCorlay:
7.x is getting really stable and this helps adoption... I would really like that dependency to be cut... This is especially annoying for people who use the conda-environment kernel manager and find themselves with the notebook server installed in many environment.

We should work together on the 8.0 objectings while I am in NYC. I would like ipywidgets to become less ipykernel-specific and more amenable to xeus-python and other implementations of the required APIs.

maartenbreddels:
I see this issue popping up in many packages, you want (for convenience, without having to read docs) some default dependency, but you want the option of having the minimal requirements. pip(and/or setuptools) supports additional dependencies, like `pip install voila[test]`, but I'd really like to see `pip install ipywidgets[minimal]`, which will drop dependencies. I'm not sure this is on the radar, or if we should put this on anyone's radar, or start pushing for something. I don't think we are the only people that have these 'convenience' vs 'minimal' install wishes.

jasongrout:
That works really well for pip, and I support adding something like that. It doesn't work for conda, where you can't have variants of a package like that.

maartenbreddels:
The solution I took with vaex, is that `vaex` is a 'meta'-package. It has all the other vaex sub packages as dependecies. Installing `vaex` is a whole-kitchensink-I-don't-care-make-it-work package, for when you care, you can install `vaex-core`, `vaex-hdf5`, `vaex-jupyter`... 

For ipywidgets we could do the same, split up in
 * ipywidgets: requires all the 3 below (does not contain code, but can be strict on the version requirements)
 * ipywidgets-python: requires python (only python code)
 * jupyter-widgets-notebook: require notebook (include the nbextension bundle)
 * jupyter-widgets-lab: requires lab (the lab bundle as .tar.gz)

We can furthermore think about splitting of base from controls everywhere (ipyvuetify only requires base on python and js side), but that requires release orchestration software I think.

vidartf:
@maartenbreddels the disadvantage of the plan you outline is that everyone would have to update all their import statements to be `from ipywidgets_python import ...` / `import ipywidgets_python`, etc.

vidartf:
We discussed this at the workshop (#2641) and decided the benefits of this outweighed the detriments.

PR #2591, opened 2019-10-21
TagsInput widget

martinRenou:
tagsinput

### DIFF
--- a/ipywidgets/widgets/widget_tagsinput.py
+++ b/ipywidgets/widgets/widget_tagsinput.py
@@ -1,0 +1,105 @@
+# Copyright(c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+"""TagsInput class.
+
+Represents a list of tags.
+"""
+
+from traitlets import (
+    CaselessStrEnum, CInt, CFloat, Bool, Unicode, List, TraitError, validate
+)
+
+from .widget_description import DescriptionWidget
+from .valuewidget import ValueWidget
+from .widget_core import CoreWidget
+from .widget import register
+from .trait_types import Color, NumberFormat
+
+
+class TagsInputBase(DescriptionWidget, ValueWidget, CoreWidget):
+    _model_name = Unicode('TagsInputBaseModel').tag(sync=True)
+    value = List().tag(sync=True)
+    allowed_tags = List().tag(sync=True)
+    allow_duplicates = Bool(True).tag(sync=True)
+
+    @validate('value')
+    def _validate_value(self, proposal):
+        if ('' in proposal['value']):
+            raise TraitError('The value of a TagsInput widget cannot contain blank strings')
+
+        if len(self.allowed_tags) == 0:
+            return proposal['value']
+
+        for tag_value in proposal['value']:
+            if tag_value not in self.allowed_tags:
+                raise TraitError('Tag value {} is not allowed, allowed tags are {}'.format(tag_value, self.allowed_tags))
+
+        return proposal['value']
+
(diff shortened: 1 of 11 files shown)


vidartf:
This looks like an interesting widget, but it isn't obvious to me if this belongs in controls or in a third-party package. Do we have a policy here?

martinRenou:
There is more in https://github.com/martinRenou/tagsinput. I can update this PR if needed

martinRenou:
> This looks like an interesting widget, but it isn't obvious to me if this belongs in controls or in a third-party package. Do we have a policy here?

I don't know myself, and I guess it will be the same question for the TextSlider PR. Maybe @jasongrout @SylvainCorlay you can answer this?

ipywidgets/widgets/widget_tagsinput.py
```diff
@@ -0,0 +1,105 @@
...
+    _view_name = Unicode('TagsInputView').tag(sync=True)
+
+    value = List(Unicode(), help='List of string tags').tag(sync=True)
+    tag_style = CaselessStrEnum(
+        values=['primary', 'success', 'info', 'warning', 'danger', ''], default_value='',
+        help="""Use a predefined styling for the tags.""").tag(sync=True)
+
+
+@register
+class ColorsInput(TagsInputBase):
```
SylvainCorlay: ColorTags / TextTags / IntTags / FloatTags?

SylvainCorlay:
Vidar: Neutral for inclusion
Jason: Neutral
Sylvain: Slightly in favor (don't see a better place for it).

Line of divide for inclusion?
------------------
 - browser standard control -> yes
 - "very common UX control" ?

Alexboiboi:
Hi there, 

thanks @martinRenou, I really like the widget and I am looking forward to see it in the main package!
@jasongrout, any chance this is going to be merged soon?

SylvainCorlay:
@martinRenou, could you rebase this one? We decided to go ahead with this PR.

packages/controls/src/widget_tagsinput.ts
```diff
@@ -0,0 +1,844 @@
...
+  /**
+   * Create the Color tag
+   */
+  createTag(value: string, index: number, selected: boolean): HTMLDivElement {
+    const tag = document.createElement('div');
+    const color = value;
+    const darkerColor: string = d3Color
+      .color(value)!
+      .darker()
+      .toString();
```
SylvainCorlay, line 679: Ping @martinRenou I had to add `!` because `d3Color.color(value)` may be null.

Thoughts?

martinRenou: Hmm I don't understand that code, does this actually compile?

SylvainCorlay: Yes, it returns `null` if `d3Color.color(value)` is null.

martinRenou: What is `d3Color.color(value)!`? Shouldn't this raise a `SyntaxError: Unexpected token`?

SylvainCorlay: `d3Color.color(value)!.darker()`

martinRenou: Sorry, this does not make sense to me

martinRenou: Just for the record, this is a TypeScript operator: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

PR #2616, opened 2019-11-16
Add comments specifying valid values for `icon` for `Button` and `ToggleButton`

rramphal:
Currently (as far as I could find), both `Button` and `ToggleButton` seem to take an `icon` value, but there is no documentation as to what the possible values are other than the example usage of `check`. This PR adds two small comments to help the user figure out what the valid values are without digging in the source code.


pbugnion:
This is great!

I think we should move this out of a comment into a new cell below, and add extra information:
- we should state that it's FontAwesome 4
- we could include a link to [this](https://fontawesome.com/v4.7.0/icons/).

SylvainCorlay:
Thanks @rramphal

Issue #2492, opened 2019-07-10
Password field (ipywidgets.Password) is not properly styled


earshinov:
**Describe the bug**

Password field (`ipywidgets.Password`) is not properly styled.

Here it is in comparison with a simple text field (`ipywidgets.Text`):

![text vs password field](https://user-images.githubusercontent.com/52021/60767635-48165200-a0c3-11e9-980c-9a11892f82f5.png)

A bunch of styles is defined for `.widget-text input[type="text"], .widget-text input[type="number"], .widget-textarea textarea`, but as you can see, `input[type="password"]` is not in the list.

**To reproduce**

In a Jupyter notebook:

```python
import ipywidgets as widgets

display(widgets.VBox([
  widgets.Text(description='Username')
  widgets.Password(description='Password')
]))
```

**Expected behavior**

Password fields should have the same size, paddings, border color etc. as ordinary text fields.

**Environment:**

- ipywidgets  7.5.0

Original issue filed to JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/6777

SylvainCorlay:
Thanks @earshinov !

This is certainly a bug. I have marked it as a "first good issue" for anyone interested in fixing it.

jasongrout:
Fixed by #2619.

PR #2619, opened 2019-11-20
Fixed style of password field #2492

ThierrySpetebroot:
This pull request addresses the issue referenced in #2492 . 

It also sets the same way as for the `input[type="text"]`:
 - the opacity (when `disabled=True`)
 - the padding

_Example:_

**Before**:
ipywidgets_fix_pwd_before

**After**:
ipywidgets_fix_pwd_after

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -416,16 +416,16 @@
     width: var(--jp-widgets-inline-width);
 }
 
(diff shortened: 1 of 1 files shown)


jasongrout:
Thank you! Would you mind put before/after screenshots so the record is clear what changed?

ThierrySpetebroot:
Sure! (see edit above)

jasongrout:
I tested this out. It looks and works great! Thanks!

PR #2626, opened 2019-11-26
New HTML attr.: Tabindex

zerline:
This PR adds a `tabindex` HTML attribute to a bunch of ipywidgets widgets: Text, Textarea, Button, ToggleButton, Combobox, Checkbox, Dropdown and Select.

This new attribute is applied on the `DOMWidget` class level, allowing generalization to all widgets.

Explicitely setting tabindex is then possible, although that's not a very good practice in general. Yet, pecifying if a control  widget should `tabbable` or not is a most useful use case for web interfaces.

### DIFF
--- a/ipywidgets/widgets/domwidget.py
+++ b/ipywidgets/widgets/domwidget.py
@@ -63,13 +63,13 @@ class DOMWidget(Widget):
         """
         self.tabindex = i
 
-    def set_tabbable(self):
+    def make_tabbable(self):
         """Make this DOM element reachable
         to keyboard tabulation navigation.
         """
         self.set_tabindex(0)
 
-    def set_untabbable(self):
+    def make_untabbable(self):
         """Make this DOM element unreachable
         to keyboard tabulation navigation.
         """
--- a/ipywidgets/widgets/domwidget.py
+++ b/ipywidgets/widgets/domwidget.py
@@ -62,3 +62,15 @@ class DOMWidget(Widget):
             Order in the keyboard tabulation.
         """
         self.tabindex = i
+
+    def set_tabbable(self):
+        """Make this DOM element reachable
+        to keyboard tabulation navigation.
+        """
+        self.set_tabindex(0)
+
+    def set_untabbable(self):
+        """Make this DOM element unreachable
+        to keyboard tabulation navigation.
+        """
+        self.set_tabindex(-1)
--- a/ipywidgets/widgets/domwidget.py
+++ b/ipywidgets/widgets/domwidget.py
@@ -3,7 +3,7 @@
 
 """Contains the DOMWidget class"""
 
-from traitlets import Unicode
+from traitlets import Int, Unicode
 from .widget import Widget, widget_serialization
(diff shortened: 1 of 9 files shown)


packages/controls/src/widget_bool.ts
```diff
@@ -101,6 +103,13 @@ class CheckboxView extends DescriptionView {
         this.label.style.display = indent ? '' : 'none';
     }
 
+    updateTabindex() {
+        if (!this.checkbox) return; // we might be constructing the parent
+        let tabindex = this.model.get('tabindex');
+        if (tabindex) this.checkbox.setAttribute('tabIndex', tabindex);
```
vidartf: Won't this `if (tabindex)` be false if `tabindex` is 0? Is this intended?

ipywidgets/widgets/domwidget.py
```diff
@@ -48,3 +49,28 @@ def _repr_keys(self):
...
+        (see https://developer.paciellogroup.com/blog/2014/08/using-the-tabindex-attribute/)
+
+        Parameters
+        ----------
+        i: integer
+            Order in the keyboard tabulation.
+        """
+        self.tabindex = i
+
+    def set_tabbable(self):
```
vidartf: When I see a method `set_x`, I'm assuming it is going to set the value of `x`. Maybe call is `make_tabbable`? Maybe we should instead add a `tabbable = Bool()` trait, thereby avoiding the `tabindex > 0` case all together?

packages/controls/src/widget_bool.ts
zerline: @vidartf That's right. Comparing to `null` will be better (in case the original value is a non zero integer).

ipywidgets/widgets/domwidget.py
zerline: ok for `make_` which is more consistent.

About avoiding the `tabindex > 0` : I'll have to ask an accessibility specialist.

zerline:
@vidartf I've made a new branch https://github.com/zerline/ipywidgets/tree/zerline/tabbable.
With `_tabindex` made private and a `tabbable` boolean as you suggested. 
But this would need some synchronization between the two. With a danger of infinite loop? I tend to leave this undecided for the moment.

vidartf:
> With `_tabindex` made private and a `tabbable` boolean as you suggested.

I would maybe not use `_tabindex` either, just sync `tabbable`. Or are you saying you want to keep `_tabindex` in case the user *definitely* wants to set the index themselves? I think exposing even `_tabindex` might be incorrect for multiple views of the same widget.

zerline:
@vidartf yes that's a good reason. I'll do that.

ipywidgets/widgets/domwidget.py
zerline: see https://github.com/jupyter-widgets/ipywidgets/pull/2640

zerline:
see https://github.com/jupyter-widgets/ipywidgets/pull/2640

zerline:
Closing this one, in favor of #2640

PR #2627, opened 2019-11-26
New HTML attr: Tooltip

zerline:
This PR implements tooltips on more widgets. Along with small code factorizations, it tries to unify `description_tooltip` -- already implemented for DescriptionWidget-based widgets -- with `tooltip` -- already implemented for buttons and checkboxes.


zerline:
Cf #2680

PR #2630, opened 2019-11-27
New HTML attr: Focus

zerline:
Programmatically focus/blur on widgets.
NB: Case you have more that one instance of the widget view, it will focus on the last one.


zerline:
See #_

PR #2640, opened 2019-12-13
Tabbable or not tabbable

zerline:
Use HTML attribute `tabIndex` just to make a widget `tabbable` or _not.
This pull request does a little refactoring as well, introducing a class SelectionView between DescriptionView and the controls.


jasongrout:
@zerline - I was looking through this again, in a review of what we have in 8.0 in preparation for moving forward again. Since the controls implemented here all by default are tabbable (I think?), such as listboxes, textboxes, etc., it seems that what this adds is the ability to set the tabindex to -1, i.e., to take the control out of tab ordering. Is that the main usecase you have for it?

I'm trying to think of how to explain what this new tabbable feature does to 8.0 custom widget authors and users, and more importantly, why it is there and what a typical usecase for it is. What are you using it for?

vidartf:
> Since the controls implemented here all by default are tabbable (I think?)

@jasongrout Initially the PR was only for "a bunch of ipywidgets widgets: Text, Textarea, Button, ToggleButton, Combobox, Checkbox, Dropdown and Select" (https://github.com/jupyter-widgets/ipywidgets/pull/2626). I don't think all of them are tabbable by default? I think I was the one that suggested this be exposed on the DOMWidget class instead of selectively only on those that did not have it by default.

zerline:
@jasongrout @vidartf sorry I was offline.
I want to use that feature in  ```sage-explorer```, for example to make tabulation go through some clickable ```HTML``` widgets, which are not ```tabbable``` by default. Sometimes I also want to _disable_ tabulation on ```Text``` widgets, especially when they are part of a sometimes large table of widgets that is itself part of a larger interface and the (possibly blind) user might get lost  in that table. Hope I'm clear enough ..

PR #2651, opened 2019-12-27
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.


jasongrout:
Thanks. Let's provisionally try running tests on arm. We haven't talked about supporting it, but if it is not a burden, might as well see if running tests there help catch issues.

PR #2656, opened 2019-12-31
Initial github action configuration

jasongrout:
This PR enables GitHub actions for testing for everything except the examples/web1 js test.

Right now these are run alongside the Travis tests. It seems that these tests run faster than travis, and are in general a little simpler to manage, I think.

What do others think? Should we switch to GitHub actions instead of Travis?

One nice thing is that we can run tests on windows and mac (though that is not enabled here)


jasongrout:
Merging since this is only adding additional checks.

PR #2657, opened 2020-01-03
Move to ESLint and Prettier

jasongrout:
This is a WIP PR to move to eslint and prettier for code linting.


jtpio:
Another thing that would be great to do would be to unify the naming of variables, function parameters and so on, to have a consistent style.

At the moment, the TypeScript code uses a mix of camel case and snake case.

This could also be done in a separate PR (if so we can open a new issue). And can be a quite big breaking change if the API (`save_changes` as an example) is also modified.

jasongrout:
The one thing left is a failing test in web3, which exposes that we are not consistent at all in how we define and use our widget manager display_view function. Let's talk about this tomorrow.

vidartf:
Current failing lints will be fixed in master to avoid further conflicts :)

Issue #2301, opened 2019-01-21
Binder not working


WesleyTheGeolien:
Hello,

I have just tried out the binder associated with this repository and unfortunately, the notebooks cannot be rerun. 

From my understanding of the Binder project, it looks in the repository for a `requirements.txt` which is not present leading to the python environment inside the docker container being non-usable. Is there a way to override this?

As a fix maybe adding the requirements needed would be a solution but I also realise that this repository is a library for the ipywidgets and that the requirements needed to run the notebooks are not necessarily the same as the requirements needed to for the package. If the former is chosen I would be happy to look through the notebooks and write the `requirements.txt`

mgeier:
I've created #2659, but sadly, it doesn't work yet. If you have an idea how to fix it, please chime in!

pbugnion:
Hi both! Thanks for raising this.

I can't reproduce the issue you mention. The binder link works fine for me, at least in the classic notebook:

<img width="1226" alt="Screenshot 2020-01-07 at 13 21 16" src="https://user-images.githubusercontent.com/1392879/71898165-8662fb80-3150-11ea-8fb3-63a380d6c10b.png">

The binder URL ends with query string `filepath=/docs/source/examples`, which does have a `requirements.txt`.

JupyterLab doesn't work because it is missing the JupyterLab extension.

I'm going to close this, but feel free to re-open if you think I'm wrong.

WesleyTheGeolien:
Thanks @pbugnion, the Binder now seems to be working however most of the default tutorials use libraries that are nto installed (numpy, matplotlib, leaflet) would it be forseeable to install these using an option in the setup.py say requirements-demo.txt

I'd be more than happy to do the work.

mgeier:
> I'd be more than happy to do the work.

I've already started the work in #2659 and #2701.
Both contain the necessary dependencies.
But both have problems installing the correct version of the widgets for JupyterLab.

If you have any ideas how to proceed there, that would be great!

mgeier:
FYI: now #2701 works.

PR #2659, opened 2020-01-05
Add settings for Binder

mgeier:
Sadly, this doesn't work yet.

There is an error when running `./dev-install.sh` on Binder:

```
ValueError:                                                                                                           
"@jupyter-widgets/jupyterlab-manager@1.1.0" is not compatible with the current JupyterLab                             
Conflicting Dependencies:                                                                                             
JupyterLab                        Extension      Package                                                              
>=1.2.1 <1.3.0                    >=2.0.0-beta.1 <3.0.0@jupyterlab/application                                        
>=1.2.2 <1.3.0                    >=2.0.0-beta.1 <3.0.0@jupyterlab/notebook                                           
>=1.2.1 <1.3.0                    >=2.0.0-beta.1 <3.0.0@jupyterlab/rendermime                                         
>=1.5.0 <1.6.0                    >=2.0.0-beta.1 <3.0.0@jupyterlab/rendermime-interfaces                              
>=4.2.0 <4.3.0                    >=5.0.0-beta.1 <6.0.0@jupyterlab/services                                           
```

Apparently, the JupyterLab and `jupyter-widgets/jupyterlab-manager` versions have to match (see https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/jupyterlab-manager/README.md), but I don't know how to handle this in a development installation ...

Any ideas?

See #2301.

### DIFF
--- a/.binder/postBuild
+++ b/.binder/postBuild
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+./dev-install.sh
--- a/.binder/README
+++ b/.binder/README
@@ -1,0 +1,10 @@
+This directory holds configuration files for https://mybinder.org/.
+
+The interactive notebooks can be accessed with this link:
+https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/master?filepath=docs/source/examples
+
+To check out a different version, just replace "master" with the desired
+branch/tag name or commit hash.
+
+To use JupyterLab, use:
+https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/master?urlpath=lab/tree/docs/source/examples
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -1,0 +1,15 @@
+name: ipywidgets_binder
+channels:
+  - conda-forge
+dependencies:
+  - bqplot
+  - ipyleaflet
+  - jupyterlab
+  - matplotlib
+  - networkx
+  - numpy
+  - pandas
+  - scikit-image
+  - scikit-learn
+  - sympy
+  - yarn
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - bqplot
   - ipyleaflet
-  - jupyterlab
+  - jupyterlab=1
   - matplotlib
   - networkx
   - numpy


pbugnion:
Hmm... As far as I can tell, the classic notebook binder link works -- see my comment on issue #2301.

The JupyterLab instance that is created doesn't support widgets, though. I've opened a new issue to address this [here](https://github.com/jupyter-widgets/ipywidgets/issues/2676). I think your PR could be adapted fairly easily to support this.

.binder/postBuild
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
+
+./dev-install.sh
```
jtpio: Since the Binder is user facing, how about using a released version of `ipywidgets` instead?

This would mean adding `ipywidgets=7.5` to `environment.yml`.

jtpio:
For JupyterLab, specifying `jupyterlab=1` in `environment.yml` should do the trick.

.binder/postBuild
mgeier: You can create a Binder image for any given commit in the repo. I would expect each of those to have the appropriate version of `ipywidgets` and `jupyter-widgets/jupyterlab-manager` installed.

If all goes well (and I'm currently not sure that it will), I would like to do the same thing for the docs on RTD, which would mean that the `latest` docs are actually using the `master` version of the software.

And then, we could create Binder links on the RTD pages which link from a specific version of the docs to the exact same version running on Binder.

mgeier:
Thanks @jtpio!

Now the above-mentioned error doesn't appear anymore, but the Binder build is hanging at some point.

This is the link I tried:
https://mybinder.org/v2/gh/mgeier/ipywidgets/binder-environment

I've kept the page open for more than an hour now and it's still not finished.

Any ideas what the problem could be?

mgeier:
UPDATE: I've just started the Binder build again (and again), and after a few minutes it now gives me the original error.

.binder/postBuild
jtpio: OK, for this use case then it makes sense to do a dev install.

pbugnion:
Thanks for working on this!

I agree with @jtpio . I don't *think* we should do a `dev-install` for the public facing Binder link for ipywidgets.

I think the right way to make progress on improving Binder support is to include the JupyterLab manager and aim to close #2676.

mgeier:
> I agree with @jtpio . I don't _think_ we should do a `dev-install` for the public facing Binder link for ipywidgets.

But it looks like @jtpio changed his mind, see https://github.com/jupyter-widgets/ipywidgets/pull/2659#discussion_r364212489.

> I think the right way to make progress on improving Binder support is to include the JupyterLab manager and aim to close #2676.

Sure. The question is only: which version of the JupyterLab manager?

I'm arguing that it should be the version corresponding to whatever commit is used in Binder.

What if you want to try out a new feature (probably from some feature branch, probably even from a fork) in Binder?
This won't work if the `jupyter-widgets/jupyterlab-manager` used by Binder is outdated (w.r.t the commit containing the new feature).

jtpio:
Sorry @mgeier for the confusion. My comment was directed to the particular use case of being able to pin a Binder to a specific commit.

jtpio:
By the way, is this PR still needed or replaced by https://github.com/jupyter-widgets/ipywidgets/pull/2701?

mgeier:
Well I would prefer #2701, but if you prefer using `conda`, I could try to get this PR working as well.

But as soon as #2701 is merged, this PR becomes obsolete.

PR #2661, opened 2020-01-05
DOC: Add thread.join() in Output example

mgeier:
This waits for the thread to finish, even when executed with `nbconvert --execute`.

There still seems to be a problem when creating the docs, though. In that case the output is shown twice for some reason ...

### DIFF
--- a/docs/source/examples/Output Widget.ipynb
+++ b/docs/source/examples/Output Widget.ipynb
@@ -383,7 +383,8 @@
     "thread = threading.Thread(\n",
     "    target=thread_func,\n",
     "    args=(\"some text\", out))\n",
-    "thread.start()"
+    "thread.start()\n",
+    "thread.join()"
    ]
   }
  ],
--- a/docs/source/examples/Output Widget.ipynb
+++ b/docs/source/examples/Output Widget.ipynb
@@ -383,7 +383,15 @@
     "thread = threading.Thread(\n",
     "    target=thread_func,\n",
     "    args=(\"some text\", out))\n",
-    "thread.start()\n",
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
> This waits for the thread to finish, even when executed with nbconvert --execute.

I think that adding `join` kind of defeats the purpose of the example because it makes the cell blocking, and equivalent to simply calling `thread_func("some text", out)`...

However, I see how this can be an issue in the context of `nbexecute` because outputs are simply not  properly populated at all.

mgeier:
> I think that adding `join` kind of defeats the purpose of the example because it makes the cell blocking, and equivalent to simply calling `thread_func("some text", out)`...

No it's not, because then `thread_func()` would not run in a separate thread, which is the whole point of this example.

But I agree that it looks strange to have `thread.start()` directly followed by `thread.join()`.

What about putting `thread.join()` into a separate cell (7a28b43400696d252dda1e0721a16ab457629f32)?

If you really want to stress that things are happening in a different thread, while it's possible to continue executing more cells, you should do so in the example.
If you use time-stamped messages you should be able to demonstrate that when the first cell has outputs with later times than the second cell.

But this is really not what this PR is about.

SylvainCorlay:
Thanks, I prefer having it in the next cell!

> In that case the output is shown twice for some reason ...

Is this due to the double inclusion of the widgets embedder or an independent bug?

mgeier:
Yeah, that's because of the double inclusion. #2660 has fixed this.

SylvainCorlay:
Thanks @mgeier this is perfect. Merging!

PR #2662, opened 2020-01-06
WidgetManagerBase: Improve create_view return type

martinRenou:
Given a video widget model `VideoModel` which has an associated view `VideoView`.  The following code does not compile:

```TypeScript
createVideo(videoModel: VideoModel) {
    this.create_child_view(videoModel).then((videoView: VideoView) => {
        // Do something
    });
}
```

With the following error:
```
Argument of type '(videoView: VideoView) => VideoView' is not assignable to parameter of type '(value: DOMWidgetView) => VideoView | PromiseLike<VideoView>'.
  Types of parameters 'videoView' and 'value' are incompatible.
    Type 'DOMWidgetView' is not assignable to type 'VideoView'.
```

This PR allows doing:
```TypeScript
createVideo(videoModel: VideoModel) {
    this.create_child_view<VideoView>(videoModel).then((videoView: VideoView) => {
        // Do something
    });
}
```

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -691,9 +691,11 @@ class WidgetView extends NativeView<WidgetModel> {
     /**
      * Create and promise that resolves to a child view of a given model
      */
-    create_child_view(child_model: WidgetModel, options = {}): Promise<DOMWidgetView> {
+    create_child_view<VT extends DOMWidgetView = DOMWidgetView>(child_model: DOMWidgetModel, options?: any): Promise<VT>;
+    create_child_view<VT extends WidgetView = WidgetView>(child_model: WidgetModel, options?: any): Promise<VT>;
+    create_child_view<VT extends WidgetView = WidgetView>(child_model: WidgetModel, options = {}): Promise<VT> {
         options = { parent: this, ...options};
-        return this.model.widget_manager.create_view(child_model, options)
+        return this.model.widget_manager.create_view<VT>(child_model, options)
             .catch(utils.reject('Could not create child view', true));
     }
 
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -89,8 +89,9 @@ export abstract class ManagerBase<T> implements IWidgetManager {
      * Make sure the view creation is not out of order with
      * any state updates.
      */
-    create_view(model: DOMWidgetModel, options: any): Promise<DOMWidgetView>;
-    create_view(model: WidgetModel, options = {}): Promise<WidgetView> {
+    create_view<VT extends DOMWidgetView = DOMWidgetView>(model: DOMWidgetModel, options?: any): Promise<VT>;
+    create_view<VT extends WidgetView = WidgetView>(model: WidgetModel, options?: any): Promise<VT>;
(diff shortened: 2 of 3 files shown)


pbugnion:
Is this still relevant given the discussion on removing `createView` this morning? Is there still value in merging this in the interim?

jasongrout:
We're removing display_model/display_view in favor of keeping and using create_view.

martinRenou:
@jasongrout would it be fine to merge this PR?

packages/base/src/widget.ts
```diff
@@ -689,9 +689,9 @@ class WidgetView extends NativeView<WidgetModel> {
     /**
      * Create and promise that resolves to a child view of a given model
      */
-    create_child_view(child_model: WidgetModel, options = {}): Promise<DOMWidgetView> {
+    create_child_view<VT extends DOMWidgetView = DOMWidgetView>(child_model: WidgetModel, options = {}): Promise<VT> {
```
jasongrout: I think this should probably have the same overrides as the create_view? Not all children views will be dom widget views.

packages/base/src/widget.ts
```diff
@@ -689,9 +689,9 @@ class WidgetView extends NativeView<WidgetModel> {
...
         options = { parent: this, ...options};
-        return this.model.widget_manager.create_view(child_model, options)
+        return this.model.widget_manager.create_view<VT>(child_model, options)
```
jasongrout: I think once you have the overrides, you won't need to specify the type parameter here, as it should select the right override.

martinRenou: Yes! Thanks you

martinRenou: Actually it does not seem to compile without it:

```
src/widget.ts:696:9 - error TS2322: Type 'Promise<DOMWidgetView>' is not assignable to type 'Promise<VT>'.
  Type 'DOMWidgetView' is not assignable to type 'VT'.
    'DOMWidgetView' is assignable to the constraint of type 'VT', but 'VT' could be instantiated with a different subtype of constraint 'WidgetView'.

696         return this.model.widget_manager.create_view(child_model, options)
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
697             .catch(utils.reject('Could not create child view', true));
```

I feel like this compilation error is not legit though, as `DOMWidgetView` is a subclass of `WidgetView`, what do you think?

packages/base/src/widget.ts
```diff
@@ -691,9 +691,10 @@ class WidgetView extends NativeView<WidgetModel> {
     /**
      * Create and promise that resolves to a child view of a given model
      */
-    create_child_view(child_model: WidgetModel, options = {}): Promise<DOMWidgetView> {
+    create_child_view<VT extends DOMWidgetView = DOMWidgetView>(child_model: WidgetModel, options?: any): Promise<VT>;
```
jasongrout:
```suggestion
    create_child_view<VT extends DOMWidgetView = DOMWidgetView>(child_model: DOMWidgetModel, options?: any): Promise<VT>;
```

martinRenou: Good catch, thanks a lot

PR #2666, opened 2020-01-06
Fix Upload widget counter value

### LINKED ISSUE #2480: FileUpload counter with multiple=False
It looks a bit odd that the `FileUpload` counter is increasing every time a file is uploaded, even when `multiple=False` is set:

fileupload-counter

When `multiple=False` is set shouldn't the counter equal to 1?

Can be tested with:

```python
from ipywidgets import FileUpload, Output

out = Output()

@out.capture()
def show_content(change):
    print(change['new'].keys())

w = FileUpload(multiple=False)
w.observe(show_content, 'value')
with out:
    display(w)
out
```

jtpio:
Fixes #2480.

It looks like the `_counter` value was used to detect a change of the value.

Instead we could use `data` to achieve the same result, since its value should change when new files are being uploaded.

fileupload-counter-fixed

### Breaking changes

- Rename `data` to `_data`
- Rename `metadata` to `_metadata`
- Make `_counter`, `_metadata` and `_data` read-only
- `value` is a list of dicts. Example:

```python
[{ 'metadata': {}, 'content': <memoryview>}]
```

### DIFF
--- a/ipywidgets/widgets/widget_upload.py
+++ b/ipywidgets/widgets/widget_upload.py
@@ -52,9 +52,6 @@ class FileUpload(DescriptionWidget, ValueWidget, CoreWidget):
 
     @observe('_counter')
     def on_incr_counter(self, change):
-        """
-        counter increment triggers the update of trait value
-        """
         res = {}
         msg = 'Error: length of metadata and data must be equal'
         assert len(self.metadata) == len(self.data), msg
--- a/ipywidgets/widgets/widget_upload.py
+++ b/ipywidgets/widgets/widget_upload.py
@@ -34,6 +34,10 @@ class FileUpload(DescriptionWidget, ValueWidget, CoreWidget):
     _model_name = Unicode('FileUploadModel').tag(sync=True)
     _view_name = Unicode('FileUploadView').tag(sync=True)
     _counter = Int(read_only=True).tag(sync=True)
+    _metadata = List(Dict(), read_only=True, help='List of file metadata').tag(sync=True)
+    _data = List(Bytes(), read_only=True, help='List of file content (bytes)').tag(
+        sync=True, from_json=content_from_json
+    )
 
     accept = Unicode(help='File types to accept, empty string for all').tag(sync=True)
     multiple = Bool(help='If True, allow for multiple files upload').tag(sync=True)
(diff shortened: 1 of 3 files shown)


pbugnion:
Thanks for this.

I agree that it's a little strange that we're using `_counter` both to decide whether there's been an upload event and to update the description.

I'm not sure I agree that switching to observing the `data` trait solves our problems though. For instance, what happens if someone uploads:
1. `A.txt` with content `hello`
2. `B.txt` with content `hello`

I don't *think* `data` changes in this case? (I haven't checked this, so I may well be wrong)

An alternative could be to keep `_counter`, but not use it in the button label? We could have a new model property (that doesn't necessarily make its way all the way to the Python layer) called `fileCount` that measures the current number of files?

jtpio:
> I don't _think_ `data` changes in this case? (I haven't checked this, so I may well be wrong)

Thanks for raising that point. We should be able to check that behavior.

It looks it's possible for `data` to not change, but for `metadata` to change. In that case `value` would indeed have to be recalculated.

> An alternative could be to keep `_counter`, but not use it in the button label? We could have a new model property (that doesn't necessarily make its way all the way to the Python layer) called `fileCount` that measures the current number of files?

Right, with the current change `_counter` is only used in the button label and doesn't need to live in the kernel. We would still need a way to detect a state change to be able to set `value` accordingly.

Another way could be to compute `value` on the frontend, and push it back to the kernel? So the `_data_changed` method would not be needed anymore.

pbugnion:
> It looks it's possible for data to not change, but for metadata to change. In that case value would indeed have to be recalculated.

Thanks for checking.

> Right, with the current change _counter is only used in the button label and doesn't need to live in the kernel.

Agreed.

> Another way could be to compute value on the frontend, and push it back to the kernel? So the _data_changed method would not be needed anymore.

To me, that sounds like the right call, except `data` and `metadata` are annoyingly exposed as top level traits in the Python-side. Somewhat weirdly, `data` and `metadata` are not read-only. I'm not sure what would happen if someone sets them server-side.

We could break backwards compatibility and just keep `value` as a read-only trait?

jtpio:
> We could break backwards compatibility and just keep `value` as a read-only trait?

It's possible that `data` and `metadata` were kept that way to leverage the binary serialization for `data`, so that the `value` dict can be computed on the Python side.

pbugnion:
> It's possible that data and metadata were kept that way to leverage the binary serialization for data

Ah yeah, that makes sense.

jtpio:
Pushed some changes to:

- make `data` and `metadata` private and read only
- use a client side `_file_count` for the Upload button view
- keep `_counter` as a way to trigger the `value` update

fileupload-counter2

pbugnion:
I think this looks good. I think it'd be useful to discuss with people explicitly removing `data` and `metadata` from the public API.

jtpio:
Thanks @pbugnion for adding the comment in https://github.com/jupyter-widgets/ipywidgets/pull/2666/commits/656e2ed16aadd30b1d09a2ad25b54d91ee1ccb1e.

For reference and to add a bit more context: without this serializer, we indeed hit the default serializer, which performs a deep copy of the value via a `JSON.parse` / `JSON.stringify`: https://github.com/jupyter-widgets/ipywidgets/blob/44fb6f81ed798592e9286ae65087e8f3718f1792/packages/base/src/widget.ts#L443

This would result in the `ArrayBuffer` (value of `content`) being serialized to an empty object (`{}` ).

PR #2668, opened 2020-01-06
Delete travis tests that are duplicated by github actions.

jasongrout:
We leave the one test that github actions does not test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,44 +4,17 @@ dist: xenial
 os: linux
 arch:
   - amd64
-  - arm64
 python:
-    - 3.5
-    - 3.6
-    - 3.7
     - 3.8
 addons:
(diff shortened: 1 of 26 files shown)


jasongrout:
The docs error is occuring because:

1. nbformat package version 5.0 was released, which makes much more precise validation against specific notebook format versions
2. The sphinx build process is creating notebooks with format 4.1
3. However, ipywidgets *requires* notebooks to be format 4.2 or later, since it uses the mimebundle type `application/vnd.jupyter.widget-view+json` as a key for json data. This change came in notebook format 4.2: https://nbformat.readthedocs.io/en/latest/changelog.html#id5 (see also https://github.com/jupyter/nbformat/pull/139#issuecomment-571415692)

PR #2670, opened 2020-01-06
Drag and Drop widgets

wolfv:
Rebased version of @btel PR for the sprint.


wolfv:
cc @jtpio as you also had a quick look today.

jtpio:
Thanks @wolfv for rebasing it.

We'll need to first fix a couple of things since [it doesn't seem to be building](https://github.com/jupyter-widgets/ipywidgets/pull/2670/checks?check_run_id=376127221) at the moment.

jtpio:
xref https://github.com/jupyter-widgets/ipywidgets/pull/2363

btel:
Thanks for rebasing @wolfv. Let me know if there is anything I can help with!

wolfv:
yep, there is some typescript attributes that i haven't properly fixed yet :) 

@btel good to see you here! we're having the ipywidgets sprint in Paris right now! Maybe you want to drop by if you have time? I'd like to help get this PR over the finish line while all the ipywidgets people are here and we can reach consensus relatively quickly.

btel:
I can't be there in person, but I will be happy to join a videocall.

wolfv:
I hope I got the rebase right this time :) Everything except the last example seems to still work fine (the last example might be broken due to bqplot, actually)

wolfv:
@btel I am currently thinking about the final API and I have the following simplifications in mind:

I would like to remove the DraggableLabel for this PR, instead I would like to keep only the DraggableBox and DropBox. Both should be changed to accept at most one item.

One remaining issue is how to set the default "drag data".

My current idea is to set the drag data to the `value` attribute of the child of the DraggableBox (if it has this property). This would make it possible to define a function like this to get the current functionality:

```
def DraggableLabel(value, draggable=True):
    outer = DraggableBox([Label(value)])
    return outer
```

btel:
It sounds good to me. DraggableLabel was just a proof-of-concept, that
should be easily implemented with DraggableBox. It would be nice tough, to
include an example with a draggable label in the docs.

wolfv:
yes, for sure. What I am trying to do right now is to rewrite your excellent notebook with just DraggableBox and DropBox.

PR #2673, opened 2020-01-07
Remove last Travis test

jasongrout:
This converts the last Travis test to GitHub actions. It fails currently, probably because we have some problem with the firefox headless that we're using to run the test in actions, or something.

It would be great if someone could help push this forward!


SylvainCorlay:
@jasongrout I think that you are missing a `yarn run build:examples` call in `.github/workflows/testjs.yml`.

PR #2679, opened 2020-01-07
Drop support for Mapping types as Selection options

### LINKED ISSUE #1958: Drop support for mutable types as dropdown options

jtpio:
Fixes #1958.
Continuation of #2130. 

### Breaking change

- This removes support for `Mapping` types when constructing a new `Dropdown`.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -106,12 +106,6 @@ def _make_options(x):
     * an iterable of (label, value) pairs
     * an iterable of values, and labels will be generated
     """
-    # Check if x is a mapping of labels to values
-    if isinstance(x, Mapping):
-        import warnings
-        warnings.warn("Support for mapping types has been deprecated and will be dropped in a future release.", DeprecationWarning)
-        return tuple((str(k), v) for k, v in x.items())
-
     # only iterate once through the options.
     xlist = tuple(x)
 
@@ -132,10 +126,10 @@ def findvalue(array, value, compare = lambda x, y: x == y):
 class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
     """Base class for Selection widgets
 
-    ``options`` can be specified as a list of values, list of (label, value)
-    tuples, or a dict of {label: value}. The labels are the strings that will be
-    displayed in the UI, representing the actual Python choices, and should be
-    unique. If labels are not specified, they are generated from the values.
+    ``options`` can be specified as a list of values or list of (label, value)
+    tuples. The labels are the strings that will be displayed in the UI,
+    representing the actual Python choices, and should be unique.
+    If labels are not specified, they are generated from the values.
 
     When programmatically setting the value, a reverse lookup is performed
     among the options to check that the value is valid. The reverse lookup uses
@@ -149,7 +143,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
     index = Int(None, help="Selected index", allow_none=True).tag(sync=True)
 
     options = Any((),
-    help="""Iterable of values, (label, value) pairs, or a mapping of {label: value} pairs that the user can select.
+    help="""Iterable of values or (label, value) pairs that the user can select.
 
     The labels are the strings that will be displayed in the UI, representing the
     actual Python choices, and should be unique.
@@ -183,9 +177,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
 
     @validate('options')
     def _validate_options(self, proposal):
-        # if an iterator is provided, exhaust it
-        if isinstance(proposal.value, Iterable) and not isinstance(proposal.value, Mapping):
-            proposal.value = tuple(proposal.value)
+        proposal.value = tuple(proposal.value)
         # throws an error if there is a problem converting to full form
         self._options_full = _make_options(proposal.value)
         return proposal.value
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -54,12 +54,11 @@ _doc_snippets['multiple_selection_params'] = """
         The options for the dropdown. This can either be a list of values, e.g.
         ``['Galileo', 'Brahe', 'Hubble']`` or ``[0, 1, 2]``, a list of
         (label, value) pairs, e.g.
-        ``[('Galileo', 0), ('Brahe', 1), ('Hubble', 2)]``,
-        or a dictionary mapping the labels to the values, e.g. ``{'Galileo': 0,
-        'Brahe': 1, 'Hubble': 2}``. The labels are the strings that will be
-        displayed in the UI, representing the actual Python choices, and should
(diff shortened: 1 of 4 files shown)


pbugnion:
Thanks!

Currently, the `options` traitlet has type `Any()`. I wonder if we can change it to a `traitlets.List` or some such?

I think we can probably cut more code from `_Selection` in general.

jtpio:
That's a good point, let's specify the type even more.

jtpio:
> Currently, the `options` traitlet has type `Any()`. I wonder if we can change it to a `traitlets.List` or some such?

`traitlets.List` might be slightly too specific, since `options` should also support other iterables such as `Tuple`.

pbugnion:
There is no reasonable traitlet supertype that includes general Python iterables, as far as I can see. 

For instance, the following currently works:

```python
import numpy as np
options = np.array(['hello', 'world'])
widgets.Dropdown(options=options)
```

As far as I can tell, no refinement of `Any` would let us keep this behaviour, so I vote in favour of keeping `Any`.

ipywidgets/widgets/widget_selection.py
```diff
@@ -133,9 +127,9 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
     """Base class for Selection widgets
 
     ``options`` can be specified as a list of values, list of (label, value)
-    tuples, or a dict of {label: value}. The labels are the strings that will be
-    displayed in the UI, representing the actual Python choices, and should be
-    unique. If labels are not specified, they are generated from the values.
+    tuples. The labels are the strings that will be displayed in the UI,
+    representing the actual Python choices, and should be unique.
+    If labels are not specified, they are generated from the values.
```
pbugnion: I think this is missing an `or`:

```rst
``options`` can be specified as a list of values or a list of (label, value) tuples.
```

pbugnion:
I think some of the entries in `doc_snippets` also need updating (e.g. multiple_selection_params).

ipywidgets/widgets/widget_selection.py
jtpio: Thanks for catching that!

jtpio:
> I think some of the entries in `doc_snippets` also need updating (e.g. multiple_selection_params).

For the `_MappingSelection`, don't we want to keep support for Mapping / dict? At least for now, and potentially drop it in a separate PR.

pbugnion:
> For the _MultipleSelection, don't we want to keep support for Mapping / dict? At least for now, and potentially drop it in a separate PR.

Ah yes, you're completely right, sorry.

Maybe we should introduce a deprecation ahead of 8?

pbugnion:
Actually, since `_MultipleSelection` also relies on `_make_options`, doesn't this PR break support for dict-like objects in widgets that inherit from `_MultipleSelection`?

pbugnion:
Somewhat unexpectedly, for a dictionary `d = { "a": 5, "b": 8}`, `tuple(d)` returns the dictionary keys:

```
>>> d = {"a": 5, "b": 6}
>>> tuple(d)
#=> ("a", "b")
```

This means that, for someone who's not paid attention to the deprecation warning and still has a dict, their widget won't crash -- it'll just behave really unexpectedly: the dropdown entries will be the same, but the values will just be ignored. I think this is going to be really surprising.

We can either:
1. not care
2. keep the `if isinstance(x, Mapping):` in `_make_options` but raise a `TypeError` if it's x is a `Mapping`.

jasongrout:
> keep the `if isinstance(x, Mapping):` in `_make_options` but raise a `TypeError` if it's x is a `Mapping`.

I'm fine with this, say for one more version.

jtpio:
Thanks @pbugnion for investigating this further.

> Actually, since `_MultipleSelection` also relies on `_make_options`, doesn't this PR break support for dict-like objects in widgets that inherit from `_MultipleSelection`?

It looks like it does yes. We should be able to catch that in the tests though. 

In the long run we'll probably want to drop support for mutable types for `MultipleSelection` widgets too. But it would be better to introduce a `DeprecationWarning` first.

> 2. keep the if isinstance(x, Mapping): in _make_options but raise a TypeError if x is a Mapping.

This could be a good compromise in the meantime.

jasongrout:
Just to be clear here - we are talking about dropping support for [*mapping* types](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping), not mutable types.

jasongrout:
@jtpio - what is the current status of this PR?

jtpio:
It was changed back to a "WIP" to check the points raised by @pbugnion above and see if they can be covered in the tests too:

> Actually, since _MultipleSelection also relies on _make_options, doesn't this PR break support for dict-like objects in widgets that inherit from _MultipleSelection?

jtpio:
We can indeed see that with a `SelectMultiple` widget:

```python
from ipywidgets import SelectMultiple

options = {'a': 'Apples', 'o': 'Oranges', 'p': 'Pears'}

select = SelectMultiple(
    options=options,
    description='Fruits'
)
select
```

Which gives back the list of keys as values: 

image

We might then want to drop support for `Mapping` types as `_MultipleSelection`options too. The `DeprecationWarning` added in #2130 should also be triggered when using a `SelectMultiple` widget in 7.5:

image

jtpio:
> keep the if isinstance(x, Mapping): in _make_options but raise a TypeError if x is a Mapping.

This would at least make the change more visible instead of silently converting the `Mapping` to a `tuple`.

jtpio:
> Currently, the `options` traitlet has type `Any()`. I wonder if we can change it to a `traitlets.List` or some such?

We can also follow this approach and be explicit about the types being accepted. For example with `Union([List(), Tuple(), Set()])`.

jtpio:
The last commit implements the ideas mentioned above:

- raise a `TypeError` for `Mapping` types, since the `DeprecationWarning` was already given on other selection widgets
- this prevents implicit casts to a `tuple`
- `numpy` arrays are still supported

jasongrout:
@jtpio - thanks! @pbugnion - do you have time to review the current changes?

pbugnion:
Great, thank you!

I'll review this in the next couple of days.

pbugnion:
I added a couple of tests to verify the behaviour when setting `.options`, clarified a docstring, and added a note on changes in ipywidgets 8.

RTD is failing because of #2768, but the rest looks good. I'm going to merge this.

Thank you!

jtpio:
Thanks @pbugnion!

PR #2680, opened 2020-01-07
Tooltips everywhere

zerline:
Tooltips everywhere. With unified name (`tooltip`).


packages/controls/src/widget_description.ts
```diff
@@ -69,19 +67,17 @@ class DescriptionView extends DOMWidgetView {
...
         } else {
             this.label.innerHTML = description;
             this.typeset(this.label);
             this.label.style.display = '';
         }
-        this.label.title = description_tooltip;
+    }
+
+    updateTooltip() {
+        this.label.title = this.model.get('tooltip');
```
vidartf: Should this still fall back to using the `description` value if `tooltip` is unset?

ipywidgets/widgets/domwidget.py
```diff
@@ -11,11 +11,20 @@
...
+    tooltip: str
+       tooltip caption
+    layout: InstanceDict(Layout)
+       widget layout
+    """
 
     _model_name = Unicode('DOMWidgetModel').tag(sync=True)
     _dom_classes = TypedTuple(trait=Unicode(), help="CSS classes applied to widget DOM element").tag(sync=True)
     tabbable = Bool(help="Is widget tabbable?", allow_none=True, default_value=None).tag(sync=True)
+    tooltip = Unicode(None, allow_none=True, help="A tooltip caption.").tag(sync=True)
```
vidartf: I think we can leave the default as empty string (the default `default` value), and not specify `allow_none=True`.

vidartf:
The tests fail for `DatePickerView` in a relevant way.

packages/controls/src/widget_description.ts
```diff
@@ -57,10 +56,11 @@ class DescriptionView extends DOMWidgetView {
         this.label.style.display = 'none';
 
         this.listenTo(this.model, 'change:description', this.updateDescription);
-        this.listenTo(this.model, 'change:description_tooltip', this.updateDescription);
         this.listenTo(this.model, 'change:tabbable', this.updateTabindex);
+        this.listenTo(this.model, 'change:tooltip', this.updateTooltip);
```
vidartf: This should be removed I think.

packages/base/src/widget.ts
```diff
@@ -835,6 +836,8 @@ class DOMWidgetView extends WidgetView {
         this.listenTo(this.model, 'comm_live_update', () => {
             this._comm_live_update();
         });
+        this.listenTo(this.model, 'change:tooltip', this.updateTooltip);
```
vidartf: I think this can also be removed

vidartf: My mistake.

PR #2685, opened 2020-01-08
Fix regression on spinning icons

### LINKED ISSUE #2477: Button Icons/FontAwesome regression no longer supports fa-spin
since this was added:
https://github.com/jupyter-widgets/ipywidgets/blob/master/ipywidgets/widgets/widget_button.py#L65-L73

we are no longer able to pass in additional fa classes to specify fa-spin or fa-lg etc.

I really liked my spinning gear icon button before this change...

i.e.
```
from ipywidgets import Button
Button(description='Running...', icon='fa-gear fa-spin fa-lg')
```

it will error on the add of the DOMTokenList due to the spaces in the string
https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_button.ts#L91

martinRenou:
This should fix #2477.

spinning

### DIFF
--- a/ipywidgets/widgets/widget_button.py
+++ b/ipywidgets/widgets/widget_button.py
@@ -39,7 +39,7 @@ class Button(DOMWidget, CoreWidget):
     tooltip: str
        tooltip caption of the toggle button
     icon: str
-       font-awesome icon name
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
LGTM

PR #2698, opened 2020-01-08
Add Font Awesome 5.12 to the JupyterLab Manager

martinRenou:
And update Font Awesome in the html-manager

### DIFF
--- a/packages/html-manager/package.json
+++ b/packages/html-manager/package.json
@@ -35,6 +35,7 @@
     "test:unit:firefox:headless": "npm run test:unit:default -- --browsers=FirefoxHeadless"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-free": "^5.12.0",
     "@jupyter-widgets/base": "^2.0.2",
     "@jupyter-widgets/controls": "^1.5.3",
     "@jupyter-widgets/output": "^2.0.1",
@@ -44,7 +45,6 @@
     "@jupyterlab/rendermime-interfaces": "^2.0.0-beta.2",
     "@lumino/widgets": "^1.6.0",
     "ajv": "^6.10.0",
-    "font-awesome": "^4.7.0",
     "jquery": "^3.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


packages/jupyterlab-manager/src/index.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
+import '@fortawesome/fontawesome-free/css/all.css';
+import '@fortawesome/fontawesome-free/css/v4-shims.css';
+
```
martinRenou: I wasn't sure where to put it, I'm of course open to better suggestions

packages/html-manager/package.json
```diff
@@ -35,6 +35,7 @@
     "test:unit:firefox:headless": "npm run test:unit:default -- --browsers=FirefoxHeadless"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-free": "^5.12.0",
```
jasongrout: FYI, this is not a typo. The org actually is fortawesome: https://github.com/FortAwesome/Font-Awesome/blob/d8db3d032d22e7a5e21f3b40905b7867be19bb02/js-packages/%40fortawesome/fontawesome-free/package.json#L51

packages/jupyterlab-manager/src/index.ts
jasongrout: We are importing other css in plugin.ts. Can we move these imports to there?

jasongrout:
Can you use the minified css? https://github.com/FortAwesome/Font-Awesome/tree/d8db3d032d22e7a5e21f3b40905b7867be19bb02/js-packages/%40fortawesome/fontawesome-free/css

martinRenou:
As discussed offline and in [this PR](https://github.com/jupyterlab/jupyterlab/pull/7754), it might not be a good idea to remove Font-Awesome from JupyterLab and expect third-party extensions to ship Font-Awesome from now on. It might create Font-Awesome version conflicts, for instance, Font-Awesome suggest not to mix v4 and v5 (see https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4).

I will close this PR and open a PR on the JupyterLab repository, upgrading Font-Awesome to 5.12 there.

PR #2699, opened 2020-01-08
prevent npm install, yarn install should be used

mbektasbbg:
prevent running `npm install`.
error message will contain `please-use-yarn`


jasongrout:
Thanks!

PR #2707, opened 2020-01-09
Auto-types

vidartf:
When `types` is not defined in tsconfig, all types from `@types` will be autodetected by Typescript.


jasongrout:
Seems to work well. Thanks!

PR #2709, opened 2020-01-09
Clean up TS project references

vidartf:
We've been having some issues with compilation. This should help avoid some errors, and speed up compilation due to using incremental builds smarter.

Ref https://www.typescriptlang.org/docs/handbook/project-references.html


vidartf:
CC @jasongrout review post merge? 👍

PR #2710, opened 2020-01-09
Split base manager into separate packages

### LINKED ISSUE #2561: Break up the base package
We had some pain recently where we updated the base widget manager, but not the base widget classes, which forced widget authors to update their dependency on the base package.

From https://github.com/jupyter-widgets/ipywidgets/pull/2472#issuecomment-507320306

I think moving forward we can:
* break base into base-widget and base-manager
* make a restricted interface for the manager exposed on a widget that lives in base-widget
* version base-widget and base-manager separately
* not sure where to put the jlab registry token- probably base-widget since that is the only thing we want people to depend on

Then base-manager depends on base-widget, and third-party libraries depend on base-widget, and we have *two* version numbers to communicate to two separate groups of people.

And @vidartf points out that we probably want to bump the compilation target to es2017 as well.

I think this would be an excellent change for 8.0 (it is a js-only thing, so we *could* probably do it before 8.0 if we weren't bumping the compilation target)

We could probably even leave the base-widget package just called base, and separate out the manager. That would make `@jupyter-widgets/base` 3.0, and base-manager depend on base 3.0. That would ease upgrading (custom widget authors would just need to update to `^1 || ^2 || ^3`)

vidartf:
Resolves #2561 .

### DIFF
--- a/.github/workflows/testjs.yml
+++ b/.github/workflows/testjs.yml
@@ -34,6 +34,10 @@ jobs:
         yarn run test:unit:firefox:headless
         popd
 
+        pushd packages/base-manager
+        yarn run test:unit:firefox:headless
+        popd
+
         pushd packages/controls
         yarn run test:unit:firefox:headless
         popd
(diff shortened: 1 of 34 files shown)


jasongrout:
Thanks, this works great!

PR #2712, opened 2020-01-09
[WIP] noUiSlider implementation

### LINKED ISSUE #630: Non-jquery slider
Looking at slider options, here is a simple slider that appears to satisfy our needs: http://refreshless.com/nouislider/ (basic typescript definitions here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/nouislider/nouislider.d.ts)

@SylvainCorlay, @jdfreder, @afshin - what do you guys think of moving to use nouislider?

Another possibility is to use the browser sliders (they are pretty much standardized now, though the UI customizations vary greatly: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/. The problem there is that we wouldn't have range sliders, which would be a regression.

ibdafna:
WIP PR for the new slider implementation.

Fixes #630

Todo: 
1. Additional testing of single-handle slider logic
2. Kernel-side valiation of model.value so it is consistent with min/max and step (as discussed)
3. Styling (current visuals are different from the JQuery-UI implementation
4. Rewrite dual-handle slider functions to use noUiSlider (almost done)

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -166,7 +166,7 @@ class FloatSlider(_BoundedFloat):
     """
     _view_name = Unicode('FloatSliderView').tag(sync=True)
     _model_name = Unicode('FloatSliderModel').tag(sync=True)
-    step = CFloat(0.1, help="Minimum step to increment the value").tag(sync=True)
+    step = CFloat(0.1, allow_none=True, help="Minimum step to increment the value").tag(sync=True)
     orientation = CaselessStrEnum(values=['horizontal', 'vertical'],
         default_value='horizontal', help="Vertical or horizontal.").tag(sync=True)
     readout = Bool(True, help="Display the current value of the slider next to it.").tag(sync=True)
@@ -207,7 +207,7 @@ class FloatLogSlider(_BoundedLogFloat):
     """
     _view_name = Unicode('FloatLogSliderView').tag(sync=True)
     _model_name = Unicode('FloatLogSliderModel').tag(sync=True)
-    step = CFloat(0.1, help="Minimum step in the exponent to increment the value").tag(sync=True)
+    step = CFloat(0.1, allow_none=True, help="Minimum step in the exponent to increment the value").tag(sync=True)
     orientation = CaselessStrEnum(values=['horizontal', 'vertical'],
         default_value='horizontal', help="Vertical or horizontal.").tag(sync=True)
     readout = Bool(True, help="Display the current value of the slider next to it.").tag(sync=True)
@@ -350,7 +350,7 @@ class FloatRangeSlider(_BoundedFloatRange):
     """
     _view_name = Unicode('FloatRangeSliderView').tag(sync=True)
     _model_name = Unicode('FloatRangeSliderModel').tag(sync=True)
-    step = CFloat(0.1, help="Minimum step to increment the value").tag(sync=True)
+    step = CFloat(0.1, allow_none=True, help="Minimum step to increment the value").tag(sync=True)
     orientation = CaselessStrEnum(values=['horizontal', 'vertical'],
         default_value='horizontal', help="Vertical or horizontal.").tag(sync=True)
     readout = Bool(True, help="Display the current value of the slider next to it.").tag(sync=True)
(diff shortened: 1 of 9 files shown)


jasongrout:
With this: `widgets.IntSlider(orientation='vertical')`, it used to be that min was at the bottom and max was at the top. Now it's reversed so that min is at the top and the max is at the bottom.

jasongrout:
`widgets.IntSlider(style={'handle_color': 'red'})` no longer styles the handle red.

jasongrout:
@ibdafna - I'm just playing around with this and making notes where I see issues.

jasongrout:
It feels like the slider thumb is just slightly off-center from its track. It feels like it is a pixel or two too high.
Screen Shot 2020-02-11 at 12 35 11 PM

jasongrout:
@ibdafna has already pointed this out, but I'll note it here to make sure we take another look: `widgets.FloatLogSlider()` is extremely laggy compared to before. Grab the thumb and drag it around, and you see the thumb is very slow to update.

jasongrout:
Stylistically, the range region used to be much thicker with `widgets.IntRangeSlider()`

Before:
Screen Shot 2020-02-11 at 12 39 34 PM

After:
Screen Shot 2020-02-11 at 12 39 54 PM

jasongrout:
Just like with the FloatLogSlider, `widgets.SelectionRangeSlider(options=range(10), value=(2,4))` is very laggy when dragging a thumb compared to before.

jasongrout:
For the lagginess, I tried turning off animation in some cases:

```diff
diff --git a/packages/controls/src/widget_float.ts b/packages/controls/src/widget_float.ts
index 2aee38fb8..94503c90f 100644
--- a/packages/controls/src/widget_float.ts
+++ b/packages/controls/src/widget_float.ts
@@ -147,13 +147,14 @@ class FloatLogSliderView extends BaseIntSliderView {
             log_value = min;
         }
 
-        return log_value
+        return log_value;
     }
 
     createSlider() {
         console.log("Twitter");
         noUiSlider.create(this.$slider, {
             start: this.logCalc(this.model.get('value')),
+            animate: this.model.get('max') % 2 === 0,
             connect: this.connect,
             range: {
                 'min': this.model.get('min'),
diff --git a/packages/controls/src/widget_int.ts b/packages/controls/src/widget_int.ts
index 43665d2be..cabbee91f 100644
--- a/packages/controls/src/widget_int.ts
+++ b/packages/controls/src/widget_int.ts
@@ -228,6 +228,7 @@ abstract class BaseIntSliderView extends DescriptionView {
         noUiSlider.create(this.$slider, {
             start: this.model.get('value'),
             connect: this.connect,
+            animate: this.model.get('max') % 2 === 0,
             range: {
                 'min': this.model.get('min'),
                 'max': this.model.get('max')
diff --git a/packages/controls/src/widget_selection.ts b/packages/controls/src/widget_selection.ts
index 4369564bb..f2335f4f8 100644
--- a/packages/controls/src/widget_selection.ts
+++ b/packages/controls/src/widget_selection.ts
@@ -686,6 +686,7 @@ class SelectionSliderView extends DescriptionView {
         noUiSlider.create(this.$slider, {
             start: this.model.get('index'),
             connect: true,
+            animate: max % 2 === 0,
             range: {
                 'min': min,
                 'max': max
```
It made a *world* of difference (just try doing `FloatSlider(max=9)` vs `FloatSlider(max=10)`). I know the animation looks cool, but I think we ought to turn it off until we can figure out how to avoid the lagginess.

ibdafna:
Just did another round of testing. Outstanding issues:
1. Finish fixing horizontal orientation CSS. The alignment seems to change too between some of the slider types - will investigate in the morning.
2. Fix disabled mode CSS. It is not clear from the styling that the handles are disabled, which may cause confusion.
3. Not passing a step value still seems to default to a step of 0.1 on the frontend.  Will also investigate in the morning.

All other functionality seems to be working fine!

jasongrout:
> Not passing a step value still seems to default to a step of 0.1 on the frontend. Will also investigate in the morning.

I think that is intentional? The attribute specification says that 0.1 is the default step for a float slider.

jasongrout:
Also finish fixing the thickening of the connection between range slider handles for the horizontal case.

ibdafna:
Changed some of the function signatures to accept the updated_view attribute. This is now used to validate whether the slider should update its value when the updateSliderValue callback is invoked. This now enables smooth animations with no lags - shall we revisit this before merging?

jasongrout:
> shall we revisit this before merging?

Sure, let's look at it tomorrow. Thanks for looking into it!

jasongrout:
Thanks so much for your hard work on this!

SylvainCorlay:
Congrats @ibdafna !

martinRenou:
Thanks a lot @ibdafna ! Awesome work

PR #2715, opened 2020-01-10
Add datetime and time pickers

### LINKED ISSUE #2548: Time and Datetime picker widgets
Seems like there is decent support for time and datetime pickers in browsers now (xref #1393), except for Safari:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time

Would we be interested in having controls widgets based on those?

vidartf:
Resolves #2548

### DIFF
--- a/ipywidgets/widgets/tests/test_datetime_serializers.py
+++ b/ipywidgets/widgets/tests/test_datetime_serializers.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Vidar Tonaas Fauske.
+# Distributed under the terms of the Modified BSD License.
+
+import pytest
+
+import datetime
+import pytz
+
+from traitlets import TraitError
+
+from ..trait_types import (
+    time_to_json,
+    time_from_json,
+    datetime_to_json,
+    datetime_from_json,
+)
+
+
+def test_time_serialize_none():
+    assert time_to_json(None, None) == None
+
+
+def test_time_serialize_value():
+    t = datetime.time(13, 37, 42, 7000)
+    assert time_to_json(t, None) == dict(
+        hours=13, minutes=37, seconds=42, milliseconds=7
+    )
+
+
+def test_time_deserialize_none():
+    assert time_from_json(None, None) == None
+
+
+def test_time_deserialize_value():
+    v = dict(hours=13, minutes=37, seconds=42, milliseconds=7)
+    assert time_from_json(v, None) == datetime.time(13, 37, 42, 7000)
+
+
+def test_datetime_serialize_none():
+    assert datetime_to_json(None, None) == None
+
+
+def test_datetime_serialize_value():
+    t = datetime.datetime(2002, 2, 20, 13, 37, 42, 7000, pytz.utc)
+    assert datetime_to_json(t, None) == dict(
+        year=2002,
+        month=1,  # Months are 0-based indices in JS
+        date=20,
+        hours=13,
+        minutes=37,
+        seconds=42,
+        milliseconds=7,
+    )
+
+
+def test_datetime_serialize_non_utz():
+    # Non-existant timezone, so it wil never be the local one:
+    tz = pytz.FixedOffset(42)
+    t = datetime.datetime(2002, 2, 20, 13, 37, 42, 7000, tz)
+    assert datetime_to_json(t, None) == dict(
+        year=2002,
+        month=1,  # Months are 0-based indices in JS
+        date=20,
+        hours=12,
+        minutes=55,
+        seconds=42,
+        milliseconds=7,
+    )
+
+
+def test_datetime_deserialize_none():
+    assert datetime_from_json(None, None) == None
+
+
+def test_datetime_deserialize_value():
+    tz = pytz.FixedOffset(42)
+    v = dict(
+        year=2002,
+        month=1,  # Months are 0-based indices in JS
+        date=20,
+        hours=13,
+        minutes=37,
+        seconds=42,
+        milliseconds=7,
+    )
+    assert datetime_from_json(v, None) == datetime.datetime(
+        2002, 2, 20, 14, 19, 42, 7000, tz
+    )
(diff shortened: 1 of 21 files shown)


vidartf:
Ready for review! :)

pbugnion:
This looks great! I've played around with the interface. I've not looked at the code at all.

- In both the TimePicker and the DatetimePicker docstring, the type of `value` can be None. It might be worth clarifying that.
- In the TimePicker, the default value of `step` is 60 (at least on Chrome?). When you set "any", it defaults to 60? Is it worth specifying this? Also, you're allowed decimal values, which is not obvious. I think both these behaviours are sensible, but it may be worth clarifying this in the docstring description of the `step` parameter?
- In the TimePicker, the max value is inclusive, which is maybe counter-intuitive. Is it worth clarifying this in the docstring?
- In both docstrings, the examples still refer to ipydatetime.

In the DatetimePicker, there's some strange behaviour around zoned date times and offset-native datetimes. As far as I can tell (I haven't looked at the code), setting the value by playing around with the UI control sets a time-zone aware datetime (at least on Chrome). However, I can then set the value manually via an assignment to `value`, with an offset-native datetime.

The most straightforward way to exploit this discrepancy to cause an exception is:
1. Create and display a DatetimePicker

```python
import ipywidgets as widgets
import datetime
dt = widgets.DatetimePicker()
dt
```

2. Set the control manually using the picker. `dt.value` is now offset-aware:

```python
dt.value.tzname()  # => 'GMT' for me
```

3. Assign an offset-native datetime to min or max:

```python
dt.max = datetime.datetime(2018, 1, 1)
```

gives:

```
~/oss/ipywidgets/ipywidgets/widgets/widget_datetime.py in _validate_max(self, proposal)
     82         if self.min and max < self.min:
     83             raise TraitError("setting max < min")
---> 84         if self.value and max < self.value:
     85             self.value = max
     86         return max

TypeError: can't compare offset-naive and offset-aware datetimes
```

It feels like the only sensible way to deal with this is to either only allow zoned datetimes or offset-native datetimes, but not a mix. From your entry in the documentation, it looks like you expect the user to select a zoned datetime, not an offset-native one. Maybe we could introduce a new traitlet, or allow customization of the current one?

vidartf:
> It feels like the only sensible way to deal with this is to either only allow zoned datetimes or offset-native datetimes, but not a mix. 

Seems reasonable. I've also added a naive picker in `ipydatetime`, so I might just add validators to both of them to enforce it (should it coerce silently, or fail noisily?).

vidartf:
@pbugnion I've added a separate picker for naive datetimes (`NaiveDatetimePicker`), and made `DatetimePicker` strictly timezone aware.

vidartf:
Now green ✔️

jasongrout:
From ipywidgets meeting today: we should have further discussion about serialization formats. CC @SylvainCorlay and @vidartf

epifanio:
Hi, I was looking for a `DatetimePicker` in IPywidgets and I came across this :) what is the status? will it be implemented?

vidartf:
It is currently available in the package ipydatetime. The hope/plan is to get this into core ipywidgets (since it is a "standard" browser control), but I need to update the PR with the latest changes from that package.

davidbrochart:
@vidartf I rebased this PR on master and opened a PR in your repo: https://github.com/vidartf/ipywidgets/pull/1

vidartf:
This should be ready now. @SylvainCorlay I didn't touch datetime serialization format. Let's discuss any changes in a followup issue/PR. For context: we ere discussing using a different format for serialization, either Unix timestamp or Isoformat string (full version or limited to Python stdlib supported version).

vidartf:
Don't merge until I've fixed the schemas becoming binary ⚠️

jasongrout:
> Don't merge until I've fixed the schemas becoming binary ⚠️

Marked as draft for now, then.

vidartf:
Fixed 👍

ipywidgets/widgets/tests/test_widget_datetime.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Vidar Tonaas Fauske.
```
jtpio, line 4: We probably want to follow the other copyright headers? (caught the eye while glancing at the diff)

```suggestion
# Copyright (c) Jupyter Development Team.
```

jasongrout:
@jtpio - is this good to merge?

jtpio:
If anyone wants to try this branch on Binder:

https://mybinder.org/v2/gh/vidartf/ipywidgets/datetime?urlpath=lab/tree/docs/source/examples/

image

packages/controls/src/widget_time.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.update2();
+  }
+
+  /**
+   * Update the contents of this view
+   *
+   * Called when the model is changed. The model may have been
+   * changed by another view or by a state update from the back-end.
+   */
+  update2(model?: Backbone.Model, options?: any): void {
```
jtpio, line 127: Should we use a more descriptive name than `update2` for this method?

ipywidgets/widgets/tests/test_datetime_serializers.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) Vidar Tonaas Fauske.
```
jtpio, line 4: Looks like a couple of these headers here and below could be updated to match the existing ones.

jtpio:
@vidartf mind rebasing on the latest `master` to pick up the update to TypeScript 4.3 to check everything compiles correctly?

vidartf:
@jtpio  Rebased, and all green ✔️

jasongrout:
@jtpio, is this good to go in?

jtpio:
I think so yes. In the dev meeting last week we had decided to wait until a couple of dependency update PRs were merged and a new alpha release cut. So it should be good to go now.

jtpio (approved):
Thanks!

PR #2717, opened 2020-01-10
Install test extras_require in dev_install

jtpio:
To avoid an extra `python -m pip install ".[test]"` afterwards to run the tests locally.


jasongrout:
Thanks!

PR #2718, opened 2020-01-10
Fix call to readout_overflow for sliders

jtpio:
Fixes the following issue:

readout_overflow_error


jasongrout:
Good catch. Thanks!

PR #2720, opened 2020-01-10
add drag and drop based on @btel's work

wolfv:


packages/controls/src/widget_dragdrop.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export
+class DraggableBoxView extends DragDropBoxViewBase {
+    /** 
+     * Draggable mixin.
+     * Allows the widget to be draggable
+     *
+     * Note: In order to use it, you will need to add
+     * handlers for dragstartevent in the view class
+     * also need to call dragSetup at initialization time
```
pbugnion: I'm not sure I understand this paragraph.

Also, I think we normally follow typedoc documentation and put API documentation above what it's documenting?

pbugnion:
I thought this was great, thank you! I added a comment inline.

packages/controls/src/widget_dragdrop.ts
wolfv: Thanks for looking through this. Indeed, I haven't gotten around to update the docs! We don't have a mixin anymore, just a "simple" widget. Will change that :)

ipywidgets/widgets/widget_dragdrop.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        self._drop_handlers.register_callback(callback, remove=remove)
+
+    def drop(self, data):
+        """
+        Programmatically trigger a drop event.
+        This will call the callbacks registered to the  drop event.
+        """
+
+        if data.get('application/vnd.jupyter.widget-view+json'):
```
SylvainCorlay: Small coding style comment:

Maybe `application/vnd.jupyter.widget-view+json` should be in an all-caps variable that we reuse everywhere the mime type is used?

wolfv: I am actually not following the spec here correctly (which should be a small dictionary with `{model: XYZ, major: 0, minor: 2}` -- the reason is that setData only allows for strings. Obviously we could encode the JSON as a string, and decode it on the python side but I didn't think it would be worth it...

SylvainCorlay: cc @jasongrout you may have an opinion on whether we should write the JSON into a string on merely pass the id.

SylvainCorlay: (regardless if we use that string constant in multiple place it should probably be in a variable.)

jasongrout: > cc @jasongrout you may have an opinion on whether we should write the JSON into a string on merely pass the id.

I'm a bit confused by what this data is, but certainly a model id is much simpler. I strongly feel that if we use the mimetype, we should be consistent and use it exactly as we have specified.

SylvainCorlay: Thedrag and drop API specifies a mime type key but the value is a string.

wolfv: Ok I switched to the real JSON type now, and do de-serialization from JSON in the backend.

packages/controls/src/widget_dragdrop.ts
wolfv: I removed this outdated docs, and refreshed the docs for Python.

ipywidgets/widgets/widget_dragdrop.py
```diff
@@ -0,0 +1,119 @@
...
+class DropWidget(DOMWidget, CoreWidget):
+    """Base widget for the single-child DropBox and DraggableBox widgets"""
+
+    draggable = Bool(default=False).tag(sync=True)
+    drag_data = Dict().tag(sync=True)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._drop_handlers = CallbackDispatcher()
+        self._dragstart_handlers = CallbackDispatcher()
```
SylvainCorlay: rm

packages/controls/src/widget_dragdrop.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    on_dragover(event: DragEvent): void {
+        event.preventDefault();
+        event.stopPropagation();
+        if (event.dataTransfer) {
+            event.dataTransfer.dropEffect = 'copy';
+        }
+    }
+}
```
SylvainCorlay: nl

SylvainCorlay:
This looks good to me :)

docs/source/examples/Drag and Drop.ipynb
```diff
@@ -0,0 +1,547 @@
...
+   "metadata": {},
+   "source": [
+    "## Draggable Label"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "`DraggableLabel` is a label that can be dragged and dropped to other fields."
```
jtpio: It looks like `DraggableLabel` can be removed from the example notebook.

jtpio: Never mind that, I thought it was initially imported from `ipywidgets`

wolfv: Indeed, the initial implementation had a `DraggableLabel`, but we removed it in favor of the more general `DraggableBox`.

packages/controls/src/widget_dragdrop.ts
```diff
@@ -0,0 +1,170 @@
...
+
+import {
+    CoreDOMWidgetModel
+} from './widget_core';
+
+import {
+    DOMWidgetView, unpack_models, WidgetModel, WidgetView, JupyterLuminoPanelWidget, reject
+} from '@jupyter-widgets/base';
+
+import * as _ from 'underscore';
```
jtpio: Don't we also need jQuery for JupyterLab?

```javascript
import $ from 'jquery';
```

wolfv: the drag'n'drop code only uses browser api's so I think for this particular code it shouldn't be necessary.

jtpio: Isn't it needed here?

https://github.com/jupyter-widgets/ipywidgets/blob/b8890e8da3b57b1351545499533608dbe00893b1/packages/controls/src/widget_dragdrop.ts#L63

wolfv: Wow, good catch. I copied this code from the `widget_box.ts`. I am not even sure wether it's necessary. Will investigate, thanks!

jtpio: The line above should be enough to make it work in JupyterLab.

jtpio:
Really neat:

drag-drop-output-view

SylvainCorlay:
I am in favor of merging this. Leaving it open for a little bit more so that others get a chance to comment.

jtpio:
@wolfv the fix for JupyterLab: https://github.com/wolfv/ipywidgets/pull/1

packages/controls/src/widget_dragdrop.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    initialize(parameters: WidgetView.IInitializeParameters): void {
+        super.initialize(parameters);
+        this.add_child_model(this.model.get('child'));
+        this.listenTo(this.model, 'change:child', this.update_child);
+
+        this.pWidget.addClass('jupyter-widgets');
+        this.pWidget.addClass('widget-container');
+        this.pWidget.addClass('widget-draggable-box');
```
jtpio: Is this class needed? It doesn't seem to be used.

wolfv: I think I just tried to follow conventions: https://github.com/jupyter-widgets/ipywidgets/blob/7cb20cdf2df8ad54fbc0897934b3075998a9a9b5/packages/controls/src/widget_box.ts#L83-L85 

someone who wants to style the drag drop widgets might want to add CSS to this?
On the other hand, the two leaf classes should probably add another class as well.

jtpio:
@wolfv a quick pass on the example notebook here, let me know what you think: https://github.com/wolfv/ipywidgets/pull/3

packages/controls/src/widget_dragdrop.ts
jtpio: > someone who wants to style the drag drop widgets might want to add CSS to this?

That could indeed be useful.

For the case of the widget_box, there is actually some default css:

https://github.com/jupyter-widgets/ipywidgets/blob/7cb20cdf2df8ad54fbc0897934b3075998a9a9b5/packages/controls/css/widgets-base.css#L98-L103

jtpio:
Maybe we'll want to look into adding the `dragenter` and `dragleave` events in a separate PR? It could be useful to give visual feedback of what and where widgets can be dropped:

drag-enter-leave

wolfv:
One thing that I am afraid of with these additional events is that depending on the latency, things might get out of order on slow connections ... 

E.g. if we had a drag_start event and allow to modify drag_data, we somehow need to guarantee that the new drag_data is set _before_ the widget is dropped etc.

jtpio:
Yes it sounds like there could be some race conditions indeed.

If we want to still be able to add a visual feedback, an alternative could be to keep these events on the frontend and provide a way to customize the hover effect (for example by setting `hover_style`).

wolfv:
Agree, adding a CSS class on hover would be nice! I was thinking to keep this PR minimal to make sure it gets in, then we can build on top :)

btel:
LGTM. Thanks for updating to code and adding awesome examples @wolfv @jtpio !

jtpio:
Another iteration on the example notebook, to add a "Dashboard Builder" section using the `AppLayout` widget:

dashboard-builder

cc @wolfv: it's in https://github.com/wolfv/ipywidgets/pull/4

wolfv:
Impressive!

Alexboiboi:
This looks so promising, can't wait to use it in my projects ;)

btel:
That's insane! Thanks @jtpio. I didn't even know that you can create a window with a widget view. It's only a step away from an interactive builder for voila dashboards ;-)

wolfv:
I rebased & regenerated the spec :)

jtpio:
Thanks @wolfv!

The docs failure seems relevant and related to https://github.com/jupyter-widgets/ipywidgets/pull/2746 (new in master since the last rebase).

wolfv:
Tried to fix it :)

jtpio:
Looks good :+1:

ipywidgets/widgets/widget_dragdrop.py
```diff
@@ -0,0 +1,118 @@
...
+        self._drop_handlers.register_callback(callback, remove=remove)
+
+    def drop(self, data):
+        """ Programmatically trigger a drop event.
+        This will call the callbacks registered to the  drop event.
+        """
+
+        if data.get('application/vnd.jupyter.widget-view+json'):
+            widget_mime = json.loads(data['application/vnd.jupyter.widget-view+json'])
+            data['widget'] = widget_serialization['from_json']('IPY_MODEL_' + widget_mime['model_id'])
```
vidartf, line 43: I'm thinking about this. Why is this method exposed as an API method? If the user is meant to use it from the kernel side, should we not handle the deserialization of the widget reference before calling this? This way, any kernel use could simply pass a widget reference as `widget`, and not have to worry about serializing the widget properly first.

maij:
This is a really nice addition! Is the on_drop callback extensible to handle more than widgets in the future, e.g. dragging an out-of-browser file into a widget?

giswqs:
When will this drag and drop functionality be available? I can't wait to try this out.

aditya-giri:
Any ETA on this? This would be very useful for me.

epifanio:
> Any ETA on this? This would be very useful for me.

Wondering about the same. such a feature would be very useful for building dynamic dashboards.

gregfreeman:
I rebased this PR and fixed some minor issues.    [gf_drag_drop](https://github.com/gregfreeman/ipywidgets/tree/gf_drag_drop).   I think this is a really cool feature.   Shall I create another PR?

wolfv:
yes, please! :)

Issue #2721, opened 2020-01-11
Consolidate the file upload widget value metadata and content?


jasongrout:
We've recently discussed the format for the file upload value attribute. It looks like what was settled on was something like: 
```python
[{'metadata': {'name': 'Untitled1.ipynb',
   'type': '',
   'size': 1734,
   'lastModified': 1552316652863},
  'content': <memory at 0x10eeadd00>}]
```
In playing around with it a bit, it seems much simpler to just combine it into one dictionary, sort of like how the [Jupyter contents api](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/notebook/master/notebook/services/api/api.yaml#/contents/get_api_contents__path_) does it:
```python
[{'name': 'Untitled1.ipynb',
   'type': '',
   'size': 1734,
   'lastModified': 1552316652863,
   'content': <memory at 0x10eeadd00>}]
```
I think it's simpler to not have to go through the metadata key, like:

```python
for f in widget.value:
    print(f['name'])
    print(f['size'])
    print(f["content"].tobytes().decode("utf-8"))
```
Thoughts?

pbugnion:
I agree having a single top-level dictionary is simpler. I think this was just forgotten about when we pushed #2666.

jasongrout:
Also, I notice in the documentation from #2714, we have an example of converting the contents to a string: `uploaded_file["content"].tobytes().decode("utf-8")`. As far as I can tell, this creates an extra copy (`tobytes` does a copy). It would probably be better to do:

```python
import codecs
codec.decode(uploaded_file["content"], 'utf-8')
```
which does not do an extra copy by working on the memory view directly.

pbugnion:
I've addressed the docs issue in #2722.

jtpio:
Agree a single dict is simpler.

PR incoming.

pbugnion:
Fixed by #2724 and #2722.

PR #2722, opened 2020-01-11
Avoid unnecessary copies in snippets for file upload

pbugnion:
Prior to this commit, we recommended extracting a string representation of the uploaded file with:

```python
uploaded_file.tobytes().decode("utf-8")
```

This causes a copy of the memoryview to be created when it's converted to bytes. This copy is immediately discarded when it is decoded. Switching to:

```python
codecs.decode(uploaded_file, encoding="utf-8")
```

means that we apparently don't create the intermediate copy (according to Jason in #2721 ). I have not explicitly verified this, though the function signatures at least seem to indicate this.

This commit also adds an example of using the uploaded file in an image widget.

With this PR, the file upload documentation looks like:

Screenshot 2020-01-11 at 12 17 19

Screenshot 2020-01-11 at 12 17 43


jasongrout:
After #2723, we won't need the tobytes in the image part of the example either.

docs/source/examples/Widget List.ipynb
```diff
@@ -1053,7 +1054,15 @@
     "To convert the uploaded file into a Pandas dataframe, you need to convert it to a StringIO object:\n",
     "\n",
     "```python\n",
-    "pd.read_csv(io.StringIO(uploaded_file[\"content\"].tobytes().decode(\"utf-8\")))\n",
+    "import io\n",
+    "import pandas as pd\n",
+    "pd.read_csv(io.StringIO(codecs.decode(uploaded_file[\"content\"], encoding=\"utf-8\")))\n",
```
jasongrout: You can use BytesIO:

```python
pd.read_csv(io.BytesIO(uploaded_file["content"]))
```

docs/source/examples/Widget List.ipynb
```diff
@@ -1036,10 +1036,11 @@
     "#=> b'This is the content of example.txt.\\n'\n",
     "```\n",
     "\n",
-    "To get the contents as a string:\n",
+    "You can get the contents as a string if the file is a text file (more precisely, if the file can be decoded as UTF-8):\n",
```
jasongrout:
```suggestion
    "If the file is a text file, you can get the contents as a string by decoding it:\n",
```

pbugnion:
Great catch on using a `BytesIO` object to load the content into a pandas dataframe.

I've made the changes, so I think this is now ready for another review.

docs/source/examples/Widget List.ipynb
```diff
@@ -1053,10 +1054,18 @@
     "    fp.write(uploaded_file[\"content\"])\n",
     "```\n",
     "\n",
-    "To convert the uploaded file into a Pandas dataframe, you need to convert it to a StringIO object:\n",
+    "To convert the uploaded file into a Pandas dataframe, you need to convert it to a [BytesIO object](https://docs.python.org/3/library/io.html#binary-i-o):\n",
```
jasongrout:
```suggestion
    "To convert the uploaded file into a Pandas dataframe, you can use a [BytesIO object](https://docs.python.org/3/library/io.html#binary-i-o):\n",
```

pbugnion:
> Thanks, looking good! I added two more very small tweaks, then I think it is good to go.

Thanks for reviewing!

PR #2723, opened 2020-01-11
Change media widgets to use memory views.

jasongrout:
This can be more efficient with fewer copies in some cases, such as uploading a file and setting the value from the file content. Also, the deserialization does not involve a copy anymore.

CC @pbugnion, @maartenbreddels

### DIFF
--- a/ipywidgets/widgets/tests/test_widget_image.py
+++ b/ipywidgets/widgets/tests/test_widget_image.py
@@ -149,8 +149,9 @@ def test_value_repr_length():
     with get_logo_png() as LOGO_PNG:
         with open(LOGO_PNG, 'rb') as f:
             img = Image.from_file(f)
-            assert len(img.__repr__()) < 120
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
We should check the impact of this on ipywebrtc.

pbugnion:
As part of the work on this, we should change the docs for the file upload widget to avoid an unnecessary `.tobytes()` when embedding the uploaded content in an image widget (x-ref #2722).

jasongrout:
> We should check the impact of this on ipywebrtc.

@maartenbreddels - are you able to do this?

jasongrout:
> We should check the impact of this on ipywebrtc.

It may be easiest to do this with an alpha release of ipywidgets with this merged in? Can we merge this in, and revert later if there are problems?

jasongrout:
Merging - if it is an issue, especially with ipywebrtc or some other downstream library, we can revert.

PR #2729, opened 2020-01-13
Styling labels and text inputs

zerline:
Adding `background-color`, `color`, `font-size` to labels and all text inputs.
Adding `font-family`,  `font-style`, `font-variant`, `text-color`, `text-decoration` to labels.


packages/controls/src/widget_string.ts
```diff
@@ -21,6 +25,105 @@ import * as _ from 'underscore';
...
+            _model_name: 'TextStyleModel',
+            _model_module: '@jupyter-widgets/controls',
+            _model_module_version: JUPYTER_CONTROLS_VERSION,
+        });
+    }
+
+    public static styleProperties = {
+        ...DescriptionStyleModel.styleProperties,
+        background_color: {
+            selector: '.input',
```
jasongrout: I think this should just be `'input'`, not `'.input'`? In CSS, prefixing the word with a dot means to look for a CSS class `input`, whereas not having a dot means to look for a tag with name `input`.

zerline: yes, that was meant to be `input` as a CSS class (which I have also added in the HTML display).
We could also consider the tag. If that works (I have not tested that), and if we can promise to have such an input tag for each and every text input.

zerline: ... which is not the case with `textarea.`

zerline:
See #2728 or #2730.

PR #2732, opened 2020-01-13
[WIP] Deferred init

wolfv:
This is a PR made with help from @jasongrout to allow for deferred init for widgets when e.g. dynamically loaded into JLab using https://github.com/wolfv/jupyterlab-dynext


jasongrout:
Looks like there are some lint errors. Run `yarn run lint` to fix some of them and to tell you the rest.

packages/jupyterlab-manager/src/semvercache.ts
```diff
@@ -3,22 +3,54 @@
...
     } else {
-      throw `Version ${version} of key ${key} already registered.`;
+      throw `Version ${version} of name ${name} already registered.`;
     }
+    this._retry();
   }
 
-  get(key: string, semver: string): T | undefined {
+  _retry(): void {
+    clearTimeout(this._timeout);
```
vidartf: Can the pollers in lumino help with this logic?

wolfv: Thanks for checking out this PR! :) Sounds like they could!

wolfv: hmm .. i don't know... our logic is pretty simple and the functions are quite straight forward. I don't know if using the polling api would make it more easy to understand -- and also I do not understand which part I should really use.

jasongrout: We just thought this through, and we think a small change in the logic would let us use the throttler without too much confusion or hassle.

jasongrout: Done!

jasongrout:
@wolf - I went ahead and renamed it SemVerRegistry to take care of the confusion you had around the word "cache", and reworked the code to have better names, etc.

wolfv:
great! thanks for doing this.

jasongrout:
@vidartf, do you want to take a look at this again?

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (this._pendingRequests.length > 0) {
+      this._throttle.invoke();
+    }
+  }
+
+  private _registry = new Map<string, Map<string, T>>();
+  private _pendingRequests: IRequest<T>[] = [];
+  private _throttle = new Throttler(() => {
+    this._retry();
+  }, TIMEOUT);
```
vidartf: I really prefer initializers with actual logic to be specified in the constructor 😕 (I prefer most initializers to be in the constructor actually, but when there is good typing they are less important).

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,147 @@
...
+    ArrayExt.removeAllWhere(
+      this._pendingRequests,
+      ({ result, timestamp }) => {
+        if (timestamp === null) {
+          // Request fulfilled, so remove the entry.
+          return true;
+        } else if (now - timestamp > TIMEOUT) {
+          // Request expired, so give up and remove the entry.
+          result.resolve(undefined);
+          return true;
```
vidartf: Does this logic mean that if the resolution takes more than `TIMEOUT` (300 ms) wall time, we give up?

wolfv: yep. 

Right now we basically immediately give up (no later loading of widgets supported).

vidartf:
(there is also a merge conflict to resolve here)

packages/jupyterlab-manager/src/semverregistry.ts
vidartf: Just to make sure I understand, what problem does this PR resolve?

packages/jupyterlab-manager/src/semverregistry.ts
jasongrout: Noted. I prefer what can be moved to the definition to be there.

packages/jupyterlab-manager/src/semverregistry.ts
jasongrout: Good question.

In general, the widget manager depends on the notebook, and custom widgets depend on the notebook manager, so we aren't guaranteed to have custom widgets loaded before a notebook is instantiated just from the dependency injection system. We noticed a problem when using @wolfv's new dynamic extension loader, which can load a custom widget like pythreejs or bqplot at app startup, but loads them asynchronously and in an unspecified order.

What this PR does is give extensions a chance to load custom widgets asynchronously near to when they are needed. We did pick kind of an arbitrary timeout to try to balance between being able to load something from the network and process it, and giving up being responsive about showing an error to the user.

jasongrout:
Fixed the merge conflict, and then it seemed like the constructor arguments were proliferating to the point where an options object was appropriate.

packages/jupyterlab-manager/src/semverregistry.ts
vidartf: Sounds like a tricky problem. Let me see if I understand the timings:
- All extensions are activated (their promises pending).
- The layout restorer starts fetching its data.
- The notebook extension resolves its tracker.
- The widget manager registers a placeholder with the rendermime registry, and resolves its registry
- Extensions that supply widgets register with the registry
- The dynamic widget extension starts fetching remote scripts.
- The `app.restored` promise resolves.
- The settings registry starts fetching schemas to update settings of other extensions
- The layout restorer starts restoring trackers (e.g. notebooks)
- Widgets are instantiated
- Settings resolve for the dynamic widget extension, or its fetch results come back, and it now registers a widget type, too late.

vidartf: Would it make sense for the dynamic widget extension to not resolve its activate function before its network fetches resolve? It can potentially slow down the app start up, but it seems like the right place to wait for remote JS fetches. If so, the only place that could be a source of a race would be the settings fetch.

vidartf: Ah, I guess not, as I assume the remote URLs are also stored in settings..

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,144 @@
...
+   * registering the object.
+   */
+  async get(name: string, semVer: string): Promise<T | undefined> {
+    const mod = this._getSync(name, semVer);
+    if (mod) {
+      return mod;
+    }
+    const result = new PromiseDelegate<T | undefined>();
+    const timestamp = Date.now();
+    this._pendingRequests.push({ name, semVer, result, timestamp });
```
wolfv: @jasongrout do you think we need to do something here for the case where you don't have e.g. bqplot, and it's never going to load? In that case, we currently might delay the notebook execution by 300ms every time a widget is shown (while IMO we shoudl only delay it the first time bqplot is requested).

packages/jupyterlab-manager/src/semverregistry.ts
wolfv: I don't think we can change the logic much but in the worst case scenario the user should see `Loading widget model...` for ~300ms max (which is blocking the notebook execution) but he should _only_ see it for the first time. I think that should be OK from a UX perspective, and buys us enough time to register the dynamically loaded widgets so restoration works as it should – otherwise simply re-executing the cell should do it.

Although that gives me an idea ... would it be possible to figure out wether a widget is being created from a restoration vs interactively? Because if yes, we should never wait interactively, and wait as long as possible in the restoration mode ... would that be possible cc @jasongrout

packages/jupyterlab-manager/src/semverregistry.ts
jasongrout: Well, we only delay responses (which I think you are right, will delay notebook execution, fair enough) when you request a widget that is not there. If a widget is shown that is there, there is no delay (other than an async promise that immediately resolves).

If there is a complicated bqplot object, than we might have 20 bqplot models that will be requested right off. We can't delay just the first and error out the other 19, because if bqplot does load, we want to resolve all 20 correctly. What are you imagining should happen in this case?

wolfv: The issue would be that we also delay notebook execution when interactively executing the notebook. I'll give this a shot + fix it if it's actually an issue.

jasongrout: >  I'll give this a shot + fix it if it's actually an issue.

We're getting closer to needing to release 8.0, so either we need to move this PR forward or bump it to a later release.

wolfv: Hi @jasongrout I just checked again with four bqplot widgets in a notebook. It remains "snappy" so I don't think our current logic is an issue. Looks good to merge to me!

wolfv: Ok turns out I was hitting some cache which made it load the old version or something like that.

It doesn't work at all right now, as it's impossible to `invoke()` the throttler when it's in `invoked` state, meaning the throttler only runs exactly _once_ right now, not enough to clear the timeouts.

I'll look into either switching to the poller in Lumino ...

wolfv:
@jasongrout I have fixed the issue with the throttler by using the debounce (it's almost the same implementation except you can re-trigger when invoked).

Alos I have added a `_already_retried` list which get's checked so we retry and hold up execution only once.

I have used a simple string-concatenation for the `name:version`. It would probably be more correct to use a tuple with equality operator but that seems difficult?!

I could write our own class that stores a list of lists with a different `contains` check ... let me know if you want me to do that.

wolfv:
@jasongrout now I am telling hte semverregistry wether we are in the kernel restore phase. if yes, 4x more time (so restoring works properly when refreshing the page).
We could remove the retry completely from the case where it's not actually restoring.

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,151 @@
...
+   * not registered.
+   *
+   * #### Notes
+   * Resolution may be delayed for a short timeout to allow for some delay in
+   * registering the object.
+   */
+  async get(
+    name: string,
+    semVer: string,
+    isKernelRestore: boolean
```
jasongrout: This shouldn't know anything about the concept of kernels. How about this parameter be a timeout (in ms) on the request, defaulting to 0? Then the thing calling get() can say to use a specific timeout on the request, like 500ms. Then this would mirror standard socket apis which have a timeout parameter.

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,151 @@
...
+  ): Promise<T | undefined> {
+    const mod = this._getSync(name, semVer);
+    if (mod || this._already_retried.indexOf(name + ':' + semVer)) {
```
jasongrout: This means this concern is true:

> If there is a complicated bqplot object, than we might have 20 bqplot models that will be requested right off. We can't delay just the first and error out the other 19, because if bqplot does load, we want to resolve all 20 correctly. What are you imagining should happen in this case?

If I request a bqplot class twice, the first time it will push a pending request. However, the second time, it will just return undefined. That doesn't sound right to me.

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,151 @@
...
+    // If we still have pending requests, try again later.
+    if (this._pendingRequests.length > 0) {
+      this._debounce.invoke();
+    }
+  }
+
+  private _already_retried = new Array<string>();
+  private _registry = new Map<string, Map<string, T>>();
+  private _pendingRequests: IRequest<T>[] = [];
+  private _debounce = new Debouncer(() => {
```
jasongrout: I don't think we want a debouncer. A debouncer will delay calling the function until TIMEOUT ms have passed, which means you can regularly call a debouncer to indefinitely delay the function call. We really do want a throttle, meaning that no matter how many times we call the retry function, it only executes at most once every TIMEOUT ms (but it *does* indeed execute every TIMEOUT ms or so, unlike the debouncer).

jasongrout: If we go with the idea of a `timeout` argument to the get function as suggested above, then the timeout on this throttle essentially controls the resolution of that timeout argument.

packages/jupyterlab-manager/src/semverregistry.ts
```diff
@@ -0,0 +1,151 @@
...
+    const mod = this._getSync(name, semVer);
+    if (mod || this._already_retried.indexOf(name + ':' + semVer)) {
+      return mod;
+    }
+    this._already_retried.push(name + ':' + semVer);
+    const result = new PromiseDelegate<T | undefined>();
+    let timestamp = Date.now() + TIMEOUT;
+    if (isKernelRestore) timestamp += TIMEOUT * 4;
+    this._pendingRequests.push({ name, semVer, result, timestamp });
+    this._debounce.invoke();
```
jasongrout: If `this._throttle.invoke();` doesn't work because we are invoking from the invoked state, we can always invoke it asynchronously: `setTimeout(()=>this._debounce.invoke, 0)` or something.

jasongrout: I filed an issue at https://github.com/jupyterlab/lumino/issues/47

vidartf:
BTW: I think this PR would be a case where using the network throttling options in Chrome devtools can be useful.

packages/jupyterlab-manager/src/semverregistry.ts
wolfv: hmm.... in my tests it seemed to work fine. I am wondering... since the JS is anyways blocking on the await in the widgets manager, we're still kind-of waiting for the first bqplot instance to be returned, right?

jasongrout: hmmm...perhaps it has something to do with the debouncer delaying resolution until all requests are made?

martinRenou:
Do we still need this? This PR is less critical now that JupyterLab 3 and federated extensions are a thing. I am wondering if there are still use cases for this?

jasongrout:
This came up when wolf and I were trying to support loading widgets on demand in jupyterlab-dynext. That project is seeing new life, so this may still be important.

PR #2733, opened 2020-01-13
Update the prettier configuration

jasongrout:
This adds lint-staged, which will run prettier on staged files automatically.


jasongrout:
This should be merged before #2734

PR #2735, opened 2020-01-14
Adding examples to the example doc how to change the size of the dropdown/radiobuttons

adam-ah:
Based on the discussion on https://github.com/jupyter-widgets/ipywidgets/issues/2050, a small update to the doc.

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -557,9 +557,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "Box(\n",
+    "widgets.Box(\n",
     "    [\n",
-    "        Label(value='Pizza topping with a very long label:'), \n",
+    "        widgets.Label(value='Pizza topping with a very long label:'), \n",
     "        widgets.RadioButtons(options=['pepperoni', 'pineapple', 'anchovies', 'and the long name that will fit fine'])\n",
     "    ]\n",
     ")"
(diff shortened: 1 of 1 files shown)


docs/source/examples/Widget List.ipynb
```diff
@@ -537,12 +537,34 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "widgets.Box(\n",
+    "    [\n",
+    "        widgets.Label(value='Pizza topping with a very long label:'), \n",
+    "        widgets.RadioButtons(options=['pepperoni', 'pineapple', 'anchovies', 'and the long name that will fit fine'])\n",
```
jasongrout: Do you want `layout={'width': 'max-content'}` somewhere in here? For me, this text fits fine with the default width, so maybe make the text a little longer so it's clear that it is working?

jasongrout:
Thanks!

docs/source/examples/Widget List.ipynb
adam-ah: That's the whole point; if you use boxes (custom layout), you don't need to set the `width` at all.

jasongrout: > That's the whole point; if you use boxes (custom layout), you don't need to set the `width` at all.

It doesn't work for me with the latest widgets release, though I have to make the string a bit longer to see it:
Screen Shot 2020-01-14 at 3 52 03 AM

adam-ah: Great catch, thanks! It seems like it will fit in more, but not all of the text. I added the same layout changes to the Box example.

jasongrout:
Thanks!

PR #2742, opened 2020-01-17
Drop underscore usage

SylvainCorlay:
Even though underscore is already a dependency of backbone...

### DIFF
--- a/packages/controls/src/widget_audio.ts
+++ b/packages/controls/src/widget_audio.ts
@@ -5,11 +5,10 @@ import { DOMWidgetView } from '@jupyter-widgets/base';
 
 import { CoreDOMWidgetModel } from './widget_core';
 
-import * as _ from 'underscore';
-
 export class AudioModel extends CoreDOMWidgetModel {
   defaults(): Backbone.ObjectHash {
-    return _.extend(super.defaults(), {
+    return {
+      ...super.defaults(),
       _model_name: 'AudioModel',
       _view_name: 'AudioView',
       format: 'mp3',
@@ -17,7 +16,7 @@ export class AudioModel extends CoreDOMWidgetModel {
       loop: true,
       controls: true,
       value: new DataView(new ArrayBuffer(0))
-    });
+    };
   }
 
   static serializers = {
--- a/packages/controls/src/widget_selection.ts
+++ b/packages/controls/src/widget_selection.ts
@@ -9,7 +9,6 @@ import { DescriptionView, DescriptionStyleModel } from './widget_description';
 
 import { uuid } from './utils';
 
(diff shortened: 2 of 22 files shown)


packages/controls/src/widget_selection.ts
```diff
@@ -314,10 +313,7 @@ export class RadioButtonsView extends DescriptionView {
    */
   update(options?: any): void {
     const items: string[] = this.model.get('_options_labels');
-    const radios = _.pluck(
-      this.container.querySelectorAll('input[type="radio"]'),
-      'value'
-    );
+    const radios = this.container.querySelectorAll('input[type="radio"]').map(x => x.value);
```
jasongrout: Does this work?

```suggestion
    const radios = Array.from(this.container.querySelectorAll('input[type="radio"]')).map(x => x.value);
```

packages/controls/src/widget_selection.ts
```diff
@@ -502,7 +499,7 @@ export class ToggleButtonsView extends DescriptionView {
     const tooltips = this.model.get('tooltips') || [];
     const disabled = this.model.get('disabled');
     const buttons = this.buttongroup.querySelectorAll('button');
-    const values = _.pluck(buttons, 'value');
+    const values = buttons.map(x => x.value);
```
jasongrout: Maybe this?
```suggestion
    const values = Array.from(buttons).map(x => x.value);
```

packages/controls/src/widget_audio.ts
```diff
@@ -5,19 +5,18 @@ import { DOMWidgetView } from '@jupyter-widgets/base';
 
 import { CoreDOMWidgetModel } from './widget_core';
 
-import * as _ from 'underscore';
-
 export class AudioModel extends CoreDOMWidgetModel {
   defaults(): Backbone.ObjectHash {
-    return _.extend(super.defaults(), {
+    return {
+      ...super.defaults(),
```
jasongrout: The disadvantage to this is that it creates a new object at every level of the class hierarchy, every time a class is created. The other would create one defaults dictionary at the root and modify it for each subclass.

However, perhaps that was just premature optimization and we shouldn't worry about it.

jasongrout: If this really does become an issue, we can use `Object.assign` directly instead of underscore these days.

jasongrout:
This works great; thanks!

SylvainCorlay:
Thanks, I would have fixed it though :)

jasongrout:
How about you review https://github.com/jupyter-widgets/ipywidgets/pull/2732 or https://github.com/jupyter-widgets/ipywidgets/pull/2532 instead? :)

PR #2746, opened 2020-01-22
Make selection container titles a tuple of strings

jasongrout:
`get_title` and `set_title` have long been somewhat a wart in our API, requiring people to put several statements after creating a selection container to label each container one by one. This PR makes the titles a proper tuple so titles can be set at creation time or all at once.

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -1223,9 +1223,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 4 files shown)


jasongrout:
@pbugnion or @vidartf - any chance you could review this?

pbugnion:
The interface is much better! Thank you!

The only thing that's a little strange is that we don't enforce that `.children` and `titles` is the same length. I'm guessing enforcing that they are the same length would make it hard to change the number of children: you'd have to hold trait notifications while you changed both. Not enforcing length consistency therefore seems sensible.

I think this is good to merge.

PR #2749, opened 2020-01-23
Move @types/backbone to devDependencies

southpolesteve:
Best I can tell there is no reason this need to be in production deps. Pulling it in is causing some @types conflicts in one of my projects.

### DIFF
--- a/packages/jupyterlab-manager/package.json
+++ b/packages/jupyterlab-manager/package.json
@@ -56,12 +56,12 @@
     "@lumino/properties": "^1.1.0",
     "@lumino/signaling": "^1.2.0",
     "@lumino/widgets": "^1.3.0",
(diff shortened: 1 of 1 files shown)


jasongrout:
We have publicly exported interfaces that use the Backbone types in that package. The Typescript manual [this recommends](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies) including these typings as dependencies.

What conflicts are you seeing?

southpolesteve:
@types/backbone depends on @types/jquery which is causing the conflict. I see the dilemma now with the exported types. I've previously solved this by ["mirroring" the type](https://github.com/Azure/azure-cosmos-js/pull/175/files) but that looks like it would be hard to do in this case since the Backbone type is so complex. Going to close this and see if I can find another workaround.

PR #2752, opened 2020-01-25
Delete display_model and display_view

### LINKED ISSUE #2751: display_view and display_model used inconsistently
We discovered in https://github.com/jupyter-widgets/ipywidgets/pull/2657#issuecomment-571852389 that display_view is used inconsistently throughout the codebase. In discussion at the widgets sprint in Paris, we decided to explore getting rid of display_view (and display_model?), and instead just having create_view.

jasongrout:
Fixes #2751

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -55,33 +55,7 @@ export interface IBase64Buffers extends PartialJSONObject {
 /**
  * Manager abstract base class
  */
-export abstract class ManagerBase<T> implements IWidgetManager {
-  /**
-   * Display a DOMWidgetView for a particular model.
-   */
-  display_model(
-    msg: services.KernelMessage.IMessage | null,
-    model: DOMWidgetModel,
-    options: any = {}
-  ): Promise<T> {
-    return this.create_view(model, options)
-      .then(view => this.display_view(msg, view, options))
-      .catch(reject('Could not create view', true));
-  }
-
-  /**
-   * Display a DOMWidget view.
-   *
-   * #### Notes
-   * This must be implemented by a subclass. The implementation must trigger the view's displayed
-   * event after the view is on the page: `view.trigger('displayed')`
-   */
-  abstract display_view(
-    msg: services.KernelMessage.IMessage | null,
-    view: DOMWidgetView,
-    options: any
-  ): Promise<T>;
-
+export abstract class ManagerBase implements IWidgetManager {
   /**
    * Modifies view options. Generally overloaded in custom widget manager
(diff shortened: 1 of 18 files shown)


jasongrout:
ReadTheDocs seems to be having issues, but all other tests pass.

jasongrout:
I think this is ready for review, @vidartf @pbugnion @martinRenou (or whoever else wants to review it). The failing test is an unrelated readthedocs issue.

packages/base-manager/src/manager-base.ts
```diff
@@ -55,33 +55,7 @@ export interface IBase64Buffers extends PartialJSONObject {
...
-   * This must be implemented by a subclass. The implementation must trigger the view's displayed
-   * event after the view is on the page: `view.trigger('displayed')`
-   */
-  abstract display_view(
-    msg: services.KernelMessage.IMessage | null,
-    view: DOMWidgetView,
-    options: any
-  ): Promise<T>;
-
+export abstract class ManagerBase implements IWidgetManager {
```
vidartf: I like how the generic arg `T` disappears!

packages/html-manager/src/libembed.ts
```diff
@@ -57,47 +59,44 @@ export function renderWidgets(
...
+  const tags = element.querySelectorAll(
+    'script[type="application/vnd.jupyter.widget-view+json"]'
+  );
+  await Promise.all(
+    Array.from(tags).map(async viewtag => {
       const widgetViewObject = JSON.parse(viewtag.innerHTML);
       const valid = view_validate(widgetViewObject);
       if (!valid) {
-        console.error('View state has errors.', view_validate.errors);
+        throw new Error(`View state has errors: ${view_validate.errors}`);
```
vidartf: Note that `Promise.all` has a fail-fast behavior (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all), so if one view has an issue, this change will cause *all* views to not be displayed.

vidartf: Ah, actually, the fail-fast will just cause the outer promise to reject, it will still keep processing the other promises. My bad :)

jasongrout: So we're all good, then?

jasongrout: ping @vidartf

jasongrout: ah, I didn't see that you marked this conversation as "resolved", which indicates we are all good on this after all.

jasongrout:
Thanks for the review, @vidartf.

Issue #2727, opened 2020-01-13
Should we convert uploaded file metadata into Python types?


pbugnion:
Following PRs #2724 and #2666 , the `FileUpload.value` traitlet looks like:

```python
[{'name': 'example.csv',
  'type': 'text/csv',
  'size': 2548,
  'lastModified': 1536574751152,
  'content': <memory at 0x11d51ab88>}]
```

It is defined as `List(Dict())`.

We could consider the following, in increasing order of complexity:

1. change `lastModified` to be a datetime object. We should check browser compatibility on whether it's always expressed as a timestamp.
2. define a namedtuple or a Python class that exposes the attributes, rather than a dict. This would make the attributes more discoverable.

notifying @jtpio , @jasongrout who have been working on this.

jasongrout:
> define a namedtuple or a Python class that exposes the attributes, rather than a dict. This would make the attributes more discoverable.

Why not have both? Use a Bunch! :)

https://github.com/ipython/traitlets/blob/master/traitlets/utils/bunch.py

(That's what we use for the change argument in traitlets change notifications)

jtpio:
Both suggestions sound good, especially if we want to stay close to the contents api: https://jupyter-notebook.readthedocs.io/en/stable/extending/contents.html#filesystem-entities

vidartf:
As an xref, this might tie in with the datetime serialization in #2715. There, it is serialized along the pattern of `DatePicker`, as a dict of year, month, day, etc. Maybe we should switch to an ISO timestamp everywhere instead?

pbugnion:
I'll try and address this issue.

jasongrout:
Fixed in #2767.

PR #2767, opened 2020-02-02
Change upload value to contain Python types

pbugnion:
Prior to this PR, the `.value` traitlet for the `FileUpload` widget was a list of dictionaries. Each dictionary contained the raw metadata generated by the browser, as well as the content as a memoryview.

This PR changes the return value to be a tuple of `Bunch` instances. Attribute names inside the bunch are converted to snake case and the last modified is converted to a zoned datetime.

For instance:

```python
# f is a FileUpload instance
>>> f.value
({'name': 'some-name.png',
  'type': 'image/png',
  'size': 29841,
  'content': <memory at 0x118001940>,
  'last_modified': 1578491179205},)
```

This PR also removes syncing an 'error' attribute per file. This was not exposed in the kernel, and it was never set to anything apart from an empty string in the frontend. Note that the global error attribute remains.

Note that this PR has some overlap with #2715 : both share datetime serialization / deserialization considerations. In here, I have done the least amount of work -- I just send a timestamp over the wire, and manage conversion in the kernel. Once 2548 is merged, it may be worth overhauling serialization to have a common interface. That won't be a breaking change.

Closes issue #2727 .


pbugnion:
This is now ready for review.

ipywidgets/widgets/widget_upload.py
```diff
@@ -16,12 +17,107 @@
 from .widget_button import ButtonStyle
 from .widget import register, widget_serialization
 from .trait_types import InstanceDict
+from traitlets.utils.bunch import Bunch
```
jasongrout: I think you can just do `from traitlets import Bunch` to not rely on the internal traitlets structure.

ipywidgets/widgets/widget_upload.py
```diff
@@ -16,12 +17,107 @@
...
+
+def _deserialize_value(js, _):
+    return [_deserialize_single_file(entry) for entry in js]
+
+
+def _serialize_single_file(uploaded_file):
+    js = {}
+    for attribute in ['name', 'type', 'size', 'content']:
+        js[attribute] = uploaded_file[attribute]
+    js['lastModified'] = int(uploaded_file['last_modified'].timestamp() * 1000)
```
jasongrout: Our general convention, at least in default controls, has been snake_case for the attributes in the protocol messages. Should we keep it snake_case here, as `last_modified`?

ipywidgets/widgets/widget_upload.py
```diff
@@ -16,12 +17,107 @@
...
+
+    style: widgets.widget_button.ButtonStyle, optional
+        Style configuration for the button.
+
+    value: List[Dict], optional
+        The value of the last uploaded file or set of files. See the
+        documentation for details of how to use this to retrieve file
+        content and metadata:
+        https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20List.html#File-Upload
+    error: str, optional
```
jasongrout: Missing blank line

ipywidgets/widgets/widget_upload.py
```diff
@@ -32,10 +128,11 @@ class FileUpload(DescriptionWidget, ValueWidget, CoreWidget):
     icon = Unicode('upload', help="Font-awesome icon name, without the 'fa-' prefix.").tag(sync=True)
     button_style = CaselessStrEnum(
         values=['primary', 'success', 'info', 'warning', 'danger', ''], default_value='',
-        help="""Use a predefined styling for the button.""").tag(sync=True)
+        help='Use a predefined styling for the button.').tag(sync=True)
     style = InstanceDict(ButtonStyle).tag(sync=True, **widget_serialization)
     error = Unicode(help='Error message').tag(sync=True)
-    value = List(Dict(), help="The file upload value").tag(sync=True)
+    value = List(Dict(), help='The file upload value').tag(
```
jasongrout: Let's make this a TypedTuple so it is read-only.

jasongrout:
This looks great! A couple of inline comments above, and then I think it's ready to go.

ipywidgets/widgets/widget_upload.py
jasongrout: Another option is to just call it `modified`, or even `mtime`, which mirrors the `stat()` structure and how you address it in python and presumably lots of other languages that let you make a stat call.

pbugnion: `mtime` sounds good.

pbugnion:
Thanks for the feedback. I made all the changes discussed.

jasongrout:
Thanks! This looks really great.

PR #2775, opened 2020-02-07
fix misnamed attribute in _MultipleSelection

hickmanw:
`self._full_options` looks like it should be `self._options_full` because:
- `self._full_options` doesn't appear anywhere else
- `self._options_full` appears elsewhere in the class, including one place where it is assigned the result of `_make_options()`, as is done in the line I modified

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -166,6 +166,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
 
         # Select the first item by default, if we can
         if 'index' not in kwargs and 'value' not in kwargs and 'label' not in kwargs:
+            options = self._options_full
             nonempty = (len(options) > 0)
             kwargs['index'] = 0 if nonempty else None
             kwargs['label'], kwargs['value'] = options[0] if nonempty else (None, None)
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -304,7 +304,7 @@ class _MultipleSelection(DescriptionWidget, ValueWidget, CoreWidget):
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
         options = _make_options(kwargs.get('options', ()))
-        self._full_options = options
+        self._options_full = options
         self.set_trait('_options_labels', tuple(i[0] for i in options))
         self._options_values = tuple(i[1] for i in options)
 
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -16,7 +16,7 @@ from .widget_style import Style
 from .trait_types import InstanceDict, TypedTuple
 from .widget import register, widget_serialization
 from .docutils import doc_subst
-from traitlets import (Unicode, Bool, Bunch, Int, Any, Dict, TraitError, CaselessStrEnum,
+from traitlets import (Unicode, Bool, Int, Any, Dict, TraitError, CaselessStrEnum,
                        Tuple, Union, observe, validate)
 
 _doc_snippets = {}
@@ -94,6 +94,13 @@ _doc_snippets['slider_params'] = """
 """
 
 
+def _exhaust_iterable(x):
+    """Exhaust any non-mapping iterable into a tuple"""
+    if isinstance(x, Iterable) and not isinstance(x, Mapping):
+        return tuple(x)
+    return x
+
+
 def _make_options(x):
     """Standardize the options tuple format.
 
@@ -161,7 +168,8 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
         # We have to make the basic options bookkeeping consistent
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
-        kwargs['options'] = self._validate_options(Bunch(value = kwargs.get('options', ())))
+        kwargs['options'] = _exhaust_iterable(kwargs.get('options', ()))
+        self._options_full = _make_options(kwargs['options'])
         self._propagate_options(None)
 
         # Select the first item by default, if we can
@@ -177,8 +185,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
     @validate('options')
     def _validate_options(self, proposal):
         # if an iterator is provided, exhaust it
-        if isinstance(proposal.value, Iterable) and not isinstance(proposal.value, Mapping):
-            proposal.value = tuple(proposal.value)
+        proposal.value = _exhaust_iterable(proposal.value)
         # throws an error if there is a problem converting to full form
         self._options_full = _make_options(proposal.value)
         return proposal.value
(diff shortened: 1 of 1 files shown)


ipywidgets/widgets/widget_selection.py
```diff
@@ -304,7 +304,7 @@ def __init__(self, *args, **kwargs):
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
         options = _make_options(kwargs.get('options', ()))
-        self._full_options = options
+        self._options_full = options
```
jasongrout: Good catch! Can you also add the _make_options call, like when it is assigned other places?
```suggestion
        self._options_full = _make_options(options)
```

jasongrout: Oh, never mind, `_make_options` is called in the line above.

jasongrout:
I was trying to figure out how this worked. If I'm reading (and remembering) correctly, the problem was that options are initialized in a nondeterministic order, and the `.value`, `.index`, and `.label` options all have validation checks against `._option_values` and `._option_labels`. So what was happening was, for example, the options and index would be given in the initial constructor arguments, but if the index was set (and hence validated) before the options was set, we'd get a validation error. So we had to essentially ensure that `options` would be processed first, so we could set `_option_values` and `_option_labels` before the value, index, or label argument was processed. So it didn't really matter what `_full_options` was, since that was going to be set anyway in the observe for the `options` argument, and wasn't used in the validation functions.

I also noticed we have a bug right now related to exhausting iterators if an iterator is given for the options. In the following case, the select does not have any options.

```
from ipywidgets import *
Select(label='b', options=((i, i) for i in 'abcdef'))
```

Perhaps a cleaner way to do this is to explicitly call the options validation and propagation functions, rather than essentially copying them to this constructor. That lets options be things like iterators as well (which is handled in the options validation). Can you call the validation/propagation functions explicitly, like this?

```diff
diff --git a/ipywidgets/widgets/widget_selection.py b/ipywidgets/widgets/widget_selection.py
index ab8db5674..d9b2ee2ad 100644
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -16,7 +16,7 @@ from .widget_style import Style
 from .trait_types import InstanceDict, TypedTuple
 from .widget import register, widget_serialization
 from .docutils import doc_subst
-from traitlets import (Unicode, Bool, Int, Any, Dict, TraitError, CaselessStrEnum,
+from traitlets import (Unicode, Bool, Bunch, Int, Any, Dict, TraitError, CaselessStrEnum,
                        Tuple, Union, observe, validate)
 
 _doc_snippets = {}
@@ -161,10 +161,8 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
         # We have to make the basic options bookkeeping consistent
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
-        options = _make_options(kwargs.get('options', ()))
-        self._options_full = options
-        self.set_trait('_options_labels', tuple(i[0] for i in options))
-        self._options_values = tuple(i[1] for i in options)
+        kwargs['options'] = self._validate_options(Bunch(value = kwargs.get('options', ())))
+        self._propagate_options(None)
 
         # Select the first item by default, if we can
         if 'index' not in kwargs and 'value' not in kwargs and 'label' not in kwargs:
@@ -303,10 +301,8 @@ class _MultipleSelection(DescriptionWidget, ValueWidget, CoreWidget):
         # We have to make the basic options bookkeeping consistent
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
-        options = _make_options(kwargs.get('options', ()))
-        self._options_full = options
-        self.set_trait('_options_labels', tuple(i[0] for i in options))
-        self._options_values = tuple(i[1] for i in options)
+        kwargs['options'] = self._validate_options(Bunch(value = kwargs.get('options', ())))
+        self._propagate_options(None)
 
         super().__init__(*args, **kwargs)
         self._initializing_traits_ = False
```

jasongrout:
> Can you call the validation/propagation functions explicitly, like this?

I just pushed this commit to this PR.

jasongrout:
CC @jupyter-widgets/jupyter-widgets for a final look-over, since I added some logic.

jasongrout:
CC @vidartf for review.

vidartf:
Note that traitlets master has removed all its utils from export (https://github.com/ipython/traitlets/blob/32ced29b0ab06914ff92e9bdcd8806536633d701/traitlets/traitlets.py#L69-L87) including `Bunch`. Although the future of traitlets master is unknown, I don't think we should reuse `Bunch` just to duck-type a proposal value for this reason. I'll do a commit to clean up.

ipywidgets/widgets/widget_selection.py
```diff
@@ -161,7 +168,8 @@ def __init__(self, *args, **kwargs):
         # We have to make the basic options bookkeeping consistent
         # so we don't have errors the first time validators run
         self._initializing_traits_ = True
-        kwargs['options'] = self._validate_options(Bunch(value = kwargs.get('options', ())))
+        kwargs['options'] = _exhaust_iterable(kwargs.get('options', ()))
+        self._options_full = _make_options(kwargs['options'])
```
jasongrout: My point here was that instead of repeating the logic of assigning self._options_full, we should consolidate the logic and call _validate_options when we needed. I think I would rather have _validate_options access `proposal` as a dict, and keep `_options_full` only being assigned there. Thoughts?

jasongrout:
> Note that traitlets master has removed all its utils from export (https://github.com/ipython/traitlets/blob/32ced29b0ab06914ff92e9bdcd8806536633d701/traitlets/traitlets.py#L69-L87) including `Bunch`. Although the future of traitlets master is unknown, I don't think we should reuse `Bunch` just to duck-type a proposal value for this reason.

Let's fix it upstream: see https://github.com/ipython/traitlets/pull/546

vidartf:
> Let's fix it upstream: see ipython/traitlets#546

This was a deliberate change in https://github.com/ipython/traitlets/pull/416 . Ref https://github.com/jupyter-widgets/traittypes/issues/28

ipywidgets/widgets/widget_selection.py
vidartf: My main concern was that someone might be overriding `_validate_options` due to this pattern being enforced previously, and in the traitlets docs it says:

```
The proposal dictionary must hold the following keys:

    * ``owner`` : the HasTraits instance
    * ``value`` : the proposed value for the modified trait attribute
    * ``trait`` : the TraitType instance associated with the attribute
```

The `must` is what caused me to change it. We can probably add another method that gets called by both init and the validators though? That would also fix us currently repeating the logic in the subclasses' validators.

jasongrout: > We can probably add another method that gets called by both init and the validators though? That would also fix us currently repeating the logic in the subclasses' validators.

That sounds like a better way. +1.

jasongrout: On second thought, I like the way you did things. I'll merge this PR as-is. Thanks again!

PR #2779, opened 2020-02-11
Convert a little more promise .then syntax to async/await for clarity.

jasongrout:
No hurry on merging this. I realized after doing a bit of this that converting the promise syntax to async/await throughout the code would be a good exercise for someone wanting to learn about promises and async/await.

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -183,7 +183,7 @@ export abstract class ManagerBase implements IWidgetManager {
    *                          required and additional options are available.
    * @param  serialized_state - serialized model attributes.
    */
-  new_widget(
+  async new_widget(
     options: IWidgetOptions,
     serialized_state: JSONObject = {}
   ): Promise<WidgetModel> {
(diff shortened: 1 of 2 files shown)


jasongrout:
Setting as a good first issue since there are plenty more cases in the code where we use promise syntax that could be converted to async/await.

jmk9000:
@jasongrout is this still something you'd like help with? Looks like a good learning opportunity for me

jasongrout:
> @jasongrout is this still something you'd like help with? Looks like a good learning opportunity for me

Help with this would be awesome!

jasongrout:
Given the conflicts, it may be easier to start over and use this PR as an initial guide.

jmk89:
Ok sounds good. I've been working locally from the branch in your personal repository to get comfortable with developing in a conda environment. A lot of the initial work has been done already, just a matter of replicating it in a new branch.

packages/base-manager/src/manager-base.ts
```diff
@@ -333,11 +328,13 @@ export abstract class ManagerBase implements IWidgetManager {
...
-    return resolvePromisesDict(this._models).then(models => {
-      Object.keys(models).forEach(id => models[id].close());
-      this._models = Object.create(null);
-    });
+  async clear_state(): Promise<void> {
+    await Promise.all(
+      Object.keys(this._models).map(async id =>
+        (await this._models[id]).close()
+      )
+    );
```
vidartf, line 336: Note: This has a small difference in nuance, which I think opens up a new corner case. If one of the unresolved models still being created rely on the models that it references not being closed, this change might cause it to throw an error where it previously did not. To clarify:
- The old method waited for all models currently in `this._models` to resolve, then closed all of them in one go (synchronously).
- The new method will close each model as soon as they resolve. The ID will still be present in the map, but the model it points to is closed.

That said, I don't really see `clear_state` being used anywhere, so it us likely to actually affect anything.

PR #2793, opened 2020-02-25
Change fontawesome support to work for fontawesome 4 and 5.

jasongrout:
JupyterLab and notebook are upgrading to fontawesome 5 (with the v4 backwards compatibility shim). However, in some places, our CSS is very specific to version 4. This commit modifies these places to use syntax that should for both fontawesome 4 and 5. It involves some DOM structure changes, but that is a private implementation detail, so is backwards compatible with our public API.

These changes should be forward-ported to master.


jasongrout:
Since this only changes the js/css, it does not require a python release.

PR #2801, opened 2020-02-27
Change css variable names to match Lab

### LINKED ISSUE #2062: button font color different in classic notebook and jupyterlab
For button with styles primay, success, info, warning, danger, the text appears **black** in jupyterlab and **white** in the classic notebook, as in the [doc](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Predefined-styles) so presumably the intended color, and clearly the best option if you ask me.

Unfortunately the color attribute is not accessible, only the 'button_color' (background-color) says the doc.  So it would be good to add color to the button style keys.

That would more than solve the slight inconsistency between JupyterLab and the classic notebook.
What do you think ?

jeffyjefflabs:
This will fix #2062 and https://github.com/jupyterlab/jupyterlab/issues/7861.

Lab renamed the css variable for inverse font colors [here](https://github.com/jupyterlab/jupyterlab/commit/66c9b95aee4e69885c83bf6301faca6302d13ffe#diff-3c7737d5780b8a993fbe339eafca37cdL67-R70), which causes styled buttons  (e.g. `button_style='danger'`) in ipywidgets to have the wrong font color (black instead of white) in Lab and Voila.  As mentioned in those two issues, [documentation](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Styling.html#Predefined-styles) and code comments suggest white is the right color.

There are possible [downstream effects](https://github.com/search?q=--jp-inverse-ui-font-color0&type=Code) but most of the search results look like snapshots and forks.

### DIFF
--- a/packages/controls/css/labvariables.css
+++ b/packages/controls/css/labvariables.css
@@ -91,10 +91,10 @@ all of MD as it is not optimized for dense, information rich UIs.
      These will typically go from light to darker, in both a dark and light theme
    */
 
-  --jp-inverse-ui-font-color0: rgba(255, 255, 255, 1);
-  --jp-inverse-ui-font-color1: rgba(255, 255, 255, 1);
-  --jp-inverse-ui-font-color2: rgba(255, 255, 255, 0.7);
-  --jp-inverse-ui-font-color3: rgba(255, 255, 255, 0.5);
+  --jp-ui-inverse-font-color0: rgba(255, 255, 255, 1);
+  --jp-ui-inverse-font-color1: rgba(255, 255, 255, 1);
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
This looks good to me!

SylvainCorlay:
Thanks! I am really happy about this.

jasongrout:
Should this be backported to the 7.x branch?

jasongrout:
I think it should. This change was in JupyterLab from a long time ago, so it doesn't work in jlab 1, so I think it counts as a bugfix.

PR #2807, opened 2020-03-05
Add jupyter_widgets

martinRenou:
As discussed in https://github.com/jupyter-widgets/ipywidgets/issues/2806

This extension installs the nbextension as well as the labextension as static assets


jasongrout:
Let's discuss this a bit further, perhaps over a video call? I raised some concerns about how the JLab 2.0 transition would have gone had we had this package in place over on the issue.

SylvainCorlay:
Hey @martinRenou we were discussing this with @jasongrout: here are somethings we would like to do

 - actually name the front-end package `jupyter_widgets` .
 - use `extra_requires` to make `jupyter_widgets` depend on (specific version ranges) notebook and jupyterlab, so that one can do `pip install jupyter_widgets[jupyterlab]` and be guaranteed that the jupyterlab version that is installed is compatible with the jupyter_widgets version installed. When making a conda package, we can use the `run_constraints`  to provide bounds to the JupyterLab version...
 - rebase on master

martinRenou:
Thanks a lot for looking again at this. I will do all this.

martinRenou:
@SylvainCorlay I took your comments into account

jasongrout:
I'm preparing for the JupyterLab 3.0 release, where JLab extensions are now encouraged to be distributed as python packages as well as npm packages. Revisiting this, I'm curious - why are we trying to combine the classic notebook and jupyterlab extensions into one package? Perhaps we should have two different packages (i.e., one for each frontend), and perhaps the jupyter_widgets package depends on them both if we want a single package for each frontend.

The nice thing about having separate packages for each frontend is that they can be independently versioned and upgraded.

So how about making a jupyterlab_widgets package, and perhaps leaving the widgetsnbextension package like it is (perhaps breaking the notebook dependency like we've discussed). Then use the pip constrains keyword, especially since pip is moving forward with their new dependency solver.

maartenbreddels:
> and perhaps the jupyter_widgets package depends on them both if we want a single package for each frontend.

I think that's a good idea. We can always split it up later in jupyterlab_widgets and jupyter_notebook_widgets if we feel we gain value by having the versions upgraded independently. I think short term this is already a huge improvement, and we can always do the split later without breaking anything.

jasongrout:
What I'm proposing right now is a new `jupyterlab_widgets` python package for widgets 7.x, which basically just wraps the jlab 3.0 widget manager. This leaves the widgetsnbextension package alone (on the 7.x branch), or maybe there's a minor upgrade to widgetsnbextension to accommodate the bump in base (in 7.x) for the jlab services upgrade.

maartenbreddels:
I think it's a lot of work releasing that (without tooling), so I suggest we postpone that full/better plan to the future when we need it.

jasongrout:
> I think it's a lot of work releasing that (without tooling), so I suggest we postpone that full/better plan to the future when we need it.

I'm almost done?

I'm basing it on the recently updated jlab extension cookiecutter, so it really hasn't been that bad...

maartenbreddels:
Grea!. I was referring to releasing all those packages to pypi and conda-forge is a lot of work. But if you are working on it, and almost done, I'm not gonna say no to it 😄

jasongrout:
> I'm almost done?
> 
> I'm basing it on the recently updated jlab extension cookiecutter, so it really hasn't been that bad...

PR at https://github.com/jupyter-widgets/ipywidgets/pull/2973 for the 7.x branch

martinRenou:
I guess we can close this then! Thanks for your work @jasongrout

PR #2816, opened 2020-03-09
Change fontawesome support to work for fontawesome 4 and 5.

### LINKED ISSUE #2794: Forward-port fontawesome 5 changes
We should forward-port #2793 to master.

jasongrout:
JupyterLab and notebook are upgrading to fontawesome 5 (with the v4 backwards compatibility shim). However, in some places, our CSS is very specific to version 4. This commit modifies these places to use syntax that should for both fontawesome 4 and 5. It involves some DOM structure changes, but that is a private implementation detail, so is backwards compatible with our public API.

This is a forward-port of https://github.com/jupyter-widgets/ipywidgets/pull/2793 and 5941cba85fb0959572d59af5301b34d89e18e653

Fixes #2794

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -391,17 +391,17 @@
   font-size: var(--jp-widgets-font-size);
 }
 
-.widget-valid i:before {
+.widget-valid i {
   line-height: var(--jp-widgets-inline-height);
   margin-right: var(--jp-widgets-inline-margin);
   margin-left: var(--jp-widgets-inline-margin);
 }
 
-.widget-valid.mod-valid i:before {
+.widget-valid.mod-valid i {
   color: green;
 }
 
-.widget-valid.mod-invalid i:before {
+.widget-valid.mod-invalid i {
   color: red;
 }
 
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -395,23 +395,13 @@
   line-height: var(--jp-widgets-inline-height);
   margin-right: var(--jp-widgets-inline-margin);
   margin-left: var(--jp-widgets-inline-margin);
-
-  /* from the fa class in FontAwesome: https://github.com/FortAwesome/Font-Awesome/blob/49100c7c3a7b58d50baa71efef11af41a66b03d3/css/font-awesome.css#L14 */
-  display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome;
-  font-size: inherit;
-  text-rendering: auto;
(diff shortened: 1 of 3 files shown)


packages/controls/src/widget_bool.ts
```diff
@@ -301,13 +301,14 @@ export class ValidView extends DescriptionView {
   /**
    * Called when view is rendered.
    */
-  render(): void {
+  render() {
```
vidartf: Do you need a rebase?

vidartf: xref https://github.com/jupyter-widgets/ipywidgets/pull/2808

jasongrout: Well, the merge would have taken care of it, presumably, but I rebased nonetheless.

jasongrout: Well, except it wouldn't take care of an explicit change... I'll fix it in a new commit.

jasongrout: This should be resolved now.

packages/controls/css/widgets-base.css
```diff
@@ -395,23 +395,13 @@
...
-  display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome;
-  font-size: inherit;
-  text-rendering: auto;
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
 }
 
 .widget-valid.mod-valid i:before {
-  content: '\f00c';
```
vidartf: Should this selector be for the icon instead? Same below.

vidartf: I guess it will still work since it is a child though.

vidartf: Maybe just remove the `:before` ?

jasongrout: It needs to be before. That is where font awesome is adding the content via classes:
Screen Shot 2020-03-17 at 9 02 24 AM

To be sure, here it is working with this PR:

Screen Shot 2020-03-17 at 9 02 36 AM

jasongrout: Actually, the fontawesome examples don't use :before, they inherit from the surrounding elements. It's simpler without the :before, so I removed it. Thanks!

vidartf:
> It involves some DOM structure changes, but that is a private implementation detail, so is backwards compatible with our public API.

Were you planning on backporting, or was this more FYI/changelog info?

jasongrout:
I think it is fine to not backport the `:before` change above.

PR #2817, opened 2020-03-09
Update for jlab 2.0.1 final

jasongrout:
This prepares us for an 8.0 alpha release that supports the jlab 2.0.1 final release.


jasongrout:
CC @jupyter-widgets/jupyter-widgets for comment/review

vidartf:
Looks good to me. For reference, are the lumino dependency updates "opportunistic" (i.e. latest version published), or the minimal version required? Either should probably be ok for a major version update, just want to stay on top of it.

jasongrout:
Right now they are opportunistic. There were issues pulling in the jupyterlab 2.0.1 packages and leaving the Lumino versions as-is, since jlab requires more recent lumino versions (it wouldn't compile because some interfaces were missing the icon things added recently).

PR #2824, opened 2020-03-18
Documentation Fixes

umerhasan17:
Does not have an associated issue. These are just some spelling errors I found while reading through the docs.


docs/source/migration_guides.md
```diff
@@ -7,7 +7,7 @@ widgets.
 Migrating from 6.0 to 7.0
 -------------------------
 
-For example migrations, see these PRs:
+For example, migrations, see these PRs:
```
jasongrout: I think this was okay. It's saying that example migrations are listed below, i.e., it was saying for (example migrations), not (for example) migrations.

jasongrout:
Thanks! This is a huge help for readability. I noted in one comment a change that I think should be reverted or rephrased.

PR #2827, opened 2020-03-19
Fix the logic for tab selection when the children changes to select a tab when possible

### LINKED ISSUE #2818: No tab selected by default
In ipywidgets 8.0.0a0, when creating a tab widget, by default no tab is activated, which is really confusing. I think the first tab should be selected.

```python
tab_nest = widgets.Tab()
tab_nest.children = [accordion, accordion]
tab_nest.titles = ('An accordion', 'Copy of the accordion')
tab_nest
```

Screen Shot 2020-03-10 at 8 20 42 AM

jasongrout:
Fixes #2818

### DIFF
--- a/ipywidgets/widgets/widget_selectioncontainer.py
+++ b/ipywidgets/widgets/widget_selectioncontainer.py
@@ -52,6 +52,22 @@ class Tab(_SelectionContainer):
             kwargs['selected_index'] = 0
         super(Tab, self).__init__(**kwargs)
 
+    @observe('children')
+    def _observe_children(self, change):
+        # if there are no tabs, then none should be selected
+        if len(change.new) == 0:
+            self.selected_index = None
+
+        # if there are tabs, but none is selected, select the first one
+        elif self.selected_index == None:
+            self.selected_index = 0
+
+        # if there are tabs and a selection, but the selection is no longer
+        # valid, select the last tab.
+        elif len(change.new) < self.selected_index:
+            self.selected_index = len(change.new) - 1
+
+
 @register
 class Stacked(_SelectionContainer):
     """Displays only the selected child."""


jasongrout:
CC @vidartf

vidartf:
I guess the interesting question is what happens for code like this:
```python
tab = widgets.Tab()
with tab.hold_trait_notifications():
    tab.children = [foo, bar]
    tab.selected_index = 1  # default to bar
```

vidartf:
And the answer to that is nothing you wouldn't expect 👍

vidartf:
I went a little fast I think, there's still this scenario:

```python
# Init in one go to avoid UI flicker:
tab = widgets.Tab(children = [foo, bar, bat], selected_index=2)

<something happens here>

# Hold notifications to avoid flicker during change:
with tab.hold_trait_notifications():
    tab.children = [foo, bar]
    tab.selected_index = 1  # Select bar
```

vidartf:
For that case, I think `tab.selected_index == 0` ends up being the result.

vidartf:
Of course, the work-around for that scenario is to use `tab.hold_sync` instead of `tab.hold_trait_notifications`, but there might be other scenarios where `hold_trait_notifications` is the only alternative.

jasongrout:
> ```python
> # Init in one go to avoid UI flicker:
> tab = widgets.Tab(children = [foo, bar, bat], selected_index=2)
> 
> <something happens here>
> 
> # Hold notifications to avoid flicker during change:
> with tab.hold_trait_notifications():
>     tab.children = [foo, bar]
>     tab.selected_index = 0  # Select foo
> ```

How is that a problem? What I would think would happen is that the children gets set to a 2-element list, and the index gets set to 0, then the observe handler would run, and would see that the selected index is valid and would not change it. How is this getting set to 1?

jasongrout:
> What I would think would happen is that the children gets set to a 2-element list, and the index gets set to 0, then the observe handler would run, and would see that the selected index is valid and would not change it. How is this getting set to 1?

Indeed, when I test it, I get selected index 0, as expected (conda-forge python 3.8.2)

vidartf:
> What I would think would happen is that the children gets set to a 2-element list, and the index gets set to 0, then the observe handler would run, and would see that the selected index is valid and would not change it. 

You are right, my brain gets tangled up over validators.

jasongrout:
> You are right, my brain gets tangled up over validators.

Mine too. Of course, this is *not* a validator, so it's a little easier to reason about, in that it happens after all validation. Should it be a cross validator instead, though?

vidartf:
> Mine too. Of course, this is *not* a validator, so it's a little easier to reason about, in that it happens after all validation. Should it be a cross validator instead, though?

Possibly 🤷‍♂

PR #2829, opened 2020-03-21
titles were not being assigned

afonit:
I modified the line to assign titles as the other way left them blank.


jasongrout:
Currently in master, you set titles by doing `w.title = ...`, not with `.set_title()` anymore. If you are using widgets 7.x, you need to look at the documentation for the 7.x release: https://github.com/jupyter-widgets/ipywidgets/blob/7.x/docs/source/examples/Widget%20List.ipynb?short_path=b86a60c#L1107

afonit:
Ah - sorry about the noise, I did not realize that when I was reading the documentation it defaulted to latest and not stable. (at least from the search I had done which brought up that page).

PR #2830, opened 2020-03-22
Remove old items from MANIFEST.in and update content of setup.py

SylvainCorlay:


SylvainCorlay:
This is removing non-existing files from the manifest, and replacing unused email with the same boilderplate as other Jupyter packages. Merging.

PR #2831, opened 2020-03-22
WIP - Include example notebooks as data files in package

SylvainCorlay:
Still fixing the docs build.


vidartf:
The issues states what it does, but not really why. Care to help me connect the dots? :)

docs/Makefile
```diff
@@ -29,8 +29,10 @@ help:
 
 clean:
 	rm -rf $(BUILDDIR)/*
+	rm -rf ./source/notebooks
 
 html:
+	rm -rf ./source/notebooks/ && mkdir ./source/notebooks && cp -r ./../share/jupyter/notebook_templates/ipywidgets/*  ./source/notebooks/
```
SylvainCorlay: I am not very happy that we need to refer to the current (makefile) directory as `.`. There is no variable storing the location of the makefile.

vidartf:
Note that this would likely also need a change to the `make.bat` file corresponding to the docs Makefile changes. An alternative could be to use https://github.com/vidartf/nbsphinx-link :)

SylvainCorlay:
> Note that this would likely also need a change to the make.bat file corresponding to the docs Makefile changes.

Yes

Issue #2832, opened 2020-03-23
Freeze Range in SelectionRangeSlider


agregory96:
When using the SelectionRangeSlider, it would be nice to have a capability similar to Plotly's where you can click between the two current points set on the range slider and drag the bar to scroll over a that range. 

An example would be if the SelectionRangeSlider is over datetimes with a tick for each hour over multiple days. If you set the first point at 3pm on the first day and the second point at 3am on the second day, it would be nice to be able to click the slider between those two points and drag the selected range across the slider to see updates in 12-hour chunks (4pm-4am, 5pm -5am, etc.) as opposed to having to manually re-set the start and end points.

I realize this is really only applicable to functions sliding over time, but would still like to suggest it as a functionality. Alternatively, if anyone knows a workaround using click events or something similar, please let me know!

jasongrout:
We don't have this turned on, but it is supported by the underlying slider: https://refreshless.com/nouislider/behaviour-option/

Perhaps we should either expose this property in the slider widgets so the user can select whether they want this behavior, or perhaps we should just turn on "drag-tap" behavior by default? @ibdafna, what do you think?

jasongrout:
To make this the default behavior, we would just need to add a `behavior` option when we create the sliders at https://github.com/jupyter-widgets/ipywidgets/blob/b91a6d9844d1e4e2571f1e7d2fa5416abf468106/packages/controls/src/widget_int.ts#L206-L221 and https://github.com/jupyter-widgets/ipywidgets/blob/b91a6d9844d1e4e2571f1e7d2fa5416abf468106/packages/controls/src/widget_float.ts#L150-L164

ibdafna:
> We don't have this turned on, but it is supported by the underlying slider: https://refreshless.com/nouislider/behaviour-option/
> 
> Perhaps we should either expose this property in the slider widgets so the user can select whether they want this behavior, or perhaps we should just turn on "drag-tap" behavior by default? @ibdafna, what do you think?

@jasongrout Indeed, that functionality is available, but not exposed. I am happy to have "drag-tap" available by default if you think it's reasonable. Will add a PR shortly.

jasongrout:
> I am happy to have "drag-tap" available by default if you think it's reasonable. Will add a PR shortly.

I think that would be great! Like the original poster, I too have missed having this feature. Thanks for a PR!

PR #2683, opened 2020-01-07
Update dependencies to JupyterLab 2.0b2 packages

jasongrout:
This updates the lumino and jlab dependencies, and fixes one outstanding error we had in requesting comm info.

PR #2834, opened 2020-03-24
Add dragging behaviour properties to sliders

ibdafna:
This PR adds functionality requested in #2832 . The drag behaviour API has been exposed on the python side and now allows four selections:

1. Tap (current default behaviour)
2. Snap (new option)
3. Drag-tap (new default behaviour for all sliders - no change in behaviour for non-range sliders)
4. Drag-snap

There are no breaking API changes, but a new `behaviour` attribute has been added to the python side.

devw_drag_behaviour1

### DIFF
--- a/ipywidgets/widgets/widget_float.py
+++ b/ipywidgets/widgets/widget_float.py
@@ -159,8 +159,8 @@ class FloatSlider(_BoundedFloat):
         default is 'horizontal', orientation of the slider
     readout : {True, False}
         default is True, display the current value of the slider next to it
-    behaviour : str
-        slider handle and connector dragging behaviour. Default is 'drag-tap'.
+    behavior : str
+        slider handle and connector dragging behavior. Default is 'drag-tap'.
     readout_format : str
         default is '.2f', specifier for the format function used to represent
         slider value for human consumption, modeled after Python 3's format
@@ -177,8 +177,8 @@ class FloatSlider(_BoundedFloat):
     continuous_update = Bool(True, help="Update the value of the widget as the user is holding the slider.").tag(sync=True)
     disabled = Bool(False, help="Enable or disable user changes").tag(sync=True)
     style = InstanceDict(SliderStyle).tag(sync=True, **widget_serialization)
-    behaviour = CaselessStrEnum(values=['drag-tap', 'drag-snap', 'tap', 'drag', 'snap'],
-        default_value='drag-tap', help="Slider dragging behaviour.").tag(sync=True)
+    behavior = CaselessStrEnum(values=['drag-tap', 'drag-snap', 'tap', 'drag', 'snap'],
+        default_value='drag-tap', help="Slider dragging behavior.").tag(sync=True)
 
 
 @register
@@ -203,8 +203,8 @@ class FloatLogSlider(_BoundedLogFloat):
         default is 'horizontal', orientation of the slider
     readout : {True, False}
         default is True, display the current value of the slider next to it
-    behaviour : str
-        slider handle and connector dragging behaviour. Default is 'drag-tap'.
+    behavior : str
+        slider handle and connector dragging behavior. Default is 'drag-tap'.
     readout_format : str
         default is '.3g', specifier for the format function used to represent
         slider value for human consumption, modeled after Python 3's format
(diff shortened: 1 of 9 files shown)


jasongrout:
> This PR adds functionality requested in #2683 .

#2832, right?

I was thinking it might be better to start out with just changing the default behavior to `'drag-tap'`, which addresses the user request, and not expose changing it to the users just yet. The tradeoff here is that increasing API surface means we ought to support it in the future (even if we change slider), so it becomes a maintenance burden. On the other hand, it might be useful enough to users to have a tap/snap/drag-snap behavior available. Are there usecases people actually see that do *not* want drag-tap behavior?

jasongrout:
If exposing this behavior to the users is the way we want to go, then I think you've done a great job here (just reading the code, I haven't run it yet)

ibdafna:
> > This PR adds functionality requested in #2683 .
> 
> #2832, right?

Indeed. Updated the PR note above with the correct one, thanks for spotting!

> I was thinking it might be better to start out with just changing the default behavior to `'drag-tap'`, which addresses the user request, and not expose changing it to the users just yet. The tradeoff here is that increasing API surface means we ought to support it in the future (even if we change slider), so it becomes a maintenance burden. On the other hand, it might be useful enough to users to have a tap/snap/drag-snap behavior available. Are there usecases people actually see that do _not_ want drag-tap behavior?

Potentially yes. Off the top of my head, let's say I'm using the range slider to control the spot range of a vanilla option payoff at expiry. I would probably not want to drag both handles of the slider at the same time given I am only looking to focus on the strike area. But then again, it might be convenient to drag the full range if I got the original parameters wrong and need to shift my window left or right.

> Thoughts? I know I've become very conservative about extending the core ipywidgets api because it aims to be very stable over a long term, so I'm wary of adding maintenance burdens.

I'm definitely with you on robustness. I added the snap behaviour because I often expected the slider handles to behave that way and thought other users might also appreciate that option. If you think we are incurring potential technical debt down the line then I'm happy to remove the changes!

vidartf:
> If you think we are incurring potential technical debt down the line then I'm happy to remove the changes!

We could also consider to "expose" it as an underscore param `_behavior` ("use at own risk"). 🤷‍♂

agregory96:
> > This PR adds functionality requested in #2683 .
> 
> #2832, right?
> 
> I was thinking it might be better to start out with just changing the default behavior to `'drag-tap'`, which addresses the user request, and not expose changing it to the users just yet. The tradeoff here is that increasing API surface means we ought to support it in the future (even if we change slider), so it becomes a maintenance burden. On the other hand, it might be useful enough to users to have a tap/snap/drag-snap behavior available. Are there usecases people actually see that do _not_ want drag-tap behavior?
> 
> Thoughts? I know I've become very conservative about extending the core ipywidgets api because it aims to be very stable over a long term, so I'm wary of adding maintenance burdens.

This was precisely what I was looking for, thanks much to everyone for the quick response, review, and implementation!

ibdafna:
@jasongrout How should we proceed?

jtpio:
For folks wanting to try this PR on Binder:

https://mybinder.org/v2/gh/ibdafna/ipywidgets/drag_behaviour?urlpath=lab/tree/docs/source/examples/

ipywidgets/widgets/widget_float.py
```diff
@@ -159,6 +159,8 @@ class FloatSlider(_BoundedFloat):
         default is 'horizontal', orientation of the slider
     readout : {True, False}
         default is True, display the current value of the slider next to it
+    behaviour : str
```
jtpio, line 162: Commenting here as this came up during the weekly meeting today, with @ianhi spotting this attribute was not in American English:

```suggestion
    behavior : str
```

jtpio: Ah that was probably because of `noUISlider` using `behaviour` in the options:

https://refreshless.com/nouislider/slider-options/#section-limit

image

jtpio:
FYI @ibdafna I rebased and added a commit to rename `behaviour` to `behavior` (American English)

ibdafna:
@jtpio thank you 😄

PR #2897, opened 2020-06-05
Changing == to self.equals in _propagate_value

kefirbandi:
Hi,

it seems to me that in order to be consistent, we should not directly use '==' here, but self.equals.
At least this is how I got my code working.

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -254,7 +254,7 @@ class _Selection(DescriptionWidget, ValueWidget, CoreWidget):
     def _propagate_label(self, change):
         if change.new is None:
             index = None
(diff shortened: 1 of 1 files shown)


vidartf:
As the labels are supposed to be strings, I don't see what isn't working. Would you mind sharing a minimal example of code that doesn't currently work, but that this PR would fix?

kefirbandi:
You're right, I messed up the code lines, now submitted a new version.
What I'm trying to fix is this code sample:
```
from ipywidgets.widgets import Dropdown
import pandas as pd
df1 = pd.DataFrame()
df2 = pd.DataFrame()
options = [['A',df1],['B',df2]]
dd=Dropdown(options=options, equals=lambda x,y : x is y)
```

vidartf (approved):

PR #2903, opened 2020-06-17
Add SingleOutput Widget

Disatiler:


nthiery:
Hi @jasongrout! This is a sequel of the discussion we had in Paris last January

zerline:
Great! I would add this code to `widget_output.py` instead of creating a new `widget_singleoutput.py` file.

Disatiler:
Hi everyone!

Small change in the SingleOutput widget. We include the class in Output widget file (same thing with its test).

Its docs is also in Output doc file "Output Widget.ipynb".

Have a nice day!

PR #2938, opened 2020-08-03
create issue templates

ianhi:
Closes: https://github.com/jupyter-widgets/ipywidgets/issues/2929

I created three issue templates that would correspond to:
1. Bug reports
2. Feature Requests
3. Custom Widget issues

with the intention of 1 and 2 being used by users of ipywidgets and 3 being a catchall for anyone with questions or issues for developing a custom widget. I wasn't totally sure what to put into the template for 3 so I just left some comments with very general guidelines. @cabreraalex you've opened a few issues of the type I'm envisioning for 3. Do you have any thoughts on what would have been most helpful for you in an issue template?

### DIFF
--- a/.github/ISSUE_TEMPLATE/custom_widget_issue.md
+++ b/.github/ISSUE_TEMPLATE/custom_widget_issue.md
@@ -1,0 +1,15 @@
+---
+name: Custom Widget
+about: Issues or questions related to developing a custom widget
+labels: type:custom-widget
+---
+
+<!-- 
+If this is a:
+
+Bug Report: Please add the relevant pieces of information from the bug report template to this issue  (https://github.com/jupyterlab/jupyterlab/issues/new?template=bug_report.md)
+
+Question: Great! Thanks for asking someone else will be able to find the answer more easily in the future. Please try to provide as much detail as you can about what you would like to accomplish and what you have tried.
+
+Feature Request: Please follow the template from https://github.com/jupyterlab/jupyterlab/issues/new?template=feature_request.md 
+-->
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/custom_widget_issue.md
```diff
@@ -0,0 +1,15 @@
...
+about: Issues or questions related to developing a custom widget
+labels: type:custom-widget
+---
+
+<!-- 
+If this is a:
+
+Bug Report: Please add the relevant pieces of information from the bug report template to this issue  (https://github.com/jupyterlab/jupyterlab/issues/new?template=bug_report.md)
+
+Question: Great! Thanks for asking someone else will be able to find the answer more easily in the future. Please try to provide as much detail as you can about what you would like to accomplish and what you have tried.
```
jasongrout, line 12:
```suggestion
Question: Great! Thanks for asking here, since someone else will be able to find the answer more easily in the future. Please try to provide as much detail as you can about what you would like to accomplish and what you have tried.
```

jasongrout:
One grammatical suggestion, otherwise these look good. Thanks!

ianhi:
Done. Some of these reference issue labels that I don't think exist yet. Do these need to be manually created, or is having them in issue template sufficient?

ianhi:
It may also be interesting to explore other issue templates. e.g. sklearn also has options for documentation improvement, and usage questions:
https://github.com/scikit-learn/scikit-learn/tree/master/.github/ISSUE_TEMPLATE

happy to include those here if that seems useful, or adding them in a subsequent pr.

cabreraalex:
LGTM! I feel like the custom widget questions/requests are too broad to have a well-defined template, especially given the current docs and API.

vidartf:
Thanks! Let's try these as-is, and then iterate if there are any problems 😃

PR #2939, opened 2020-08-03
Add comments to the lock bot

ianhi:
The stated goal of the lock bot on this repo is ([ref](https://github.com/jupyter-widgets/ipywidgets/pull/2876#issue-417536199))

> This encourages people to open a new issue with full details of their situation rather than resurrecting one that has been closed and resolved.

This makes that explicit to a user encountering a locked thread.

@jasongrout you originally added the lock, do you oppose this?

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -14,12 +14,11 @@ exemptLabels: []
 lockLabel: resolved-locked
 
 # Comment to post before locking. Set to `false` to disable
-lockComment: false
 
-# lockComment: >
-#   This thread has been automatically locked since there has not been
-#   any recent activity after it was closed. Please open a [new issue](https://github.com/jupyterlab/jupyterlab/issues/new/choose)
-#   for related discussion.
+lockComment: >
+   This thread has been automatically locked since there has not been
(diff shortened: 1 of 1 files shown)


jasongrout:
While theoretically having a message is great, what we found in JupyterLab is that the messages triggered notifications to all the people watching the repo (including the core developers), overwhelming our inboxes with messages that did not require any attention (ironically, were only there to say that no attention should be given). So we opted to turn off the messages, relying instead on the label description to convey this information, as well as documentation about the contribution process and the reason for the lock.

If there was a way to have such a message posted without notifying anyone watching the repos, i.e., a silent message of some type, that would be ideal.

jasongrout:
(Of course, I just realized I could filter these messages in my email client, which might solve the issue there, but wouldn't solve the issue with other systems for managing GitHub issues.)

If others feel strongly about enabling this message, I'm happy to try it again.

ianhi:
> overwhelming our inboxes with messages that did not require any attention (ironically, were only there to say that no attention should be given)

Oh huh that is not great :/ - I think you made the right call on jupyterlab as the benefit to that message probably is not so huge. I'll close this

PR #2954, opened 2020-08-28
modify outputs to use comm

### LINKED ISSUE #2953: outputs widget contextmanager fails on metakernel
This is an issue found in  Calysto/metakernel#217

The context manager in the outputs widget requires a call to get_ipython and therefore fails on non-ipython kernels.  The other mecchanisms for access outputs widget appear to work,

joequant:
This modifies the outputs widget to use comm rather than calling
ipython directly.  This fixes #2953 and Calysto/metakernel#217


jasongrout:
Thanks!

Adding to the queue for review...

ipywidgets/widgets/widget_output.py
```diff
@@ -108,23 +108,39 @@ def __enter__(self):
...
-        ip = get_ipython()
+        kernel = None
         if etype is not None:
+            ip = get_ipython()
             if ip:
+                kernel = ip
                 ip.showtraceback((etype, evalue, tb), tb_offset=0)
+            elif self.comm is not None and self.comm.kernel is not None:
+                kernel = self.comm.kernel
+                kernel.send_response(kernel.iopub_socket,
```
martinRenou, line 130: Not sure using those kernel APIs is a good idea.

I wonder if simply re-raising the error would work? And let the kernel catch it and publish error traceback. If this doesn't work, I don't know a better solution than the current PR.

jasongrout: The point behind this code is to publish the error message to the output widget (i.e., before `self.msg_id` is reset below). If we just raised the error, we wouldn't be trapping it for the output widget.

It's debatable whether trapping and showing the error in the output widget is better than raising it to the enclosing scope, but that would also be a change in behavior, and I think this PR is trying to be backwards compatible with the behavior.

jasongrout:
I resolved the conflict with master (this conflicted with https://github.com/jupyter-widgets/ipywidgets/pull/3199).

@martinRenou or @ibdafna (or anyone else) - can you review?

ibdafna:
Will review in the morning!

ibdafna (approved):
LGTM

jasongrout:
Itay, feel free to merge if you are comfortable with the code.

jtpio:
Looks like the test failure was not relevant, and should be fixed in #3227

PR #2960, opened 2020-09-02
Feat: ErrorWidget as fallback when widgets models or views fail.

maartenbreddels:
This makes sure that we always render something, with a descriptive
error message. This makes it easier for users to diagnose or report
issues.

JupyterLab screenshot:

image

Note that in all cases, the Button and Hbox are rendered.

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -2,6 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import * as services from '@jupyterlab/services';
+import * as widgets from '@jupyter-widgets/base';
 
 import { JSONObject, PartialJSONObject } from '@lumino/coreutils';
 
@@ -90,7 +91,7 @@ export abstract class ManagerBase implements IWidgetManager {
     const viewPromise = (model.state_change = model.state_change.then(
       async () => {
         try {
-          const ViewType = (await this.loadClass(
+          const ViewType = (await this.loadViewClass(
             model.get('_view_name'),
             model.get('_view_module'),
             model.get('_view_module_version')
@@ -295,11 +296,11 @@ export abstract class ManagerBase implements IWidgetManager {
     serialized_state: any = {}
   ): Promise<WidgetModel> {
     const model_id = options.model_id;
-    const model_promise = this.loadClass(
+    const model_promise = this.loadModelClass(
       options.model_name,
       options.model_module,
       options.model_module_version
-    ) as Promise<typeof WidgetModel>;
+    );
     let ModelType: typeof WidgetModel;
     try {
       ModelType = await model_promise;
@@ -475,6 +476,42 @@ export abstract class ManagerBase implements IWidgetManager {
     moduleName: string,
     moduleVersion: string
   ): Promise<typeof WidgetModel | typeof WidgetView>;
+
+  protected async loadModelClass(
+    className: string,
+    moduleName: string,
+    moduleVersion: string
+  ): Promise<typeof WidgetModel> {
+    try {
+      const promise: Promise<typeof WidgetModel> = this.loadClass(
+        className,
+        moduleName,
+        moduleVersion
+      ) as Promise<typeof WidgetModel>;
+      await promise;
+      return promise;
+    } catch (error) {
+      return widgets.createErrorWidget(error);
+    }
+  }
+
+  protected async loadViewClass(
+    className: string,
+    moduleName: string,
+    moduleVersion: string
+  ): Promise<typeof WidgetView> {
+    try {
+      const promise: Promise<typeof WidgetView> = this.loadClass(
+        className,
+        moduleName,
+        moduleVersion
+      ) as Promise<typeof WidgetView>;
+      await promise;
+      return promise;
+    } catch (error) {
+      return widgets.createErrorWidgetView(error);
+    }
+  }
 
   /**
    * Create a comm which can be used for communication for a widget.
--- a/packages/base/src/errorwidget.ts
+++ b/packages/base/src/errorwidget.ts
@@ -1,0 +1,52 @@
+import {
+  WidgetModel,
+  DOMWidgetModel,
+  DOMWidgetView,
+  WidgetView
+} from './widget';
+import { JUPYTER_WIDGETS_VERSION } from './version';
+
+// create a Widget Model that captures an error object
+export function createErrorWidget(error: Error): typeof WidgetModel {
+  class ErrorWidget extends DOMWidgetModel {
+    constructor(attributes: any, options: any) {
+      attributes = {
+        ...attributes,
+        _view_name: 'ErrorWidgetView',
+        _view_module: '@jupyter-widgets/base',
+        _model_module_version: JUPYTER_WIDGETS_VERSION,
+        _view_module_version: JUPYTER_WIDGETS_VERSION,
+        failed_module: attributes._model_module,
+        failed_model_name: attributes._model_name,
+        error: error
+      };
+      super(attributes, options);
+      console.log(attributes);
+    }
+  }
+  return ErrorWidget;
+}
+
+export function createErrorWidgetView(error?: Error): typeof WidgetView {
+  return class ErrorWidgetView extends DOMWidgetView {
+    render() {
+      console.log('render');
+      const module = this.model.get('_model_module');
+      const name = this.model.get('_model_name');
+      const viewModule = this.model.get('_view_module');
+      const viewName = this.model.get('_view_name');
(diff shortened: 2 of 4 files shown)


maartenbreddels:
This replaces https://github.com/voila-dashboards/voila/pull/667 but is a better version because we can distinguish between loading models and views.

packages/base-manager/src/manager-base.ts
```diff
@@ -476,6 +477,42 @@ export abstract class ManagerBase implements IWidgetManager {
...
+        className,
+        moduleName,
+        moduleVersion
+      ) as Promise<typeof WidgetView>;
+      await promise;
+      return promise;
+    } catch (error) {
+      return widgets.createErrorWidgetView(error);
+    }
+  }
```
martinRenou, line 514: `loadModelClass` and `loadViewClass` are really alike, would it make sense to create only one template function `loadClass<T extends WidgetModel | WidgetView>` (not sure about the TypeScript notation here)?

maartenbreddels: DRY vs KISS, for 2 repeats with low complexity KISS wins :)
Feel free to disagree though.

martinRenou: I wouldn't say `loadClass<T>` is more complex, but that's a personal opinion. I don't mind the small repetition.

maartenbreddels: Ah maybe I misunderstood you. The point is that they are different (1 returns a View, the other a Model). I don't see how we can put this in 1 method (like we had before).

martinRenou: Ah I missed there was a `loadClass` already, yeah never mind. I should not treat TypeScript like it's C++ :P I've got assumption about the language that might not be correct.

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,52 @@
...
+        _view_name: 'ErrorWidgetView',
+        _view_module: '@jupyter-widgets/base',
+        _model_module_version: JUPYTER_WIDGETS_VERSION,
+        _view_module_version: JUPYTER_WIDGETS_VERSION,
+        failed_module: attributes._model_module,
+        failed_model_name: attributes._model_name,
+        error: error
+      };
+      super(attributes, options);
+      console.log(attributes);
```
martinRenou, line 24: Trailing console.log

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+  return ErrorWidget;
+}
+
+export function createErrorWidgetView(error?: Error): typeof WidgetView {
+  return class ErrorWidgetView extends DOMWidgetView {
+    render() {
+      console.log('render');
```
martinRenou, line 33: here as well :)

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,52 @@
...
+        error: error
+      };
+      super(attributes, options);
+      console.log(attributes);
+    }
+  }
+  return ErrorWidget;
+}
+
+export function createErrorWidgetView(error?: Error): typeof WidgetView {
```
martinRenou, line 30: Do we actually need a WidgetView implementation? It's unlikely to be displayed on the page, or is it?

maartenbreddels: It is, that's what you see in the screenshot. When the widget fails, you will see it on the page.

vidartf:
So previously we used to use the various other mimetypes in the bundle to present an error message (but we stopped doing that). Would you mind mentioning why this is better?

martinRenou:
> So previously we used to use the various other mimetypes in the bundle to present an error message

I am really interested in this. Was it really what was happening? I feel like the widget mimetype was always rendered, only that this mimetype was showing `Model not found` or similar errors when the widget manager fails to render.
It would be really nice in `ipympl` to have a fallback image mimetype when the widget manager fails to render, but it does not seem to work.

vidartf:
Context in https://github.com/jupyter-widgets/ipywidgets/pull/2007. We seemed to agree then that the best long term solution was "_a short HTML entry with a link to an entry in the docs describing the possible causes of seeing the message"._ Doing something smarter on the JS side might also be a good thing, but I think there is also value in having a shared/similar way of doing things for various different errors that can occur. I.e. I don't think this will solve the error message if the user forgot to install the jupyter widget manager (or the install is broken). tl;dr: not opposed, but if another solution solves more problems, we should at least compare them.

vidartf:
Ugh, it seems the notebook issue that was preventing us from using html mimetype is still open: https://github.com/jupyter/notebook/issues/2980

maartenbreddels:
> Would you mind mentioning why this is better?

The error in this PR give more information because they are aware of the issue (missing Model, missing View). The mimetype fallback msg will only give a general error msg, and not a stacktrace.

The errors given now in Jupyter Lab are:
image

and classical:
image

SylvainCorlay:
I would like to get this in for 8.0 if possible.

packages/base-manager/src/manager-base.ts
```diff
@@ -476,6 +477,42 @@ export abstract class ManagerBase implements IWidgetManager {
...
+    moduleVersion: string
+  ): Promise<typeof WidgetModel> {
+    try {
+      const promise: Promise<typeof WidgetModel> = this.loadClass(
+        className,
+        moduleName,
+        moduleVersion
+      ) as Promise<typeof WidgetModel>;
+      await promise;
+      return promise;
```
SylvainCorlay, line 492: Why do you need to await the promise before returning it?

maartenbreddels: to catch a possible exception

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,50 @@
...
+      const viewName = this.model.get('_view_name');
+      const errorMessage = String(error || this.model.get('error').stack);
+      this.el.innerHTML = `Failed to create WidgetView '${viewName}' from module '${viewModule}' for WidgetModel '${name}' from module '${module}', error:<pre>${errorMessage}</pre>`;
+    }
+  };
+}
+
+export class ErrorWidgetView extends DOMWidgetView {
+  render() {
+    console.log('render');
```
davidbrochart, line 44: Leftover?

davidbrochart:
@maartenbreddels is it supposed to work with classical notebook? It prints:
```
Failed to load widget 'DOMWidgetModel' from module 'badwidget', error:
Error: Script error for "badwidget"
http://requirejs.org/docs/errors.html#scripterror
    at makeError (http://localhost:8888/static/components/requirejs/require.js?v=d37b48bb2137faa0ab98157e240c084dd5b1b5e74911723aa1d1f04c928c2a03dedf922d049e4815f7e5a369faa2e6b6a1000aae958b7953b5cc60411154f593:168:17)
    at HTMLScriptElement.onScriptError (http://localhost:8888/static/components/requirejs/require.js?v=d37b48bb2137faa0ab98157e240c084dd5b1b5e74911723aa1d1f04c928c2a03dedf922d049e4815f7e5a369faa2e6b6a1000aae958b7953b5cc60411154f593:1735:36)
```

packages/base/src/errorwidget.ts
maartenbreddels: correct

maartenbreddels:
I think I have tested this, is this the js console?

davidbrochart:
No, the cell output.

maartenbreddels:
That should be exactly what it does, or do you refer to the short stack trace?
Looking at https://github.com/voila-dashboards/voila/pull/667 it seems to do the same as what I got.

davidbrochart:
Yes, I was wondering why we don't have the same stack trace [as in jlab](https://github.com/jupyter-widgets/ipywidgets/pull/2960#issue-477782459).

maartenbreddels:
might be due to the transpiling target, I think in the notebook we still use something older (ES5?) and we lose the stack trace, but not 100% sure.

davidbrochart:
OK, thanks.
I tried your PR on jlab3 by rebasing it on #3086, it works fine.

jasongrout:
We're pushing forward to ipywidgets 8. Is there someone that wants to either champion this for 8.0 very soon, or can we move it to 8.1?

jasongrout:
Moving to 8.1 since a champion has not come forth.

trungleduc:
Hello @maartenbreddels, I'm trying to finish this PR. How do you think about this behavior of failback widget?

https://user-images.githubusercontent.com/4451292/138243743-34a49956-6d54-4aab-ba7d-12713a12c762.mp4

PR #2785, opened 2020-02-16
HTML sanitizer for descriptions.

zerline:
An HTML sanitizer for descriptions. On the Python side: that is, we sanitize as early as possible. We may have to think about jslinks, though. Should address issue #2636

PR #2972, opened 2020-09-17
Fix character escapes in combobox options

### LINKED ISSUE #2940: Combobox does not work if options contain double quotes (")
The second combobox doesn't offer any drop-down options:

combobox

Code to reproduce the above:
```python
import ipywidgets
print(ipywidgets.__version__)
display(ipywidgets.Combobox(options=['OptionA','OptionB']))
display(ipywidgets.Combobox(options=['"Option A"','"Option B"']))
```

@vidartf pointed out the issue is here, with html escaping:

https://github.com/jupyter-widgets/ipywidgets/blob/6be18d9b75353f7b4a1c328c6ea06d8959f978f6/packages/controls/src/widget_string.ts#L534-L536

vidartf:
Fixes #2940. Independent of work in #2785 since these are limited to attribute values.

### DIFF
--- a/packages/controls/src/widget_string.ts
+++ b/packages/controls/src/widget_string.ts
@@ -531,10 +531,13 @@ export class ComboboxView extends TextView {
     this.isInitialRender = false;
 
     const opts = this.model.get('options') as string[];
-    const optLines = opts.map(o => {
-      return `<option value="${o}"></option>`;
-    });
-    this.datalist.innerHTML = optLines.join('\n');
+    const optionFragment = document.createDocumentFragment();
+    for (const v of opts) {
+      const o = document.createElement('option');
+      o.value = v;
+      optionFragment.appendChild(o);
+    }
+    this.datalist.appendChild(optionFragment);
   }
 
   isValid(value: string): boolean {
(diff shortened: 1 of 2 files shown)


jasongrout:
Instead of splicing strings together and using innerHTML, what do you think of constructing the option elements, adding them to a document fragment, then adding that at the end? Something like:

```typescript
    const opts = this.model.get('options') as string[];
    const optionFragment = document.createDocumentFragment();
    opts.forEach(v => {
      const o = document.createElement('option');
      o.value = v;
      optionFragment.appendChild(o);
    })
    this.datalist.appendChild(optionFragment);
```

vidartf:
When I originally wrote it, I found `createElement` loops to be less performant for large amounts of options (~10k+), but at the time I don't think I tried using document fragments. I've just ran some checks in chrome with 100k options and profiling and the performance seems to be at least same order of magnitude for the two, so I think your point about the code being much simpler should be the deciding factor here 👍

jasongrout:
To be fair, I think we should be testing up to maybe 100 options. If someone is putting 10k options in, they have more serious problems and should probably be using a different filtering/selection strategy.

jasongrout:
Thanks!

PR #2973, opened 2020-09-17
Update for JupyterLab 3

jasongrout:
This is specific to the 7.x branch. We'd need to make a separate PR to master, I think.

Basically I generated an extension from the jlab 3 extension cookiecutter, then merged it with the existing jlab manager extension.


jasongrout:
Turns out that we already had a [jupyterlab-widgets](https://pypi.org/project/jupyterlab-widgets/) package from a long time ago, so reusing that name was really convenient.

PR #2975, opened 2020-09-18
Bugfix: fix return key submission of IntText widgets

### LINKED ISSUE #2959: BoundedIntText: Pressing ENTER should trigger change event
Pressing `ENTER` after changing value of `BoundedIntText` does not trigger event change. Works for `BoundedFloatText`. To reproduce, run snippet below and try pressing `ENTER` after changing the value. Pressing `Tab` does what's expected.
```
num_rows = widgets.BoundedIntText(min=1, max=2**16, step=1, description='num_rows')
def on_value_change(change):
    print(change['new'])
num_rows.observe(on_value_change, names='value')
num_rows
```

agoose77:
Fixes #2959

### DIFF
--- a/packages/controls/src/widget_int.ts
+++ b/packages/controls/src/widget_int.ts
@@ -600,6 +600,10 @@ export class IntTextView extends DescriptionView {
    */
   handleKeypress(e: KeyboardEvent): void {
     if (/[e,.\s]/.test(String.fromCharCode(e.keyCode))) {
+      // Handle return key submission
+      if (e.keyCode === 13) {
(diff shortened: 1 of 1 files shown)


jasongrout:

vidartf:
In handleKeyUp there is still this: Is that a problem?
```ts
    value = value.replace(/[e,.\s]/g, '');
```

jasongrout:
> In handleKeyUp there is still this: Is that a problem?

I think that is fine. It is just stripping out whitespace from the value, not blocking the keystroke.

PR #2984, opened 2020-09-29
fix: when CDN is used as fallback, no module is returned

maartenbreddels:
git bisect revealed an issue in https://github.com/jupyter-widgets/ipywidgets/pull/2792


SylvainCorlay:
Thanks for trackking this down!

PR #2993, opened 2020-10-05
Add license statements from the bundled js packages

jasongrout:
These license files were hand-copied from each package’s distribution in the node_modules directory. Even better would be using something like https://www.npmjs.com/package/license-webpack-plugin, or something else that can automatically do this copying.


jasongrout:
We aren't using the webpack license plugin right now because:

(a) the webpack compilation is one from JupyterLab, so we don't have direct control over it
(b) that particular plugin does not support webpack 5 yet

I'll open an issue in jlab to see about using this plugin when it is available.

vidartf:
This is a good point. We should probably do this for widget cookie-cutters as well, as the Python nbextension also typically bundles dependencies.

PR #2995, opened 2020-10-13
Remove JupyterLab dependency in jupyterlab_widgets

martinRenou:
Even though jupyterlab_widgets is useless without JupyterLab, it is only a small Python package (that don't need JupyterLab) and some static files installed in the right path (share/jupyter/labextensions). I would then expect it to **not** depend on JupyterLab.

This will allow other widgets extension to depend on the jupyterlab_widget package so that those widgets extensions can work on JupyterLab out of the box (given that they install their static files as well), without them depending on JupyterLab.

cc. @ianhi

### DIFF
--- a/jupyterlab_widgets/setup.py
+++ b/jupyterlab_widgets/setup.py
@@ -12,7 +12,7 @@ import setuptools
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 # The name of the project
-name="jupyterlab_widgets"
+name = "jupyterlab_widgets"
 
 # Ensure a valid python version
 ensure_python(">=3.6")
@@ -40,7 +40,8 @@ data_files_spec = [
     ("share/jupyter/lab/schemas/%s" % labext_name, os.path.join(lab_path, 'schemas', labext_name), "*.*"),
 ]
 
-cmdclass = create_cmdclass("jsdeps", 
+cmdclass = create_cmdclass(
+    "jsdeps",
     package_data_spec=package_data_spec,
     data_files_spec=data_files_spec
 )
@@ -66,11 +67,9 @@ setup_args = dict(
     description="A JupyterLab extension.",
     long_description= long_description,
     long_description_content_type="text/markdown",
-    cmdclass= cmdclass,
+    cmdclass=cmdclass,
     packages=setuptools.find_packages(),
-    install_requires=[
-        "jupyterlab~=3.0.0rc4",
-    ],
+    install_requires=[],
     zip_safe=False,
     include_package_data=True,
     python_requires=">=3.6",


vidartf:
I think we *do* need some system to say which version(s) of lab the JS is compatible with though...

martinRenou:
We might be able to do this using conda-forge's `run_constrained` https://conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime

Concerning pip, idk.

jasongrout:
pip is introducing a new solver this year that I think will let it do things like conda's `run_constrained`, so this might be good timing.

I went back and forth on whether jupyterlab_widgets should depend on jlab or not. It's a continuation of our conversation from the workshop earlier this year, of where to draw the cutoff in the dependency.

jasongrout:
Sounds like the community sentiment is to not have the dependency, so let's go for it!

ianhi:
:O horrraayyyy

So if I understand correctly now widgets (e.g. ipympl) can just require `jupyterlab-widgets` without forcibly installing jlab. This allows much simpler install instuctions which is rad. But what about:

> I think we do need some system to say which version(s) of lab the JS is compatible with though...

Based on 

> pip is introducing a new solver this year that I think will let it do things like conda's run_constrained, so this might be good timing.

I went a fairly unproductive deep dive of pip documentation and that wasn't totally my conclusion, very happy to be wrong though.

jasongrout:
> I went a fairly unproductive deep dive of pip documentation and that wasn't totally my conclusion, very happy to be wrong though.

Something similar is implemented in https://pip.pypa.io/en/stable/user_guide/#constraints-files, though not quite mirroring a setup.py runtime constraint. The new solver, IIRC, keeps track of dependencies in an environment and makes sure the environment stays consistent. So I think the underlying capability is there to have something like run_constrained, even if a similar setup.py-specified constraint is not currently implemented.

jasongrout:
I released this as `jupyterlab_widgets` 1.0.0a6, so people can test making it a dependency and see if that works out well.

ianhi:
> I released this as jupyterlab_widgets 1.0.0a6 (on pypi and conda-forge), so people can test making it a dependency and see if that works out well.

Just tried it with ipympl, worked great! I could use it with just notebook without installing jlab, then once I installed jlab3 it worked immediately.

martinRenou:
@jasongrout should we go further and make `jupyterlab_widgets` an ipywidgets dependency? So that `conda install -c conda-forge ipywidgets` or `pip install ipywidgets` is enough for having widgets working in JupyterLab.

jasongrout:
> @jasongrout should we go further and make `jupyterlab_widgets` an ipywidgets dependency? So that `conda install -c conda-forge ipywidgets` or `pip install ipywidgets` is enough for having widgets working in JupyterLab.

I'm okay with that. That would mean that other widget packages actually don't have to depend on it - they just depend on ipywidgets.

martinRenou:
Yes! That would be great.

PR #3008, opened 2020-11-05
Add jupyterlab_widgets dependency

martinRenou:


jasongrout (approved):

jasongrout:
Note that in general we won't want to constrain this version too much, since it will have a runtime constraint depending on the version of jlab the user has installed. For example, jupyterlab_widgets 1 will depend on jlab 3, but jupyterlab_widgets 2 presumably will depend on jlab 4. We want the user's environment solver to automatically pick jupyterlab_widgets 1 or 2, depending on what lab version they have, so we don't want to constrain the version too much here.

Since jupyterlab_widgets has an old, old version available, though, we need to constrain to be at least 1.0 (when 1.0 is published).

PR #3020, opened 2020-11-19
perf: use class_traits with a cache to speed up creation of widgets

maartenbreddels:
Benchmark results for creating a button `widgets.Button()`
```
------------------------------------------------ benchmark: 1 tests -----------------------------------------------
Name (time in ms)              Min      Max    Mean  StdDev  Median     IQR  Outliers       OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------
test_performance_create     2.8579  36.1314  3.3029  2.0412  3.0983  0.2583      1;22  302.7645     266           1
-------------------------------------------------------------------------------------------------------------------
```

After this PR, in combination with https://github.com/ipython/traitlets/pull/639

```
------------------------------------------------ benchmark: 1 tests -----------------------------------------------
Name (time in ms)              Min      Max    Mean  StdDev  Median     IQR  Outliers       OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------
test_performance_create     2.0695  62.7899  2.9496  3.6363  2.3238  0.8066      2;27  339.0254     290           1
-------------------------------------------------------------------------------------------------------------------
```

Which is a 30% performance increase.

Note that `.class_traits` is the same as `.traits` but is a classmethod, so that we can cache on a class level. Note that add_trait will create a new class, so will play nice with the cache. This is tested in case this changes.

### DIFF
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -10,6 +10,7 @@ from contextlib import contextmanager
 from collections.abc import Iterable
 from IPython import get_ipython
 from ipykernel.comm import Comm
+import functools
 from traitlets import (
     HasTraits, Unicode, Dict, Instance, List, Int, Set, Bytes, observe, default, Container,
     Undefined)
@@ -367,7 +368,7 @@ class Widget(LoggingHasTraits):
 
     @default('keys')
     def _default_keys(self):
-        return [name for name in self.traits(sync=True)]
+        return [name for name in self.class_traits(sync=True)]
 
     _property_lock = Dict()
     _holding_sync = False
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Note, without the traitlets PR this result is:
```
------------------------------------------------ benchmark: 1 tests -----------------------------------------------
Name (time in ms)              Min      Max    Mean  StdDev  Median     IQR  Outliers       OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------
test_performance_create     2.2499  38.9077  2.6620  2.0604  2.4685  0.2669      1;22  375.6595     316           1
-------------------------------------------------------------------------------------------------------------------
```

ipywidgets/widgets/widget.py
```diff
@@ -547,6 +548,13 @@ def add_traits(self, **traits):
                 self.keys.append(name)
                 self.send_state(name)
 
+    @classmethod
+    @functools.lru_cache(None)
```
rmorshea, line 552: I feel like an unlimited cache size is a bit dangerous. Maybe it's not in practice though?

PR #3021, opened 2020-11-19
Fetch the full widget state via a control Comm

maartenbreddels:
Related issues
 * https://github.com/jupyter-widgets/ipywidgets/issues/3019
 * https://github.com/voila-dashboards/voila/issues/764

More info https://github.com/voila-dashboards/voila/pull/766

### DIFF
--- a/ipywidgets/widgets/tests/test_send_state.py
+++ b/ipywidgets/widgets/tests/test_send_state.py
@@ -3,7 +3,7 @@
 
 from traitlets import Bool, Tuple, List
 
-from .utils import setup, teardown
+from .utils import setup, teardown, DummyComm
 
 from ..widget import Widget
 
@@ -23,3 +23,12 @@ def test_empty_hold_sync():
     with w.hold_sync():
         pass
     assert w.comm.messages == []
+
+
+def test_control():
+    comm = DummyComm()
+    Widget.close_all()
+    w = SimpleWidget()
+    Widget.handle_control_comm_opened(comm, {})
+    Widget.handle_control_comm_msg({'type': 'models-request'})
+    assert comm.messages
--- a/ipywidgets/widgets/widget.py
+++ b/ipywidgets/widgets/widget.py
@@ -25,8 +25,9 @@ from json import loads as jsonloads, dumps as jsondumps
 
 from base64 import standard_b64encode
 
-from .._version import __protocol_version__, __jupyter_widgets_base_version__
+from .._version import __protocol_version__, __control_protocol_version__, __jupyter_widgets_base_version__
 PROTOCOL_VERSION_MAJOR = __protocol_version__.split('.')[0]
+CONTROL_PROTOCOL_VERSION_MAJOR = __control_protocol_version__.split('.')[0]
 
 def _widget_to_json(x, obj):
     if isinstance(x, dict):
@@ -290,6 +291,7 @@ class Widget(LoggingHasTraits):
     # Class attributes
     #-------------------------------------------------------------------------
     _widget_construction_callback = None
+    _control_comm = None
 
     # widgets is a dictionary of all active widget objects
     widgets = {}
@@ -301,7 +303,6 @@ class Widget(LoggingHasTraits):
     def close_all(cls):
         for widget in list(cls.widgets.values()):
             widget.close()
-
 
     @staticmethod
     def on_widget_constructed(callback):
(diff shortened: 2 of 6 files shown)


martinRenou (commented):
LGTM! This works very nicely with https://github.com/voila-dashboards/voila/pull/766 and brings a very good speedup to Voila when there are many widgets.

It's also a simple addition to ipywidgets and doesn't change the normal behavior.

martinRenou (approved):

SylvainCorlay:
This was discussed today at the Jupyter widgets call.

For this to get merged, we need to add documentation for this in the protocol documentation.

ipywidgets/widgets/widget.py
```diff
@@ -317,6 +317,26 @@ def _call_widget_constructed(widget):
...
+                'model_name': widget._model_name,
+                'model_module': widget._model_module,
+                'model_module_version': widget._model_module_version,
+                'state': widget.get_state(drop_defaults=drop_defaults),
+            }
+        full_state, buffer_paths, buffers = _remove_buffers(full_state)
+        # the message is also send as buffer, so it does not get handled by jupyter_server
+        msg = jsondumps([full_state, buffer_paths]).encode('utf8')
+        buffers.insert(0, msg)
+        comm.send(buffers=buffers)
```
vidartf, line 338: What is the reasoning behind this sending the entire state as a "buffer" instead of sending as regular JSON? Is it to sidestep the decoding that happens in the server?

ipywidgets/widgets/tests/test_send_state.py
```diff
@@ -23,3 +23,11 @@ def test_empty_hold_sync():
...
         pass
     assert w.comm.messages == []
+
+
+def test_control():
+    comm = DummyComm()
+    Widget.close_all()
+    w = SimpleWidget()
+    Widget.handle_comm_opened_control(comm, {})
+    assert comm.messages
```
vidartf, line 33: Ideally the tests would be inspecting the messages a little closer, e.g. check the buffer decoding etc.

jasongrout:
Can we add a version metadata field for the comm_open message, like we do for the jupyter.widget channel? See https://github.com/jupyter-widgets/ipywidgets/pull/3021. This will let us experiment more freely and communicate at the start what the protocol version will be.

jasongrout:
> This will let us experiment more freely and communicate at the start what the protocol version will be.

For example, the working around the server performance problems by packaging the JSON in a binary buffer is probably something that can go away in a future iteration.

ipywidgets/widgets/widget.py
martinRenou: Yes. It has a big performance impact. Although I would agree we don't want to do that and we would want to improve the server instead.

blois:
Would it make sense to have the control channel message be a request to have the widget state be broadcast on the iopub channel? Since the data can be changing while the messages are in flight there can be a race between the response here and comm messages on iopub. Colab has had similar issues between shell and iopub- even though they are one WebSocket to the client they are still separate ZMQ sockets from the kernel and the messages can get reordered.

ipywidgets/widgets/widget.py
maartenbreddels: I'm not sure that is the right solution either, I think the fact that the jupyter protocol speaks JSON and the messages we send (the widget protocol) are JSON are a mere coincidence. We should not embed our widget protocol message inside the jupyter protocol as JSON, the JSON from the widget protocol should just be a string or binary buffer in the jupyter message. That way the server would not parse the JSON (it would just be a string). In this case I've chosen a buffer for performance reasons, but a single string would also suffice (i.e. json.dumps(..)). But I don't want to go too much off topic here.

ipywidgets/widgets/widget.py
```diff
@@ -317,6 +318,27 @@ def _call_widget_constructed(widget):
         if Widget._widget_construction_callback is not None and callable(Widget._widget_construction_callback):
             Widget._widget_construction_callback(widget)
 
+    @classmethod
+    def handle_comm_opened_control(cls, comm, msg):
```
vidartf, line 322: I think the general consensus from the dev meeting is that this PR makes sense, but that the logic should not happen on comm open. Rather, the new comm channel can be considered a channel for the widget manager, where "get all the current widget state" would be one type of message that could be sent on it. That allows us to also use this channel for other purposes in the future without having to transmit all the state every time.

martinRenou: Done :) Thanks for the review

maartenbreddels: Totally agree with that, doing it on comm open was just a simple approach, and what @vidartf describes is much better! And thanks for implementing that @martinRenou !

ipywidgets/widgets/widget.py
```diff
@@ -317,6 +318,39 @@ def _call_widget_constructed(widget):
         if Widget._widget_construction_callback is not None and callable(Widget._widget_construction_callback):
             Widget._widget_construction_callback(widget)
 
+    @classmethod
+    def handle_control_comm_opened(cls, comm, msg):
+        version = msg.get('metadata', {}).get('version', '')
+        if version.split('.')[0] != CONTROL_PROTOCOL_VERSION_MAJOR:
+            raise ValueError("Incompatible widget control protocol versions: received version %r, expected version %r"%(version, __control_protocol_version__))
```
martinRenou, line 325: Should we close the comm manually here?

martinRenou (commented):
Needs docs for the models message format

martinRenou:
Added docs + made the messaging format/naming closer to the other channel

ipywidgets/widgets/widget.py
martinRenou: This is done automatically by ipykernel, we don't need to close it

ipywidgets/widgets/widget.py
```diff
@@ -317,6 +318,47 @@ def _call_widget_constructed(widget):
...
+
+        cls._control_comm = comm
+        cls._control_comm.on_msg(cls.handle_control_comm_msg)
+
+    @classmethod
+    def handle_control_comm_msg(cls, msg):
```
vidartf, line 331: This could probably be a "private" method, i.e. `_handle_control_comm_msg`

ipywidgets/widgets/widget.py
```diff
@@ -317,6 +318,47 @@ def _call_widget_constructed(widget):
...
+        if cls._control_comm is None:
+            raise RuntimeError('Control comm has not been properly opened')
+
+        data = msg['content']['data']
+        method = data['method']
+
+        if method == 'request_states':
+            # Send back the full widgets state
+            cls.get_manager_state()
+            widgets = Widget.widgets.values()
```
vidartf, line 342: I see that the other methods here are static (e.g. `handle_comm_opened`), and then explicitly use `Widget.` to access other static / class attributes. I kind of agree that it makes more sense as a classmethod, but just noticed this as a difference. I'm just curious why there is `Widget.widgets` here, instead of `cls.widgets`?

vidartf (commented):
Logic seems good to me, but I have some minor comments on code/conventions.

I also note that it might be good to have a method on the JS base manager to request the state and handle and/or parse the reply (e.g. put the buffers back in, update the state correctly). Even if we don't use it, it might be useful to have a "reference" implementation, especially if it is written such that it could be used directly by other projects (i.e. voila). It would probably make sense to add this in a follow-up PR though, so no need to add it now.

ipywidgets/widgets/widget.py
martinRenou: Good catch! I'll fix it, thanks

martinRenou:
Makes sense! Thanks for the review, I took your comments into account

martinRenou:
Thanks!

martinRenou:

PR #3033, opened 2020-12-09
Make INotebookTracker optional in the manager plugin

jtpio:
This should help make the plugin more flexible.

For example when it is loaded as a federated extension on a page that doesn't provide the `@jupyterlab/notebook-extension:tracker`.

### DIFF
--- a/jupyterlab_widgets/src/plugin.ts
+++ b/jupyterlab_widgets/src/plugin.ts
@@ -157,8 +157,8 @@ export function registerWidgetManager(
  */
 const plugin: JupyterFrontEndPlugin<base.IJupyterWidgetRegistry> = {
   id: '@jupyter-widgets/jupyterlab-manager:plugin',
-  requires: [INotebookTracker, IRenderMimeRegistry, ISettingRegistry],
-  optional: [IMainMenu, ILoggerRegistry],
+  requires: [IRenderMimeRegistry, ISettingRegistry],
+  optional: [INotebookTracker, IMainMenu, ILoggerRegistry],
   provides: base.IJupyterWidgetRegistry,
   activate: activateWidgetExtension,
   autoStart: true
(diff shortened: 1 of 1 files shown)


vidartf (approved):

vidartf:
@jtpio The change in itself is ok, but as Jason points out, there probably needs to be additional changes to allow for providing the context. To make it more reusable for other things, it would probably also make sense to make changes to the manager to separate:
- the code that only need to know about the kernel/session connection from the code
- the code that needs to know about the notebook model (for loading/saving state, marking the model as dirty on changes)

jtpio:
> Just curious, how are you registering the widget manager in the case where you do not have a tracker?

For now the main motivation is when the `jupyterlab_widgets` extension is loaded as federated on the page, in a lab-based app that does not necessarily have the notebook tracker (for example the tree page in [JupyterLab Classic](https://github.com/jtpio/jupyterlab-classic)). So having the token optional would make it more like a no-op instead of logging an error in the dev tools console.

> there probably needs to be additional changes to allow for providing the context. To make it more reusable for other things, it would probably also make sense to make changes to the manager to separate

Yes absolutely. It sounds like it would also make it easier to provide widget support for code consoles: https://github.com/jupyter-widgets/ipywidgets/pull/3004 
Or even arbitrary jupyterlab widgets backed by a kernel.

jasongrout:
Merging since this doesn't hurt anything, and helps slightly in the classic case (or other cases where there is not a notebook tracker).

jasongrout:
I'll note that master (8.0) does not have this issue, I think courtesy of https://github.com/jupyter-widgets/ipywidgets/pull/2528

vidartf:
Also note that I had completely forgotten that the work to split the kernel manager to not need document context was already done (by me, even!) in master: https://github.com/jupyter-widgets/ipywidgets/blob/3108308d93c423845901500b108f45b4d7e9fc5b/jupyterlab_widgets/src/manager.ts#L416

jtpio:
Ah nice!

We probably didn't realize since this PR was against the 7.x branch.

PR #3040, opened 2020-12-18
Docs: Remove reference to resolved bug

carlobeltrame:
#1811 was apparently resolved last January, I think this notice can be removed from the docs.


vidartf:
Thanks!

PR #3060, opened 2021-01-07
Fix debouncing and throttling code

davidbrochart:
The `Timer` from `threading` only starts if you call `.start()`. There were other minor bugs. I checked that it works with both timers (async and threaded).


SylvainCorlay:
Thanks!

sindhu51188:
Can we stop this emails please

jasongrout:
@sindhu51188 - as it explains in the footer to the message you quoted, you are receiving emails because you are subscribed to the issues. You can follow the link to unsubscribe.

PR #3074, opened 2021-02-01
Bugfix/parameters in the from file method to be passed along in the media class 

kedarisetti:
Extra parameters in the from_url() method do not get passed on to the super class.
this fix passes the parameters along.


jasongrout:
Thanks! Can we separate this change from the change in #3075?

kedarisetti:
Done

jasongrout:
Thanks!

PR #3075, opened 2021-02-01
Existence of attributes in html video tags makes the attribute to be true.

kedarisetti:
The correct way of setting/unsetting the tags in the html video element for 'looping', 'autoplay' and 'controls'
is to make the tag available (with any value) or removing it.

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
https://stackoverflow.com/questions/14161516/html5-video-completely-hide-controls

### DIFF
--- a/packages/controls/src/widget_video.ts
+++ b/packages/controls/src/widget_video.ts
@@ -85,9 +85,30 @@ export class VideoView extends DOMWidgetView {
     }
 
     // Video attributes
-    this.el.loop = this.model.get('loop');
-    this.el.autoplay = this.model.get('autoplay');
-    this.el.controls = this.model.get('controls');
+    const loop  = this.model.get('loop');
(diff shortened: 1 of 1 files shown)


jasongrout:
Is it not working for you? It seems to work fine for me as-is, without this change, i.e., setting loop to False in python disables looping.

We're using the actual DOM attributes, which are documented as being true or false, like https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/autoplay

For example, this disables the controls just fine for me without this change:

```python
Video(value=b'https://upload.wikimedia.org/wikipedia/commons/4/40/Reaction_Diffusion_Varying.webm', format='url', controls=False)
```

jasongrout:
What browser and OS are you using?

kedarisetti:
#3074 fixes this issue for me. Therefore closing.

PR #3080, opened 2021-02-02
Move widgetsnbextension, jupyterlab_widgets from install_requires to extras_require

mkoeppe:
A longstanding TODO in the `ipywidgets` `setup.py`: "Dynamically add [the dependency on `widgetsnbextension`] only if notebook 4.x is installed in this interpreter, to allow ipywidgets to be installed on bare kernels."

Such conditional dependencies, of course, are incompatible with modern Python packaging.

Recent versions of `ipywidgets` have added another dependency that is in the way of the goal of installs on bare kernels: an `extras_require` that pulls in all of JupyterLab via `jupyterlab_widgets`: `':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0']`

In this pull request, we change these problematic dependencies to `extra_requires`. 

Cross-ref: https://trac.sagemath.org/ticket/31278

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,19 +115,15 @@ install_requires = setuptools_args['install_requires'] = [
     # Requiring nbformat to specify bugfix version which is not required by
     # notebook.
     'nbformat>=4.2.0',
-    # TODO: Dynamically add this dependency
-    # only if notebook 4.x is installed in this
-    # interpreter, to allow ipywidgets to be
-    # installed on bare kernels.
-    'widgetsnbextension~=3.5.0'
 ]
 
 extras_require = setuptools_args['extras_require'] = {
     ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
     ':python_version>="3.3"': ['ipython>=4.0.0'],
-    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0'],
     'test:python_version=="2.7"': ['mock'],
     'test': ['pytest>=3.6.0', 'pytest-cov'],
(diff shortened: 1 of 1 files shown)


martinRenou:
We discussed during yesterday's meeting removing the Notebook dependency from widgets nbextension.

Concerning jupyterlab-widgets, it does not pull jupyterlab anymore, since https://github.com/jupyter-widgets/ipywidgets/pull/2995.

mkoeppe:
> Concerning jupyterlab-widgets, it does not pull jupyterlab anymore, since #2995.

Thanks, that's good news, I had missed that.

martinRenou:
Having widgetsnbextension and jupyterlab-widgets as install dependencies makes it much more user-friendly because people can just do `pip install ipywidgets` and they have a working setup for both Jupyter Notebook and JupyterLab. I understand that for bare kernel installations that might not always make sense, but those extensions only bring static assets that it installs on the machine, it should do no harm (apart from the Notebook that is pulled currently).

mkoeppe:
If the notebook dependency is indeed dropped from `widgetsnbextension` (when is this going to happen?), then I would agree that the proposed change in this PR is not necessary.

willingc:
Is this something we can move into 8.0 instead of 7.x? I'm not sure that changing behavior here makes good sense for existing users. There are a lot of breaking changes in 8.0 and this seems better suited to that.

jasongrout:
> If the notebook dependency is indeed dropped from `widgetsnbextension` (when is this going to happen?), then I would agree that the proposed change in this PR is not necessary.

That is the current plan for 8.0.

Closing as requested, then.

PR #3092, opened 2021-02-06
Update locking workflow

jasongrout:
The old workflow seems to not be working anymore. [This repo](https://github.com/dessant/lock-threads) looks like the natural successor.

I upped the time for locking inactive resolved issues and PRs to 90 days since ipywidgets has a slower cadence than many other Jupyter projects.


willingc (approved):
Looks fine to me. Thanks @jasongrout

PR #3110, opened 2021-02-10
Traitlets 4 vs 5 changes defaults in message spec

jasongrout:
We get these by these commands from the release docs:

```
python ./packages/schema/generate-spec.py -f json-pretty > packages/schema/jupyterwidgetmodels.latest.json
python ./packages/schema/generate-spec.py -f markdown > packages/schema/jupyterwidgetmodels.latest.md
```

I noticed that these gave updates that were not already in the files. It seems that our updated packages now give updated formatting to the json, so it's probably more instructive to look at the two small changes in the .md file.

### DIFF
--- a/packages/schema/jupyterwidgetmodels.latest.json
+++ b/packages/schema/jupyterwidgetmodels.latest.json
@@ -310,7 +310,13 @@
       {
         "allow_none": true,
         "default": null,
-        "enum": ["contain", "cover", "fill", "scale-down", "none"],
+        "enum": [
+          "contain",
+          "cover",
+          "fill",
+          "scale-down",
+          "none"
+        ],
         "help": "The object-fit CSS attribute.",
         "name": "object_fit",
         "type": "string"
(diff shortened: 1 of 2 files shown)


jasongrout:
It looks like these differences are caused by traitlets 4 vs traitlets 5. That is something we need to look into, but not a blocker for an alpha release, I think.

jasongrout:
Converting to draft since this needs more discussion before merging (or closing)

SylvainCorlay:
**Traitlets 4:**
Screenshot from 2021-03-18 15-58-27

**Traitlets 5:**
Screenshot from 2021-03-18 16-03-35

vidartf:
Note that:
```py
>>> traitlets.__version__
'5.0.5'
>>> Tuple().make_dynamic_default()
()
>>> Tuple(traitlets.Unicode()).make_dynamic_default()
<blank = None>
```

vidartf:
I added a fix for this to #3174 (since it is painful to work with specs on Windows without it).

Issue #3063, opened 2021-01-14
Playing an Audio widget programmatically?


deeplook:
I tried to play an Audio widget programmatically, but I don't see a `play()`-like method to do that, like:

```python
from ipywidgets import Audio
w = Audio.from_file("audio/a.mp3", autoplay=False, loop=False)
w.play()
# AttributeError: 'Audio' object has no attribute 'play'
```

Is it possible?

ianhi:
It's not currently possible. But I think this is a reasonable feature request. The steps to accomplish this are:
1. Add a `playing` `Bool` attribute to the python side 
https://github.com/jupyter-widgets/ipywidgets/blob/bd8ec4509cdfee0a3609b30a4bc9ea5e398b856b/ipywidgets/widgets/widget_media.py#L214-L217

2. On the javascript side add a `this.model.on("change:playing", <some fxn>)` callback to update call `.play()` or `.pause()`
This should probably go here: https://github.com/jupyter-widgets/ipywidgets/blob/bd8ec4509cdfee0a3609b30a4bc9ea5e398b856b/packages/controls/src/widget_audio.ts#L39

https://www.w3schools.com/Jsref/met_audio_pause.asp


See how the play button widget works for inspiration: https://github.com/jupyter-widgets/ipywidgets/blob/51322341d4f6d88449f9dbf6d538c60de6d23543/packages/controls/src/widget_int.ts#L922

SylvainCorlay:
@deeplook I think the best tool for this is ipywebrtc, not the media widgets of ipywidgets.

willingc:
Let's close this issue as @SylvainCorlay's suggestion makes good sense and requires no additional development here. @deeplook and ianhi please close this issue if you agree. Thanks!

deeplook:
I'll surely try ipywebrtc, but it does have a slight flavor of having to use an overkill solution for a little missing feature in the existing Audio widget that would simply make it a bit more of a fully implemented thingy.

hai-schrodinger:
I agree with @deeplook.

ianhi:
My sense is that the ability to start and stop playing is essential functionality for an `audio` widget. So if ipywidgets is going to provide an audio widget it is incomplete without that. If ipywebrtc is the superior solution then perhaps the audio widget should be deprecated (although I'm pro keeping the audio widget with a playing trait).

ianhi:
[...]

PR #3112, opened 2021-02-11
Add play/pause methods to media widgets

martinRenou:
As discussed in #3063

I did not make a `playing` attribute as part of the model, the reason is that if we do this, changing the `playing` would trigger an `update()` call on the front-end view, resetting the video to its start.

### DIFF
--- a/docs/source/examples/Media widgets.ipynb
+++ b/docs/source/examples/Media widgets.ipynb
@@ -42,7 +42,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "video1"
+    "video1.pause()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "video1.play()"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


jasongrout:
I forget: can you use `model.changedAttributes()` to determine if only the playing attribute has changed, then short-circuit the update method if so?

martinRenou:
Oh wow, I didn't know about `changedAttributes` thanks! I'll give it a try.

jasongrout:
Yeah, I think we use it in a similar manner here to short-circuit even calling the update method if all we are doing is updating the view count: https://github.com/jupyter-widgets/ipywidgets/blob/d23eddc735ef5303e10e6a152f14ad2c34d66891/packages/base/src/widget.ts#L636-L640

jasongrout:
> Oh wow, I didn't know about `changedAttributes` thanks! I'll give it a try.

Docs here: https://backbonejs.org/#Model-changedAttributes

marscher:
Thanks for adding the play/pause feature. I wonder if it would be possible to expose the time/frame information as well. Some people tried to link Audio/Video widget with matplotlibs AxesWidget, e.g. to share the time-axis both widgets possess.

A request for this can be found on [(Gitter)](https://gitter.im/jupyter-widgets/Lobby?at=60a6bade6fe44c28cf24f71b) and I'm also looking for this ability. The jslink function can only be used on attributes linked with the backend as it seems.

The "playbackRate" seems also interesting. I guess a generic interface dynamically defining DOM-events, synchronized backend traits ain't possible?

martinRenou:
> I wonder if it would be possible to expose the time/frame information as well

The current video frame changes *a lot* and syncing with the back-end might be annoying. Maybe this kind of feature would better go in ipywebrtc instead.

PR #3117, opened 2021-02-12
Change graph example to receive a tuple instead of a dict

### LINKED ISSUE #3089: Bug in Exploring Graphs example code
## Description

In the [`Exploring Graphs.ipynb`](https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Exploring%20Graphs.ipynb) file in the docs/source/examples directory, the last cell does not work in 8.0 since it uses a dict for `generator`. Instead, the code should use a list of tuples.

```python
interact(plot_random_graph, n=(2,30), m=(1,10), k=(1,10), p=(0.0, 1.0, 0.001),
         generator={
             'lobster': random_lobster,
             'power law': powerlaw_cluster,
             'Newman-Watts-Strogatz': newman_watts_strogatz,
             u'Erdős-Rényi': erdos_renyi,
         });
```
Alternatively, we could introduce dicts as a possibility again now that their order is guaranteed.

marimeireles:
Fixes #3089 
I also changed the `p` value to `0.99` because I'd get an error saying the value should be smaller than 1.
I also got some errors saying I needed to have scipy installed when I tried to generate some graphs super fast. I wonder if it's a weird edge case or if we should add it to the notebook.

### DIFF
--- a/docs/source/examples/Exploring Graphs.ipynb
+++ b/docs/source/examples/Exploring Graphs.ipynb
@@ -74,21 +74,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "interact(plot_random_graph, n=(2,30), m=(1,10), k=(1,10), p=(0.0, 1.0, 0.001),\n",
-    "         generator={\n",
-    "             'lobster': random_lobster,\n",
-    "             'power law': powerlaw_cluster,\n",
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks! Merging now!

PR #3118, opened 2021-02-15
Rename pWidget to lmWidget

### LINKED ISSUE #3087: Rename .pWidget to .lmWidget
To complete our transition to Lumino, perhaps we should rename the `.pWidget` attribute to `.lmWidget`? Or is that a bit gratuitous?

davidbrochart:
Fixes #3087

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -886,6 +886,16 @@ export class DOMWidgetView extends WidgetView {
     this.updateTooltip();
   }
 
+  /**
+   * Getter for backward compatibility.
+   *
+   * pWidget is deprecated and will be removed in the future,
+   * please use lmWidget instead.
+   */
+  get pWidget(): Widget {
+    return this.lmWidget;
+  }
+
   setLayout(layout: LayoutModel, oldLayout?: LayoutModel): void {
     if (layout) {
       this.layoutPromise = this.layoutPromise.then(oldLayoutView => {
@@ -904,12 +914,12 @@ export class DOMWidgetView extends WidgetView {
                 // Post (asynchronous) so layout changes can take
                 // effect first.
                 MessageLoop.postMessage(
-                  this.pWidget,
+                  this.lmWidget,
                   Widget.ResizeMessage.UnknownSize
                 );
               });
               MessageLoop.postMessage(
-                this.pWidget,
+                this.lmWidget,
                 Widget.ResizeMessage.UnknownSize
               );
               return view;
@@ -1034,21 +1044,21 @@ export class DOMWidgetView extends WidgetView {
   }
 
   _setElement(el: HTMLElement): void {
-    if (this.pWidget) {
-      this.pWidget.dispose();
+    if (this.lmWidget) {
+      this.lmWidget.dispose();
     }
 
     this.$el = el instanceof $ ? el : $(el);
     this.el = this.$el[0];
-    this.pWidget = new JupyterLuminoWidget({
+    this.lmWidget = new JupyterLuminoWidget({
       node: el,
       view: this
     });
   }
 
   remove(): any {
-    if (this.pWidget) {
-      this.pWidget.dispose();
+    if (this.lmWidget) {
+      this.lmWidget.dispose();
     }
     return super.remove();
   }
@@ -1063,9 +1073,9 @@ export class DOMWidgetView extends WidgetView {
 
   private _comm_live_update(): void {
     if (this.model.comm_live) {
-      this.pWidget.removeClass('jupyter-widgets-disconnected');
+      this.lmWidget.removeClass('jupyter-widgets-disconnected');
     } else {
-      this.pWidget.addClass('jupyter-widgets-disconnected');
+      this.lmWidget.addClass('jupyter-widgets-disconnected');
     }
   }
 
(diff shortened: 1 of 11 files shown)


willingc:
@davidbrochart Consider renaming prefix from `lm` to `lum`. I keep reading it as `I M` instead of `L M`. For users and new contributors `lm` could be confusing.

davidbrochart:
Yes, it can be confusing. I don't have any problem with `lumWidget`, should we go ahead with this?

jtpio:
Maybe the original suggestion was `lmWidget` to follow a similar naming as for the CSS classes? (`p-Widget` and `lm-Widget`)

For example: https://github.com/jupyterlab/lumino/blob/4f2432290e874b040310d428593abe4afc336066/packages/widgets/style/widget.css#L11-L12

willingc:
@davidbrochart Given @jtpio's description of the source, let's leave as is in the PR. Thanks!

willingc (approved):

packages/base/src/widget.ts
```diff
@@ -886,6 +886,10 @@ export class DOMWidgetView extends WidgetView {
     this.updateTooltip();
   }
 
+  get pWidget(): Widget {
```
jtpio, line 889: Should there be a docstring to mention that using `pWidget` is deprecated?

jasongrout:
> @davidbrochart Consider renaming prefix from `lm` to `lum`. I keep reading it as `I M` instead of `L M`. For users and new contributors `lm` could be confusing.

I also keep reading it as "I M" - note to self about never using `l` to start a name when it might be used in a camelcase situation!

There is precedence for using `lmWidget`, so I'm fine with moving forward with it, but if in the future we decide to do something different, I would rather go all the way and do `luminoWidget` over `lumWidget`.

packages/base/src/widget.ts
davidbrochart: Done, thanks.

packages/controls/src/widget_selectioncontainer.ts
```diff
@@ -209,7 +209,7 @@ export class AccordionView extends DOMWidgetView {
   }
 
   children_views: ViewList<DOMWidgetView> | null;
-  pWidget: Accordion;
+  lmWidget: Accordion;
```
jasongrout, line 212: I think we also need `pWidget: Accordion` until it is removed (since it is right now inherited as type `Widget`)

packages/controls/src/widget_selectioncontainer.ts
```diff
@@ -423,7 +423,7 @@ export class TabView extends DOMWidgetView {
 
   updatingTabs = false;
   childrenViews: ViewList<DOMWidgetView> | null;
-  pWidget: JupyterLuminoTabPanelWidget;
+  lmWidget: JupyterLuminoTabPanelWidget;
```
jasongrout, line 426: Here pWidget also needs the typing until it is removed.

packages/html-manager/src/output.ts
```diff
@@ -61,13 +61,13 @@ export class OutputView extends outputBase.OutputView {
...
+    this.lmWidget.insertWidget(0, this._outputView);
+    this.lmWidget.addClass('jupyter-widgets');
+    this.lmWidget.addClass('widget-output');
     this.update();
   }
 
   model: OutputModel;
   private _outputView: OutputArea;
-  pWidget: Panel;
+  lmWidget: Panel;
```
jasongrout, line 72: Here pWidget also needs the typing until it is removed.

packages/controls/src/widget_box.ts
```diff
@@ -136,7 +136,7 @@ export class BoxView extends DOMWidgetView {
   }
 
   children_views: ViewList<DOMWidgetView> | null;
-  pWidget: JupyterLuminoPanelWidget;
+  lmWidget: JupyterLuminoPanelWidget;
```
jasongrout, line 139: Here pWidget also needs the typing until it is removed.

packages/controls/src/widget_controller.ts
```diff
@@ -429,5 +429,5 @@ export class ControllerView extends DOMWidgetView {
   axis_box: Panel;
   button_box: Panel;
   model: ControllerModel;
-  pWidget: JupyterLuminoPanelWidget;
+  lmWidget: JupyterLuminoPanelWidget;
```
jasongrout, line 432: Here pWidget also needs the typing until it is removed.

packages/controls/src/widget_selectioncontainer.ts
jasongrout: Unless there is some way to say that `pWidget` returns `typeof this.lmWidget` - I don't know if that will work with subclassing and redefining `lmWidget`, though.

jasongrout (commented):
Looks good, with one comment about preserving `pWidget` typings in subclasses.

packages/controls/src/widget_selectioncontainer.ts
davidbrochart: I didn't find a way to type `pWidget` in the base class as the type of `lmWidget` in the subclass, so I added back `pWidget`'s typing in each subclass as you initially suggested. Thanks!

jasongrout:
Thanks!

jasongrout:
We had generally positive sentiment in the widget dev meeting today to rename `.lmWidget` to `.luminoWidget`. I've made that #3137.

PR #3119, opened 2021-02-15
Update repo-detection

### LINKED ISSUE #3090: Update repo-detection code for jupyterlab_widgets
We have some code (that actually doesn't work, I think?) at https://github.com/jupyter-widgets/ipywidgets/blob/5532c91d892e01731e16dffa54cf34dd2628928a/jupyterlab_widgets/setup.py#L49-L57 (and soon in master after #3086) to detect if we are building `jupyterlab_widgets` in the repo, to avoid recompiling js files many times.

The updated jlab extension cookiecutter has better code at 

https://github.com/jupyterlab/extension-cookiecutter-ts/blob/b13df389efd10fbccba441537aa198e64f623fa4/%7B%7Bcookiecutter.python_name%7D%7D/setup.py#L53-L57

I think we should switch our code to using code like is currently in the jlab extension cookiecutter.

davidbrochart:
Fixes #3090

### DIFF
--- a/jupyterlab_widgets/setup.py
+++ b/jupyterlab_widgets/setup.py
@@ -6,6 +6,7 @@ import os
 from jupyter_packaging import (
     create_cmdclass, install_npm, ensure_targets,
     combine_commands, ensure_python, get_version,
+    skip_if_exists
 )
 import setuptools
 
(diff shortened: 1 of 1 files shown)


jtpio:
Nice, thanks!

I wonder whether we should add some CI jobs like in `voila` and `bqplot` that check we can install from the `sdist`, and the lab extension is correctly installed.

For example this one: https://github.com/voila-dashboards/voila/blob/master/.github/workflows/packaging.yml

The downside would be that it would require copy pasting the workflow here in this repo too. But iirc it helped catch some packaging issues for example when distributing lab extensions on conda-forge.

jtpio:
> I wonder whether we should add some CI jobs like in `voila` and `bqplot` that check we can install from the `sdist`, and the lab extension is correctly installed.

We can do that in a separate PR.

jtpio (approved):
Thanks!

PR #3120, opened 2021-02-15
Use python_requires >=3.5

### LINKED ISSUE #3067: Use python_requires to check the python version  so py2.py3 wheels are not built and pip will work correctly.
### Problem

* The wheels being built for ipywidgets claim it is python 2 compatible. eg ipywidgets-7.6.3-py2.py3-none-any.whl
* Modern versions of pip do not understand that ipywidgets is not python 2 compatible. setup.py is not run when downloading a wheel.

So it installs it without error on python 2.

I believe the  py2.py3 wheels are being built because the _python_requires_ argument is not being passed to setup but I do not know that for sure.

### Proposed Solution

Add a python_requires argument to the setup call to solve it.

EG Add this to [setup_args in setup.py](https://github.com/jupyter-widgets/ipywidgets/blob/90c1c14211a6be685c6077646f40973f96faced1/setup.py#L67) 

```python
python_requires = '>=3.5',
```

Then you can remove this
```python
#-----------------------------------------------------------------------------
# Minimal Python version sanity check
#-----------------------------------------------------------------------------

import sys

v = sys.version_info
if v[:2] < (3, 5):
    error = "ERROR: %s requires Python version 3.5 or above." % name
    print(error, file=sys.stderr)
    sys.exit(1)

```

This issue from another project has a very good description of the problem and how to solve it

https://github.com/PLOS/allofplos/issues/20

davidbrochart:
Fixes #3067

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,21 +30,10 @@ Usage
 """
 
 #-----------------------------------------------------------------------------
-# Minimal Python version sanity check
-#-----------------------------------------------------------------------------
-
-import sys
-
-v = sys.version_info
-if v[:2] < (3, 5):
-    error = "ERROR: %s requires Python version 3.5 or above." % name
-    print(error, file=sys.stderr)
-    sys.exit(1)
-
-#-----------------------------------------------------------------------------
 # get on with it
 #-----------------------------------------------------------------------------
 
+import sys
 import os
 from distutils.core import setup
 from distutils.command.build_py import build_py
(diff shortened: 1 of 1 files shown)


willingc (approved):

willingc:
Version 8.0 of widgets should drop 3.5 as it has reached [End-of-Life](https://devguide.python.org/#status-of-python-branches) and lacks good support for async and does not have f-strings.

jasongrout:
> Version 8.0 of widgets should drop 3.5 as it has reached [End-of-Life](https://devguide.python.org/#status-of-python-branches) and lacks good support for async and does not have f-strings.

This is now https://github.com/jupyter-widgets/ipywidgets/pull/3131

PR #3122, opened 2021-02-16
Privatize Widgets.widget

### LINKED ISSUE #2944: Widgets are not unique when stored in a dict in sub-class
This may be a python feature rather than an ipywidgets bug, but thought it was worth pointing out in case someone else has a similar issue and can't figure it out.

I am subclassing `HBox` and `VBox` to create self-contained "combo-widgets". However, I noticed that when I made independent instances of my combo-widgets, the widgets they contained seemed to be connected. It turns out that because I was storing the widgets in a `dict` the widgets were pointing to the same address.

Here is an MCVE:
```python
import ipywidgets as widgets
class myBox(widgets.HBox):
    def __init__(self):
        self.widgets['innerbox']=widgets.Text()
        super().__init__(children=[self.widgets['innerbox']])
b1 = myBox()
print(id(b1.widgets['innerbox']))
b2 = myBox()
print(id(b1.widgets['innerbox']), id(b2.widgets['innerbox']))
```
Result is
```
140583624309712
140583623982352 140583623982352
```
It doesn't happen if I store the inner widget like so
```python
        self.innerbox=widgets.Text()
```
if I give the inner widgets different arguments every time they are instantiated they still end up pointing to the same place
```python
import ipywidgets as widgets
import uuid
class myBox(widgets.HBox):
    def __init__(self):
        self.id = str(uuid.uuid1())
        self.widgets['innerbox']=widgets.Text(description=str(self.id))
        super().__init__(children=[self.widgets['innerbox']])
b1 = myBox()
print(b1.id, b1.widgets['innerbox'])
b2 = myBox()
print(b2.id, b2.widgets['innerbox'])
print(b1.id, b1.widgets['innerbox'])
print(id(b1.widgets['innerbox']), id(b2.widgets['innerbox']))
```
```
dfa1e66c-d865-11ea-bd85-fa163e8ac59c Text(value='', description='dfa1e66c-d865-11ea-bd85-fa163e8ac59c')
dfa587f4-d865-11ea-a034-fa163e8ac59c Text(value='', description='dfa587f4-d865-11ea-a034-fa163e8ac59c')
dfa1e66c-d865-11ea-bd85-fa163e8ac59c Text(value='', description='dfa587f4-d865-11ea-a034-fa163e8ac59c')
140583624267536 140583624267536
```

ianhi:
Closes: https://github.com/jupyter-widgets/ipywidgets/issues/2944

Using such a generic name could lead to collisions with user code as in the linked issue. Here it is changed to a name far less likely to collide with a user subclass and privatized. I'm less sure about making it private but it seems as though doing this will increase future flexibility and clean up what shows up when tab completing on a widget subclass.

I also gave the same treatment to `widget_types` for consistency.

### DIFF
--- a/ipywidgets/embed.py
+++ b/ipywidgets/embed.py
@@ -129,7 +129,7 @@ def _get_recursive_state(widget, store=None, drop_defaults=False):
 
 def add_resolved_links(store, drop_defaults):
     """Adds the state of any link models between two models in store"""
-    for widget_id, widget in Widget.widgets.items(): # go over all widgets
+    for widget_id, widget in Widget._active_widgets.items(): # go over all widgets
         if isinstance(widget, Link) and widget_id not in store:
             if widget.source[0].model_id in store and widget.target[0].model_id in store:
                 store[widget.model_id] = widget._get_embed_state(drop_defaults=drop_defaults)
@@ -207,7 +207,7 @@ def embed_data(views, drop_defaults=True, state=None):
         view_specs: a list of widget view specs
     """
     if views is None:
-        views = [w for w in Widget.widgets.values() if isinstance(w, DOMWidget)]
+        views = [w for w in Widget._active_widgets.values() if isinstance(w, DOMWidget)]
     else:
         try:
             views[0]
--- a/ipywidgets/tests/test_embed.py
+++ b/ipywidgets/tests/test_embed.py
@@ -29,7 +29,7 @@ class CaseWidget(Widget):
 class TestEmbed:
 
     def teardown(self):
-        for w in tuple(Widget.widgets.values()):
+        for w in tuple(Widget._active_widgets.values()):
             w.close()
 
     def test_embed_data_simple(self):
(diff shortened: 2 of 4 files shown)


jasongrout:
Thanks!

vidartf:
So is this still safe to be considered "protected" then? I.e. subclasses can rely on this property being available with the current variable name until the next major release ?

vidartf:
I mean, the fact that the embed code now access a "private" variable seems hacky.

ianhi:
> So is this still safe to be considered "protected" then? I.e. subclasses can rely on this property being available with the current variable name until the next major release ?

> I mean, the fact that the embed code now access a "private" variable seems hacky.


My understanding of private variables was that it's ok for anything in the library they are in to use them, including subclasses. In which case it is not hacky for the embed code to access it. Is that different than the canonical view? I'm totally happy to revisit this if you have a suggestion for what a better approach is.

pllim:
Looks like this was "privatized" without any deprecation period. Broke our code downstream. FYI.

jasongrout:
> Looks like this was "privatized" without any deprecation period. Broke our code downstream. FYI.

Thanks for pointing this out. At the very least, we should note this in the migration documentation.

maartenbreddels:
Yeah, I think `.widgets` is used widely, and this will at least break a lot of my code. I understand the change, but maybe we could at least provide a class based getter such as in:
https://github.com/widgetti/react-ipywidgets/commit/6fb07845d143f02165c11405ed5bec20391b09c5

If a subclass decides to use `.widgets', yes it will break code that assumed it to the dict of widgets, but it will make most existing code out there still work.

PR #3123, opened 2021-02-16
Add a workflow to test the packaging

jtpio:
Follow-up to #3119 

This adds a new workflow on CI to test installing `ipywidgets` and `jupyterlab_widgets` with the whell and the sdist.

This can help check the extensions are correctly bundled and installed.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -42,6 +42,13 @@ jobs:
         cd jupyterlab_widgets
         python setup.py sdist bdist_wheel
         cp ./dist/* ../dist
+    - name: Build PyPI distributions for widgetsnbextension
+      run: |
+        jlpm
+        jlpm run build
+        cd widgetsnbextension
+        python setup.py sdist bdist_wheel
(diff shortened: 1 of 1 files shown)


jasongrout:
It seems the matrix was not actually working the way we thought - the dist was overwriting itself, instead of adding more possibilities. I think I fixed it...

jasongrout:
@jtpio - I think this is good now. I've added a bunch to the PR, so it would be great if you could look over it again. If it looks good to you, then +1 from me if you want to merge it.

jasongrout (approved):

jasongrout:
Summary of changes:

- made sure that the ipywidgets wheel and sdist, as well as widgetsnbextension and jupyterlab_widgets sdists were tested
- Streamlined to just windows and linux, python 3.9, except we also test the ipywidgets sdist on python 3.6 windows and linux. I figured the macOS was similar enough to the linux we didn't need to test that as well.
- Added caching for the installs

jtpio:
Thanks Jason, looks good!

PR #3124, opened 2021-02-17
Add event listener for resize events

ibdafna:
This PR adds an event listener so that `lumino` based widgets rendered in a classic notebook can receive a resize message when a window resize event triggers.

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -915,18 +915,6 @@ export class DOMWidgetView extends WidgetView {
           oldLayoutView.remove();
         }
 
-        const resizeFunc = () => {
-          MessageLoop.postMessage(
-            this.pWidget,
-            Widget.ResizeMessage.UnknownSize
-          );
-        };
-
-        window.addEventListener('resize', resizeFunc);
-        this.once('remove', () => {
-          window.removeEventListener('resize', resizeFunc);
-        });
-
         return this.create_child_view(layout)
           .then(view => {
             // Trigger the displayed event of the child view.
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -915,6 +915,18 @@ export class DOMWidgetView extends WidgetView {
           oldLayoutView.remove();
         }
 
+        const resizeFunc = () => {
+          MessageLoop.postMessage(
+            this.pWidget,
+            Widget.ResizeMessage.UnknownSize
+          );
+        };
+
+        window.addEventListener('resize', resizeFunc);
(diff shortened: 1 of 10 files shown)


jasongrout:
Fixes https://github.com/jupyter-widgets/ipywidgets/issues/2467

packages/base/src/widget.ts
```diff
@@ -895,6 +895,18 @@ export class DOMWidgetView extends WidgetView {
           oldLayoutView.remove();
         }
 
+        const resizeFunc = () => {
```
jtpio, line 898:
```suggestion
        const resizeFunc = (): void => {
```

jasongrout:
@ibdafna, thanks for looking into this.

I think we should register the handler in the classic notebook widget manager *once*, then when it is triggered, have the classic notebook manager go through each of the top-level widgets and trigger the lumino event. And probably do the same for the html manager.

The way things are set up right now in this PR, it looks like every widget will register a handler with window (whether it needs to or not).

@ibdafna - I'm happy to connect over video to walk through this if you want.

ibdafna:
> 
> 
> @ibdafna, thanks for looking into this.
> 
> I think we should register the handler in the classic notebook widget manager _once_, then when it is triggered, have the classic notebook manager go through each of the top-level widgets and trigger the lumino event. And probably do the same for the html manager.
> 
> The way things are set up right now in this PR, it looks like every widget will register a handler with window (whether it needs to or not).
> 
> @ibdafna - I'm happy to connect over video to walk through this if you want.

Thank you - yes, I'll reach out!

packages/html-manager/src/htmlmanager.ts
```diff
@@ -32,6 +34,16 @@ export class HTMLManager extends ManagerBase {
       },
       0
     );
+
+    this._viewList = new Set<DOMWidgetView>();
+    window.addEventListener('resize', () => {
+      this._viewList.forEach(view => {
+        MessageLoop.postMessage(
+          view.lmWidget,
+          Widget.ResizeMessage.UnknownSize
```
jasongrout, line 43: Likely this will also need `LuminoWidget.Widget` instead of just `Widget` like in the code below

jasongrout (commented):
Thanks! I found a few things while testing, noted below.

jasongrout:
FYI, I rebased on master

jasongrout:
@ibdafna - I rebased and fixed a few issues with the examples, and checked manually that the resize event is now happening for the classic notebook and the html manager. Thanks again!

jasongrout (approved):

PR #3127, opened 2021-02-17
Make sure buffer is a DataView

### LINKED ISSUE #2588: Factor out state deserialization logic
See https://github.com/jupyter-widgets/ipywidgets/pull/2570#discussion_r335501207 for more details.

davidbrochart:
Fixes #2588

### DIFF
--- a/packages/base/src/utils.ts
+++ b/packages/base/src/utils.ts
@@ -100,18 +100,25 @@ export function reject(message: string, log: boolean) {
 export function put_buffers(
   state: Dict<BufferJSON>,
   buffer_paths: (string | number)[][],
-  buffers: DataView[]
+  buffers: any[]
 ): void {
   for (let i = 0; i < buffer_paths.length; i++) {
     const buffer_path = buffer_paths[i];
-    // say we want to set state[x][y][z] = buffers[i]
+    // make sure the buffers are DataViews
+    let buffer = buffers[i];
+    if (!(buffer instanceof DataView)) {
+      buffer = new DataView(
+        buffer instanceof ArrayBuffer ? buffer : buffer.buffer
+      );
+    }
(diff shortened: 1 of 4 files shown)


vidartf:
This code should probably be made DRY as well then?

https://github.com/jupyter-widgets/ipywidgets/blob/51322341d4f6d88449f9dbf6d538c60de6d23543/packages/base-manager/src/manager-base.ts#L160-L168

packages/base/src/utils.ts
```diff
@@ -100,18 +100,25 @@ export function reject(message: string, log: boolean) {
 export function put_buffers(
   state: Dict<BufferJSON>,
   buffer_paths: (string | number)[][],
-  buffers: DataView[]
+  buffers: any[]
```
vidartf, line 103: `any` seems overly permissive. AFAICT, this should be a union of:
- `DataView`
- `ArrayBuffer`
- a generic type that includes a `buffer` attribute of the correct type (`ArrayBuffer`). This would catch TypedArrays.

davidbrochart: Thanks, we also need to add `ArrayBufferView` in the union.

davidbrochart:
> This code should probably be made DRY as well then?
> 
> https://github.com/jupyter-widgets/ipywidgets/blob/51322341d4f6d88449f9dbf6d538c60de6d23543/packages/base-manager/src/manager-base.ts#L160-L168

Good catch, and here too:
https://github.com/jupyter-widgets/ipywidgets/blob/51322341d4f6d88449f9dbf6d538c60de6d23543/packages/base/src/widget.ts#L219-L228

vidartf (approved):
Looks good! Thanks 👍

jasongrout:
Thanks @davidbrochart, and thanks for the review @vidartf.

PR #3131, opened 2021-02-18
Drop python 3.5 support

jasongrout:
As @willingc said in https://github.com/jupyter-widgets/ipywidgets/pull/3120#issuecomment-779353535: “Version 8.0 of widgets should drop 3.5 as it has reached [End-of-Life](https://devguide.python.org/#status-of-python-branches) and lacks good support for async and does not have f-strings.”


jasongrout:
The spec failure is discussed #3110

PR #3142, opened 2021-02-25
Add handle_color style key to SelectionSlider

jpn--:
The SelectionSlider widget has a very similar interface to the FloatSlider and IntSlider, but is missing the handle_color style.  This adds it.  Seems to work for me, but I'm not super familiar with UI coding and don't know how to write tests for this.  I don't see tests for this feature implemented similarly on the FloatSlider.  If this needs a test, please point me to a test for something similar and I can probably figure it out.  Thanks!

### DIFF
--- a/ipywidgets/widgets/widget_selection.py
+++ b/ipywidgets/widgets/widget_selection.py
@@ -15,6 +15,7 @@ from .widget_core import CoreWidget
 from .widget_style import Style
 from .trait_types import InstanceDict, TypedTuple
 from .widget import register, widget_serialization
+from .widget_int import SliderStyle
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Thanks! This looks good to me. To make the tests pass, you need to update the latest model spec for the selection slider in https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema.

SylvainCorlay:
(could you also fix the JSON schema file?)

jpn--:
Seems like it worked.  Hooray!

ibdafna (approved):
LGTM. Thank you!

PR #3146, opened 2021-03-03
Rename widget>jupyter-widget, but maintaining compatibility

### LINKED ISSUE #2896: More specific namespacing for CSS classes?
Right now our widget CSS classes are namespaced as `.widget-*`. `.widget` is pretty generic. For example, #2893 showed at least one conflict with another library.

Perhaps we should change the namespace to something like `.jw-*`, which seems more specific to 'jupyter widgets' and less generic than `.widget`

marimeireles:
Hey volks
This relates to https://github.com/jupyter-widgets/ipywidgets/issues/2896

In the issue Jason says we should rename it to `jp-widget`, but I thought `jupyter-widget` conformed more with the rest of the names we had. If it's an issue, I can fix it.

I don't really know how to inform users that we're changing the name of the CSS classes. Should I add it somewhere in the docs?

Also, I'm far from being knowledgeable in CSS but I noticed that using the vanilla stuff is actually a bit annoying to maintain. I opened an [issue](https://github.com/jupyter-widgets/ipywidgets/issues/3150) so we can discuss using sass.


jtpio:
Thanks Mariana!

packages/controls/css/lumino.css
```diff
@@ -63,23 +63,27 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  * We've scoped the rules so that they are consistent with exactly our code.
  */
 
-.jupyter-widgets.widget-tab > .p-TabBar {
+.jupyter-widgets.widget-tab > .p-TabBar,
+.jupyter-widgets.jupyter-widget-tab > .p-TabBar {
```
jtpio, line 67: Wondering whether we want to deprecate the previous use of `.widget`.

For reference Lumino added some `<DEPRECATED>` comments for the switch from `p-Widget` to `lm-Widget`: 

https://github.com/jupyterlab/lumino/blob/e29f3c7172d2a256046e7fc509d8bacbb9093aa5/packages/widgets/style/widget.css#L11

marimeireles: I can do that.
I've never deprecated css before, so... don't really have better ideas.

jasongrout:
This looks great, thanks!

I merged in master to take care of some conflicts, and fixed two typos, and now tests all pass.

PR #3148, opened 2021-03-04
Add action to comment on PRs with a binder link

ianhi:
This PR will add an action that (hopefully) will comment on PRs with a binder link to allow easy testing of the PR.


jasongrout:
Awesome, let's try it out!

jtpio:
Nice, thanks!

Should we also provide a `postBuild` file to explicitely install `jupyterlab_widgets` and `widgetsnbextension` from source?

It looks like they are installed from PyPI otherwise?

https://github.com/jupyter-widgets/ipywidgets/blob/98e6c86c82b62437a28693581671ee4ca99ac874/setup.cfg#L42-L43

PR #3149, opened 2021-03-04
test binder bot

ianhi:
test to see if https://github.com/jupyter-widgets/ipywidgets/pull/3148 worked


ianhi:
looks like it works! hooray

image

I'm going to delete the branch now so that link wont actually work in the future, but I promise it really did work.

PR #3155, opened 2021-03-10
Pin to ipykernel<6 for the docs

### LINKED ISSUE #3154: Doc build Failing
## Description
See https://readthedocs.org/projects/ipywidgets/builds/13200690/

jtpio:
Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3154

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 bqplot
-ipykernel
+ipykernel<6
 jupyter_client
 matplotlib
 nbsphinx>=0.5


jtpio:
Looking at the RTD logs, it looks like it was pulling `ipykernel==6.0.0a0`:

image

jtpio:
cc @ianhi I think we could keep this pin until `ipykernel` stabilizes.

ianhi:
Using ipywidgets fails for me with ipykernel 6a. So maybe this isn't the correct fix?

ianhi:
See https://github.com/jupyter-widgets/ipywidgets/issues/3157

Should we merge this anyway in order to have the docs work on any new PRs?

jtpio:
> Using ipywidgets fails for me with ipykernel 6a. So maybe this isn't the correct fix?

Isn't it the same issue? Since RTD was pulling `ipykernel==6.0.0a0` (probably because of `--pre` somewhere)

> Should we merge this anyway in order to have the docs work on any new PRs?

It should be ready to go yes, and we can continue tracking this in https://github.com/jupyter-widgets/ipywidgets/issues/3157.

ianhi:
> Isn't it the same issue? Since RTD was pulling ipykernel==6.0.0a0 (probably because of a --pre somewhere)

Yeah the same issue, but my thought was just that this was more so fixing a symptom rather than the root cause. 



> >    Should we merge this anyway in order to have the docs work on any new PRs?
> 
> It should be ready to go yes, and we can continue tracking this in #3157.

right on. thanks!

jtpio:
> Yeah the same issue, but my thought was just that this was more so fixing a symptom rather than the root cause.

Sure. I think the root cause is being investigated in https://github.com/ipython/ipykernel/issues/602

PR #3162, opened 2021-03-15
Update to TypeScript 4.3

jtpio:
Update the `typescript` dependency to the latest version.

Leaving as a draft for now as there might be some breaking changes to fix.


jtpio:
One of the changes introduced in TypeScript `4.0` is causing compilation to fail with:

```bash
Error: src/widget.ts(187,14): error TS2790: The operand of a 'delete' operator must be optional.
Error: src/widget.ts(193,12): error TS2790: The operand of a 'delete' operator must be optional.
Error: src/widget.ts(760,12): error TS2790: The operand of a 'delete' operator must be optional.
Error: src/widget.ts(798,12): error TS2790: The operand of a 'delete' operator must be optional.
```

https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#operands-for-delete-must-be-optional

Which corresponds to:

https://github.com/jupyter-widgets/ipywidgets/blob/f085739f9ed9706ace94158f13c03430393dc40c/packages/base/src/widget.ts#L187

jtpio:
Adding to the 8.0 milestone as this might result in changes in public interfaces and typings.

jtpio:
> One of the changes introduced in TypeScript `4.0` is causing compilation to fail with:

Another change is related to properties overriding accessors (TS2610) in 4.0:

https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#properties-overriding-accessors-and-vice-versa-is-an-error

For example:

```
Error: src/widget_box.ts(139,3): error TS2610: 'pWidget' is defined as an accessor in class 'DOMWidgetView', but is overridden here in 'BoxView' as an instance property.
```

jasongrout:
I updated to TS 4.3 and fixed the errors.

jasongrout:
@jtpio: your commits look good. Can you review my commits, and merge if they look good?

jasongrout:
(or @vidartf or anyone else, feel free to review all of our commits and merge if they look good :)

jtpio:
Thanks Jason for picking this up!

widgetsnbextension/package.json
```diff
@@ -29,7 +29,7 @@
     "@jupyterlab/services": "^6.0.0",
     "@lumino/messaging": "^1.3.3",
     "@lumino/widgets": "^1.11.1",
-    "backbone": "1.2.3"
+    "backbone": "1.4.0"
```
jtpio, line 32: Just to double check: this `backbone` update was required to grab the new `preinitialize` method? (and then fix the TS error)

jtpio: Yes looks like it's mentioned in the change log on https://backbonejs.org:

image

jtpio:
Just checked the diff and it looks good :+1: Thanks!

Leaving it open one more day in case someone else also wants to have a look, and then we should be able to merge it.

jtpio:
Thanks!

PR #3167, opened 2021-03-23
Update Output Widget.ipynb

gsteele13:
Added some documentation about using Output widgets with interact, see this issue:

https://github.com/jupyter-widgets/ipywidgets/issues/1919

### DIFF
--- a/docs/source/examples/Output Widget.ipynb
+++ b/docs/source/examples/Output Widget.ipynb
@@ -391,6 +391,79 @@
    "source": [
     "thread.join()"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Using the output widget with matplotlib\n",
+    "\n",
+    "You can also use the output widget it self to build your own interaction with graphical outputs such as matplotlib. \n",
+    "\n",
+    "When using matplotlib, it is important that your update function clears the output widget state using the `wait=True` flag to prevent flickering of the display while interacting. A simple example:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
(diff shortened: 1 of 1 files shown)


basnijholt:
Thanks! This helped me avoiding displaying multiple new plots.

basnijholt (approved):
LGTM!

ianhi:
Overall this look nice - my one worry is that it steers the reader away from things like `ipympl` which for my money is the best way to use matplotlib in the notebook.  But its not clear to me what the policy, if any, is for mentioning 3rd party widget libraries in the main docs. @jtpio is that something we should try to avoid?

jtpio:
Indeed there doesn't seem to be a clear guideline on whether the docs should include references to third-party widgets.

However the `ipywidgets` docs already have some examples with `bqplot` in https://ipywidgets.readthedocs.io/en/stable/examples/Layout%20Templates.html

`matplotlib` is a very popular library which is also used a lot in notebooks, so it could sense to have a small example like in this PR. And maybe there could also be a link encouraging folks to check out `ipympl`?

gsteele13:
I would be totally fine with also mentioning that for matplotlib, ipympl is a possible more responsive alternative. 

Note that ipympl is not the full solution though: ultimately, if you want to interact with large colorscale plots, you will also start to lag, and likely want to throttle. 

(BTW, bokeh is *way* faster that the ipympl / widget driver, factor of 20 or more. But it requires many people, like me, to relearn a new plotting library API...)

And maybe a final point: *regardless* of what back end library you are trying to interact with, *eventually* you will hit lag issues for more complex interactions, and then you *really* want to throttle these. 

Ipywidgets itself is not linked to a particular thing it is interacting with, and if the thing it is interacting with is laggy (which it will be at some point) then you want to throttle. 

I am still a proponent actually of including the throttling directly into the `interact()` code itself, have been looking at it when I have spare time, but unfortunately, there is less of that than I would like to have :)

PR #3168, opened 2021-03-23
Unpin ipykernel<6 for the docs

jtpio:
Follow-up to https://github.com/jupyter-widgets/ipywidgets/pull/3155 to see if the new `6.0.0a1` release would fix the issue reported in https://github.com/jupyter-widgets/ipywidgets/issues/3154.

Reverts back to how it was before the pin added in https://github.com/jupyter-widgets/ipywidgets/pull/3155

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 bqplot
-ipykernel<6
+ipykernel
 jupyter_client
 matplotlib
 nbsphinx>=0.5


jtpio:
Same kind of failure as in #3154: https://readthedocs.org/api/v2/build/13318670.txt

jtpio:
Now passing with the latest `ipykernel` prerelease.

jtpio:
Looks like we can then remove the pin

PR #3169, opened 2021-03-23
Complete docstring for `interactive`.

fperez:
Include description of `auto_display` that is also accepted, and add description of defaults.


jasongrout:
Thanks!

Issue #3170, opened 2021-03-25
Image Widget memory leak when a lot of updates


joseph2rs:
## Description

Browser memory leak caused by Image Widget not revoking the ObjectURLs.

## Reproduce

- Create an Image Widget from an array (not from a url)
- Perform a lot of updates of the image (for example in a loop)

## Context

- ipywidgets version 7.6.3
- Operating System and version: Ubuntu 20, Windows 10
- Browser and version: Firefox 86, Chomium 89

Bug is here: https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/controls/src/widget_image.ts#L65

At each update, a `URL.createObjectURL` is performed. As a result, the browser is "storing" the blob in memory. But the memory is never cleaned, because `URL.revokeObjectURL` is never called

```js
    // Clean up the old objectURL
    const oldurl = this.el.src;
    this.el.src = url;
    if (oldurl && typeof oldurl !== 'string') {
      URL.revokeObjectURL(oldurl);
    }
```

`oldurl` is always a string, so the `revokeObjectURL` is never called.

### If using JupyterLab
- JupyterLab version: 3.0.10

ianhi:
Fixed by https://github.com/jupyter-widgets/ipywidgets/pull/3171

PR #3171, opened 2021-03-25
fix: [Widget Image] Remove oldurl string check since it's always a string and prevent the ObjectURL cleaning resulting in a memory leak.

joseph2rs:
See #3170

### DIFF
--- a/packages/controls/src/widget_audio.ts
+++ b/packages/controls/src/widget_audio.ts
@@ -62,8 +62,9 @@ export class AudioView extends DOMWidgetView {
     // Clean up the old objectURL
     const oldurl = this.el.src;
     this.el.src = url;
-
-    URL.revokeObjectURL(oldurl);
+    if (oldurl) {
+      URL.revokeObjectURL(oldurl);
+    }
 
     // Audio attributes
     this.el.loop = this.model.get('loop');
--- a/packages/controls/src/widget_audio.ts
+++ b/packages/controls/src/widget_audio.ts
@@ -62,9 +62,8 @@ export class AudioView extends DOMWidgetView {
     // Clean up the old objectURL
     const oldurl = this.el.src;
     this.el.src = url;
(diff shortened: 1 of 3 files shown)


jasongrout:
Thanks!

This check was added in 1b1639118c1b8bde71e8691b5aa6e59887af9e4f (#1676) to distinguish between that src being an external url or an internal object url.

However, it looks like [step 2](https://w3c.github.io/FileAPI/#dfn-revokeObjectURL) of the revoking process checks to see if the url has scheme `blob:`, so I think we can call revoke on any URL (including external URLs) and it will silently return if the url is not a blob url.

I see that we have this same logic in the video and audio widgets. Can you fix it in those files too?

jasongrout:
FYI, on current firefox and chrome, `URL.revokeObjectURL('http://google.com')` just silently returns without an error message, so I think we don't even have the error message the standard warns may happen.

jasongrout:
Other places are at https://github.com/jupyter-widgets/ipywidgets/blob/1a121620aea08b2df4bcf92e851bd961317dc1db/packages/controls/src/widget_audio.ts#L65-L67 and https://github.com/jupyter-widgets/ipywidgets/blob/1a121620aea08b2df4bcf92e851bd961317dc1db/packages/controls/src/widget_video.ts#L68-L70

joseph2rs:
FYI i removed also the check on `oldurl` since passing undefined or null fail silently.

jasongrout:
> FYI i removed also the check on `oldurl` since passing undefined or null fail silently.

Reading the standard, we aren't guaranteed passing in undefined or null will always fail silently, just that a url that is not an object url will. In the standard, the first step is to parse the url, and the parsing standard seems to really want a string.

Can you put back in the oldurl test so we're following the standard more closely? That will make it more certain it will work across browsers, at least ones that implement the spec as written.

jasongrout:
Thanks!

PR #3172, opened 2021-03-25
Visual regression testing using Galata

mbektas:
- adds visual regression testing support using https://github.com/jupyterlab/galata
- test suite uploads and runs a notebook, that has code cells creating different widget instances, in JupyterLab using headless browser
- cell output image captures are taken and compared to reference images
- test output artifacts are saved and accessible on workflow page

(https://github.com/mbektasbbg/ipywidgets/pull/4 demonstrates detecting visual regression caused by change to widget label rendering)

@ellisonbg @blink1073 @afshin FYI, this is the first integration of Galata in a Jupyter project! More to come.


mbektasbbg:
Sample UI test result report generated as artifact
ipywidgets galata report

and diff related to failed test
widgets_cell_0

blink1073 (approved):
Love it!

jtpio:
Looks great, thanks @mbektasbbg for this!

ui-tests/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "ipywidgets-ui-tests",
```
jtpio, line 2: Wondering whether this new package should be added to the existing `lerna` packages? Then we might be able to do without `ui-tests/package-lock.json` and keep dependencies in a top-level `yarn.lock`.

Or is it better to keep the `ui-tests` package a bit more separated from the rest?

mbektasbbg: @jtpio good point. Initially I tried adding as dependency to top level package.json. I had some build issues then since rest of the project is using `chai` for unit testing and that was causing name clashes with `jest` that Galata uses (names like `it`, `test`, `expect`). Also, having a separate `ui-tests` directory & package decouples things nicely. Only the developers that add UI tests need to install the dependencies (which include browser binaries) and UI tests can work with different jupyterlab versions than ipywidgets is dependent on.

mbektasbbg: I will add some documentation to guide on adding new UI test cases, building and testing locally.

jtpio: Ah right, there can indeed be some clashes (also since galata is built for a specific version of lab).

.github/workflows/tests.yml
```diff
@@ -143,3 +143,79 @@ jobs:
...
+        jupyter labextension develop . --overwrite
+        jupyter labextension list
+    - name: Install Galata
+      run: |
+        cd ui-tests
+        npm install
+    - name: Launch JupyterLab
+      run: |
+        cd ui-tests
+        npm run start-jlab:detached
```
jtpio, line 202: Since `yarn` is currently used for the dev setup of ipywidgets, should we also use it for the `ui-tests` directory?

jtpio: Just pushed 1fe8d6d1366239e278929a6e9d6857f3fcc7fb60 for this.

jtpio (approved):
Thanks!

Issue #3178, opened 2021-04-13
More helpful semver range message


jasongrout:
### Problem

When this error is shown at the console: https://github.com/jupyter-widgets/ipywidgets/blob/e0d41f6f02324596a282bc9e4650fd7ba63c0004/jupyterlab_widgets/src/manager.ts#L270
it would be helpful to know what versions of that module actually *are* registered in the frontend, instead of just seeing that the requested version is not registered.

### Proposed Solution

Add a method to the semver cache at https://github.com/jupyter-widgets/ipywidgets/blob/master/jupyterlab_widgets/src/semvercache.ts to list all available versions of a given key, i.e., could just return `Object.keys(this._cache[key])` or something like that (and would probably have to deal with the case where `key` was not in `this._cache`). Then add this list of available versions to the error message above.

bsyouness:
I'll get started on this now! :)

jtpio:
Thanks @bsyouness!

vidartf:
Closed in #3185

PR #3185, opened 2021-04-22
More helpful semver range message

bsyouness:
As per #3178


jupyterlab_widgets/pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["jupyter_packaging~=0.7.9", "jupyterlab~=3.0", "setuptools>=40.8.0", "wheel"]
```
ianhi, line 2: Why these changes? I believe that the pin on `jupyter_packaging` is important as the API for that package is subject to change.

bsyouness: The build was broken before these changes (see commit 2b9a2d74). I didn't want to pin a version of jupyter_packaging but 7.9 doesn't have get_version_info, so I just left it out. I can add a TODO to figure out what version of jupyter_packaging to use given that it's out of scope for this PR. (I think get_version_info was introduced with 9.0.)

jtpio: Maybe we could then update to jupyter-packaging 0.9 in a separate PR so it's easier to track?

jtpio: Opened https://github.com/jupyter-widgets/ipywidgets/pull/3194

bsyouness: Thanks!

bsyouness:
@jasongrout now that #3194 is merged, this should be ready to be merged as well.

jasongrout:
Thanks. Adding it to my queue (but @jtpio and @ianhi, feel free to merge if you've already reviewed this)

jupyterlab_widgets/src/semvercache.ts
```diff
@@ -28,6 +28,12 @@ export class SemVerCache<T> {
     }
   }
 
+  getAllModules(key: string): Record<string, any> | undefined {
```
jasongrout, line 31: Can we change this to `getAllVersions`? I think that would be a bit clearer about what is being returned.

jasongrout (approved):
This looks good, with one change to the API to be a bit clearer (in my opinion - if you don't like the change, please discuss).

bsyouness:
There is something broken with the auto-linting. I only pushed my changes with the new name for `getAllVersions` and it triggered an auto-linting event and the linter check complains about the changes that were made by the auto-linter...

jtpio:
> There is something broken with the auto-linting. I only pushed my changes with the new name for `getAllVersions` and it triggered an auto-linting event and the linter check complains about the changes that were made by the auto-linter...

`prettier` was updated in https://github.com/jupyter-widgets/ipywidgets/pull/3162. So a git clean and reinstalling the dependencies from scratch should hopefully help.

bsyouness:
Should be good to merge now.

jtpio (approved):
Thanks!

PR #3194, opened 2021-05-04
Update to Jupyter Packaging 0.10

jtpio:
Similar to the update in the TypeScript cookiecutter: https://github.com/jupyterlab/extension-cookiecutter-ts/pull/143/files

We might also want to update `widgetnbextension`:

 https://github.com/jupyter-widgets/ipywidgets/blob/e0d41f6f02324596a282bc9e4650fd7ba63c0004/widgetsnbextension/pyproject.toml#L4

And: https://github.com/jupyter-widgets/ipywidgets/blob/master/widgetsnbextension/setup.py

Originally brought in https://github.com/jupyter-widgets/ipywidgets/pull/3185#pullrequestreview-650731089

### DIFF
--- a/widgetsnbextension/setup.py
+++ b/widgetsnbextension/setup.py
@@ -5,12 +5,9 @@
 # Distributed under the terms of the Modified BSD License.
 
 from jupyter_packaging import (
-    combine_commands,
-    create_cmdclass,
-    ensure_targets,
-    get_version,
-    install_npm,
-    skip_if_exists
+    wrap_installers,
+    npm_builder,
+    get_data_files
 )
 from pathlib import Path
 from setuptools import setup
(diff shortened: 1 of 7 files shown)


jtpio:
Need to check whether the test failure is related to https://github.com/jupyterlab/extension-cookiecutter-ts/issues/146

martinRenou:
We'll need to do that on all other widgets

jtpio:
Updating to `0.10` which added the `skip-if-exists` option back.

bsyouness (approved):
Looks good to me! Are we waiting for someone else to review this before merging?

jasongrout:
Closing and reopening again to trigger the tests

widgetsnbextension/setup.py
```diff
@@ -19,26 +16,19 @@
...
-    cmdclass["jsdeps"] = js_command
-else:
-    cmdclass["jsdeps"] = skip_if_exists(js_targets, js_command)
+cmdclass = wrap_installers(post_develop=post_develop)
 
 if __name__ == '__main__':
-    setup(cmdclass=cmdclass)
+    setup(
+        cmdclass=cmdclass,
+        data_files=get_data_files(data_files_spec),
```
jasongrout, line 33: @jtpio - I've been wondering about how jupyter_packaging handles data files that don't exist until after the build process. Part of the reason for the complicated build process before was that the data files could not be calculated until after the js was built. How are the data files retrieved with `get_data_files` before they are built now?

PR #3199, opened 2021-05-08
OutputWidget: use public API to get parent header

minrk:
getting the parent now has a public API in 6.0 (alpha), so use that when available instead of the private `Kernel._parent_header`, which is deprecated in favor of the public method now that we have it.

Added in https://github.com/ipython/ipykernel/pull/661


SylvainCorlay:
Thanks. This is a longstanding issue!

PR #3223, opened 2021-06-26
Fix example web3 missing process during runtime

smeng9:
There are some missing variables after upgrading to webpack v5.

This pull request follows https://github.com/jupyter-widgets/ipywidgets/pull/3211 to fix the missing process


jasongrout:
Thanks! This was also fixed in #3124.

PR #3229, opened 2021-07-03
Automatically close answered issues

jasongrout:
Often we’ll have a question which we answer to the best of our ability, and doesn’t look like it needs further follow-up. Sometimes the original author does not close the issue and also never responds. This builds up a lot of open issues that essentially have been answered or resolved. This workflow gives us a way to signal that a question is answered and will be automatically closed as answered unless there is further followup.

The configuration here is, for any issue with the ‘answered’ label, wait until there are 30 days of no activity, then apply the ‘closing as answered’ label for 7 days, then close the issue if there is still no further activity. A triage person can also directly assign the ‘closing as answered’ label to shortcut directly to the 7-day waiting period before closing.


jtpio (approved):
Thanks!

PR #3233, opened 2021-07-10
Update linting

jasongrout:
This updates our lint packages to latest versions and reruns the lint.


jasongrout:
This also starts running the prettier check in the CI test so we don't slowly drift out of sync.

jtpio (approved):
Thanks!

PR #3234, opened 2021-07-10
Fix spec generation for traitlets 5

jasongrout:
This fixes the spec generation for traitlets 5. Our tests work fine since they currently use traitlets 4.

Note that there is no traitlets 4 package on conda-forge for python 3.9.

This PR is independent of https://github.com/jupyter-widgets/ipywidgets/pull/3110, and is needed since the spec can no longer be generated in traitlets 5 for the new date time pickers in master.


jasongrout:
Looks like the binder test failure is not relevant: `Error: Unhandled error: HttpError: Resource not accessible by integration`

jasongrout:
Ping @vidartf to review this. Thanks!

PR #3246, opened 2021-08-06
Fixed typos

giswqs:
Fixed many typos using [codespell](https://github.com/codespell-project/codespell).


vidartf:
Thanks for this! My only question would be whether we should be updating typos in the specs for previous releases, i.e. anything less than v8. I'll bring it up in next week's dev meeting, and merge/update as needed 👍

vidartf (approved):

PR #3251, opened 2021-08-10
Add the ability to fully specify CDNs for loaded libraries

bsyouness:
This PR adds the ability to pass full URLs to resolve modules that we're loading in.

TODO: add documentation to say how to use the feature, e.g. add in html: <script data-jupyter-widgets-module-cdn="bqplot, XXX/share/jupyter/nbextensions/bqplot/index.js"></script>

TODO: figure out if we want to read `data-jupyter-widgets-module-cdn` first and if it can't be resolved check if the module is present locally (right now it does the opposite, i.e. checks if the module is present in the current directory first)

### DIFF
--- a/packages/html-manager/src/libembed-amd.ts
+++ b/packages/html-manager/src/libembed-amd.ts
@@ -5,13 +5,24 @@ import * as libembed from './libembed';
 
 let cdn = 'https://cdn.jsdelivr.net/npm/';
 let onlyCDN = false;
+const modulesCDN: { [key:string]: string } = {};
 
 // find the data-cdn for any script tag, assuming it is only used for embed-amd.js
 const scripts = document.getElementsByTagName('script');
 Array.prototype.forEach.call(scripts, (script: HTMLScriptElement) => {
   cdn = script.getAttribute('data-jupyter-widgets-cdn') || cdn;
   onlyCDN = onlyCDN || script.hasAttribute('data-jupyter-widgets-cdn-only');
+  if (script.hasAttribute('data-jupyter-widgets-module-cdn')) {
+    const CDNTuple = script.getAttribute('data-jupyter-widgets-module-cdn');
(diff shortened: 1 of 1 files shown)


bsyouness:
@jasongrout FYI

jasongrout:
Note: backport this to 7.x after merging.

jasongrout:
I've been thinking about this, and reminding myself how requirejs works (which is what is used in the html manager.

Would it be enough to just give a requirejs config on the page? It looks like we first try to use requirejs to load the module as-is, so presumably we could define this mapping of module to location as a requirejs map. The CDN is only used as a fallback (unless the cdn-only option is used, of course).

As a bonus, perhaps we could make an example that shows how to use the existing requirejs maps for installed nbextensions (which are providing amd modules for various custom widget libraries).

For example, for bqplot, if we wanted to load a locally-installed nbextension for it, we might be able to just either require `nbextensions/bqplot/extension` (which sets up the require config), or just manually specify the equivalent code manually:

```js
window.require&&window.require.config({map:{"*":{bqplot:"nbextensions/bqplot/index"}}})
```

If we didn't want a locally-loaded version, we could just give a different URL to load than `nbextensions/bqplot/index`, perhaps our favorite CDN.

jasongrout:
Slightly related: https://github.com/jupyter-widgets/widget-cookiecutter/issues/99 - examining the cookiecutter's hardcoding of CDN URLs.

jasongrout:
In practice, we found that setting the requirejs map as in https://github.com/jupyter-widgets/ipywidgets/pull/3251#issuecomment-903010796 was sufficient to accomplish the goals pursued here, so I'll close this.

jasongrout:
(Please correct me if I'm misremembering, @bsyouness...)

bsyouness:
I think that's right. @mnowacki-b FYI

PR #3261, opened 2021-08-29
Allow `Output.clear_output()` to be thread-safe.

### LINKED ISSUE #3260: `Output.clear_output()` in a thread does not block until the output is cleared
## Description

I create an `Output` object on the main thread and display something on it.  Then, in another thread, I try to clear the output and display other things on it.  There is a surprising race condition that causes some of the things in the new thread to not be displayed.  Specifically, the call to `out.clear_output()` does not clear the output right away, so subsequent calls to the Output object will get dropped once `clear_output` finally takes effect.

## Reproduce

From IPython notebook:
```python
import string
import threading
from ipywidgets import Output
import time

out = Output()
out.append_stdout('hello\n')
display(out)
```
```python
def bad_clear(out):
    out.clear_output()
    for c in string.ascii_letters:
        out.append_stdout(c)
        time.sleep(.01)

thread = threading.Thread(target=bad_clear, args=[out])
thread.start()
```
For me, the output from the first cell can be `jklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ`.  `abcdefghi` got cleared!

To work around this, I currently do the following:
```python
def good_clear(out):
    out.clear_output()
    while out.outputs:
        time.sleep(0.01)  # NEW
    for c in string.ascii_letters:
        out.append_stdout(c)
        time.sleep(.01)

thread = threading.Thread(target=good_clear, args=[out])
thread.start()
```
## Expected behavior

I expect the output of `out` to be cleared in `bad_clear` immediately after `out.clear_output()` is called.  If it's too troublesome to do this, then I at least expect a note in the docstring and the example added here: https://github.com/jupyter-widgets/ipywidgets/pull/1794

## Context

Previous relevant issues: https://github.com/jupyter-widgets/ipywidgets/pull/1794 https://github.com/jupyter-widgets/ipywidgets/issues/1722 https://github.com/jupyter-widgets/ipywidgets/pull/1752
- ipywidgets version: 7.6.3

eriknw:
Fixes #3260.

### DIFF
--- a/ipywidgets/widgets/widget_output.py
+++ b/ipywidgets/widgets/widget_output.py
@@ -72,9 +72,16 @@ class Output(DOMWidget):
         wait: bool
             If True, wait to clear the output until new output is
             available to replace it. Default: False
+
+        If any parameters are given, then this should be run on the main thread,
+        because race conditions may occur if run from multiple threads.  It is
+        safe to call this function without any arguments from multiple threads.
         """
-        with self:
-            clear_output(*pargs, **kwargs)
+        if not pargs and not kwargs:
+            self.outputs = ()
+        else:
+            with self:
+                clear_output(*pargs, **kwargs)
 
     # PY3: Force passing clear_output and clear_kwargs as kwargs
     def capture(self, clear_output=False, *clear_args, **clear_kwargs):


vidartf:
There are some intricacies with this change that makes it not so straight-forward. First of all, I think a clear alternative to this change exists: Document that users can set the `outputs` trait to empty as an alternative way of clearing outputs, and explain in which situations the two ways of clearing the outputs makes the most sense.

Getting more into details, I note the following points:
1. The content of an output widget can be modified in two ways:
  a. By manipulating the `outputs` trait, either directly, or via the convenience `append_*` methods on the output widgets.
  b. By entering its context manager, and using standard Jupyter protocol messages (e.g. via `IPython.display.display()` or `IPython.display.clear_output()` etc).
2. I think that mixing the two patterns of modifying the outputs can lead to some race-conditions, even without threads. I qualify this with a "I think that" as I do not have any repros to demonstrate it.
3. The context manager only works on the main thread.
4. The current proposal in this PR would have the implementation of the `clear_output` method change from one way of modifying the outputs to another depending on the signature which it is called by, i.e. `out.clear_output()` differs from `out.clear_output(wait=False)`. It would also change the default way used in most cases.
5. The `@out.capture(clear_output=True)` decorator uses `clear_output()`, which would make it likely to mix the two patterns, increasing the risk of races.

In sum, I think this change would be likely to make the behavior more mysterious/confusing. Instead I think we should more clearly document the current pattern and suggestions. An example for of a documentation paragraph could be:

    There are two different ways of changing the outputs of an Output widget:
    - By manipulating the `outputs` trait of the widget. This includes calling the `Output.append_*` methods.
    - By using it as a context manager (in a `with:` clause), and using standard methods for manipulating outputs, i.e. `IPython.display` etc.
    
    Mixing the two ways is not recommended, as it is likely to cause race conditions, giving unpredictable behavior.
    
    Note: For multi-threaded scenarios, only the first method is going to be effective, as the context manager requires that the code is running on the main thread.

We can then also add a note in the docstring of `clear_output` saying that it will not work from another thread than the main thread. Ideally we would also add a similar note for the context manager (`__enter__` method? `__init__`? class docstring?).

PR #3269, opened 2021-09-07
Set border property upon construction

ibdafna:
This PR fixes an issue where setting the border property in the constructor of a given widget did not reflect in the rendered output (property was not actually set).


jasongrout (approved):
Thanks!

PR #3271, opened 2021-09-08
Hold sync during set_state + fix selection widgets flakiness

### LINKED ISSUE #2259: Hold sync during set_state?
While reviewing #2256, I considered that maybe a more elegant solution would be to use `hold_sync` when setting the state from an update message. This would ensure that any new state changes as a direct response to the update (i.e. synchronous changes from validators or observers) all get collapsed into a single update message to the frontend. This collapsed update could then be verified against the property lock simplifying the logic.

A possible problem is that if multiple attributes change, then there is a potential for a change in the order of JS-side change notifications. For example:
- Previously a new state cause the updates: `{foo: 5}`, then `{bar: 1}`. This would trigger the `change:foo` and `change` events on the JS side first, with only foo updated. Then `change:bar` and `change` with bar updated.
- Now, the new state cause the update: `{foo: 5, bar: 1}`. This triggers `change:foo` and `change:bar` in undefined order, then a single `change` event, all with both foo and bar updated.

This could potentially be an issue for selections, e.g. [here](https://github.com/jupyter-widgets/ipywidgets/blob/463cb0444d1a2f31c3d5cacfe6b38d067f231406/packages/controls/src/widget_selection.ts#L154-L155). In this case, the view-code would need to always update the options and index in the same go, using `hasChanged` to confirm e.g. like this:

```js
initialize(parameters) {
    super.initialize(parameters);
    this.listenTo(this.model, 'change', (model, value, options) => this._updateSelection(options));
    // Create listbox here so that subclasses can modify it before it is populated in render()
    this.listbox = document.createElement('select');
}

updateSelection(options: any = {}) {
    // Debounce set calls from ourselves:
    if (options.updated_view === this) {
        return;
    }
    const optsChanged = this.hasChanged('_options_labels');
    const idxChanged = this.hasChanged('index');
    // The attributes we care about did not change, skip
    if (!idxChanged && !optsChanged) {
        return;
    }

    // Get the index first!
    const idx = this.model.get('index');

    if (optsChanged) {
        // Need to update options:
        ... // Code as before
    }
    // Always set the index
    this.listbox.selectedIndex = index === null ? -1 : index;
}
```

vidartf:
- Hold sync during set_state events. Fixes #2259.
- Add unit test highlighting flakiness of selection widget event code.
- Fix flakiness.

Open question: All `update` events will be passed the `options` argument by backbone, if given. Currently we do not reflect this in our base classes unless we use it. Are we happy with this, or do we want to add that? Adding it could lead to typing noise for third-party widgets which implement their own `update` function, and JS is happy to discard superfluous arguments to a function.

### DIFF
--- a/ipywidgets/widgets/tests/test_set_state.py
+++ b/ipywidgets/widgets/tests/test_set_state.py
@@ -236,8 +236,34 @@ def test_property_lock():
     widget.set_state({'value': 42})
     assert widget.value == 42
 
+    # we expect no new state to be sent
+    calls = []
+    widget._send.assert_has_calls(calls)
+
+def test_hold_sync():
+    # when this widget's value is set to 42, it sets the value to 2, and also sets a different trait value
+    class AnnoyingWidget(Widget):
+        value = Float().tag(sync=True)
+        other = Float().tag(sync=True)
+
+        @observe('value')
+        def _propagate_value(self, change):
+            print('_propagate_value', change.new)
+            if change.new == 42:
+                self.value = 2
+                self.other = 11
+
+    widget = AnnoyingWidget(value=1)
+    assert widget.value == 1
+
+    widget._send = mock.MagicMock()
+    # this mimics a value coming from the front end
+    widget.set_state({'value': 42})
+    assert widget.value == 2
+    assert widget.other == 11
+
     # we expect only single state to be sent, i.e. the {'value': 42.0} state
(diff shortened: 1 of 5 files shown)


martinRenou:
@maartenbreddels for awareness. Backward in https://github.com/jupyter-widgets/ipywidgets/pull/2256 and https://github.com/jupyter-widgets/ipywidgets/issues/2259

martinRenou (approved):
LGTM

PR #3272, opened 2021-09-08
Fix tagsinput initial display

### LINKED ISSUE #3230: tagsinput widget not visible
## Description

Currently on master, the tagsinput widget seems to be not visible in jupyterlab: 
Screen Shot 2021-07-03 at 2 51 21 AM

## Reproduce

1. Install dev ipywidgets using the dev-install script
2. Try to create a tagsinput widget and display it. Notice it does not show up. The DOM structure also seems off:
Screen Shot 2021-07-03 at 2 53 44 AM

## Expected behavior

See the tagsinput widget

## Context

- ipywidgets version: current master
- Operating System and version: macos catalina
- Browser and version: Firefox 89

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

No problems in the server log

No errors in the browser js console

### If using JupyterLab
- JupyterLab version: 3.0.16, with just ipywidgets dev installed

martinRenou:
Fix #3230

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -985,6 +985,48 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Tags input widget"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `TagsInput` widget is useful to for selecting/creating a list of tags. You can drag and drop tags to reorder them, limit them to a set of allowed values, or even prevent making duplicate tags."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tags = widgets.TagsInput(\n",
+    "    value=['pizza', 'fries'],\n",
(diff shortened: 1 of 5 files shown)


martinRenou:
Turning this into draft.

This fixes the issue as you can now interact with the tags input if it was empty at creation.

Though it still looks empty and "invisible" even though it's there. We should maybe add a placeholder inviting the user to add the first tag manually.

martinRenou:
I reused the CSS of the Text widget so that the input is visible.

I also added the ability to add a placeholder.

tagsinput

jtpio (approved):
Thanks!

Looks good when testing on Binder :+1:

jtpio:
By the way as a follow-up, maybe we could add the `TagsInput` and other widgets to the widget list notebook for convenience? 

https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html

Which are for now living in a separate example notebook: https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Widget%20TagsInput.ipynb

martinRenou:
Yes good idea. I will do that.

martinRenou:
Done @jtpio  :)

PR #3279, opened 2021-09-14
Update to the new Galata

### LINKED ISSUE #3274: Update to the latest Galata
With Galata now in the core JupyterLab repo (https://github.com/jupyterlab/jupyterlab/pull/10796), we should update the UI Tests setup here:

https://github.com/jupyter-widgets/ipywidgets/tree/master/ui-tests

To the new one as documented in: https://github.com/jupyterlab/jupyterlab/tree/master/galata

jtpio:
Fixes #3274 

Update to the latest `@jupyterlab/galata` now living in the JupyterLab repo: https://github.com/jupyterlab/jupyterlab/blob/master/galata

This makes full use of `playwright` and helps simplify the UI Tests setup.

### TODO

- [x] Update to `@jupyterlab/galata@~4.0.2`
- [x] Update CI workflow
- [x] Update reference snapshots
- [x] Update docs

The test report can be downloaded from the summary page:

image

When the CI check fails, new reference snapshots are generated and uploaded:

image


ui-tests/tests/notebooks/widgets.ipynb
```diff
@@ -6,15 +6,8 @@
...
-    "import ipywidgets as widgets"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
+    "import ipywidgets as widgets\n",
```
jtpio, line 9: Moving the import to the cell below, since Galata would otherwise block on trying to fetch the cell output with `galata.notebook.getCellOutput()` when there is no output.

jtpio:
Adding back to the 8.0 for consideration since it should now be ready to go.

.github/workflows/tests.yml
```diff
@@ -151,27 +151,30 @@ jobs:
...
-        python -m pip install jupyterlab==3.0.3
+        python -m pip install -U jupyterlab~=3.2 jupyter-packaging~=0.10
 
-    - name: Use Node.js 12.x
-      uses: actions/setup-node@v1
+    - name: Set up Node
+      uses: actions/setup-node@v2
       with:
-        node-version: 12.x
+        node-version: 14.x
```
vidartf, line 172: Is Node 12 no longer supported?

jtpio: It could be that node 12 was causing issues for this workflow. Although other workflows still use that version.

We could try to switch back to 12 and see if the check fails.

jtpio: The CI check failed after switching to node 12 in this commit: https://github.com/jupyter-widgets/ipywidgets/runs/4313698147?check_suite_focus=true

With:

image

And passes after switching back to node 14: https://github.com/jupyter-widgets/ipywidgets/runs/4313796358?check_suite_focus=true

martinRenou (approved):
LGTM :)

ui-tests/tests/widgets.test.ts
```diff
@@ -1,90 +1,47 @@
...
-  });
-
-  test('Close notebook widgets.ipynb', async () => {
-    await galata.notebook.close(true);
-  });
-
-  test('Open home directory', async () => {
-    await galata.filebrowser.openHomeDirectory();
-  });
+    await page.notebook.runCellByCell();
```
martinRenou, line 39: Are you executing the Notebook a second time here? Or did line 30 not execute the Notebook?

jtpio: Ah good catch, looks like we should be able to remove that line which is likely just a leftover from a previous iteration.

jtpio:
```suggestion
```

PR #3280, opened 2021-09-15
Widgetsnbextension: throw error on failure to render

martinRenou:
This goes in pair with https://github.com/jupyter/notebook/pull/6181

### DIFF
--- a/widgetsnbextension/src/extension.js
+++ b/widgetsnbextension/src/extension.js
@@ -21,6 +21,16 @@ require('./embed_widgets');
 var LuminoWidget = require('@lumino/widgets');
 var LuminoMessaging = require('@lumino/messaging');
 
+var NOTEBOOK_VERSION_INFO = Jupyter.version.split('.');
+var NOTEBOOK_MAJOR = parseInt(NOTEBOOK_VERSION_INFO[0]);
+var NOTEBOOK_MINOR = parseInt(NOTEBOOK_VERSION_INFO[1]);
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
+1 on this.

This will enable viewing alternative mime types when widgets cannot be rendered. It would be interesting to have a independent release of widgetsnbextension with that change.

widgetsnbextension/src/extension.js
```diff
@@ -123,15 +123,12 @@ function register_events(Jupyter, events, outputarea) {
       Jupyter.notebook.kernel &&
       Jupyter.notebook.kernel.widget_manager;
     if (!manager) {
-      node.textContent =
-        'Error rendering Jupyter widget: missing widget manager';
-      return;
+      throw 'Error rendering Jupyter widget: missing widget manager';
```
vidartf, line 126: Note: In general in Javascript you'd want to throw an Error, and not a string directly. Was this intentional?

martinRenou: Thanks for the review, I will update this

martinRenou:
TODO

- What happens for current Notebook?
- Should we do a Notebook version check?

martinRenou:
Now doing a Notebook version check to **not** change the behavior for Notebook 6.4.4 and before.

SylvainCorlay:
Discussed during the ipywidgets call. Merging as decided in the call.

SylvainCorlay:
@meeseeksdev backport to 7.x

PR #3284, opened 2021-09-20
Select: Do not force a selection if there is currently no selection and the options list changes

### LINKED ISSUE #2230: Changing options of Select widget triggers Callback
@jasongrout we should have a way to change the options of a Select widget without triggering the callback function. Whenever the options are changed for the widget the value attribute is automatically set which forces the callback to be triggered. See code below;

```python
from ipywidgets import Select

# callback
def cb(change):
    print(change)

# widget
s_w = Select(value=None, options=['a','b'])
s_w.observe(cb, names='value')

# triggers no callback
s_w.value = None
# triggers callback
s_w.options = ['1', '2', '3']
```

martinRenou:
Fix https://github.com/jupyter-widgets/ipywidgets/issues/2230


jtpio (approved):
Thanks!

PR #3286, opened 2021-09-24
Add support for localization to the lab extension

jtpio:
At the moment the "Save Widget State Automatically" command label is not translated when using JupyterLab in a different display language, in the following case in French:

image

This PR adds an optional `ITranslator` to the lab plugin to handle this.

We also use a custom `jupyterlab_widgets` domain since the extension lives outside of the core JupyterLab repo right now.

### DIFF
--- a/jupyterlab_widgets/package.json
+++ b/jupyterlab_widgets/package.json
@@ -63,6 +63,7 @@
     "@jupyterlab/rendermime-interfaces": "^3.0.0",
     "@jupyterlab/services": "^6.0.0",
     "@jupyterlab/settingregistry": "^3.0.0",
+    "@jupyterlab/translation": "^3.0.0",
     "@lumino/algorithm": "^1.1.0",
     "@lumino/coreutils": "^1.3.0",
     "@lumino/disposable": "^1.1.1",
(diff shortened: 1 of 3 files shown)


jtpio:
This can be backported to 7.x.

vidartf:
This mostly looks good to me, but to make sure I understand: If a translator wanted to add an entry for this, would they add the entry to the same language pack that JupyterLab uses? How would the versioning of our keys compare to that of JupyterLab? I.e. if we change something, how will we ensure compatibility between an old ipywidgets version and new lab and/or new ipywidgets and old lab?

jtpio:
Since `jupyterlab_widgets` is a regular lab extension, it can follow the same process as other extensions which are currently living in https://github.com/jupyterlab/language-packs

Also this is only for the lab extension and should be independent of the main `ipywidgets` package.

I've just opened https://github.com/jupyterlab/language-packs/pull/109 to add `jupyterlab_widgets` to the list.

fcollonval:
> but I think we are not confident that it all works on integration level just yet?

Yep for now the merge of source strings is working. But there is no information about the version from which the strings are coming. So there is no easy way to handle the lifecycle of the source strings for extensions in the JupyterLab language packs.

jasongrout (approved):

PR #3287, opened 2021-09-24
Prevent Tab to fail to render if there are not titles

martinRenou:
Tabs used to render even if there were no titles specified, but they now fail with "Cannot read index X of undefined".

This PR fixes this issue

### DIFF
--- a/packages/controls/src/widget_selectioncontainer.ts
+++ b/packages/controls/src/widget_selectioncontainer.ts
@@ -156,7 +156,7 @@ export class AccordionView extends DOMWidgetView {
    */
   update_titles(): void {
     const collapsed = this.luminoWidget.collapseWidgets;
-    const titles = this.model.get('titles');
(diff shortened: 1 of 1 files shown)


vidartf (approved):
These seems good to me, but do you know which change caused this to start happening?

martinRenou:
Looking into it, it seems to have been introduced by this change: https://github.com/jupyter-widgets/ipywidgets/commit/e88f8dd07ca64cbad2663e0c2524bd47af19a809#diff-9828a4224e590035f2c3ba7d52fa21efadcf6eedea653e5904b2244de33c46c2R35 which changed it from `titles` to `_titles` which might have been wrong.

So I guess the right fix would be to change it back to `titles`?

martinRenou:
Oh ok I see, thanks @jasongrout. I will open a PR fixing it.

PR #3288, opened 2021-09-27
Do not wait for "displayed" before applying the layout

### LINKED ISSUE #2605: setLayout before attaching to DOM
Currently, we wait till the DOMWidgetView object is displayed (i.e. attached to DOM), before we set the layout:
https://github.com/jupyter-widgets/ipywidgets/blob/92d7d42c00a1b0d9ce921533acb08beefdea3eb2/packages/base/src/widget.ts#L810

This originally comes from this commit:
https://github.com/jupyter-widgets/ipywidgets/commit/9c3de0980968f42ce92feddc6c00dbeb01dd41a5

However, this leads to unneeded resize events. First the DOM elements get attached, and the element get a particular size, then the layout may influence the sizing, leaving several libraries (ipyvolume, bqplot, probably also ipyleaflet) to redraw itself.

I think we should set the layout, style and css _dom_classes before attaching.

martinRenou:
See https://github.com/jupyter-widgets/ipywidgets/issues/2762

Also fix https://github.com/jupyter-widgets/ipywidgets/issues/2605

And trigger a "layout-applied" event

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -897,11 +897,9 @@ export class DOMWidgetView extends WidgetView {
       }
     );
 
-    this.displayed.then(() => {
-      this.update_classes([], this.model.get('_dom_classes'));
-      this.setLayout(this.model.get('layout'));
-      this.setStyle(this.model.get('style'));
-    });
+    this.update_classes([], this.model.get('_dom_classes'));
+    this.setLayout(this.model.get('layout'));
+    this.setStyle(this.model.get('style'));
 
     this._comm_live_update();
     this.listenTo(this.model, 'comm_live_update', () => {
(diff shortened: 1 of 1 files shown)


packages/base/src/widget.ts
```diff
@@ -920,23 +918,25 @@ export class DOMWidgetView extends WidgetView {
...
-              view.trigger('displayed');
-              this.listenTo(view.model, 'change', () => {
-                // Post (asynchronous) so layout changes can take
-                // effect first.
-                MessageLoop.postMessage(
-                  this.luminoWidget,
-                  Widget.ResizeMessage.UnknownSize
-                );
-              });
+            this.trigger('layout-applied');
```
martinRenou, line 923: Adding a "layout-applied" event to the widget, we need to document this

vidartf (changes requested):
From reading the code in the core ipywidgets repo, I made these notes:
1. The `layoutPromise` and `stylePromise` are currently typed as `Promise<any>`, but at least the `stylePromise` is used under the assumption that it is a `Promise<undefined | StyleView>`: https://github.com/jupyter-widgets/ipywidgets/blob/c7bd32948ba677ab564e4cada84ca3306340c40e/packages/controls/src/widget_selection.ts#L561-L565
2. By not waiting for the `displayed` promise before calling `setLayout`, there is an opportunity for the `StyleView` to be initialized before the `render` method of the owning widget is called. If this happens, the child elements of the widget might not be created yet, which means that any styles applied via query strings will fail to be applied (e.g. DescriptionWidget, IntSlider, etc.). This will not currently happen for the example widgets, since they create their child elements before any `await` calls in the `render` method, but there is no guarantee that this is true for all third-party widgets.

So in conclusion: If we decide to go ahead with this pattern, point 1. needs to be addressed. On the other hand, I think we should ensure that `setStyle` (and probably also `setLayout` for consistency) does not get called until the initial `render` method of the view has resolved. The `displayed` promise seems to be the best bet for this, so I would much prefer a solution that keeps this pattern.

PR #3301, opened 2021-10-20
Move Python packages to the `python` folder

jtpio:
Since the ipywidgets repo is also a monorepo of Python packages, this change moves the Python packages to the `py/`:

- `py/ipywidgets`
- `py/widgetsnbextension`
- `py/jupyterlab_widgets`

This came up during the weekly meeting yesterday when discussing the release process and adtopting the release:

- https://github.com/jupyter-widgets/team-compass/issues/1#issuecomment-947370890
- https://github.com/jupyter-widgets/ipywidgets/pull/3298

While there are still a couple of open questions before being able to adopt the releaser, doing this move should be safer and could be merged before.


jasongrout:
@jtpio, I fixed a few more places where we needed to update the paths. Do my changes look good to you? In particular, does the dev-install.sh script work for you to install dev versions of the packages?

jasongrout (approved):

jtpio:
Ah thanks @jasongrout for catching this, changes look good.

It's interesting that we were not calling `./dev-install.sh` script on CI as it should normally have failed otherwise.

jasongrout:
I'm rebasing my new devinstall github workflow to be before my fixes to make sure that it fails, then I'll force push my fix again.

jasongrout:
Yay, the new dev-install workflow fails without the fix!

jasongrout:
Thanks for pushing this forward @jtpio!

jtpio:
Nice!

PR #3302, opened 2021-10-21
Variable Inspector example used a wrong callback argument signature

NichtJens:
I ran into an issue described here https://github.com/ipython/ipython/issues/12214 for the Variable Inspector example.

The `inspector.close()` line would crash with the slightly cryptic message:

```
ValueError: Function <bound method VariableInspectorWindow._fill of <__main__.VariableInspectorWindow object at 0x2b7148c38a58>> is not registered as a post_run_cell callback
```

... despite the callback clearly being registered.

The fix is to a accept a second argument in the callback.


jasongrout:
Closed/reopened to trigger another test run.

PR #3304, opened 2021-10-27
ErrorWidget as fallback when widgets models or views fail - Following up 

trungleduc:
## References

Initially started in #2960 by @maartenbreddels, this PR resolves the conflicts in #2960  and updates the behavior of the failback widget.

## Code changes
- Based on the `loadModelClass` and `loadViewClass` logic on old PR, this PR adds some more `try-catch` blocks to catch the errors in model creation or view rendering.
- Add tests for the new error widget.

## User-facing changes
- An error widget is shown on the failed widget, users can click on it to show the error messages. 
- The layout of the parent widget is preserved in case of failed children widgets.

https://user-images.githubusercontent.com/4451292/139028390-d71d302b-b823-433f-8985-ddd120c11691.mp4

## Backwards-incompatible changes
N/A


packages/base-manager/src/manager-base.ts
```diff
@@ -330,35 +342,57 @@ export abstract class ManagerBase implements IWidgetManager {
...
       options.model_module_version
-    ) as Promise<typeof WidgetModel>;
+    );
     let ModelType: typeof WidgetModel;
+
+    const makeErrorModel = (error: any, msg: string) => {
+      const Cls = widgets.createErrorWidget(error, msg);
+      const widget_model = new Cls({
+        _model_module: options.model_module,
+        _model_name: options.model_name,
```
martinRenou, line 356: Do we need to pass those options? IIUC those are for creating the actual widget, not the error widget.

packages/base-manager/src/manager-base.ts
```diff
@@ -330,35 +342,57 @@ export abstract class ManagerBase implements IWidgetManager {
...
+      });
+      widget_model.name = options.model_name;
+      widget_model.module = options.model_module;
```
martinRenou, line 359: Same question here I guess, do we need to set those parameters?

packages/base-manager/src/manager-base.ts
```diff
@@ -520,6 +554,45 @@ export abstract class ManagerBase implements IWidgetManager {
...
+    try {
+      const promise: Promise<typeof WidgetModel> = this.loadClass(
+        className,
+        moduleName,
+        moduleVersion
+      ) as Promise<typeof WidgetModel>;
+      await promise;
+      return promise;
+    } catch (error) {
+      const msg = `Failed to load model class '${className}' from module '${moduleName}'`;
```
martinRenou, line 571: Should we add a `console.log(error)` here as well? Like you did when failing to create a view.

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,98 @@
...
+  </g>
+</g>
+</svg>`;
+
+// create a Widget Model that captures an error object
+export function createErrorWidget(
+  error: Error,
+  msg?: string
+): typeof WidgetModel {
+  class ErrorWidget extends DOMWidgetModel {
```
martinRenou, line 24: Suggestion: Should we put this class definition outside of the `createErrorWidget` function? Same suggestion for `createErrorWidgetView`.

You could maybe pass `msg` and `error` as extra constructor parameters instead of using them to define the class?

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,98 @@
...
+    <path  d="M 8.2494641,21.074514 V 5.6225142 c 0,-0.314 0.254,-0.567 0.57,-0.567 H 29.978464 c 2.388,0 9.268,5.8269998 9.268,8.3029998 v 5.5835 l -3.585749,4.407396 -2.772971,-3.535534 -5.126524,3.414213 -5.944543,-3.237436 -5.722718,3.06066 z m 30.9969999,3.8675 v 15.5835 c 0,0.314 -0.254,0.567 -0.57,0.567 H 8.8194641 c -0.315,0.002 -0.57,-0.251 -0.57,-0.566 v -15.452 l 7.8444949,2.628449 5.656854,-2.65165 4.24264,3.005204 5.833631,-3.237437 3.712311,3.944543 z" style="fill:url(#linearGradient3448);stroke:#888a85"  />
+    <path d="m 30.383464,12.110514 c 4.108,0.159 7.304,-0.978 8.867,1.446 0.304,-3.9679998 -7.254,-8.8279998 -9.285,-8.4979998 0.813,0.498 0.418,7.0519998 0.418,7.0519998 z" style="fill:url(#linearGradient3445);stroke:#868a84" />
+    <path enable-background="new" d="m 31.443464,11.086514 c 2.754,-0.019 4.106,-0.49 5.702,0.19 -1.299,-1.8809998 -4.358,-3.3439998 -5.728,-4.0279998 0.188,0.775 0.026,3.8379998 0.026,3.8379998 z" style="opacity:0.36930003;fill:none;stroke:url(#linearGradient3442)" />
+  </g>
+</g>
+</svg>`;
+
+// create a Widget Model that captures an error object
+export function createErrorWidget(
```
martinRenou, line 20: Suggestion: `createErrorWidgetModel`?

packages/base/src/errorwidget.ts
```diff
@@ -0,0 +1,98 @@
...
+<g >
+  <g transform="translate(0.24520123,0.93464292)">
+    <path  d="M 8.2494641,21.074514 V 5.6225142 c 0,-0.314 0.254,-0.567 0.57,-0.567 H 29.978464 c 2.388,0 9.268,5.8269998 9.268,8.3029998 v 5.5835 l -3.585749,4.407396 -2.772971,-3.535534 -5.126524,3.414213 -5.944543,-3.237436 -5.722718,3.06066 z m 30.9969999,3.8675 v 15.5835 c 0,0.314 -0.254,0.567 -0.57,0.567 H 8.8194641 c -0.315,0.002 -0.57,-0.251 -0.57,-0.566 v -15.452 l 7.8444949,2.628449 5.656854,-2.65165 4.24264,3.005204 5.833631,-3.237437 3.712311,3.944543 z" style="fill:url(#linearGradient3448);stroke:#888a85"  />
+    <path d="m 30.383464,12.110514 c 4.108,0.159 7.304,-0.978 8.867,1.446 0.304,-3.9679998 -7.254,-8.8279998 -9.285,-8.4979998 0.813,0.498 0.418,7.0519998 0.418,7.0519998 z" style="fill:url(#linearGradient3445);stroke:#868a84" />
+    <path enable-background="new" d="m 31.443464,11.086514 c 2.754,-0.019 4.106,-0.49 5.702,0.19 -1.299,-1.8809998 -4.358,-3.3439998 -5.728,-4.0279998 0.188,0.775 0.026,3.8379998 0.026,3.8379998 z" style="opacity:0.36930003;fill:none;stroke:url(#linearGradient3442)" />
+  </g>
+</g>
+</svg>`;
```
vidartf, line 17: Maybe keep as an external file that is imported?

packages/base-manager/src/manager-base.ts
trungleduc: Yes you are correct, these are the relics of old implementation when the error messages are generated inside `ErrorWidget`

packages/base-manager/src/manager-base.ts
trungleduc: Added

packages/base/src/errorwidget.ts
trungleduc: No, we can't. The message and error need to be hard-coded into a model or view class since we can not control the instantiation of the error widget. `loadClass` method of managers only requests the class and not the object.

packages/base/src/errorwidget.ts
trungleduc: Updated

packages/base/src/errorwidget.ts
trungleduc: Updated

packages/base/src/errorwidget.ts
martinRenou: But do we need to create a new class every time `createErrorWidgetView` gets called?

trungleduc: Yes, in order to display the different error messages.

martinRenou: Ok I see. It's annoying that we can't pass this error message as a constructor parameter.

martinRenou:
Note: I made a half-baked backport of this PR to the 7.x branch, putting [the link here](https://github.com/martinRenou/ipywidgets/tree/build_embed) in case we need it later. Meeseeks will not be able to backport it automatically.

packages/base/src/utils.ts
```diff
@@ -242,3 +242,13 @@ export function remove_buffers(
   const new_state = remove(state, []) as JSONObject;
   return { state: new_state, buffers: buffers, buffer_paths: buffer_paths };
 }
+
+export const SVG_ICON = `<svg style="height:50%;max-height: 50px;" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
```
martinRenou, line 246: It's unclear from the variable name `SVG_ICON` what is this icon. Maybe rename it to `FILE_ERROR_SVG_ICON` or `BROKEN_FILE_SVG_ICON` or just `ERROR_SVG_ICON`?

PR #3314, opened 2021-11-17
7.x: Fix control comm channel test

martinRenou:
It's odd the CI didn't catch the test was not working after the latest reviews. We need to check that the CI runs the tests properly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
         python -m pip install jupyter_packaging
     - name: Build the extension
       run: |
-        pip install .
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -45,6 +45,7 @@ jobs:
       run: |
         python -m pip install jupyterlab~=3.0
         python -m pip install jupyter_packaging
+        python -m pip install pytest
```
martinRenou, line 48: Should be removed, and later do `pip install -e .[test]`

martinRenou:
Merging as discussed during last meeting

PR #3329, opened 2021-12-22
Drop Python 3.6 support

jasongrout:
Python 3.6 hits end-of-life in a few days. This drops support for Python 3.6.

It also upgrades our testing to use the current NodeJS LTS, version 16.


jasongrout:
Requesting review from @SylvainCorlay and @ibdafna since I know they are working on other PRs before the RC, but please, anyone feel free to review this.

ibdafna (approved):

PR #3335, opened 2022-01-03
Use control comm target in base Manager

martinRenou:
Port of https://github.com/voila-dashboards/voila/pull/766 for JupyterLab and classic Jupyter Notebook

Use the `jupyter.widget.control` comm target for fetching widget states in the JupyterLab manager and the widgetsnbextension, this is used *e.g.* when refreshing the JupyterLab page on an already running kernel that contains widgets.

It will also be used by Voila in https://github.com/voila-dashboards/voila/pull/846.

I would really like this to be considered for the coming 8.0.0 release. It would also be nice to backport it to the 7.x branch if we keep making releases from there.

User facing changes
---------------------

This improves the performances when refreshing the JupyterLab page or when connecting to an already existing kernel that contains widgets.

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -4,7 +4,11 @@
 import * as services from '@jupyterlab/services';
 import * as widgets from '@jupyter-widgets/base';
 
-import { JSONObject, PartialJSONObject } from '@lumino/coreutils';
+import {
+  JSONObject,
+  PartialJSONObject,
+  PromiseDelegate,
+} from '@lumino/coreutils';
 
 import {
   DOMWidgetView,
@@ -31,6 +35,21 @@ import { removeMath, replaceMath } from './latex';
 import sanitize from 'sanitize-html';
 
 const PROTOCOL_MAJOR_VERSION = PROTOCOL_VERSION.split('.', 1)[0];
+
+/**
+ * The control comm target name.
+ */
+export const CONTROL_COMM_TARGET = 'jupyter.widget.control';
+
+/**
+ * The supported version for the control comm channel.
+ */
+export const CONTROL_COMM_PROTOCOL_VERSION = '1.0.0';
+
+/**
+ * Time (in ms) after which we consider the control comm target not responding.
+ */
+export const CONTROL_COMM_TIMEOUT = 4000;
 
 /**
  * Sanitize HTML-formatted descriptions.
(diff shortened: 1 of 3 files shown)


martinRenou:
Making this WIP for now as I'll be doing the same for the widgetsnbextension

martinRenou:
@meeseeksbot backport to 7.x

martinRenou:
meeseeksdev backport to 7.x

martinRenou:
Comment: could this be part of the base manager?

martinRenou:
We could probably do this to check for the target:

```typescript
const reply = await kernel.requestCommInfo({target_name: this.comm_target_name});
```

The documentation is not entirely clear on the behavior of this when the target is no registered (to be improved):

https://jupyter-client.readthedocs.io/en/stable/messaging.html#comm-info

martinRenou:
Question: what happens if I reduce the tornado web socket message size to something very small? Does it fail? How?

python/jupyterlab_widgets/src/manager.ts
```diff
@@ -106,6 +121,106 @@ export abstract class LabWidgetManager
...
       return;
     }
+
+    const commId = uuid();
+    const initComm = await this._create_comm(
+      CONTROL_COMM_TARGET,
+      commId,
+      { widgets: null },
+      { version: CONTROL_COMM_VERSION }
+    );
```
vidartf, line 131: Does this not need to be inside the `try` clause?

jasongrout:
> Question: what happens if I reduce the tornado web socket message size limit to something very small? Does it fail? How?

My guess is that the message will just never come back from the kernel. I think it would be good to have a timeout on the request that will fall back to the older method. Separately, I think it would be good to have a check in the server zmq/websocket bridge to see if a comm message is too big, and if it is, send a comm_close message both ways (to kernel and frontend) with an error status, indicating the comm channel had an error.

python/jupyterlab_widgets/src/manager.ts
```diff
@@ -106,6 +121,106 @@ export abstract class LabWidgetManager
...
+      if (!bufferPaths[bufferPath[0]]) {
+        bufferPaths[bufferPath[0]] = [];
+      }
+      bufferPaths[bufferPath[0]].push(bufferPath.slice(1));
+    }
+
+    // Start creating all widgets
+    for (const [widget_id, state] of Object.entries(states) as any) {
+      try {
+        const comm = await this._create_comm('jupyter.widget', widget_id);
```
martinRenou, line 189: This can probably be faster with a `Promise.all` instead of a `for () {await [...];}`

martinRenou:
> We could probably do this to check for the target:
>
> ```typescript
> const reply = await kernel.requestCommInfo({target_name: this.comm_target_name});
> ```

Unfortunately this doesn't work. The kernel answers to the request with `{"status": "ok"}` even when there is no target handler registered.

martinRenou:
I updated the PR this morning:

- Moved the `loadFromKernel` logic into the base manager, as suggested by @vidartf 
- Fallback to the old implementation if the control target is not responding in time (4 seconds), as suggested by @jasongrout

Unfortunately we can't use the commInfo request to know if the target is registered kernel-side. See https://github.com/jupyter-widgets/ipywidgets/pull/3335#issuecomment-1005571385

martinRenou:
> Unfortunately this doesn't work. The kernel answers to the request with {"status": "ok"} even when there is no target handler registered.

Check again: what are the exact response messages when
- the comm target is not registered
- the comm target is registered

martinRenou:
So it's annoying, requesting the comm info will provide the same answer whether the target is registered or not:

```
{
    'comms': {},
    'status': 'ok'
}
```

No target registered:
--------------------

no_target


Target registered:
-----------------

target


So using this request will not give us the information about the comm target.

martinRenou:
Also adding a link to the `comm_info_reply` implementation in ipykernel: https://github.com/ipython/ipykernel/blob/0ab288fe42f3155c7c7d9257c9be8cf093d175e0/ipykernel/kernelbase.py#L771-L787

martinRenou:
The xeus implementation looks similar https://github.com/jupyter-xeus/xeus/blob/eab12033ed8dc0202dfba733446dcf7ced10fca1/src/xkernel_core.cpp#L303-L322

packages/base-manager/src/manager-base.ts
```diff
@@ -342,6 +361,197 @@ export abstract class ManagerBase implements IWidgetManager {
     return await modelPromise;
   }
 
+  /**
+   * Fetch all widgets states from the kernel using the control comm channel
+   * If this fails (control comm handler not implemented kernel side),
+   * it will fallback to `_loadFromKernelSlow`.
+   */
+  protected async _loadFromKernel(): Promise<void> {
```
jasongrout, line 369: Can we add a note in the function documentation here that this is a utility method for subclasses to use as they see fit? Since we don't use this method in this base class, it is a bit confusing why it is defined here without extra context.

jasongrout (commented):
Looks good! Two minor comments to address inline, and then +1 from me to merge. Thanks!

packages/base-manager/src/manager-base.ts
jasongrout: (and same comment for the other loading function defined just below)

packages/base-manager/src/manager-base.ts
```diff
@@ -342,6 +361,197 @@ export abstract class ManagerBase implements IWidgetManager {
...
+    // Try fetching all widget states through the control comm
+    let data: any;
+    let buffers: any;
+    try {
+      const initComm = await this._create_comm(
+        CONTROL_COMM_TARGET,
+        uuid(),
+        { widgets: null },
```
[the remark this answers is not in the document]
martinRenou: This might be a leftover, I'll see if I can remove it

packages/base-manager/src/manager-base.ts
martinRenou: Sure! Make sense. Thanks for the review!

packages/base-manager/src/manager-base.ts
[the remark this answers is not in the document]
martinRenou: Done :)

martinRenou:
Thanks for your review @jasongrout

PR #3342, opened 2022-01-11
HTMLManager: Do not reset lab's CSS variables if they are defined

### LINKED ISSUE #1506: Dark theme broken when notebook contains widgets in metadata
I followed the guide on https://blog.jupyter.org/the-templating-system-of-nbconvert-6-47ea781eacd2 to convert my notebooks to HTML with dark theme, using
```
jupyter nbconvert xleaflet.ipynb --to html --HTMLExporter.theme=dark
```

However some of the converted notebooks have the dark theme style for about 0.5 second, then the HTML page turns to the white theme automatically.
After some investigation I found out that it's only happening to the notebooks with widgets state in the metadata, which makes nbconvert add the following code snippet to the HTML page,
https://github.com/jupyter/nbconvert/blob/2e41429def702dd561586303356fdf1f9be7e95c/share/jupyter/nbconvert/templates/base/jupyter_widgets.html.j2#L3-L26

which in turn breaks the dark theme by updating the CSS properties.

**Nbconvert version:**

6.0.7

### LINKED ISSUE #3340: ipywidgets's HTMLManager package overrides the JupyterLab CSS in nbconvert
## Description

ipywidgets's HTMLManager package seems to override JupyterLab CSS variables when loaded

## Reproduce

1. Install `nbconvert==6.4.0` and `ipywidgets==7.6.5`
2. Create a Notebook `test.ipynb` containing widgets
3. Convert the Notebook to HTML using a dark theme, using the following command:
`jupyter nbconvert --to html --HTMLExporter.theme=dark test.ipynb`
4. Open the resulting HTML file with your browser, you should see the following:

https://user-images.githubusercontent.com/21197331/148768880-4f23b6b2-deb1-4d60-a99e-6224120e9c5b.mp4

## Expected behavior

The resulting HTML page should have a dark theme, ipywidgets should keep the CSS variables untouched when a theme is already loaded.

Ideally, ipywidgets wouldn't even ship those CSS variables.

## Extra Context

~~It seems that, by chance, Voila is not impacted by this issue. I understand that the ipywidgets's control package CSS gets loaded before Voila loads the theme. But I might be wrong.~~
Voila is not impacted by this because it doesn't use HTMLManager.

martinRenou:
Do not override the JupyterLab's CSS variables if they are already defined in the page.
This fixes the nbconvert's HTMLExporter when using a dark theme.

Testing it:

```
jupyter nbconvert --to html --HTMLExporter.theme=dark --HTMLExporter.template_name=lab --HTMLExporter.widget_renderer_url=/path/to/local/ipywidgets/packages/html-manager/dist/embed-amd.js NotebookWithWidgets.ipynb
```

Screenshot:
-----------

dark

References:
-----------

Fix #3340 fix https://github.com/jupyter/nbconvert/issues/1506

### DIFF
--- a/packages/html-manager/src/libembed.ts
+++ b/packages/html-manager/src/libembed.ts
@@ -10,7 +10,16 @@ import '@fortawesome/fontawesome-free/css/all.min.css';
 import '@fortawesome/fontawesome-free/css/v4-shims.min.css';
 
 import '@lumino/widgets/style/index.css';
-import '@jupyter-widgets/controls/css/widgets.css';
+import '@jupyter-widgets/controls/css/widgets-base.css';
+
+// If lab variables are not found, we set them (we don't want to reset the variables if they are already defined)
+if (
+  getComputedStyle(document.documentElement).getPropertyValue(
+    '--jp-layout-color0'
+  ) === ''
+) {
+  require('@jupyter-widgets/controls/css/labvariables.css');
+}
 
 // Used just for the typing. We must not import the javascript because we don't
 // want to include it in the require embedding.


martinRenou:
As a side note, I wonder if this logic should not be in the base-manager so that we get it for free in any manager (thinking about Voila for example).

maartenbreddels:
See also https://github.com/jupyter-widgets/ipywidgets/pull/2534

maartenbreddels:
Yeah, I also had trouble finding it, but I remember trying to attack a similar issue.

jasongrout:
Thanks!

martinRenou:
meeseeksdev backport to 7.x

PR #3343, opened 2022-01-11
Document protocol version 3.0.0

### LINKED ISSUE #3330: Document the widget protocol changes for echoing comm sync messages
In https://github.com/jupyter-widgets/ipywidgets/pull/3195, a significant new addition to the syncing protocol was introduced. This protocol change should be documented in https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/messages.md#widget-messaging-protocol-version-2. I think this also changes the protocol version number (2->3 or 2->2.1?), which may need to be updated in a few places.

Also, a bullet point should be added to the changelog for 8.0.

maartenbreddels:
Closes #3330

### DIFF
--- a/packages/base-manager/test/src/manager_test.ts
+++ b/packages/base-manager/test/src/manager_test.ts
@@ -168,7 +168,7 @@ describe('ManagerBase', function () {
           },
         },
         metadata: {
-          version: '2.0.0',
+          version: '3.0.0',
         },
       });
       expect(model.comm).to.equal(comm);
@@ -231,7 +231,7 @@ describe('ManagerBase', function () {
         },
         buffers: [new DataView(new Uint8Array([1, 2, 3]).buffer)],
         metadata: {
-          version: '2.0.0',
+          version: '3.0.0',
         },
       });
       expect(model.comm).to.equal(comm);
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
We discussed this PR in the Jupyter widgets call and we think that the new protocol version number should be 3.0.

Indeed, if the backend implements this new protocol, and an old client connects to it, it will interpret echo messages as regular updates and behave accordingly, leading to wrong behavior.

maartenbreddels:
it's not wrong behaviour I think, it will not break (except the jitter is annoying).
Actually, a kernel implementation knows about the version number of the protocol the front end uses, so if the front end creates a COM with protocol version <2.1 we can choose not to send the echo messages, making it fully backward compatible.

vidartf:
> it's not wrong behaviour I think, it will not break (except the jitter is annoying). Actually, a kernel implementation knows about the version number of the protocol the front end uses, so if the front end creates a COM with protocol version <2.1 we can choose not to send the echo messages, making it fully backward compatible.

And what do you propose we do in the case where the kernel is the one that opens the COM (which is that majority of cases)? In this case we will not know what protocol version the client has (unless I'm remembering the code wrong), so it cannot know to not send echos.

vidartf (approved):

SylvainCorlay:
Merging as discussed in the team call.

PR #3358, opened 2022-01-26
feat: backward compatibility with 7.x

mariobuikhuizen:
This way widget libraries accessing `JupyterPhosphorWidget` and/or `pWidget` can run with 8.x without making a 7.x incompatible build.

For example:
- https://github.com/bqplot/bqplot/search?q=pWidget
- https://github.com/jupyter-widgets/ipyleaflet/search?q=pWidget
- https://github.com/mariobuikhuizen/ipyvue/search?q=pWidget

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -1177,12 +1177,6 @@ export class DOMWidgetView extends WidgetView {
    * @deprecated Use {@link luminoWidget} instead (Since 8.0).
    */
   get pWidget(): Widget {
-    if (!DOMWidgetView.deprecationWarningDisplayed) {
-      console.warn(
-        'The use of pWidget is deprecated, use luminoWidget instead.'
-      );
-      DOMWidgetView.deprecationWarningDisplayed = true;
-    }
     return this.luminoWidget;
   }
 
@@ -1191,5 +1185,4 @@ export class DOMWidgetView extends WidgetView {
   luminoWidget: Widget;
   layoutPromise: Promise<any>;
   stylePromise: Promise<any>;
-  private static deprecationWarningDisplayed = false;
 }
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -885,6 +885,11 @@ export class JupyterLuminoWidget extends Widget {
   private _view: DOMWidgetView;
 }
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay (approved):

SylvainCorlay:
Thanks! As per the discussion at the team meeting on Tuesday, we should get this in!

packages/base/src/widget.ts
```diff
@@ -1167,9 +1172,24 @@ export class DOMWidgetView extends WidgetView {
       this.el.removeAttribute('tabIndex');
     }
   }
+
+  /**
+   * @deprecated Use {@link luminoWidget} instead (Since 8.0).
+   */
+  get pWidget(): Widget {
+    if (!DOMWidgetView.deprecationWarningDisplayed) {
+      console.warn(
```
jasongrout, line 1181: This is going to put a *lot* of warnings in consoles of users that cannot do anything about it as soon as we release. Perhaps we can start off deleting this console.warn, with just docs saying it is deprecated, then up it to a console warning later to put more pressure on migrating?

mariobuikhuizen: The intention of this code was to maximally show one warning per page load. I can't see how this would happen more often, maybe I'm missing something? If one message is still too much, I'm still fine with taking it out.

jasongrout: Sorry, I meant that lots of users will have this warning, not that it would show up a lot of times on a single page. I think it's gentler to give developers a chance to migrate before showing warnings to their users.

jasongrout (approved):
Thanks!

martinRenou:
@jasongrout @mariobuikhuizen Should we do an alias for `processPhosphorMessage` pointing to `processLuminoMessage` as well?

mariobuikhuizen:
Yeah, I'm for it. It just wasn't an issue in ipyvue.

martinRenou:
Thanks a lot @mariobuikhuizen ! I was thinking about it this morning, it's unclear to me how this would look like for someone who wants to still support ipywidgets<8 (they need to overwrite `processPhosphorMessage`) and want to support ipywidgets===8 (they need to overwrite `processLuminoMessage` as well).

PR #3360, opened 2022-01-27
Add note about workaround for editable user installs via pip.

jasongrout:
Apparently a bug in pip prevents projects with pyproject.toml from having editable `--user` installs.

See https://github.com/pypa/pip/issues/7953 and https://github.com/jupyter-widgets/ipywidgets/issues/3351

### DIFF
--- a/docs/source/dev_install.md
+++ b/docs/source/dev_install.md
@@ -29,6 +29,15 @@ cd ipywidgets
 ./dev-install.sh
 ```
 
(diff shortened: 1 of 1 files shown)


seanbreckenridge:
@jasongrout  See [setuptools issue](https://github.com/pypa/setuptools/issues/3063), is likely an issue with the `-e` (editable) flag, and not the user prefix

PR #3367, opened 2022-02-03
Fix downward arrow misalignment for combobox

xiaochen-db:
Remove the problematic line-height style that affected combobox downward arrow. 

Before the change:
Screen Shot 2022-02-02 at 5 41 22 PM

After the change:
Screen Shot 2022-02-02 at 5 48 57 PM


jasongrout:
Though I didn't see any difference visually when originally trying this out, our visual regression tests are failing. Here's what I'm seeing when I investigate manually on Chrome, which matches up with what the tests are showing (a 1-pixel difference in the screenshots).

With the line-height (status quo):
Screen Shot 2022-02-08 at 02 09 59

Without the line-height (i.e., after this PR):
Screen Shot 2022-02-08 at 02 10 24

jasongrout:
With Safari, we have the same height of that particular element whether or not the line-height is active, however when the line-height is active, we also have a scrollbar to the right of the widget, and no scrollbar when the line-height is not active the scrollbar disappears:

Safari, with line-height active:
Screen Shot 2022-02-08 at 02 51 41

Safari, without line-height active:
Screen Shot 2022-02-08 at 02 51 56

SylvainCorlay (approved):

jasongrout:
Thanks!

jasongrout:

PR #3377, opened 2022-02-11
Introduce synchronous has_model function in the manager base class.

jasongrout:
This simplifies our code that has had to work around get_model either returning undefined or a rejected promise if a model was not registered.
This paves the way for a future get_model in 8.0 that is truly asynchronous, never returning undefined.


martinRenou (approved):
It's way cleaner that way! Thanks

jasongrout:
> It's way cleaner that way! Thanks

Feel free to merge if it looks good to you!

PR #3380, opened 2022-02-11
Edit and expand docs for jupyter.widget.control comm channel

jasongrout:
This clarifies and expands on the existing documentation.

### DIFF
--- a/packages/schema/messages.md
+++ b/packages/schema/messages.md
@@ -348,11 +348,11 @@ This is implemented in ipywidgets 7.7.
 
 ### The `jupyter.widget.control` comm target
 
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
Thanks!

ibdafna:

maartenbreddels:
I think we should also have an optional argument, for which Comms to get, or which to skip. I can imagine that if the frontend connects while the kernel is running, it already received a lot of widget data, the only thing left then is to get the missing widgets/Comms. Not something that is required for 8.0, but at least something to keep in mind.

Note: I've seen this a lot doing benchmarks with voila. where we first wait for 90% of the widget data to come in (we missed the first 10%). And then we ask again for all widgets, which is a shame.

PR #3388, opened 2022-02-18
Backport loadFromKernel nodifications to 8.x

ibdafna:


packages/base-manager/src/manager-base.ts
```diff
@@ -507,7 +542,7 @@ export abstract class ManagerBase implements IWidgetManager {
 
           let msg_id = '';
           const info = new PromiseDelegate<Private.ICommUpdateData>();
-          comm.on_msg((msg: services.KernelMessage.ICommMsgMsg) => {
```
jasongrout, line 510: We lost a type annotation here that we should preserve.

packages/schema/messages.md
```diff
@@ -370,11 +370,11 @@ This is implemented in ipywidgets 7.7.
 
 ### The `jupyter.widget.control` comm target
 
-A kernel-side Jupyter widgets library registers a `jupyter.widget.control` comm target that is used for fetching all widgets states through a "one shot" comm message (one for all widget instances). Unlike the `jupyter.widget` comm target, the created comm is global to all widgets,
```
jasongrout, line 373: I think we should ignore the changes to this file in this PR.

packages/base-manager/src/manager-base.ts
```diff
@@ -375,7 +375,7 @@ export abstract class ManagerBase implements IWidgetManager {
     try {
       const initComm = await this._create_comm(
         CONTROL_COMM_TARGET,
-        uuid(),
+        utils.uuid(),
```
jasongrout, line 378:
```suggestion
        uuid(),
```

jasongrout:
I took at crack at automated porting this one commit too. Not sure if it is any more successful, but may be useful for comparison: 

First, I slightly modified master to get it looking more like the 7.x before the commit in question:

https://github.com/jasongrout/ipywidgets/commit/7d46c917f361c55e69753499c6a49ef42ff3e674

Then I applied the patch from https://github.com/jasongrout/ipywidgets/commit/0296e03e8ad2568630c73b7d626320c43f3e4d3d while ignoring whitespace (something like `git apply --ignore-whitespace diff-file.patch` or something)

https://github.com/jasongrout/ipywidgets/commit/92632cb01d2d18e269b59cf8737c87f38eed00bc

ibdafna:
@jasongrout many thanks!

PR #3337, opened 2022-01-04
[Backport 7.x] Use control comm target in LabManager

martinRenou:
Backport of #3335

PR #3389, opened 2022-02-18
Port #3337 and #3377 to master

### LINKED ISSUE #3378: Forward port has_model to the 8.x widget manager
In https://github.com/jupyter-widgets/ipywidgets/pull/3377, I introduce has_model in the base widget manager class, simplifying logic.

Right now we have a confusing `get_model` method:

- In the base widget manager, it returns undefined if the model is not registered, which means that it can be used to synchronously check if a model is registered. This means the function is not purely async.
- In the jlab widget manager, the function returns a rejected promise if the model is not registered, making the function a truly async function, but leaving us without a way to easily check synchronously if a model is registered.

Once we have `has_model` by porting #3377 to master, I think we can change the base `get_model` to be a truly async function like it is in the jlab manager now. This is a breaking change, so would need to be in widgets 8.0.

### LINKED ISSUE #3379: Forward port changes to kernel loading logic from #3337
As mentioned in https://github.com/jupyter-widgets/ipywidgets/pull/3337, there were some changes made there that should be forward ported to master around carefully handling asynchronous issues.

jasongrout:
This ports relevant parts of #3337 and #3377 to master, and makes get_model signature backwards incompatible.

Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3378
Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3379

- [x] Document "Change manager.get_model() to always return a promise" as a breaking change in the 8.0 changelog (I guess it's *technically* not breaking since it was allowed before, but it is a change in default behavior, so should be noted).
- [x] Add tests for has_model

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -84,7 +84,7 @@ See also the
 - Fetch the full widget state via a control Comm ([#3021](https://github.com/jupyter-widgets/ipywidgets/pull/3021))
 - Export LabWidgetManager and KernelWidgetManager ([#3166](https://github.com/jupyter-widgets/ipywidgets/pull/3166))
 - More helpful semver range message ([#3185](https://github.com/jupyter-widgets/ipywidgets/pull/3185))
-
+- Make the base widget manager `.get_model()` method always return a Promise, which is rejected if the requested model is not registered. To test if a model is registered, use the new `.has_model()` method ([#3389](https://github.com/jupyter-widgets/ipywidgets/pull/3389))
 
 ### Documentation improvements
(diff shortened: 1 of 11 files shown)


jasongrout:
There still is a few more things I'll need to do to make has_model pervasive throughout the system that I'll work on tomorrow.

ibdafna (approved):

jasongrout:
> ** [ibdafna](/ibdafna) ** approved these changes [ 7 minutes ago](#pullrequestreview-887037217)

Thanks!

PR #3402, opened 2022-03-02
Limit dependabot to run-time dependencies only.

vidartf:
Make dependabot only make PRs for versions that affect runtime dependencies. This would mean that we would not be getting security updates pushed automatically for dev dependencies. However, developers should still be able to receive a notice about security updates by running `yarn audit` locally (npm does this by default on `install` command, and yarn may change to that in the future as well).


jasongrout:
Thanks!

PR #3416, opened 2022-03-16
Fix doc build

jasongrout:
Rely on the default of readthedocs installing things with pip, rather than insisting on setuptools (which doesn't work anymore for us).


jasongrout:

jasongrout:
> @meeseeksdev please backport to master

Actually, looks like this is already fixed in master.

Issue #3429, opened 2022-04-05
8.0 Migration plan for frontends implementing widgets


jasongrout:
@Donjayamanne and I were discussing in the dev meeting how frontends implementing ipywidgets support should manage the transition from ipywidgets 7 to 8.

@DonJayamanne was hoping that the 7.x javascript might be able to support the 7.x subset of widgets even with the 8.0 kernel package. He said it worked with an 8.0 beta, but not with 8.0 RC. Officially this is not currently supported, but I'm not sure what exactly the breaking change is.

Ideally a frontend would be able to load either the 7.x or 8.x javascript, whichever is required by the kernel.

Regardless, just like we have a transition guide for custom widget authors, we should have a paragraph or two in the migration guide for frontends implementing ipywidgets helping them understand how they might bridge supporting ipywidgets 7 and 8.

jasongrout:
Some thoughts about this:

Widget javascript that is normally bundled with the widget manager in ipywidgets 8.0 includes models from `@jupyter-widgets/base` (model module version 2.0.0) and `@jupyter-widgets/controls` (model module version 2.0.0): See the model listing at https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/jupyterwidgetmodels.latest.md, as compared with the core controls provided with ipywidgets 7: https://github.com/jupyter-widgets/ipywidgets/blob/7.x/packages/schema/jupyterwidgetmodels.latest.md

Ideally, there would be a single widget manager in the frontend, and that widget manager would load both the 7.x core widget model js and the 8.x core widget model js, and use whichever model js was requested by the kernel. If the single frontend widget manager implemented the interface of both the 7.x widget manager and the 8.x widget manager, I think all would be well. There were changes in the widget manager interface from 7.x to 8.x, though, so we'd need to be careful.

Another way to manage things is to somehow peek into the kernel and know from it what possible core models would be requested, then load either the 7.x widget manager and js, or the 8.x widget manager and js. For example, we could have a new widget control comm request that would reply with the version of ipywidgets, or the list of registered models, and infer from that which widget manager to load. This isn't as elegant, but it may be less prone to breaking.

DonJayamanne:
> Another way to manage things is to somehow peek into the kernel and know from it what possible core models would be

The current thinking is to do this in VS Code. Query version of ipywidgets in the kernel and load the corresponding JS bundle.

jasongrout:
[...]

PR #3442, opened 2022-04-20
Split the JupyterLab extension into plugins that provide the widget manager and register widgets

jasongrout-db:
This paves the way to having separate extensions that load multiple versions of core widgets, so that ipywidgets 7.x and 8.x can be supported simultaneously, for example.

See also conversation at #3429

### DIFF
--- a/python/jupyterlab_widgets/src/plugin.ts
+++ b/python/jupyterlab_widgets/src/plugin.ts
@@ -208,7 +208,7 @@ function activateWidgetExtension(
   };
   if (settingRegistry !== null) {
     settingRegistry
-      .load(plugin.id)
+      .load(managerPlugin.id)
       .then((settings: ISettingRegistry.ISettings) => {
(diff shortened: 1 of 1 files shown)


vidartf:
@jasongrout What work remains here? Is this something that we should block 8.0 for?

jasongrout:
> @jasongrout What work remains here? Is this something that we should block 8.0 for?

I fixed the remaning errors, and I think this is ready for review.

vidartf (approved):
The code changes here are fine. Thinking about the stated goal of supporting 7.x and 8.x side-by-side, I do have some concerns w.r.t to most third party extensions declaring base and controls (if depending on controls) as singletons in their package.json's jupyterlab section.

jasongrout:
> I do have some concerns w.r.t to most third party extensions declaring base and controls (if depending on controls) as singletons in their package.json's jupyterlab section.

They should be doing that anyway.

PR #3470, opened 2022-05-25
fix: _pending_msgs was negative, iopub.status was called too often

### LINKED ISSUE #3469: Bug: Message throttling is broken after first update message from frontend
## Description

The message throttling mechanism should accumulate change in a widget model while the kernel is busy. When the kernel becomes idle again, we send the outstanding changes. To catch the kernel status changes (introduced in #1383 ) we save the previous callback in https://github.com/jasongrout/ipywidgets/blob/ae946250b0e3233da04c3a03c4f92d444b91bce3/jupyter-js-widgets/src/widget.ts#L417 and call it at https://github.com/jasongrout/ipywidgets/blob/ae946250b0e3233da04c3a03c4f92d444b91bce3/jupyter-js-widgets/src/widget.ts#L421

However, after the first call to `send_sync_message`, we never restore `callbacks.iopub.status`, and thus on the next call to `send_sync_message`, we call the function we assigned at https://github.com/jasongrout/ipywidgets/blob/ae946250b0e3233da04c3a03c4f92d444b91bce3/jupyter-js-widgets/src/widget.ts#L418 in the previous call.

This will cause `_pending_msg' to go to -1 at line https://github.com/jasongrout/ipywidgets/blob/ae946250b0e3233da04c3a03c4f92d444b91bce3/jupyter-js-widgets/src/widget.ts#L250 and the message throttling mechanism at https://github.com/jasongrout/ipywidgets/blob/ae946250b0e3233da04c3a03c4f92d444b91bce3/jupyter-js-widgets/src/widget.ts#L354 will never be used again.

## Reproduce

```
import ipywidgets as w
w.FloatSlider()
```
Move the slider once (a single click), now `_pending_msg` should be 0, but the `callbacks.iopub.status` contains the old function. Move the slider again, now the _pending_msg should be called twice when the kernel goes idle, and `_pending_msgs` should be -1. Now drag the slider, and check in the websocket messages, and you should see multiple update calls directly following each other:

In this screenshot in chrome:
image

We can see 6 (green) update messages.

## Expected behavior

We expected to see 1 update message, followed by a busy/idle pair, followed by a last single update, and another busy/idle message

maartenbreddels:
This broke message throttling.

Fixes #3469

I think this should be backported to 7.x, since behaviour now is really badly defined. For views the throttling works the first time only (because the iopub callbacks gets reused when using .touch), for models (calling save_changes), we should not see this issue.

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -333,6 +333,12 @@ export class WidgetModel extends Backbone.Model {
     if (this.comm !== void 0) {
       if (msg.content.execution_state === 'idle') {
         this._pending_msgs--;
+        if (this._pending_msgs < 0) {
+          console.error(
+            `Pending messages < 0 (=${this._pending_msgs}), which is unexpected`
+          );
+          this._pending_msgs = 0; // do not break message throttling in case of unexpected errors
+        }
         // Send buffer if one is waiting and we are below the throttle.
         if (this._msg_buffer !== null && this._pending_msgs < 1) {
           const msgId = this.send_sync_message(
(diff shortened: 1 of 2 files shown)


maartenbreddels:
As I suggested in #3469 the following works much better:
```python
import time
import ipywidgets as w

@w.interact
def test(f = 10.1):
    print(f**2)
    time.sleep(1)

```

Before this would take many seconds to finish (there is a long queue of messages that will trigger the test function), while with this PR it is only max 2 seconds (the one in flight, and the 1 pending).

Also, websocket messages behave as predicted/expected now:
image

1 update to the kernel, followed by busy/echo/idle (for 7.x that should just be busy/idle).

ibdafna (approved):

jasongrout:

jasongrout:
Backported in #3494

PR #3475, opened 2022-06-06
fix: Issue 3466 - Disconnect broken link problems

kellietay:
PR Summary:
Issue #3466 mentions 2 issues. This PR specifically only fixes issue 1 - "When restarting, the broken link icon is, well, broken in jupyterlab."

Fix Description:
- updated font-weight to 900 as FontAwesome now only allows font-weight 900 as part of its free plans.

PR #3476, opened 2022-06-07
Update broken link icon to work in FontAwesome 4 and 5

jasongrout:
#3475 fixed the broken link icon for FontAwesome 5, but unfortunately things broke for FontAwesome 4 (i.e., classic Notebook). This uses CSS that works for both FontAwesome 4 and FontAwesome 5, so the broken link works in classic notebook and JupyterLab.


ibdafna (approved):

jasongrout:

PR #3478, opened 2022-06-08
Use `dict` instead of `print` to test callable without signature

frenzymadness:
In Python 3.11, `inspect.signature(print)` no longer raises
ValueError.

Fixes: https://github.com/jupyter-widgets/ipywidgets/issues/3459

PR #3480, opened 2022-06-10
Backport PR #3478 on branch 7.x (Use dict instead of print to test callable without signature)

frenzymadness:
I have backported only the change for the test because in 7.x there is no testing matrix in Github Actions and no individual setup.{py,cfg} files – there is one in the root of the project but the Python versions there are outdated.


frenzymadness:
The failure of the CI does not seem to be related to this change.

jasongrout:
Thanks. I agree that the test is a different issue. I'll do that in a separate PR.

jasongrout:
I just released this in ipywidgets 7.7.1 on pypi. Thanks again for the contribution.

PR #3449, opened 2022-05-02
Remove nbformat dependency 

anthonyaag:
Remove nbformat dependency from ipywidgets #3418

PR #3481, opened 2022-06-10
Remove an old nbformat dependency that is no longer needed.

jasongrout:
Remove an old nbformat dependency that is no longer needed.

Backport of #3449


jasongrout:
The readthedocs issue is fixed in #3482.

PR #3484, opened 2022-06-10
Check self.keys before accessing trait in set_state()

xiaochen-db:
When widgets echo mode is on (`JUPYTER_WIDGETS_ECHO == true`) and `sync_data` contains non-trait attribute, a `TraitError` will be thrown by `self.trait_metadata(attr, 'echo_update', default=True)`.

An example is from ipyleaflet where field `_dragging` can be [sent to kernel](https://github.com/jupyter-widgets/ipyleaflet/blob/3437521/js/src/Map.js#L272) while it's not a registered trait for class [Map](https://github.com/jupyter-widgets/ipyleaflet/blob/3437521c69514a965f7d8e5c559b82bd742a0dcc/ipyleaflet/leaflet.py#L2134)

We should check if attribute is inside `self.keys` before accessing it as a trait

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_set_state.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_set_state.py
@@ -304,8 +304,8 @@ def test_echo():
     assert widget.value == 1
 
     widget._send = mock.MagicMock()
-    # this mimics a value coming from the front end
-    widget.set_state({'value': 42})
+    # this mimics a state coming from the front end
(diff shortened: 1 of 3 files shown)


jasongrout:
Awesome, thanks. Nice catch in ipyleaflet as well. I don't know that we officially support state messages having non-trait attributes, but since we guard against it in the code below this change, it makes a lot of sense to go ahead and guard against it here too.

jasongrout:

PR #3485, opened 2022-06-10
Fix message throttle by making a copy of the callbacks dict

### LINKED ISSUE #3469: Bug: Message throttling is broken after first update message from frontend

jasongrout:
Fixes #3469 in a different way from #3470.

This partially reverts code changes from #3470, but leaves the tests.

I think the underlying assumptions around send_sync_message and callbacks are:

1. send_sync_message is only ever called once per message. When the original sync happens, either we put the message on the queue or we send it immediately, and the callback modifications happen only when we actually send the message.
2. There was an implicit assumption in the send_sync_message code is that the callbacks object passed in is a new object unique to this message, i.e., we aren't modifying a global object, that then we'll modify again later.

I think assumption 2 is sometimes not true, so this guards against it by making a (2-deep) copy of the callbacks object before modifying and using it.

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -334,9 +334,10 @@ export class WidgetModel extends Backbone.Model {
     if (this.comm !== void 0) {
       if (msg.content.execution_state === 'idle') {
         this._pending_msgs--;
+        // Sanity check for logic errors that may push this below zero.
         if (this._pending_msgs < 0) {
           console.error(
-            `Pending messages < 0 (=${this._pending_msgs}), which is unexpected`
+            `Jupyter Widgets message throttle: Pending messages < 0 (=${this._pending_msgs}), which is unexpected. Resetting to 0 to continue.`
           );
           this._pending_msgs = 0; // do not break message throttling in case of unexpected errors
         }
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Hi Jason,

I tried a (lodash) clone, but that didn't work, this solution is much better.
LGTM!

cheers,

Maarten

maartenbreddels (approved):

jasongrout:

PR #3488, opened 2022-06-12
Add more overview info on docs front page

jasongrout:
This is probably still a bit wordy, but it's nice to have some high-level context on the front page.

Also I deleted the developer_docs page which seemed largely empty.


ibdafna (approved):
LGTM

jasongrout:
Awesome, thanks for the review. We can keep iterating to make the docs more approachable.

jasongrout:

PR #3492, opened 2022-06-21
fix: backbone set patch in ill defined state after exceptions in handler

maartenbreddels:
If an exception occurs during set() in backbone-patch.ts this._changing
is left true. This means that next calls to set() will skip the
trigger('change', ...) code.

This fix will leave the backbonejs model in a better defined state.

FYI: This bug is present upstream in backbonejs as well.

Note: the diff does not render nice, but it's basically a try/finally block.

cc @kecnry

### DIFF
--- a/packages/base/src/backbone-patch.ts
+++ b/packages/base/src/backbone-patch.ts
@@ -76,58 +76,65 @@ export function set(
   const changes = [];
   const changing = this._changing;
   this._changing = true;
-
-  if (!changing) {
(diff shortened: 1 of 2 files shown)


jasongrout:
> Note: the diff does not render nice, but it's basically a try/finally block.

You can see the diff better if you ignore whitespace: https://github.com/jupyter-widgets/ipywidgets/pull/3492/files?diff=split&w=1

jasongrout:

PR #3501, opened 2022-06-23
User migration docs for 7.x -> 8.0

vidartf:
This is meant as an explicit list of (the main) backwards incompatible changes. As this hasn't really been tracked throughout the development process, it is mainly based of the entries in the changelog. Other backwards incompatible changes probably exist, but I think we should keep this PR's review about the document structure and the items currently in it.

### DIFF
--- a/docs/source/user_migration_guides.md
+++ b/docs/source/user_migration_guides.md
@@ -1,0 +1,92 @@
+Migrating user code
+===================
+
+These are migration guides aimed specifically at user of ipywidgets.
+
+Migrating from 7.x to 8.0
+-------------------------
+
+For more details about the changes done for the 8.0 major version, please consult the
+[changelog](./changelog).
+
+### Code
+
+#### FileUpload
+
+The `data` and `metadata` traits have been removed, and the `value` trait revamped to
+be a list of dicts containing the file information. The keys of these dicts are:
+
+- `content`: The buffer of file data
+- `name`: The name of the file
+- `type`: The MIME type of the file content
+- `size`: The size of the buffer in bytes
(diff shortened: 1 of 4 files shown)


docs/source/user_migration_guides.md
```diff
@@ -0,0 +1,92 @@
...
+#### Tooltips
+
+As part of an effort to make it possible to
+[set tooltips for all widgets](https://github.com/jupyter-widgets/ipywidgets/pull/2680),
+the old `description_tooltip` attribute for certain widgets was removed. Now all widgets
+that inherit `DOMWidget` have the attribute `tooltip` instead.
+
+Suggested migration: Search and replace `description_tooltip` to `tooltip`.
+TBD: ipywidgets should add a `description_tooltip` keyword argument to `DescriptionWidget`s, with
+a deprecation warning.
```
vidartf, line 38: I think this might make sense to add as a separate PR to ensure a smoother migration.

docs/source/user_migration_guides.md
```diff
@@ -0,0 +1,92 @@
...
+
+- `content`: The buffer of file data
+- `name`: The name of the file
+- `type`: The MIME type of the file content
+- `size`: The size of the buffer in bytes
+- `last_modified`: A UTC datetime representing the "last modified" value reported for the file
+
+Suggested migration: Rewrite all usage of `FileUpload` to use the new structure.
+If you need to support both 7.x and 8.x, you can e.g. write functions `get_file_buffer` and similar
+to wrap reads from the widget.
```
vidartf, line 27: This seems to me to be the largest breaking change for users. I.e. it is likely to completely break code (exceptions will be raised that cannot be recovered from), and is likely to be heavily used.

Issue #3474, opened 2022-06-02
Check public path for AMD modules


jasongrout:
https://github.com/jupyter-widgets/ipywidgets/pull/3464 may need some more work. It seems that the webpack auto public path does not work inside of AMD modules, since by the time the AMD callback is called the currentScript is null. This would affect the various amd modules generated by the html manager. This is hinted at in https://github.com/jupyter-widgets/ipywidgets/pull/2191#discussion_r214508413. See also https://github.com/jupyter-widgets/ipywidgets/pull/1652, https://github.com/jupyter-widgets/ipywidgets/issues/1650.

A workaround for automatically figuring out the public path in an AMD module using the require 'module' magic dependency is at https://github.com/jupyter-widgets/widget-cookiecutter/pull/103

jasongrout:
@xiaochen-db and I confirmed that using the html manager libembed-amd.js file from requirejs is currently broken on master. We do indeed need the fix from https://github.com/jupyter-widgets/widget-cookiecutter/pull/103

PR #3508, opened 2022-07-03
Better logic to set webpack public path in AMD modules

jasongrout:
Fix for #3474.

This uses the technique from https://github.com/jupyter-widgets/widget-cookiecutter/pull/103 to get the public path for amd modules and creates an example that uses the amd modules built for html-manager.

This is another take on https://github.com/jupyter-widgets/ipywidgets/pull/3464, since that approach did not work to set the public path in AMD modules. Essentially, since AMD modules are executed asynchronously, the webpack 'auto' public path logic does not work. Instead, we use the requirejs-specific 'module' magic dependency, which allows us to get the path to the current AMD module, and use that to get the public path. We do this public path computation in a separate script to isolate it from any library code.

### DIFF
--- a/examples/embed-amd/README.md
+++ b/examples/embed-amd/README.md
@@ -1,0 +1,24 @@
+# Using the HTML widget manager as a RequireJS AMD module
+
+## Description
+
+This is an example project showing how to embed widgets in an HTML document using a RequireJS AMD module.
+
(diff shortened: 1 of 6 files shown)


jasongrout:
@vidartf @ibdafna @martinRenou - can one of you review this and approve/merge it if it looks good, and then I'll release another RC of widgets 8 with these fixes?

ibdafna (approved):
Glanced at this quickly and it looks good! Thank you!

PR #3517, opened 2022-07-10
[wip] Package example notebooks, dev packages in JupyterLite

bollwyvl:
## Elevator Pitch

This fattens up the RTD environment so it can build the packages-under-development for deployment to the jupyterlite site. This means a code change would be directly reflected in a PR on RTD.

## References

- follow-on to #3491

## Changes

- [x] uses the as-built labextension in jupyterlite
- [x] use ipywidgets wheel in pyolite


bollwyvl:
This one actually worked:

https://ipywidgets--3517.org.readthedocs.build/en/3517/

jasongrout:
I just merged this back into #3491. Thanks, this really gets us ahead in usability for using jupyterlite in docs!

bollwyvl:
as always, a work in progress... but getting it into the other PR where there is more existing discussion is probably good... if you like, i can retarget your branch (perhaps after merging `main`) so the delta is a little clearer, and treat this as a sub-PR... though i am not at a suitable computer to do any non-web-ui work at this moment...

happy to help further if you have specific questions, or if there are upstream things... that we can readily do anything about in the near term :blush:.

jasongrout:
> or if there are upstream things... that we can readily do anything about in the near term 😊.

This would be awesome: https://github.com/jupyterlite/jupyterlite/pull/734

It adds the jupyterlab-github extension to the default jupyterlite, so that it is easy for people to immediately start using jupyterlite with existing content.

PR #3522, opened 2022-07-15
Fix selection container logic around setting titles.

### LINKED ISSUE #3520: v7 widget_selectioncontainer.py needs deprecation warnings or fixes before v8 can be finalized...
## Description
This stretch of code in v8 is incompatible with v7 compatible code:
widget_selectioncontainer.py lines 50 - 64.
```
    def set_title(self, index, title):
        """Sets the title of a container page.
        Parameters
        ----------
        index : int
            Index of the container page
        title : unicode
            New title
        """
        titles = list(self.titles)
        # for backwards compatibility with ipywidgets 7.x
        if title is None:
            title = ''
        titles[index]=title
        self.titles = tuple(titles)
```
Any version 7 compatible code that calls `container.set_title(indexnum,  titlestr)` will get an `IndexError: list assignment index out of range` from line 63 (`titles[index]=title`), if the calls are made immediately after the container is loaded with a list of objects. The tuple `titles` does not appear to get set to have a length = the number of sub objects, when the container is defined.

Thus I suggest that version 7 ipywidgets begin issuing a deprecation warning that the container.set_title() syntax will be replaced with specifying a titles tuple in the initial 8.0 release. I suspect the problem is related to @ dispatching. Naively, I would have checked in this code for `len(titles) == len(children)` and then appended empty strings to titles until they matched.

There is also the related issue that positional syntax for defining the objects in one of these containers `container([...])` is no longer supported only the keyword syntax `container(children=[...])`. As the keyword syntax has been the recommendation in the documentation at least since 7.6, this is not a huge concern. However, I wonder why that was acceptable without a deprecation warning? It is also not consistent with the base Box widget which does accept a positional list of children.

## Reproduce

This code running against ipywidgets v8.0.0rc1 in Jupyter Lab will reproduce the error.
```
from ipywidgets import widgets
tab_contents = ['P0', 'P1', 'P2', 'P3', 'P4']
children = [widgets.Text(description=name) for name in tab_contents]
tab = widgets.Tab()
tab.children = children
for i in range(len(children)):
    tab.set_title(i, str(i))
tab
```

I was testing [jupyter-Pandas-GUI](https://pypi.org/project/jupyter-Pandas-GUI/0.7.0rc0/) against Google Colab, where it did not work because Colab is using ipywidgets 8+. I then tested against ipywidgets 8.0.0rc1 in Juputer Lab to isolate the error.

## Expected behavior

Expect the titles to be set.
## Context

- ipywidgets version 8.0.0rc1
- Operating System and version:Linux Ubuntu 20.04lts
- Browser and version: Firefox and Chrome

### If using JupyterLab
- JupyterLab version: 3.4.3

<details><summary>Installed Labextensions</summary>
<pre>
JupyterLab v3.4.3
/home/jonathan/.local/share/jupyter/labextensions
        jupyterlab-datawidgets v7.0.0 enabled OK

/home/jonathan/.local/share/virtualenvs/JLabtst-Ix77i15z/share/jupyter/labextensions
        jupyterlab-drawio v0.9.0 enabled OK (python, jupyterlab-drawio)
        jupyterlab-plotly v5.9.0 enabled OK
        @jupyter-widgets/jupyterlab-manager v5.0.0-rc.1 enabled OK (python, jupyterlab_widgets)

Other labextensions (built into JupyterLab)
   app dir: /home/jonathan/.local/share/virtualenvs/JLabtst-Ix77i15z/share/jupyter/lab
        @kptan86/jupyterlab-snippets-multimenus v0.1.4 enabled  X

   The following extension are outdated:
        @kptan86/jupyterlab-snippets-multimenus
</pre>
</details>

jasongrout:
Previously, the title logic was being overridden by Tab. This refactors the code so we can just override the selected index logic, without overriding the titles logic.

Fixes #3520

### DIFF
--- a/.github/workflows/binder-on-pr.yml
+++ b/.github/workflows/binder-on-pr.yml
@@ -23,7 +23,7 @@ jobs:
             issue_number: context.issue.number,
             owner: context.repo.owner,
             repo: context.repo.repo,
-            body: `[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/${PR_HEAD_USERREPO}/${PR_HEAD_REF}?urlpath=lab/tree/docs/source/examples/) :point_left: Launch a binder notebook on branch _${PR_HEAD_USERREPO}/${PR_HEAD_REF}_`
+            body: `[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/${PR_HEAD_USERREPO}/${PR_HEAD_REF}?urlpath=lab/tree/docs/source/examples/) :point_left: Launch a binder notebook on branch _${PR_HEAD_USERREPO}/${PR_HEAD_REF}_. You can also try out changes using JupyterLite in the documentation that is built in the docs/readthedocs.org check below.`
           })
       env:
         PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_selectioncontainer.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_selectioncontainer.py
@@ -5,7 +5,7 @@ from unittest import TestCase
 
 from traitlets import TraitError
 
-from ipywidgets.widgets import Accordion, HTML
+from ipywidgets.widgets import Accordion, Tab, Stack, HTML
 
 
 class TestAccordion(TestCase):
@@ -26,6 +26,11 @@ class TestAccordion(TestCase):
     def test_selected_index_out_of_bounds(self):
         with self.assertRaises(TraitError):
             Accordion(self.children, selected_index=-1)
+
+    def test_children_position_argument(self):
+        Accordion(self.children)
+        Tab(self.children)
+        Stack(self.children)
 
 
     def test_titles(self):
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_selectioncontainer.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_selectioncontainer.py
@@ -5,55 +5,164 @@ from unittest import TestCase
 
 from traitlets import TraitError
 
-from ipywidgets.widgets import Accordion, Tab, Stack, HTML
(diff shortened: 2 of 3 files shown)


gutow:
When I clone this branch and do a `pip install -e .`, things run without error any errors showing in the notebook, but javascript errors (seen in the browser console) prevent the widget from being rendered.
In JLab: `Error: Shared module @jupyter-widgets/base doesn't exist in shared scope default` and `Exception opening new comm 12 [jlab_core.fc632a0f38747f007842.js:2:1067689](http://localhost:8888/static/lab/jlab_core.fc632a0f38747f007842.js?v=fc632a0f38747f007842)
    _handleCommOpen http://localhost:8888/static/lab/jlab_core.fc632a0f38747f007842.js?v=fc632a0f38747f007842:2`

In classic notebook an error about protocol v2 appears.

It looks like ipywidgets needs some of the javascript requirements updated.

jasongrout:
There are some JS build steps needed in addition to pip install to get a dev install. I typically use the dev-install.sh script at the repo root to build the appropriate js in the right order, or do those commands manually: https://github.com/jupyter-widgets/ipywidgets/blob/master/dev-install.sh

Or you can use the "launch binder" button in the second comment to try things out: https://mybinder.org/v2/gh/jasongrout/ipywidgets/selection?urlpath=lab/tree/docs/source/examples/

Or the easiest way these days to test a change is to use the JupyterLite instance in the documentation built (evaluate the first piplite cell, then start trying out code: https://ipywidgets--3522.org.readthedocs.build/en/3522/

jasongrout:
Functional review by @gutow is at https://github.com/jupyter-widgets/ipywidgets/issues/3520#issuecomment-1186210251. I'll let someone else do a code review and merge.

vidartf:
I see this was merged via https://github.com/jupyter-widgets/ipywidgets/pull/3523

vidartf:
I guess the jupyterlite docs change was not included in that PR though, so @jasongrout you might want to do a rebase to check if there is anything left over ?

PR #3523, opened 2022-07-16
Rename Stacked widget to Stack

jasongrout:
The Stacked widget was added in #2376. Considering the name, all of our other widgets are present tense, like Accordion, Tab, Box, etc, so the grammar of "Stack" feels like it fits better than "Stacked". As a survey:

Toolkits that use "Stack":

* flutter: https://api.flutter.dev/flutter/widgets/Stack-class.html
* GTK: https://docs.gtk.org/gtk4/class.Stack.html
* MUI: https://mui.com/material-ui/react-stack/
* Kivy: https://kivy.org/doc/stable/api-kivy.uix.stacklayout.html?highlight=stack

Toolkits that use "Stacked":

* QT: https://doc.qt.io/qt-6/qstackedwidget.html
* Lumino (which in general follows QT closely)

This builds on #3522, so that should be merged first.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -27,7 +27,7 @@ See also the
 - Tagsinput widget ([#2591](https://github.com/jupyter-widgets/ipywidgets/pull/2591), [#3272](https://github.com/jupyter-widgets/ipywidgets/pull/3272))
 - Drop notebook dependency from widgetsnbextension ([#2590](https://github.com/jupyter-widgets/ipywidgets/pull/2590))
 - Cast 'value' in range sliders to a tuple ([#2441](https://github.com/jupyter-widgets/ipywidgets/pull/2441))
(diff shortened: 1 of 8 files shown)


ibdafna:
@jasongrout Is this a good idea now that we have a "stable" API?

jasongrout:
> @jasongrout Is this a good idea now that we have a "stable" API? (Assuming this goes into 8.0)

It's the last chance to change it before releasing said "stable" api - after releasing it probably won't ever change. I feel moderately about this from an API naming perspective, but if everyone else is against the name change at this point, that's fine too.

Separate from the question of timing, do you agree with the name change, or disagree, or think it doesn't matter?

jasongrout:
4/5 dev meeting attendees agree with the change, one liked it better but suggested there may be an even better choice, and no one had an objection to the timing.

jasongrout:
ping @astrofrog for comments as well.

astrofrog:
No objections

PR #3533, opened 2022-08-01
Using the new Comm Python package

martinRenou:
Related discussions:
- https://github.com/jupyter-xeus/xeus-python/issues/342 
- https://github.com/jupyter-widgets/ipywidgets/issues/3514

Currently working on a Python comm package on https://github.com/martinRenou/comm for now, but it should probably be moved in the IPython org

This has the benefit of being able to:
- remove the ipykernel dependency in ipywidgets https://github.com/jupyter-widgets/ipywidgets/pull/3533
- remove the ipykernel monkey patch in xeus-python https://github.com/jupyter-xeus/xeus-python/pull/548
- remove the ipykernel package mock in jupyterlite's pyolite 
- remove the ipykernel package mock in emscripten-forge

### DIFF
--- a/python/ipywidgets/ipywidgets/__init__.py
+++ b/python/ipywidgets/ipywidgets/__init__.py
@@ -21,28 +21,39 @@ accessible as a `value` attribute.
 from ._version import __version__, __protocol_version__, __jupyter_widgets_controls_version__, __jupyter_widgets_base_version__
 
 import os
-from IPython import get_ipython
-from .widgets import *
+
 from traitlets import link, dlink
+from IPython import get_ipython
+try:
+    from comm import get_comm_manager
+except ImportError:
+    def get_comm_manager():
+        ip = get_ipython()
+
+        if ip is not None and ip.kernel is not None:
+            return get_ipython().kernel.comm_manager
+
+from .widgets import *
+
 
 def load_ipython_extension(ip):
     """Set up Jupyter to work with widgets"""
     if not hasattr(ip, 'kernel'):
         return
-    register_comm_target(ip.kernel)
+    register_comm_target()
 
 def register_comm_target(kernel=None):
     """Register the jupyter.widget comm target"""
-    if kernel is None:
-        kernel = get_ipython().kernel
-    kernel.comm_manager.register_target('jupyter.widget', Widget.handle_comm_opened)
-    kernel.comm_manager.register_target('jupyter.widget.control', Widget.handle_control_comm_opened)
+    comm_manager = get_comm_manager()
+
+    comm_manager.register_target('jupyter.widget', Widget.handle_comm_opened)
+    comm_manager.register_target('jupyter.widget.control', Widget.handle_control_comm_opened)
 
 def _handle_ipython():
     """Register with the comm target at import if running in Jupyter"""
     ip = get_ipython()
     if ip is None:
         return
-    load_ipython_extension(ip)
+    register_comm_target()
 
 _handle_ipython()
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_widget_templates.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_widget_templates.py
@@ -229,7 +229,7 @@ class TestTwoByTwoLayout(TestCase):
         assert box.layout.grid_template_areas == ('"top-left top-right"\n' +
                                                   '"bottom-left bottom-right"')
         # check whether frontend was informed
-        send_state.assert_called_once_with(key="grid_template_areas")
+        send_state.assert_called_with(key="grid_template_areas")
 
         box = widgets.TwoByTwoLayout(top_left=button1, top_right=button3,
                                      bottom_left=None, bottom_right=button4)
@@ -240,7 +240,7 @@ class TestTwoByTwoLayout(TestCase):
         box.merge = False
         assert box.layout.grid_template_areas == ('"top-left top-right"\n' +
                                                   '"bottom-left bottom-right"')
-        send_state.assert_called_once_with(key="grid_template_areas")
+        send_state.assert_called_with(key="grid_template_areas")
 
 
 class TestAppLayout(TestCase):
--- a/python/ipywidgets/ipywidgets/widgets/tests/utils.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/utils.py
@@ -4,8 +4,14 @@
 from ipywidgets import Widget
 import ipywidgets.widgets.widget
 
-import comm
-from ipykernel.comm import Comm
+# The new comm package is not available in our Python 3.7 CI (older ipykernel version)
+try:
+    import comm
+    NEW_COMM_PACKAGE = True
+except ImportError:
+    NEW_COMM_PACKAGE = False
+
+import ipykernel.comm
 
 
 class DummyComm():
@@ -13,7 +19,7 @@ class DummyComm():
     kernel = 'Truthy'
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        super().__init__()
         self.messages = []
 
     def open(self, *args, **kwargs):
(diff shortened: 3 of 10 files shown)


python/ipywidgets/ipywidgets/__init__.py
```diff
@@ -21,28 +21,26 @@
...
+
 
 def load_ipython_extension(ip):
     """Set up Jupyter to work with widgets"""
-    if not hasattr(ip, 'kernel'):
-        return
-    register_comm_target(ip.kernel)
+    register_comm_target()
 
-def register_comm_target(kernel=None):
```
maartenbreddels, line 34: should we keep the argument for backwards compatibility?

SylvainCorlay:
> This has the benefit of being able to:
> 
> remove the ipykernel dependency in ipywidgets https://github.com/jupyter-widgets/ipywidgets/pull/3533
> remove the ipykernel monkey patch in xeus-python https://github.com/jupyter-xeus/xeus-python/pull/548
> remove the ipykernel package mock in jupyterlite's pyolite
> remove the ipykernel package mock in emscripten-forge

Wow. I think it is quite an improvement.

python/ipywidgets/ipywidgets/__init__.py
```diff
@@ -21,28 +21,36 @@
...
 
 def load_ipython_extension(ip):
     """Set up Jupyter to work with widgets"""
     if not hasattr(ip, 'kernel'):
         return
-    register_comm_target(ip.kernel)
+    register_comm_target()
 
-def register_comm_target(kernel=None):
+def register_comm_target():
```
maartenbreddels, line 42:
```suggestion
def register_comm_target(kernel=None):
```
I think we shouldn't change the API, as that would be a breaking change.

python/ipywidgets/ipywidgets/widgets/tests/utils.py
```diff
@@ -1,11 +1,14 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from ipykernel.comm import Comm
 from ipywidgets import Widget
 import ipywidgets.widgets.widget
 
-class DummyComm(Comm):
+import comm
+from ipykernel.comm import Comm
```
maartenbreddels, line 8:
```suggestion
```
import is not used right?

python/ipywidgets/ipywidgets/widgets/tests/utils.py
```diff
@@ -16,16 +19,30 @@ def __init__(self, *args, **kwargs):
...
+def dummy_get_comm_manager(**kwargs):
+    return {}
+
+
 _widget_attrs = {}
 undefined = object()
 
 def setup_test_comm():
+    comm.create_comm = dummy_create_comm
+    comm.get_comm_manager = dummy_get_comm_manager
```
maartenbreddels, line 45: should we restore this? in a try/finally?

python/ipywidgets/ipywidgets/widgets/widget_output.py
```diff
@@ -134,15 +132,6 @@ def __exit__(self, etype, evalue, tb):
...
                 ip.showtraceback((etype, evalue, tb), tb_offset=0)
-            elif self.comm is not None and self.comm.kernel is not None:
-                kernel = self.comm.kernel
-                kernel.send_response(kernel.iopub_socket,
-                                     u'error',
-                                     {
-                    u'traceback': ["".join(traceback.format_exception(etype, evalue, tb))],
-                    u'evalue': repr(evalue.args),
-                    u'ename': etype.__name__
-                    })
```
maartenbreddels, line 145: Why can we remove this?

maartenbreddels (commented):
Great work.

I do worry we do not have a test yet that exposes https://github.com/ipython/ipykernel/issues/1090

I'm pretty sure this fixes it, but #3653 also didn't pick up this issue, which leaves me a bit worries that we're vulnerable for regressions.

python/ipywidgets/ipywidgets/widgets/tests/utils.py
martinRenou: We should 👍🏽 I will restore them in the teardown function

python/ipywidgets/ipywidgets/widgets/tests/utils.py
martinRenou: It is used later in that file, in the teardown function for restoring the original Comm class

martinRenou:
Thanks a lot for your review @maartenbreddels ! I will resolve your comment and get the CI green hopefully :D

python/ipywidgets/ipywidgets/widgets/widget_output.py
martinRenou: I will bring it back. What is annoying with this code is that it's very specific to ipykernel, but I can add more checks so we don't go through this with *e.g.* xeus-python

martinRenou:
@maartenbreddels it was difficult to get green but now it seems good!

Thankfully, the Python 3.7 build requires an old version of ipykernel (latest ipykernel does not support 3.7?) so we have tests against the old versions and new versions.

martinRenou:
Would you have time to make it a review @maartenbreddels ?

python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -686,7 +693,7 @@ def notify_change(self, change):
         # Send the state to the frontend before the user-registered callbacks
         # are called.
         name = change['name']
-        if self.comm is not None and self.comm.kernel is not None:
+        if self.comm is not None and (self.comm.kernel is not None if hasattr(self.comm, "kernel") else True):
```
maartenbreddels, line 696:
```suggestion
        if self.comm is not None and getattr(self.comm, 'kernel', True) is not None:
```
do you think this look better.. ?

python/ipywidgets/ipywidgets/widgets/widget_output.py
```diff
@@ -111,9 +111,9 @@ def __enter__(self):
         kernel = None
         if ip and getattr(ip, "kernel", None) is not None:
             kernel = ip.kernel
-        elif self.comm is not None and self.comm.kernel is not None:
+        elif self.comm is not None and getattr(self.comm, "kernel", None) is not None:
```
maartenbreddels, line 114:
```suggestion
        elif self.comm is not None and getattr(self.comm, 'kernel', True) is not None:
```
then also this?

maartenbreddels (commented):
awesome work, some minor comments

python/ipywidgets/ipywidgets/widgets/tests/test_widget_templates.py
```diff
@@ -229,7 +229,7 @@ def test_update_dynamically(self, send_state): #pylint: disable=no-self-use
         assert box.layout.grid_template_areas == ('"top-left top-right"\n' +
                                                   '"bottom-left bottom-right"')
         # check whether frontend was informed
-        send_state.assert_called_once_with(key="grid_template_areas")
+        send_state.assert_called_with(key="grid_template_areas")
```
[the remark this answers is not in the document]
martinRenou: Yes, I think the test was faulty from the beginning. Though it's unclear to me as to why my PR fixes it.

Setting `bottom_left` would not only trigger [changing grid_template_areas](https://github.com/jupyter-widgets/ipywidgets/blob/6f6156c7e2421f437614b0649a5a3f37c3b4a3ec/python/ipywidgets/ipywidgets/widgets/widget_templates.py#L448) but also [a couple more properties from layout](https://github.com/jupyter-widgets/ipywidgets/blob/6f6156c7e2421f437614b0649a5a3f37c3b4a3ec/python/ipywidgets/ipywidgets/widgets/widget_templates.py#L446-L447)

martinRenou:
Thanks for the review! I resolved your comments

maartenbreddels:
@martinRenou do you know what caused that change?
I'm fine removing the newlines, but maybe good to be aware what is responsible for this change?



> Hello. Is there a plan to merge and release this soon?

Yes, I think we should get this released asap, which I could do tomorrow if we manage to get this in.

martinRenou:
> Yes, I think we should get this released asap, which I could do tomorrow if we manage to get this in.

💯

martinRenou:
@maartenbreddels all green!

maartenbreddels:
@martinRenou great work, i guess we also need to backport this to 7.x
do you want to try the bot?

martinRenou:

bollwyvl:
If seeing errors like this in `pytest`:

```py
    def _create_comm(*args, **kwargs):
        """Create a Comm.
    
        This method is intended to be replaced, so that it returns your Comm instance.
        """
>       raise NotImplementedError("Cannot ")
E       NotImplementedError: Cannot
```

Adding:

```py
import ipykernel.ipkernel  # noqa
```

Is enough to trigger the [side-effects](https://github.com/ipython/ipykernel/blob/a895ce72e4c0ecff95280e68ae96faa283e561cf/ipykernel/ipkernel.py#L66) to overload that manager.

Seems like it should go in the docs... or even in that unhelpful traceback itself, cuz 99% of the time, i reckon people are testing with `ipykernel` lying around.

PR #3534, opened 2022-08-02
Split layout and style docs, add style attribute table

### LINKED ISSUE #3257: Audit styling attributes across widgets
As brought up in discussion from https://github.com/jupyter-widgets/ipywidgets/pull/2728, it would be nice to unify some of the styling attributes various widgets have, and make sure they are used consistently across all widgets (a style `background` attribute was one thing brought up there, for example).

vidartf:
This splits the "Layout and Style" doc notebook into two, as it was quite large, and overly dominated by content on layout.

It then adds tables about which style classes are on which widget, and which attributes each style class has. ~~Note that the cells generating the tables are hidden in nbsphinx!~~

A possible change would be to also/instead do a table with widget --> style attributes supported, but that would necessarily be more verbose.

Closes #3257.


jasongrout:
@vidartf, your changes look good to me, thanks! I also merged the tagsinput file into widget list, added a few more examples to widget list, and changed the default formatting for IntInputs to look like an integer, rather than to look like a decimal. Can you review my changes?

jasongrout (approved):

jasongrout:
@ibdafna - I added you as a reviewer in case you have time before @vidartf does to review this.

vidartf:
Your changes make sense to me Jason!

PR #3537, opened 2022-08-04
Lazy load the JupyterLite example on the docs front page.

jasongrout:
Jupyterlite-sphinx now has an option to lazily load the heavy jlite bundle.


martinRenou (approved):
Looks good!

PR #3540, opened 2022-08-16
Fix jupyterlite_contents

jasongrout:
It seems we are only getting one file in the jlite contents these days, which started happening right around the time we switched to jupyterlite_sphinx 0.7.0.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -31,7 +31,7 @@ jupyterlite_dir = "."
 # "examples/Layout Example.ipynb", # bqplot/ipyleaflet dependencies require ipywidgets 7
 # "examples/Layout Templates.ipynb", # bqplot/ipyleaflet dependencies require ipywidgets 7
 # "examples/Variable Inspector.ipynb", # uses ipykernel-specific features
-jupyterlite_contents = "examples"
+jupyterlite_contents = ["examples"]
 
 # intersphinx_mapping = {
 #     'ipython': ('http://ipython.org/ipython-doc/dev/', None),
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -160,8 +160,12 @@ def on_config_inited(*args):
     IPYW = ROOT / "python/ipywidgets"
     subprocess.check_call([sys.executable, "-m", "build"], cwd=str(IPYW))
 
+    WIDG = ROOT / "python/widgetsnbextension"
+    subprocess.check_call([sys.executable, "-m", "build"], cwd=str(WIDG))
+
     JLW = ROOT / "python/jupyterlab_widgets"
     subprocess.check_call(["jupyter", "labextension", "build", "."], cwd=str(JLW))
+    subprocess.check_call([sys.executable, "-m", "build"], cwd=str(JLW))
 
 def setup(app):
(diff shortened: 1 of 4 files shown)


martinRenou (approved):
Nice catch! Thanks

jasongrout:
Docs is failing to build:

```
yarn install v1.21.1
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error An unexpected error occurred: "https://registry.npmjs.org/corser/-/corser-2.0.1.tgz: Request failed \"503 Service Unavailable\"".
```
So I'll try again in a bit.

docs/source/conf.py
```diff
@@ -31,7 +31,7 @@
 # "examples/Layout Example.ipynb", # bqplot/ipyleaflet dependencies require ipywidgets 7
 # "examples/Layout Templates.ipynb", # bqplot/ipyleaflet dependencies require ipywidgets 7
 # "examples/Variable Inspector.ipynb", # uses ipykernel-specific features
-jupyterlite_contents = "examples"
+jupyterlite_contents = "examples/"
```
jasongrout-db, line 34:
```suggestion
jupyterlite_contents = "examples"
```

jasongrout:
Looks like the actual error was that jupyterlite_contents inadvertently can't be a string anymore, so I changed it to an array. I put in a fix for jupyterlite_sphinx in https://github.com/jupyterlite/jupyterlite-sphinx/pull/73

jasongrout:
> I put in a fix for jupyterlite_sphinx in [jupyterlite/jupyterlite-sphinx#73](https://github.com/jupyterlite/jupyterlite-sphinx/pull/73)

And @martinRenou is amazingly fast at releasing, and the fix is now in jupyterlite-sphinx 0.7.2. Thanks Martin!

docs/source/jupyter_lite_config.json
```diff
@@ -68,7 +68,7 @@
       "https://files.pythonhosted.org/packages/py3/t/traitlets/traitlets-5.1.1-py3-none-any.whl",
       "https://files.pythonhosted.org/packages/py3/t/typing_extensions/typing_extensions-4.1.1-py3-none-any.whl",
       "https://files.pythonhosted.org/packages/py3/v/vega-datasets/vega_datasets-0.9.0-py3-none-any.whl",
-      "https://files.pythonhosted.org/packages/py3/w/widgetsnbextension/widgetsnbextension-4.0.0rc1-py3-none-any.whl",
+      "https://files.pythonhosted.org/packages/py3/w/widgetsnbextension/widgetsnbextension-4.0.0rc2-py3-none-any.whl",
```
vidartf, line 71: Cannot this also be done with a `../../python/...` path?

vidartf (commented):
Mostly looks good. Can we use a local path for the nbextension? If not, we will need to update that line for every release, right?

docs/source/jupyter_lite_config.json
jasongrout: Great question. I've pushed a commit to try it.

jasongrout: Seems to work great!

vidartf (approved):

PR #3543, opened 2022-08-17
Mention removing on_displayed in changelog

jasongrout:
We removed on_displayed, but forgot to mention this breaking change in the 8.0 changelog.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -156,6 +156,7 @@ Here is a short list of some of the other changes in ipywidgets 8.0.
 - ErrorWidget as fallback when widgets models or views fail - Following up ([#3304](https://github.com/jupyter-widgets/ipywidgets/pull/3304))
 - Fix matplotlib plots in interact ([#3277](https://github.com/jupyter-widgets/ipywidgets/pull/3277))
 - Fix selection equality checking ([#2897](https://github.com/jupyter-widgets/ipywidgets/pull/2897))
+- Remove the `on_displayed` Python callback mechanism ([#2021](https://github.com/jupyter-widgets/ipywidgets/pull/2021))
 
 ### Developers
 


massgh:
What are some reasonable alternatives to `on_displayed` method? I need to fire a JavaScript event when widget is displayed using Output widget and that is broken now.

jasongrout:
> What are some reasonable alternatives to `on_displayed` method? I need to fire a JavaScript event when widget is displayed using Output widget and that is broken now.

You can define a processLuminoMessage method and take action on the `after-show` message, like ipyleaflet does: https://github.com/jupyter-widgets/ipyleaflet/blob/a19fbca6b8f0f82187ae8459e76e080ecb5841a5/js/src/Map.js#L476

jasongrout:
bqplot also does something similar: https://github.com/bqplot/bqplot/blob/78c8c572156055e8d15a115f528dd59ecc0a5b59/js/src/Figure.ts#L749

massgh:
That's great. Thanks 👍

jasongrout:
Note that the code you see in ipyleaflet handles ipywidgets 7 and 8:

https://github.com/jupyter-widgets/ipyleaflet/blob/a19fbca6b8f0f82187ae8459e76e080ecb5841a5/js/src/Map.js#L450-L485

while the code I pointed out in bqplot only works in ipywidgets 8.

massgh:
> Note that the code you see in ipyleaflet handles ipywidgets 7 and 8:
> 
> https://github.com/jupyter-widgets/ipyleaflet/blob/a19fbca6b8f0f82187ae8459e76e080ecb5841a5/js/src/Map.js#L450-L485
> 
> while the code I pointed out in bqplot only works in ipywidgets 8.

A quick workaround I found equivalent to `on_displayed` is simply displaying JavaScript in Output widget, it runs JavaScript on first display, also when another view is displayed and even when browser window is refreshed. This is useful in cases where somebody is not digging deep into JavaScript side of widgets, just need something quick.

massgh:
> Note that the code you see in ipyleaflet handles ipywidgets 7 and 8:
> 
> https://github.com/jupyter-widgets/ipyleaflet/blob/a19fbca6b8f0f82187ae8459e76e080ecb5841a5/js/src/Map.js#L450-L485
> 
> while the code I pointed out in bqplot only works in ipywidgets 8.

A quick workaround I found equivalent to `on_displayed` is simply displaying JavaScript in Output widget, it runs JavaScript immediately when Output widget is displayed itself or in some Box, also when another view is displayed and even when browser window is refreshed. This is useful in cases where somebody is not digging deep into JavaScript side of widgets, just need something quick.

PR #3549, opened 2022-08-18
Deserialize state before setting it, as is done elsewhere set_state is used

jasongrout:
This is causing at least one problem in the ipyleaflet 8.0 upgrade tests, as seen in https://github.com/jupyter-widgets/ipyleaflet/pull/968

This code is similar to code from another place in this file that has similar logic: https://github.com/jupyter-widgets/ipywidgets/blob/bf6816912f7015ac6a8da0b5db26274da15be48d/packages/base-manager/src/manager-base.ts#L703-L715


martinRenou (approved):
Thanks a lot for investigating this! This looks good to me

PR #3551, opened 2022-08-19
[7.x] Pin to `jupyterlab_widgets>=1.0.0,<3`

jtpio:
As noticed in https://github.com/jupyterlab/jupyterlab/issues/12962 and https://github.com/jupyterlab/jupyterlab/pull/12967, the JupyterLab UI tests were affected by the ipywidgets 8 release even though they pin on `ipywidgets==7.6.5`.

This had to do with `jupyterlab_widgets==3` being pulled when installing `ipywidgets==7.6.5`.

This PR proposes pinning the version of `jupyterlab_widgets` for the 7.x release series to avoid mismatches.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ install_requires = setuptools_args['install_requires'] = [
 extras_require = setuptools_args['extras_require'] = {
     ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
     ':python_version>="3.3"': ['ipython>=4.0.0'],
-    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0,<2'],
+    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0,<3'],
     'test:python_version=="2.7"': ['mock'],
     'test': ['pytest>=3.6.0', 'pytest-cov'],
 }
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ install_requires = setuptools_args['install_requires'] = [
 extras_require = setuptools_args['extras_require'] = {
     ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
     ':python_version>="3.3"': ['ipython>=4.0.0'],
-    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0'],
+    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0,<2'],
     'test:python_version=="2.7"': ['mock'],
     'test': ['pytest>=3.6.0', 'pytest-cov'],
 }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install jupyterlab~=3.0
-        python -m pip install jupyter_packaging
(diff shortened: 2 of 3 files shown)


jtpio:
For reference the current `master` branch (8.x) pins on `~=3.0`:

https://github.com/jupyter-widgets/ipywidgets/blob/89026e1a2f74493774f861ca504d486486cb3155/python/ipywidgets/setup.cfg#L41

vidartf (approved):

vidartf (commented):
Actually, the 2.x version was reserved for 7.x support for lab 4. This might not be released if the lumino 2 upgrade precludes this, but it should still be safe to use `<3` in the pin/

jtpio:
Good point, I updated the pin to `jupyterlab_widgets>=1.0.0,<3` in https://github.com/jupyter-widgets/ipywidgets/pull/3551/commits/46b9d8dc23259e8c6a36ec8aae160265256b58a4

setup.py
```diff
@@ -123,7 +123,7 @@
 extras_require = setuptools_args['extras_require'] = {
     ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
     ':python_version>="3.3"': ['ipython>=4.0.0'],
-    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0'],
+    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0,<3'],
```
jasongrout, line 126: I'm a bit confused - is this saying that anything over python 3.6 should use jupyterlab_widgets 1.x? Should we add a new row now for Python 3.7 or above?
```suggestion
    ':python_version>="3.6"': ['jupyterlab_widgets>=1.0.0,<3'],
    ':python_version>="3.7"': ['jupyterlab_widgets>=3.0.0'],
```

jtpio: I guess that is the way to automatically install `jupyterlab_widgets` when installing `ipywidgets`.

If we dropped support for Python 3.6 on the 7.x branch we could also just replace the existing `:python_version>="3.6"` by `:python_version>="3.7"` directly.

jtpio: > Should we add a new row now for Python 3.7 or above?

Not sure about the new line, I think the pin on `jupyterlab_widgets>=1.0.0,<3` is still relevant on this `7.x` branch to avoid pulling `ipywidgets` 8 and `jupyterlab_widgets` 3 which don't seem compatible.

jasongrout: Ah, sorry, I missed that this was for the 7.x branch. In that case, your changes make a lot of sense.

jasongrout:
How far back should we backport this? Perhaps to 7.7 and 7.6?

jasongrout:
7.6.0 was released 24 Dec 2020, so porting back to 7.6.x gives us almost about 1.5 years of support for the old 7.x releases, and corresponds to the jlab 3.0 release the day before. I think that's probably good enough. If people need to pin farther back than that, they'd have to do it anyway based on their jlab version.

tldr; I think we should backport this to 7.6.x as well.

jtpio:
> Perhaps to 7.7 and 7.6?

Sounds good :+1:

jasongrout:
I spent some time investigating the test failure, but I'm not seeing the root cause. I believe it is just in the development install, though, so given the severity of this issue, I'll go ahead and merge and release, and we can follow up about the development install.

PR #3554, opened 2022-08-19
Tweak some typings

vidartf:
Widen the types of some methods. Using `BufferJSON` will allow objects with `DataView`s to be passed.


jasongrout:
Is this a bugfix you need in 8.0.x for something you are doing?

vidartf:
@jasongrout It is preventing me from passing valid data to `_deserialize_state` in some tests without casting to `any`. E.g. https://github.com/vidartf/ipydatawidgets/blob/51ed3ec0b469f79222df62abd6a399e9fd0ae7db/packages/jupyter-datawidgets/tests/src/common.spec.ts#L21-L32

PR #3556, opened 2022-08-20
Make the selection widgets mapping options error more helpful.

jasongrout:
Based on comments at https://github.com/jupyter-incubator/sparkmagic/issues/769, we could have more helpful error messages about a backwards-incompatible change in ipywidgets 8.

Alternatively, #3557 explores how easy it would be to add support for mapping types again, now that dicts are by default ordered.


vidartf:
Merged for now as a quick fix, and then #3557 can be discussed in more details after.

PR #3557, opened 2022-08-20
Allow Mappings to be given as the options in a selection widget again.

jasongrout:
We deprecated this long ago when dicts were not ordered (#2130, #1958, #2679), then removed it in ipywidgets 8.0. However, now dicts are ordered by default, so perhaps we can put it back in and keep people's code from breaking.

I want to place a high value on easy ways to support backwards compatibility, as many people would rather just not change their code if they can help it.

However, if we are going to stick with this decision in 8.0, #3556 makes the error more helpful. #3556 and this PR are mutually exclusive - at most one should be merged.

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
@@ -187,17 +187,33 @@ def test_dict():
         dict(a=5),
         dict(a=5, b='b', c=dict),
     ]:
-        with pytest.raises(TypeError):
-            c = interactive(f, d=d)
-
+        c = interactive(f, d=d)
+        w = c.children[0]
+        check = dict(
+            cls=widgets.Dropdown,
+            description='d',
+            value=next(iter(d.values())),
+            options=d,
+            _options_labels=tuple(d.keys()),
+            _options_values=tuple(d.values()),
+        )
+        check_widget(w, **check)
 
 def test_ordereddict():
     from collections import OrderedDict
     items = [(3, 300), (1, 100), (2, 200)]
     first = items[0][1]
     values = OrderedDict(items)
-    with pytest.raises(TypeError):
-        c = interactive(f, lis=values)
+    c = interactive(f, lis=values)
+    assert len(c.children) == 2
+    d = dict(
+        cls=widgets.Dropdown,
+        value=first,
+        options=values,
+        _options_labels=("3", "1", "2"),
+        _options_values=(300, 100, 200),
+    )
+    check_widgets(c, lis=d)
 
 def test_iterable():
     def yield_values():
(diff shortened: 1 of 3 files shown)


jasongrout:
The expected tests fail. If we decide to do this PR, I can fix the tests.

vidartf (approved):
There was a general consensus in the dev meeting today that it was probably best to add back the support now that dicts are ordered by default, but to not put any use/mention of this pattern in the docs. @maartenbreddels also argued for not mentioning it in the docstring either, but instead just leave a comment in the code that this support is here for backwards compatibility reasons.

vidartf:
Tests will then need fixing.

jasongrout:
@vidartf - rerequesting review, since tests pass now.

Also pinging @jtpio and @pbugnion, who helped remove the dict option and deprecated it in 7.x (though it appears the deprecation warning was never really seen by users?)

jtpio:
> though it appears the deprecation warning was never really seen by users?

I believe the warning was shown as suggested by the second screenshot in https://github.com/jupyter-widgets/ipywidgets/pull/2679#issuecomment-577199266?

vidartf (approved):

python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
```diff
@@ -187,17 +187,33 @@ def test_dict():
...
+    c = interactive(f, lis=values)
+    assert len(c.children) == 2
+    d = dict(
+        cls=widgets.Dropdown,
+        value=first,
+        options=values,
+        _options_labels=("3", "1", "2"),
+        _options_values=(300, 100, 200),
+    )
+    check_widgets(c, lis=d)
```
vidartf, line 216: The above check uses `**check` while this uses `lis=d`. I'm not sure if I understand the nuance of why that works, but can maybe look again on Tuesday.

vidartf:
I'll leave off the merge for now to leave some time for others to comment, but looks ok to me!

jasongrout:
> I believe the warning was shown as suggested by the second screenshot in [#2679 (comment)](https://github.com/jupyter-widgets/ipywidgets/pull/2679#issuecomment-577199266)?

Only if you reset the DeprecationWarning filter to always show deprecation warnings, as shown in that screenshot. I imagine a vanishingly small fraction of our users did this?

python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
jasongrout: Great question. One is using check_widget, the other is using check_widgets :). That got me for a bit too. I'll rename check_widgets to check_widget_children to be more clear.

jtpio:
> Only if you reset the DeprecationWarning filter to always show deprecation warnings (or ran Python with the appropriate switches to always show deprecation warnings, etc.), as shown in that screenshot. I imagine a vanishingly small fraction of our users did this?

Oh right, then yes probably not many users were able to see it. 

In that case yes it makes sense to revert it since most users probably didn't get the notice :+1:

PR #3561, opened 2022-08-22
Handle `KernelWidgetManager` in the JupyterLab `OutputModel`

jtpio:
Investigating an issue in https://github.com/voila-dashboards/voila/pull/846 where using an `interactive_output` would trigger the following error:

```
TypeError: Cannot read properties of undefined (reading 'sessionContext')
```

https://github.com/voila-dashboards/voila/pull/846 uses the JupyterLab packages but with a custom `KernelWidgetManager`, not a `WidgetManager` like in the JupyterLab extension. This is because it only has access to a kernel connection, not a document context.

However in its present state, the `OutputModel` assumes it can access the `context` from its `widget_manager` here: https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/python/jupyterlab_widgets/src/output.ts#L36 

Ideally it should be able to handle both a `WidgetManager` and `KernelWidgetManager` or even a `LabWidgetManager` directly.

---

For reference the `widget_manager` assigned to a `WidgetModel` is declared as `any` here:

https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/packages/base/src/widget.ts#L72

And assigned here:

https://github.com/jupyter-widgets/ipywidgets/blob/58adde2cfbe2f78a8ec6756d38a7c637f5e599f8/packages/base/src/widget.ts#L124

### DIFF
--- a/python/jupyterlab_widgets/src/output.ts
+++ b/python/jupyterlab_widgets/src/output.ts
@@ -7,7 +7,7 @@ import { JupyterLuminoPanelWidget } from '@jupyter-widgets/base';
 
 import { Panel } from '@lumino/widgets';
 
-import { KernelWidgetManager, WidgetManager } from './manager';
+import { LabWidgetManager, WidgetManager } from './manager';
 
 import { OutputAreaModel, OutputArea } from '@jupyterlab/outputarea';
 
@@ -63,12 +63,7 @@ export class OutputModel extends outputBase.OutputModel {
    * Reset the message id.
    */
   reset_msg_id(): void {
-    let kernel;
-    if (this.widget_manager instanceof WidgetManager) {
-      kernel = this.widget_manager.context.sessionContext?.session?.kernel;
-    } else if (this.widget_manager instanceof KernelWidgetManager) {
-      kernel = this.widget_manager.kernel;
-    }
+    const kernel = this.widget_manager.kernel;
     const msgId = this.get('msg_id');
     const oldMsgId = this.previous('msg_id');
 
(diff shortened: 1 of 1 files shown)


python/jupyterlab_widgets/src/output.ts
```diff
@@ -114,7 +117,7 @@ export class OutputModel extends outputBase.OutputModel {
     }
   }
 
-  widget_manager: WidgetManager;
+  widget_manager: LabWidgetManager;
```
jtpio, line 120: Unfortunately changing from a more specialized to a more generic type here looks like a breaking change.

jtpio: > Thinking that another approach could be for those defining their own widget manager based on a `KernelWidgetManager` to also define their own `OutputModel` and register it

Linking to https://github.com/voila-dashboards/voila/pull/846/commits/38409998b5db79a6fb7bc78dba78325f2116b305 which implements this approach (may be temporary).

vidartf: > Unfortunately changing from a more specialized to a more generic type here looks like a breaking change.

Having breaking changes and therefore a major version bump on the lab manager isn't necessarily that problematic though.

vidartf (approved):
This looks good, but will *probably* require a major version bump of the jupyterlab extension (at least on the JS side), since the typing on the OutputModel changes. That should hopefully not be a big problem though.

martinRenou:
Thanks!

stevejpurves:
This looks like this could have been released in `8.0.x` already but the milestone shows `8.1`, i'n still seeing an error in `8.0.4` and wanted to check that this was expected?

jtpio:
Since this change affects the `jupyterlab-widgets` Python package only, looks like it should have be available now? The latest release shows December 22, 2022 on PyPI:

image

PR #3565, opened 2022-08-23
Allow the 'class' attribute to pass the sanitizer

### LINKED ISSUE #3564: HTML sanitization removes class attribute in descriptions
As of Ipywidgets 8.0, Ipywidgets removes the `class` attribute from HTML in the `description` attribute (https://github.com/jupyter-widgets/ipywidgets/pull/2785). This seems like important functionality.

```
i = interactive(func, val=False)
circle='<span class="projectname_circle" style="background:#d73229"></span>' if circle is not None else ''
i.children[0].description_allow_html = True
i.children[0].description = circle+'Title'
```

The resulting markup:
```
<label class="widget-label-basic">
    <input type="checkbox">
    <span title="<span class=&quot;projectname_circle&quot; style=&quot;background:#d73229&quot;></span>Jam">
        <span style="background:#d73229"></span>Title
    </span>
</label>
```

jasongrout:
This allows the 'class' attribute in the default sanitizer.

Fixes #3564

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -70,7 +70,7 @@ function default_inline_sanitize(s: string): string {
     'ul',
   ];
   const allowedAttributes = {
-    '*': ['aria-*', 'style', 'title'],
+    '*': ['aria-*', 'class', 'style', 'title'],
     a: ['href'],
     img: ['src'],
     style: ['media', 'type'],


vidartf:
Note that the default options for sanitize-html (https://github.com/apostrophecms/sanitize-html#default-options) are quite restrictive. Do we have any sources that support the case that allowing `class` is safe? Or alternatively, so we have any sources that allowing `class` is not safe? Do we have any standard to apply when deciding which tags/attributes to allow? While the default behavior is no HTML at all, since we are maintaining our own custom sanitizer allow-list instead of using the default one, we should have some sort of standard for what we consider allowable or not.

vidartf (approved):
From today's dev meeting, the following points were noted:
- The sanitizer settings are overridden in the lab widget manager, and lab allows the class attribute, so this issue should not have been seen in lab.
- As lab allows the class attribute, and since we allow `style`, it was generally perceived as less risky. 
- We should ideally have a more well defined way of reviewing such changes in the future, and also maybe lab should have a review of its filters.

PR #3566, opened 2022-08-23
Use precommit to normalize codebase

jasongrout:
I've seen several Jupyter projects adopt precommit (thanks @blink1073!). For fun, I copied JupyterLab's precommit file and tried it out here.


Untitled.ipynb
```diff
@@ -0,0 +1,113 @@
+{
```
vidartf, line 1: I assume this file was not meant to be included? :)

jasongrout: I took it out and force-pushed a new set of commits. Thanks for catching this.

jasongrout:
It seems that 3251bc22d79cf889d8aa5d218cfdaca21d5c4be5 introduces a circular dependency or something - we can't even import `ipywidgets`

bollwyvl:
Yep, this would be good. Would be interesting to rekick this to see if anything improves the CI situation.

bollwyvl:
As discussed on #3673: potentially consider adding to `prettier` config:

```yaml
printWidth: 88     # to align with `black`
proseWrap: always  # to improve narrative diffs
```

Also:
- jupyter-black
- and _some_ way of formatting/normalizing markdown cells of notebooks... 
  - I don't yet have a good answer for this, though not for lack of trying: `prettier` still seems to give better results over `mdformat`, but is relatively lousy at many small inputs called from python
  - perhaps a `prettier-plugin-ipynb` would be an option, but out of scope for `ipywidgets`

PR #3567, opened 2022-08-24
fix: make .widget and .widget_types deprecated

### LINKED ISSUE #3562: Note in migration guide the privatization of .widgets and .widget_types
### Problem

The migration guide does not cover the privatization of `.widgets` and `.widget_types`, as noted [here](https://github.com/jupyter-widgets/ipywidgets/pull/3122#issuecomment-1222396020).

### Suggested Improvement

Put a note in the user migration guide that covers these usecases and provides a shim for accessing the variables in ipywidgets 7 and 8 if needed (though note that the variable is now private in ipywidgets 8, so references are discouraged).

maartenbreddels:
In #3122 we renamed .widget and .widget_types to ._active_widgets
and ._widget_types. That breaks code, and we did not have a deprecation
period.
This PR makes the dict and registry non-members of the Widget class
and puts in a backwards compatible way to deprecated these members.

Closes #3562

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/widget.py
+++ b/python/ipywidgets/ipywidgets/widgets/widget.py
@@ -6,8 +6,10 @@
 in the Jupyter notebook front-end.
 """
 import os
+import typing
 from contextlib import contextmanager
 from collections.abc import Iterable
+import warnings
 from IPython import get_ipython
 from ipykernel.comm import Comm
 from traitlets import (
@@ -34,6 +36,9 @@ def envset(name, default):
 PROTOCOL_VERSION_MAJOR = __protocol_version__.split('.')[0]
 CONTROL_PROTOCOL_VERSION_MAJOR = __control_protocol_version__.split('.')[0]
 JUPYTER_WIDGETS_ECHO = envset('JUPYTER_WIDGETS_ECHO', default=True)
+# we keep a strong reference for every widget created, for a discussion on using weak references see:
+#  https://github.com/jupyter-widgets/ipywidgets/issues/1345
+instances : typing.MutableMapping[str, "Widget"] = {}
 
 def _widget_to_json(x, obj):
     if isinstance(x, dict):
@@ -50,8 +55,8 @@ def _json_to_widget(x, obj):
         return {k: _json_to_widget(v, obj) for k, v in x.items()}
     elif isinstance(x, (list, tuple)):
         return [_json_to_widget(v, obj) for v in x]
-    elif isinstance(x, str) and x.startswith('IPY_MODEL_') and x[10:] in Widget._active_widgets:
-        return Widget._active_widgets[x[10:]]
+    elif isinstance(x, str) and x.startswith('IPY_MODEL_') and x[10:] in instances:
+        return instances[x[10:]]
     else:
         return x
 
@@ -259,10 +264,16 @@ class WidgetRegistry:
                             for view_name, widget in sorted(vn.items()):
                                     yield (model_module, model_version, model_name, view_module, view_version, view_name), widget
 
+
+
+# a registry of widgets by module, version, and name so we can create a Python model from widgets
+# that are constructed from the frontend.
+registry = WidgetRegistry()
+
 def register(widget):
     """A decorator registering a widget class in the widget registry."""
     w = widget.class_traits()
-    Widget._widget_types.register(w['_model_module'].default_value,
+    registry.register(w['_model_module'].default_value,
                                  w['_model_module_version'].default_value,
                                  w['_model_name'].default_value,
                                  w['_view_module'].default_value,
@@ -272,6 +283,16 @@ def register(widget):
     return widget
 
 
+class _staticproperty(object):
+    def __init__(self, fget):
+        self.fget = fget
+
+    def __get__(self, owner_self, owner_cls):
+        assert owner_self is None
+        return self.fget()
+
+
+
 class Widget(LoggingHasTraits):
     #-------------------------------------------------------------------------
     # Class attributes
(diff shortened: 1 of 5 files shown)


python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -272,22 +283,46 @@ def register(widget):
...
+        return instances
+
+    @_staticproperty
+    def _active_widgets():
+        warnings.warn("Widget._active_widgets is deprecated, use ipywidgets.widgets.widget.instances", DeprecationWarning)
+        return instances
+
+    @_staticproperty
+    def _widget_types():
+        warnings.warn("Widget._widget_types is deprecated, use ipywidgets.widgets.widget.register", DeprecationWarning)
```
jasongrout, line 315:
```suggestion
        warnings.warn("Widget._widget_types is deprecated, use ipywidgets.widgets.widget.registry", DeprecationWarning)
```

python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -272,22 +283,46 @@ def register(widget):
...
+        return registry
 
-    # _widget_types is a registry of widgets by module, version, and name:
-    _widget_types = WidgetRegistry()
+    @_staticproperty
+    def widget_types():
+        warnings.warn("Widget.widget_types is deprecated, use ipywidgets.widgets.widget.register", DeprecationWarning)
```
jasongrout, line 320:
```suggestion
        warnings.warn("Widget.widget_types is deprecated, use ipywidgets.widgets.widget.registry", DeprecationWarning)
```

jasongrout (commented):
The two conventions for module-level variables I see in pep8 are:

[Global variable names](https://peps.python.org/pep-0008/#global-variable-names):

> (Let’s hope that these variables are meant for use inside one module only.) The conventions are about the same as those for functions.

> Modules that are designed for use via from M import * should use the __all__ mechanism to prevent exporting globals, or use the older convention of prefixing such globals with an underscore (which you might want to do to indicate these globals are “module non-public”).

and [Constants](https://peps.python.org/pep-0008/#constants):

> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

What do you think about making instances  either `_instances` or `INSTANCES` (and same for registry).

python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -272,22 +283,46 @@ def register(widget):
...
     # Class attributes
     #-------------------------------------------------------------------------
     _widget_construction_callback = None
     _control_comm = None
 
-    # _active_widgets is a dictionary of all active widget objects
-    _active_widgets = {}
+    @_staticproperty
+    def widgets():
+        warnings.warn("Widget.widgets is deprecated, use ipywidgets.widgets.widget.instances", DeprecationWarning)
```
jasongrout, line 305: See https://github.com/jupyter-widgets/ipywidgets/pull/3569 for some utility functions to give deprecation warnings the user will actually see.

maartenbreddels: should we see which PR merges first, if this comes first, we can add changes to #3569 to use that

python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -272,22 +283,46 @@ def register(widget):
...
-    # _active_widgets is a dictionary of all active widget objects
-    _active_widgets = {}
+    @_staticproperty
+    def widgets():
+        warnings.warn("Widget.widgets is deprecated, use ipywidgets.widgets.widget._instances", DeprecationWarning)
+        return _instances
+
+    @_staticproperty
+    def _active_widgets():
+        warnings.warn("Widget._active_widgets is deprecated, use ipywidgets.widgets.widget._instances", DeprecationWarning)
```
jasongrout, line 310: Since _instances is now a module private variable, can we not advertise it as if it was a supported public API?

maartenbreddels: What kind of wording did you have in mind?
I'm also not 100% sure we should consider it 'private', more like 'protected'/'expert mode'. I think people can use it, but with the `_` we give a warning saying "this is internal stuff, we don't document the behavior, read the code", and "this shouldn't be normal user code, but it's there if you really need to."

FYI: I'm gonna depend on this dict a lot for the coming period :)

jasongrout: > this is internal stuff, we don't document the behavior, read the code",

That's exactly what I mean by "private". And that if the behavior changes, it's up to you to fix your code. I just think we shouldn't advertise it in the deprecation message as if it was fully supported for people to use as a public API.

Of course, this *is* python, which means that anyone can get into anything...

jasongrout: > What kind of wording did you have in mind?

```suggestion
        warnings.warn("Widget._active_widgets is deprecated", DeprecationWarning)
```

vidartf:
To fully close https://github.com/jupyter-widgets/ipywidgets/issues/3562 I think we should also add a note to the migration guide about this.

jasongrout:
@maartenbreddels - I switched the internal imports to relative imports, and added a note to the migration docs like Vidar suggested in https://github.com/maartenbreddels/ipywidgets/pull/2 - can you take a look?

bollwyvl:
This introduced a small issue, referring to `register.get` instead of `_registry.get`, which breaks creating new widgets in the client.

A prospective fix is up in #3653.

PR #3568, opened 2022-08-26
Fix slider update on readout text change

jtpio:
## Description

In ipywidgets 7 it was possible to edit the readout of a slider, press Enter and the position of the slider would be updated according to the new value:

https://user-images.githubusercontent.com/591645/186837698-817b9328-d6d1-484f-8304-92c07816fd8f.mp4

This does not seem possible anymore with the latest ipywidgets release (8.0.1 at the time of writing).

## Changes

Remove `{ updated_view: this }` from the options when the change comes from updating the text.

**Before**

https://user-images.githubusercontent.com/591645/186837747-a9db61aa-a56f-40cc-837c-acaec9273549.mp4

**After**

https://user-images.githubusercontent.com/591645/186838550-9d248dbd-fa35-417d-ada5-e709f265e003.mp4

### DIFF
--- a/packages/controls/src/widget_float.ts
+++ b/packages/controls/src/widget_float.ts
@@ -215,7 +215,7 @@ export class FloatLogSliderView extends BaseIntSliderView {
 
       if (value !== this.model.get('value')) {
         this.readout.textContent = this.valueToString(value);
-        this.model.set('value', value, { updated_view: this });
+        this.model.set('value', value);
         this.touch();
       } else {
         this.readout.textContent = this.valueToString(this.model.get('value'));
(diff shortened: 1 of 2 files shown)


jtpio:
https://github.com/jupyter-widgets/ipywidgets/pull/2712#issuecomment-585557467 might give some context on the origin of `updated_view`:

> Changed some of the function signatures to accept the updated_view attribute. This is now used to validate whether the slider should update its value when the updateSliderValue callback is invoked. This now enables smooth animations with no lags

jasongrout:
Thanks!

PR #3569, opened 2022-08-26
Calculate correct stack level for deprecation warnings to show them to the user

jasongrout:
Generating deprecation warnings with the correct stacklevel is tricky to get right. Up to now, users have not been seeing these deprecation warnings because we had incorrect stacklevels. See https://github.com/python/cpython/issues/68493 and https://github.com/python/cpython/issues/67998 for good discussions, as well as https://github.com/ipython/ipython/issues/8478#issuecomment-106119371 and https://github.com/ipython/ipython/pull/8480#issuecomment-106107638 for more context.

We see these issues directly. Normally stacklevel=2 might be enough to target the caller of a function. However, in our deprecation warning in the __init__ method, each level of subclasses adds one stack frame, so depending on the subclass, the appropriate stacklevel is different. Thus this PR implements a trick from a discussion in Python to calculate the first stack frame that is external to the ipywidgets library, and use that as the target stack frame for the deprecation warning.

This code should now produce 6 deprecation warnings in Jupyter Notebook or JupyterLab, whereas before it was producing no visible deprecation warnings.

```python
import ipywidgets as w
def f(): pass
x=w.Text(description_tooltip='fda')
x.on_submit(f)
x.description_tooltip
x.description_tooltip='asdf'
w.FileUpload(description_tooltip='asdf')
w.Button(icon='fa-gear')
```
Screen Shot 2022-08-26 at 04 35 19

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/utils.py
+++ b/python/ipywidgets/ipywidgets/widgets/utils.py
@@ -1,0 +1,35 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import sys
+import warnings
+
+# This function is from https://github.com/python/cpython/issues/67998
+# (https://bugs.python.org/file39550/deprecated_module_stacklevel.diff) and
+# calculates the appropriate stacklevel for deprecations to target the
+# deprecation for the caller, no matter how many internal stack frames we have
+# added in the process. For example, with the deprecation warning in the
+# __init__ below, the appropriate stacklevel will change depending on how deep
+# the inheritance hierarchy is.
+def external_stacklevel(internal):
+    """Find the first frame that doesn't any of the given internal strings
+
+    The depth will be 2 at minimum in order to start checking at the caller of
+    the function that called this utility method.
+    """
+    level = 2
+    frame = sys._getframe(level)
+    while frame and any(s in frame.f_code.co_filename for s in internal):
+        level +=1
+        frame = frame.f_back
+    return level
+
+def deprecation(message, internal=None):
+    """Generate a deprecation warning targeting the first external frame
+    
+    internal is a list of strings, which if they appear in filenames in the
+    frames, the frames will also be considered internal.
+    """
+    if internal is None:
+        internal = []
+    warnings.warn(message, DeprecationWarning, stacklevel=external_stacklevel(internal+['ipywidgets/widgets/']))
--- a/python/ipywidgets/ipywidgets/widgets/utils.py
+++ b/python/ipywidgets/ipywidgets/widgets/utils.py
@@ -11,25 +11,28 @@ import warnings
 # added in the process. For example, with the deprecation warning in the
 # __init__ below, the appropriate stacklevel will change depending on how deep
 # the inheritance hierarchy is.
-def external_stacklevel(internal):
-    """Find the first frame that doesn't contain any of the given internal strings
+def _external_stacklevel(internal):
+    """Find the stacklevel of the first frame that doesn't contain any of the given internal strings
 
-    The depth will be 2 at minimum in order to start checking at the caller of
-    the function that called this utility method.
+    The stacklevel will be from the perspective of the caller, and will start checking at the caller
+    of the function that called this function. Thus, the minimum stacklevel will be 1.
     """
+    # Get the level of my caller's caller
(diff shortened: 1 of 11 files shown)


python/ipywidgets/ipywidgets/widgets/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+# This function is from https://github.com/python/cpython/issues/67998
+# (https://bugs.python.org/file39550/deprecated_module_stacklevel.diff) and
+# calculates the appropriate stacklevel for deprecations to target the
+# deprecation for the caller, no matter how many internal stack frames we have
+# added in the process. For example, with the deprecation warning in the
+# __init__ below, the appropriate stacklevel will change depending on how deep
+# the inheritance hierarchy is.
+def external_stacklevel(internal):
+    """Find the first frame that doesn't any of the given internal strings
```
vidartf, line 15:
```suggestion
    """Find the first frame that doesn't contain any of the given internal strings
```

python/ipywidgets/ipywidgets/widgets/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+    The depth will be 2 at minimum in order to start checking at the caller of
```
vidartf, line 17: This logic seems to be using the assumption that `external_stacklevel` is always called from `deprecation`. If this is intended, we might as well make it private (`_external_stacklevel`).

vidartf (approved):
I had two minor lints, but other than that this looks good!

python/ipywidgets/ipywidgets/widgets/utils.py
jasongrout: I thought this default made sense for general use:

level 0: inside external_stacklevel
level 1: the line calling external_stacklevel (usually the warnings.warn line)
level 2: whoever called the function that is trying to give out a deprecation warning

In general, I think it makes sense to point to the code that is calling the function giving the deprecation warning. If I was always calling this only from the deprecate function below, I would have put the default at level 3.

vidartf: To clarify some points:
```python
import ipywidgets
def my_function():
    # what should this output?
    print(ipywidgets.widgets.utils.external_stacklevel([]))
    # and what about this?
    print(ipywidgets.widgets.utils.external_stacklevel(['ipywidgets']))
my_function()
```
This currently prints `2` and `2`, which I argue does not make sense when used directly, when you consider the docstring "Find the first frame that doesn't any of the given internal strings". If that description was correct, I would expect `0` and `1`.

vidartf: If the second sentence of the docstring is taken into account, that explains the behavior of the function, but that means that you would never really know what stack level you were at inside `my_function`.

vidartf: as such, my comment is that the current behavior only makes sense when the stack level of the calling function of `external_stacklevel` is never of interest (which will be the case inside `deprecation`, but does not make sense as a general export that is available from non `_private` paths from the root import of `ipywidgets`.

jasongrout: Ah, I see my mistake - I should adjust the number I'm giving back, since it will be used at a different level than what is was calculated on.

jasongrout:
I was about to push my tests and another fix as well - I'll work on incorporating your changes, especially your path separator fix - thanks!

jasongrout:
I see you are still pushing tests. Please let me know when you are done and I'll work on merging my tests and your tests.

vidartf:
I'm done, but please look at the bit there I removed the `-1`: AFAIU, the stack level is 1-based, not 0-based.

vidartf:
And sorry for not commenting that I was working on it!

jasongrout:
@vidartf - I merged your changes in and I think this is ready for another review from you.

maartenbreddels (approved):
Nice work, also on the traitlet path (a complexity I guess I introduced), I've also checked it on the python repl:
```
>>> w.Text(description_tooltip="fda")
<stdin>:1: DeprecationWarning: the description_tooltip argument is deprecated, use tooltip instead
Text(value='', tooltip='fda')
```
👍

PR #3570, opened 2022-08-26
Stricter typescript + some cleanups

vidartf:
This increases the typescript strictness, and makes the exceptions that we need explicit (this will also opt us in to any future TS strictness checks by default). In addition to that, the following changes were made:
- Fix assumptions about all caught errors being instances of `Error`. With strict checks, TS insists that these are `unknown`.
- Further narrow typing of output from `_deserialize` from #3554: After deserialization, we know we have a dict of something, but have no idea about what is in it, since deserializers can return anything. Ideally we should also make it explicit in `ISerializers` that the return type is unknown, but this might be a breaking change.. ?
- Change the typing of `unpack_models` so that it is compatible with `ISerializers` (newly included strictness check will also complain if not). This is because the `ISerlializers` make all argument optional to allow them to ignore arguments they don't care about, even if the arguments *must always be passed by a caller*. This is chosen to be the lesser of two evils, but this might be open for discussion.
- Narrows the typings of the `ICallbacks` callbacks. Also not clear whether this is an ok change or not.

### DIFF
--- a/packages/base/src/errorwidget.ts
+++ b/packages/base/src/errorwidget.ts
@@ -9,7 +9,7 @@ import { BROKEN_FILE_SVG_ICON } from './utils';
 
 // create a Widget Model that captures an error object
 export function createErrorWidgetModel(
-  error: Error,
+  error: unknown,
   msg?: string
 ): typeof WidgetModel {
   class ErrorWidget extends DOMWidgetModel {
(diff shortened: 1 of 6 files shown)


martinRenou (approved):
LGTM! Thanks

PR #3581, opened 2022-09-01
Datetime widget fixes

### LINKED ISSUE #3500: AttributeError raised by DatetimePicker
https://github.com/jupyter-widgets/ipywidgets/blob/84abba08bd242133b63fe992ebc65b8358b8b2f9/python/ipywidgets/ipywidgets/widgets/widget_datetime.py#L56

This line will raise an error AttributeError when I clear the picked datetime. As described in: 
https://github.com/vidartf/ipydatetime/issues/8

```
venv/lib/python3.8/site-packages/ipydatetime/datetime_widget.py in _validate_tz(self, value)
    135 
    136     def _validate_tz(self, value):
--> 137         if value.tzinfo is not None:
    138             raise TraitError('%s values needs to be timezone unaware' % (self.__class__.__name__,))
    139         return value

AttributeError: 'NoneType' object has no attribute 'tzinfo'
```

vidartf:
- Adds a missing export
- Fixes #3500 (I'm still interested in figuring out what is going wrong there, but this fix should be ok).

### DIFF
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -1161,6 +1161,26 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+       "## Naive picker\n",
+       "\n",
+       "In some cases you might want to be able to pick naive datetime objects, i.e. timezone-unaware datetimes. To quote the Python 3 docs:\n",
+       "\n",
+       "> Naive objects are easy to understand and to work with, at the cost of ignoring some aspects of reality."
+   ]
+   },
+   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+       "ipydatetime.NaiveDatetimePicker(description='Pick a Time')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "## Color picker"
    ]
   },
--- a/docs/source/examples/Widget List.ipynb
+++ b/docs/source/examples/Widget List.ipynb
@@ -1161,11 +1161,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-       "## Naive picker\n",
-       "\n",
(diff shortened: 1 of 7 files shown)


jasongrout:
Thanks. Can you add documentation in the Widget List page for NaiveDateTime?

I was looking at the api, and I think arguably the default date/time picker probably should be the naive one, just like the generic python date/time is naive by default. I think the datetime picker is most often used for filtering historical data, and I think generally users want to pick date/times according to the source data's convention. In other words, if I'm in timezone +5, and I select data from midnight of a certain date to midnight of another date, I think it is most likely I mean midnight of the data's date, not midnight in whatever timezone I happen to be in (i.e., I most likely am wanting the naive date/time). However, we've already shipped 8.0, so perhaps that ship has sailed, and we should just really clearly document the differences between the two, and note when a person might want to use one vs another.

jasongrout-db:
> However, we've already shipped 8.0, so perhaps that ship has sailed

Also, I didn't realize this matches the api in ipydatetime (of course, since you authored both :), so in a sense it is backwards compatible with this api. So yeah, I think just adding more explanation about the NaiveDateTimePicker in the docs, and when you would want to use it vs the DateTimePicker, would be great.

vidartf:
I ported the docs over from ipydatetime (simplified the code).

docs/source/examples/Widget List.ipynb
```diff
@@ -1157,6 +1157,28 @@
...
+    "This is useful if you need to compare the picked datetime to naive datetime objects, as Python will otherwise complain!"
+   ]
+   },
+   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ipywidgets.NaiveDatetimePicker(description='Pick a Time')"
```
jasongrout, line 1179:
```suggestion
    "widgets.NaiveDatetimePicker(description='Pick a Time')"
```

jasongrout:
Thanks!

PR #3582, opened 2022-09-02
Add a note about defining widget model defaults with es6 inheritance.

jasongrout:
Looks like we tell people how to move from `.extend()` to es6 classes, but forgot to tell them how to handle model defaults.


ibdafna (approved):

SylvainCorlay:
This looks good to me !

PR #3584, opened 2022-09-04
Promisetoasync

jmk89:
Some were simple and some were not

### DIFF
--- a/packages/base-manager/src/manager-base.ts
+++ b/packages/base-manager/src/manager-base.ts
@@ -138,61 +138,60 @@ export abstract class ManagerBase implements IWidgetManager {
    * any state updates.
    *
    */
-  create_view<VT extends DOMWidgetView = DOMWidgetView>(
+  async create_view<VT extends DOMWidgetView = DOMWidgetView>(
     model: DOMWidgetModel,
     options?: any
   ): Promise<VT>;
-  create_view<VT extends WidgetView = WidgetView>(
+  async create_view<VT extends WidgetView = WidgetView>(
     model: WidgetModel,
     options?: any
   ): Promise<VT>;
-  create_view<VT extends WidgetView = WidgetView>(
+  async create_view<VT extends WidgetView = WidgetView>(
     model: WidgetModel,
     options = {}
   ): Promise<VT> {
     const id = uuid();
-    const viewPromise = (model.state_change = model.state_change.then(
-      async () => {
-        const _view_name = model.get('_view_name');
-        const _view_module = model.get('_view_module');
-        try {
-          const ViewType = (await this.loadViewClass(
-            _view_name,
-            _view_module,
-            model.get('_view_module_version')
-          )) as typeof WidgetView;
-          const view = new ViewType({
-            model: model,
-            options: this.setViewOptions(options),
-          });
-          view.listenTo(model, 'destroy', view.remove);
-          await view.render();
+    await model.state_change;
+    const _view_name = model.get('_view_name');
+    const _view_module = model.get('_view_module');
+    let viewPromise: Promise<any>;
+    try {
+      const ViewType = (await this.loadViewClass(
+        _view_name,
+        _view_module,
+        model.get('_view_module_version')
+      )) as typeof WidgetView;
+      const view = new ViewType({
+        model: model,
+        options: this.setViewOptions(options),
+      });
+      view.listenTo(model, 'destroy', view.remove);
+      await view.render();
 
-          // This presumes the view is added to the list of model views below
-          view.once('remove', () => {
-            if (model.views) {
-              delete model.views[id];
-            }
-          });
-
-          return view;
-        } catch (e) {
-          console.error(
-            `Could not create a view for model id ${model.model_id}`
-          );
-          const msg = `Failed to create view for '${_view_name}' from module '${_view_module}' with model '${model.name}' from module '${model.module}'`;
-          const ModelCls = widgets.createErrorWidgetModel(e, msg);
-          const errorModel = new ModelCls();
-          const view = new widgets.ErrorWidgetView({
-            model: errorModel,
-            options: this.setViewOptions(options),
-          });
-          await view.render();
-
-          return view;
+      // This presumes the view is added to the list of model views below
+      view.once('remove', () => {
+        if (model.views) {
+          delete model.views[id];
         }
-      }
-    ));
+      });
+
+      viewPromise = Promise.resolve(view);
+    } catch (e) {
+      console.error(
+        `Could not create a view for model id ${model.model_id}`
+      );
+      const msg = `Failed to create view for '${_view_name}' from module '${_view_module}' with model '${model.name}' from module '${model.module}'`;
+      const ModelCls = widgets.createErrorWidgetModel(e, msg);
+      const errorModel = new ModelCls();
+      const view = new widgets.ErrorWidgetView({
+        model: errorModel,
+        options: this.setViewOptions(options),
+      });
+      await view.render();
+
+      viewPromise = Promise.reject(view);
+    }
+
     if (model.views) {
       model.views[id] = viewPromise;
     }
@@ -308,14 +307,13 @@ export abstract class ManagerBase implements IWidgetManager {
     // Create the model. In the case where the comm promise is rejected a
     // comm-less model is still created with the required model id.
     return commPromise.then(
-      (comm) => {
+      async (comm) => {
         // Comm Promise Resolved.
         options_clone.comm = comm;
         const widget_model = this.new_model(options_clone, serialized_state);
-        return widget_model.then((model) => {
-          model.sync('create', model);
-          return model;
-        });
+        const model = await widget_model;
+        model.sync('create', model);
+        return model;
       },
       () => {
         // Comm Promise Rejected.
@@ -327,12 +325,11 @@ export abstract class ManagerBase implements IWidgetManager {
     );
   }
 
-  register_model(model_id: string, modelPromise: Promise<WidgetModel>): void {
+  async register_model(model_id: string, modelPromise: Promise<WidgetModel>): Promise<void> {
     this._models[model_id] = modelPromise;
-    modelPromise.then((model) => {
-      model.once('comm:close', () => {
-        delete this._models[model_id];
-      });
+    const model = await modelPromise
+    model.once('comm:close', () => {
+      delete this._models[model_id];
     });
   }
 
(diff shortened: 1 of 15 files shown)


jmk89:
This is a replacement to https://github.com/jupyter-widgets/ipywidgets/pull/2779

packages/base/src/widget.ts
```diff
@@ -219,64 +218,65 @@ export class WidgetModel extends Backbone.Model {
...
-                this._expectedEchoMsgIds.has(attrName)
-              );
-              expectedEcho.forEach((attrName: string) => {
-                // Skip echo messages until we get the reply we are expecting.
-                const isOldMessage =
-                  this._expectedEchoMsgIds.get(attrName) !== msgId;
-                if (isOldMessage) {
-                  // Ignore an echo update that comes before our echo.
+        try {
+          await this.state_change;
```
jasongrout, line 228: This one is a little tricky. We always want to process messages in order, but the message processing is async. So we have to queue any new message behind the other messages that have already come in. We do this creating a promise for every message we'd like to process. The `this.state_change` attribute points to the last message we've queued. So we chain our new message processing off of `this.state_change` (that's the `then`), and then we update `this.state_change` to point to the new promise we created. It's sort of like a linked list (i.e., a promise chain) where `this.state_change` always points to the last element of the chain, so we know what we need to chain the next promise off of.

Because we need to update `this.state_change`, I don't think we can convert this function to async/await. However, we might be able to refactor the bulk of this function into another async function, then call that to get the promise we need to update `this.state_change` with.

jmk89: That makes things clearer, I see. Seems like this change to [create view](https://github.com/jupyter-widgets/ipywidgets/pull/3584/commits/a843b1d5a29a0eb5e8476f5b852fbda4feb2bf00#diff-4fd512840569881075db4057ecca74fba82b67d649c904876c71d24d8764a8d2L154) in manager base should also remain unconverted to an async function?

jmk89: Updated with some of the logic refactored into an async helper method, and reverted the async change to create_view since this was another place where promises are chained to queue state changes. Tests passed locally.

maartenbreddels:
I wonder what this means for performance (having skimmed over https://v8.dev/blog/fast-async ).
I might do a comparison before/after to see if it does matter.

jmk89:
Having some trouble figuring out what to do about the failing checks. Particularly I can't replicate the `Error: src/widget.ts(289,5): error TS2322: Type 'Dict<unknown>' is not assignable to type 'Dict<BufferJSON>'.` on my local machine when I run ./dev-install.sh or a clean/build. 

The build error is happening in a method I refactored out of _handle_comm_msg. It's nearly a copy and paste of the original promise callback function. Any advice is welcome.

ibdafna:
@jmk89 This is due to the change in the `this` context. I will add a suggestion shortly

ibdafna:
@jmk89 Here is a commit I added last week to fix the issue of not compiling: https://github.com/ibdafna/ipywidgets/pull/1/commits/c4f468eaa2714d4061b7c986fb888e712022700d

jasongrout:
As an update, we discussed this in the ipywidgets dev meeting, and a few of us will have more time to look at this next week during the widgets dev workshop. Thanks!

jmk89:
@ibdafna sounds good to me, thanks for asking 😄  . Although I'm not fully understanding the fix 😅 - it seems to me like setting the state_change to the awaited state_change breaks the "chain of state change" which was being queued. 

It's quite likely that I'm misunderstanding, if there's any enlightenment you can forward my way I'd appreciate it 😄

packages/controls/src/widget_box.ts
```diff
@@ -91,18 +93,15 @@ export class BoxView extends DOMWidgetView {
...
-      .then((views: DOMWidgetView[]) => {
-        // Notify all children that their sizes may have changed.
-        views.forEach((view) => {
-          MessageLoop.postMessage(
-            view.luminoWidget,
-            Widget.ResizeMessage.UnknownSize
-          );
-        });
-      });
+  async update_children(): Promise<void> {
```
DonJayamanne, line 96: Same comment, e.g. in line 92, we're calling this method but not awaiting on that, hence its possible the style is being set before the children are updated.
Logically that doesn't sem correct, though thats how it worked in the past.
Looking at the code now it seems incorrect to not await on this method when its called in other places.

packages/base-manager/src/manager-base.ts
```diff
@@ -327,12 +326,14 @@ export abstract class ManagerBase implements IWidgetManager {
     );
   }
 
-  register_model(model_id: string, modelPromise: Promise<WidgetModel>): void {
+  async register_model(
+    model_id: string,
+    modelPromise: Promise<WidgetModel>
+  ): Promise<void> {
```
DonJayamanne, line 332: Coulnd't this change the overall code, 
previously this wasn't async, now the caller might be tempted into awating on this when they didn't in the past,
and such an await could have other ripple effects, which could potentially lead to race conditions or the like.

For instance if `modelPromise` failed in the past, then the call to `reigster_model` would'nt fail, but now it would.

packages/base-manager/src/manager-base.ts
```diff
@@ -664,91 +663,83 @@ export abstract class ManagerBase implements IWidgetManager {
...
+      1: any call to `new_model` with state registers the model promise (e.g. with `register_model`)
+        synchronously (before it's first `await` statement).
+      2: any calls to a model constructor or the `set_state` method on a model,
+        happens asynchronously (in a `then` clause, or after an `await` statement).
+
+      Without these assumptions, one risks trying to set model state with a reference
+        to another model that doesn't exist yet!
+    */
+
+    const all_models = await Object.keys(models).map(async (model_id) => {
```
DonJayamanne, line 685: I don't think `await would have any effect here we need to change this to 
```typescript
return Promise.all(Object.keys....)
```

packages/base-manager/src/manager-base.ts
```diff
@@ -664,91 +663,83 @@ export abstract class ManagerBase implements IWidgetManager {
...
+        const comm = await this._create_comm(this.comm_target_name, model_id);
+        modelCreate.comm = comm;
+        return this.new_model(modelCreate); // No state, so safe wrt. case 1
+      } else {
+        return this.new_model(modelCreate, modelState); // case 1
+      }
     });
 
-    return all_models;
+    return Promise.all(all_models);
```
DonJayamanne, line 732: I think we can get rid of this variable and just return as the previous code was doing it

packages/base/src/viewlist.ts
```diff
@@ -68,9 +68,8 @@ export class ViewList<T> {
       this.views.length - first_removed
     );
     for (let j = 0; j < removed.length; j++) {
-      removed[j].then(function (view) {
-        remove.call(context, view);
-      });
+      const view = await removed[j];
```
DonJayamanne, line 71: Doesn't this change the behaviour of the code, previously this function would return even before the promise `remoted[j]` resolves.
However now we're changing the behaviour, and forcing these promises to resolve before we can return from this function.

This could have some undesired consequenses

packages/base/src/widget.ts
```diff
@@ -947,7 +955,9 @@ export class DOMWidgetView extends WidgetView {
   /**
    * Public constructor
    */
-  initialize(parameters: WidgetView.IInitializeParameters): void {
+  async initialize(
+    parameters: WidgetView.IInitializeParameters
+  ): Promise<void> {
```
DonJayamanne, line 960: Doesn't this change the behaviour, previously `initialize` was not a promise.
In the past no one would wait for `initialze` to complete even though there were some internal promises, and now we're changing the flow of the code be making `initialize` an async function.
From my understanding this should still remain a sync function 
@jasongrout @ibdafna /cc

packages/base/src/widget.ts
```diff
@@ -991,67 +1000,61 @@ export class DOMWidgetView extends WidgetView {
     this.updateTooltip();
   }
 
-  setLayout(layout: LayoutModel, oldLayout?: LayoutModel): void {
+  async setLayout(layout: LayoutModel, oldLayout?: LayoutModel): Promise<void> {
```
DonJayamanne, line 1003: As earlier, I'm not ssure this old public method should be converted to `async`

packages/base/src/widget.ts
```diff
@@ -991,67 +1000,61 @@ export class DOMWidgetView extends WidgetView {
...
+        this.trigger('layout-changed');
+        this.layoutPromise = Promise.resolve(view);
+      } catch {
+        utils.reject('Could not add LayoutView to DOMWidgetView', true);
+      }
     }
   }
 
-  setStyle(style: StyleModel, oldStyle?: StyleModel): void {
+  async setStyle(style: StyleModel, oldStyle?: StyleModel): Promise<void> {
```
DonJayamanne, line 1037: Same as earlier, shoudl we be chaning these to be async?
If we are changing them, shoudln't all of the calling code be made `async` as well?

DonJayamanne (commented):
I'm not entirely convinced we should be changing the signature of public methods from sync to async and not awaiting on them.
We could inadvertently be introducing a lot of race conditions because previously we didn't await on promises and now we're awaiting on them.

packages/base/src/widget.ts
jmk89: I see your point here. Especially since the outer promises in this method (layoutPromise and possibly create_child_view) are used to chain promises into a queue to be assigned to layoutPromise. Seems like this method should remain sync, however the inner promise `displayed` could be awaited. 

Therefore only the resolve of the `create_child_view` needs to be async.

packages/base-manager/src/manager-base.ts
jmk89: The call to `model.once` could be wrapped in a try catch block to prevent `register_model` from failing. 

As far as the ripple effects of converting a previously sync method to async - is that the nature of converting something to async? Maybe there is a way to signal that this async method doesn't need to be awaited? I can look around to see if this is a pain point with async-conversions. I'm new to async stuff so this is an interesting consideration to me.

packages/base/src/viewlist.ts
jmk89: I can definitely see how this could cause things to go wonky - particularly seems like it would slow down every update cycle significantly. Maybe this is one of the more obvious methods to keep as a promise.

jmk89 (commented):
@DonJayamanne thanks for taking the time to look through my PR and leave feedback. I was initially painting with some broad strokes here.

packages/base-manager/src/manager-base.ts
jmk89: Maybe I'm misunderstanding something with chaining promises, but a rejected promise forces the program execution to jump to nearest rejection handler. In the past case, there was no `.catch()` chained to `modelPromise`, therefore the call to `register_model` would also fail, wouldn't it? 

Regardless, reverting this function to non-async. Because it doesn't make sense to `await` here since it would essentially just be blocking the thread and definitely can cause unnecessary race conditions.

packages/base/src/widget.ts
jmk89: Leaving this one a sync function

packages/base/src/widget.ts
jmk89: Updated this in a similar way to above - The established method of setStyle remains sync. The inner promise is converted to async. This way prevents any confusion if `setStyle` should now be awaited or not.

Also like above, the inverse of the `if (style)` block is used to remove a layer of indenting.

PR #3589, opened 2022-09-14
Button tooltip fix - default to description

### LINKED ISSUE #2286: Button tooltips don't default to the description
Widgets with descriptions have a `description_tooltip` option that sets the tooltip, which defaults to the description if not specified: https://github.com/jupyter-widgets/ipywidgets/blob/1223d4128aebe6c8831a034a73d1546a91f5138a/packages/controls/src/widget_description.ts#L70-L73

The button tooltip defaults to the empty string, rather than defaulting to the description. Perhaps we should make this consistent and default the button tooltip to the description.

jmk89:
Fixes jupyter-widgets/ipywidgets#2286 . Tooltip on buttons no longer defaults to "null". Rather, it defaults to the description if no tooltip is set.

There had been some discussion related to description_tooltips previously, but description_tooltips has been deprecated after PR https://github.com/jupyter-widgets/ipywidgets/pull/2680 was merged. Issue jupyter-widgets/ipywidgets#421 can be closed too.

### DIFF
--- a/packages/controls/src/widget_button.ts
+++ b/packages/controls/src/widget_button.ts
@@ -106,9 +106,7 @@ export class ButtonView extends DOMWidgetView {
     const description = this.model.get('description');
     const icon = this.model.get('icon');
 
-    tooltip
-      ? this.el.setAttribute('title', tooltip)
-      : this.el.setAttribute('title', description);
+    this.el.setAttribute('title', tooltip ?? description);
 
     if (description.length || icon.length) {
       this.el.textContent = '';
--- a/packages/controls/src/widget_button.ts
+++ b/packages/controls/src/widget_button.ts
@@ -101,10 +101,15 @@ export class ButtonView extends DOMWidgetView {
   update(): void {
     this.el.disabled = this.model.get('disabled');
     this.updateTabindex();
-    this.el.setAttribute('title', this.model.get('tooltip'));
 
+    const tooltip = this.model.get('tooltip');
     const description = this.model.get('description');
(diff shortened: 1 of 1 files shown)


packages/controls/src/widget_button.ts
```diff
@@ -101,10 +101,15 @@ export class ButtonView extends DOMWidgetView {
...
     this.updateTabindex();
-    this.el.setAttribute('title', this.model.get('tooltip'));
 
+    const tooltip = this.model.get('tooltip');
     const description = this.model.get('description');
     const icon = this.model.get('icon');
+
+    tooltip
+      ? this.el.setAttribute('title', tooltip)
+      : this.el.setAttribute('title', description);
```
jasongrout, line 111: We can express very similar logic using the nice ?? operator in one line. One advantage to using ?? is that you can do `b.tooltip = ''` to turn off the tooltip if you want (i.e., setting tooltip to None will default to the description, but if tooltip is a string we use it.)

```suggestion
    this.el.setAttribute('title', tooltip ?? description);
```

jasongrout (commented):
Thanks! This looks great. I suggested one change to simplify the code and make it so that setting the tooltip to an empty string disables the tooltip. If you agree, I think we are good to merge this!

packages/controls/src/widget_button.ts
jmk89: Ah very nifty. I don't see any reason to change your suggestion. Seems like a good opportunity to use the "Commit Suggestion" for my first time :)

PR #3590, opened 2022-09-14
fix: using _comm_default shadows possible bugs

maartenbreddels:
This came up when working on  https://github.com/ipython/traitlets/pull/777, instead of `_comm_default`, this method seems to trigger more code paths, which revealed that we couldn't create the DataWidget (since the default was None because we didn't pass in `args=()`).


jasongrout:
Thanks!

maartenbreddels:

PR #3592, opened 2022-09-15
perf: allow the layout of the dom widget to be None

maartenbreddels:
This causes 1 less widget to be created, which matters when you care about performance.
For instance, creating 100 buttons will now create 300 widgets (button, layout and style).
With this change, we only create 200 widgets (button and style). The frontend already supported this, this only enables it in the Python kernel.

Note that the layout is a widget with many traits, causing a significant overhead from traitlets.

I don't think this is a breaking change and can also be backported to 7.x.


jasongrout:
I think this warrants a bump in the model version for base and core controls, right?

https://github.com/jupyter-widgets/ipywidgets/blob/633048585e55095519d153e14ea93e7510a17ad5/packages/base/src/version.ts#L6
https://github.com/jupyter-widgets/ipywidgets/blob/633048585e55095519d153e14ea93e7510a17ad5/packages/controls/src/version.ts#L8
https://github.com/jupyter-widgets/ipywidgets/blob/633048585e55095519d153e14ea93e7510a17ad5/python/ipywidgets/ipywidgets/_version.py#L10-L12

maartenbreddels:
> I think this warrants a bump in the model version for base and core controls to 2.1.0, right? This would signify it is backwards compatible, but that it is a change to the model spec.

I agree, also output since it inherits from DOMWidget.




> by default you wouldn't see these speed gains. You would only see them if you explicitly set the layout to None, right? In other words, by default we create a Layout widget for you.

Correct.

PR #3594, opened 2022-09-17
Add dev meeting info to readme

jasongrout:
I'm always looking for the dev meeting info in this repo instead of the team-compass repo, so I've now added it to the readme here.


martinRenou (approved):
Thank you!

PR #3602, opened 2022-09-29
Perf: various optimizations

maartenbreddels:
Gives a ~25-30% performance boost for creating widgets (benchmarking against 1000 buttons) by

  * not having comm a trait (no need for this)
  * doing less for every instance (and caching it per class) (.keys and .to_json lookup)
  * micro-optimization for `state_get`
     *  not using unneeded `_trait_to_json`
     *  also using the cached to_json (`_trait_to_json_dict`)
     * not checking `not PY3` in loop
     * not checking `drop_defaults` in loop

Supersedes https://github.com/jupyter-widgets/ipywidgets/pull/3020

Benchmarks are done including this branch https://github.com/ipython/traitlets/pull/777

### DIFF
--- a/ipywidgets/widgets/tests/utils.py
+++ b/ipywidgets/widgets/tests/utils.py
@@ -26,7 +26,6 @@ _widget_attrs = {}
 undefined = object()
 
 def setup_test_comm():
-    Widget.comm.klass = DummyComm
     ipywidgets.widgets.widget.Comm = DummyComm
     _widget_attrs['_ipython_display_'] = Widget._ipython_display_
     def raise_not_implemented(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


filipre:
A somewhat wild guess but I think 

> not having comm a trait (no need for this)

might fix https://github.com/jupyter-widgets/ipywidgets/issues/3448. At least it would make debugging the issue a bit easier

danielhollas:
Hi :wave: 

Just wanted to signal our interest in this PR. :blush: I've tested this branch and by my (very rough) benchmark, this would save us around 300 ms startup time for our app. :open_mouth: (the linked `traitlets` PR already saved us 600ms, thanks so much @maartenbreddels for your work!)

Not sure if I can be of any help here besides testing. I've tried to merge with the current `7.x` branch and there were only two minor conflicts that should be easily resolvable. I did not try to put this on top of 8.x branch.

maartenbreddels:
> Just wanted to signal our interest in this PR. 😊

Thank you :)

I'm experimenting with this in solara, adding it to ipyvue at runtime https://github.com/widgetti/solara/pull/389

I think the `not having comm a trait` feature might be something more for ipywidget9, as it might be a breaking change.

PR #3616, opened 2022-10-21
Update 'Develop and Build Documentation' with new instructions

kellietay:
PR Summary: 
- docs/readme.md was updated in this [commit](https://github.com/jupyter-widgets/ipywidgets/commit/b20ad59a9272ab9afb144386e8e3fc083904e287), but this change was not updated in the official readthedocs.
- this PR updates the instructions on docs/Readme.md to docs/dev_docs.md
- We also update docs/Readme.md to point to the official docs.

### DIFF
--- a/docs/Readme.md
+++ b/docs/Readme.md
@@ -1,24 +1,3 @@
 # Building ipywidgets documentation
 
-To build the ipywidgets documentation:
-
-1. Install the development version of ipywidgets with the `dev-install.sh` script in the repo root directory.
-2. Install the Python packages for building documentation with either mamba or conda:
-   ```sh
-   conda env update --file docs/environment.yml 
-   ```
-
-   or with `pip`:
-
-   ```sh
-   python -m pip install -r docs/requirements.txt
-   # Also install pandoc separately
-   ```
-
-3. Build the HTML docs with sphinx:
-   ```sh
-   cd docs/source
-   python -m sphinx -T -E -b html -d ../build/doctrees -D language=en . ../build/html
-   ```
-
-4. Open the documentation from the build directory: `docs/build/html`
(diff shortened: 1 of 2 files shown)


jasongrout:
@supriyakhandekar volunteered to review this

docs/Readme.md
```diff
@@ -1,24 +1,3 @@
...
-   ```
-
-3. Build the HTML docs with sphinx:
-   ```sh
-   cd docs/source
-   python -m sphinx -T -E -b html -d ../build/doctrees -D language=en . ../build/html
-   ```
-
-4. Open the documentation from the build directory: `docs/build/html`
+To build the ipywidgets documentation see the instructions available on - [Jupyter Widgets - Develop and Build Documentation](https://ipywidgets.readthedocs.io/en/latest/dev_install.html)
```
supriyakhandekar, line 3: We may want to link the following instead: https://ipywidgets.readthedocs.io/en/latest/dev_docs.html

supriyakhandekar: Or rather, this seems to the correct link we want to provide: https://github.com/jupyter-widgets/ipywidgets/tree/master/docs

supriyakhandekar: (Since the changes to the dev_docs.md seem to be mirroring that^)

kellietay: Instructions in the readme were removed so that we would not have to maintain two different sets of the same instructions on both the readme and the ipywidgets.readthedocs.io, but i can add back the readme instructions as well

supriyakhandekar: It makes sense to maintain one set of instructions however it seems that the instructions here https://ipywidgets.readthedocs.io/en/latest/dev_install.html is for the dev install of ipywidgets, not specifically for documentation. And if we were to link https://ipywidgets.readthedocs.io/en/latest/dev_docs.html instead, this seems to be different set of instructions from the README instructions that you updated

Issue #3597, opened 2022-09-21
Slider widgets misses value change with clicking when continuous update is False


imcovangent:
## Description

When a slider widget is used, one would expect that a value change is observed when the slider is clicked and when it is dragged. However, when `continous_update` is set to `False`, then a click on the slider to change the value is not observed.

## Reproduce

1. Make a jupyter notebook.
2. Add slider widget in first cell:
```
from ipywidgets import widgets
test_sl = widgets.IntSlider(
        min=0,
        max=100,
        step=1,
        description="Test:",
        disabled=False,
        continuous_update=False,
        orientation="horizontal",
        readout=True,
        readout_format="d",
    )
display(test_sl)
```
3. Show value of the slider in a second cell:
```
test_sl.value
```
4. Drag slider to new value
5. Rerun the second cell to see that the slider value is updated correctly.
6. Now click on the slider to change the value
7. Rerun the second cell to see that the slider value is not updated as expected.

## Expected behavior

I would expect that the slider value is updated and observed if you click a slider, also when continuous updating is disabled.

## Context

- ipywidgets version 8.0.1
- Operating System and version: Windows 10
- Browser and version: Chrome

ceriottm:
Hi! I also noticed this, which I'm pretty sure is a change in behavior relative to ipywidgets 7. Can anyone confirm whether this is deliberate or a regression?

jasongrout:
This looks like a regression to me. CC @ibdafna

maartenbreddels:
This is fixed in ipywidgets 8.0.4

PR #3617, opened 2022-10-21
Fix slider change event issue with tapping

ibdafna:
Signed-off-by: Itay Dafna <i.b.dafna@gmail.com>

Addresses #3597

### DIFF
--- a/packages/controls/src/widget_float.ts
+++ b/packages/controls/src/widget_float.ts
@@ -4,7 +4,10 @@
 import { CoreDescriptionModel } from './widget_core';
 
 import {
-  BaseIntSliderView, IntRangeSliderView, IntSliderView, IntTextView
+  BaseIntSliderView,
+  IntRangeSliderView,
+  IntSliderView,
+  IntTextView,
 } from './widget_int';
 
 import { format } from 'd3-format';
@@ -100,7 +103,7 @@ export class FloatLogSliderModel extends BoundedFloatModel {
   readout_formatter: any;
 }
 
-export class FloatRangeSliderModel extends FloatSliderModel { }
+export class FloatRangeSliderModel extends FloatSliderModel {}
 
 export class FloatSliderView extends IntSliderView {
   /**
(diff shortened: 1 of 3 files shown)


jasongrout:
@ibdafna - it does look like the linting test doesn't pass, so we do need to run the lint on these changes.

vidartf (approved):
The code changes themselves look good, but I would maybe want you to do an extra pass on the docstrings? Some of them don't match anymore, some are missing, and some could be made more precise (e.g. the update event happens for more than just the slide event, but it is maybe the most common case, and the change event happens for more than just mouse up).

packages/controls/src/widget_float.ts
```diff
@@ -237,6 +239,18 @@ export class FloatLogSliderView extends BaseIntSliderView {
...
+  /**
+   * Called when the slider handle is released after dragging,
+   * or by tapping or moving by the arrow keys.
+   */
+  handleSliderChangeEvent(values: number[], handle: number): void {
+    const base = this.model.get('base');
+    const actual_value = Math.pow(base, this._validate_slide_value(values[0]));
+    this.readout.textContent = this.valueToString(actual_value);
+
+    this.handleSliderChanged(values, handle);
```
vidartf, line 251: I guess one problem that can occur is that when you tap or move by the arrow keys, you will get two events: one from the update event, and one from the change event. If we wanted to avoid this, we could conditionally only pass the event on if `continuous_update` is false here (and other similar event handlers). I'm just not sure how bad of an issue this duplicate is, and if it can lead to some issues with multiple users, or if backbone will automatically discard the second change as a no-op.

vidartf (approved):
Looks good, but a comment for consideration either in this PR or for future issues.

ceriottm:
Just to confirm that this fixes #3597 also in our setting. Awesome. It'd be great to have this merged in the near future, as we need this to deploy an interactive course that relies heavily on sliders.

maartenbreddels:
This PR is released in ipywidgets 8.0.4

PR #3618, opened 2022-10-21
Pass change options into update calls

jasongrout:
We assume these options several places in the code, for example in the selection widget views, but we were never passing them through when we called the update method.


vidartf (approved):
The change itself looks good, and fixes an annoying bug! I think this bug is the reason that `TimeView` has an `update2` method? So that one can probably be changed back to use `update` then.

PR #3620, opened 2022-10-23
Support horizontal orientation of radio buttons

### LINKED ISSUE #1247: Make it possible to have horizontal radio buttons
From #1239:
> I needed to arrange radio buttons horizontally, couldn't seem to accomplish this with existing widget in 5.2.2 so I made my own. Not sure if this has been addressed in 6.0, perhaps I could use the ToggleButtons now.

DonJayamanne:
Fixes #1247
Before I work on the documentation, I'd like to wait and see if this is the right approach.
I've followed the same approach as the sliders, using the same option name `orientation` with the same obvious values.

The only area I'm unsure about is the CSS class name and the approach to change the orientation.

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -976,11 +976,29 @@
 }
 
 /* <DEPRECATED> */
+.widget-radio-box-vertical, /* </DEPRECATED> */
+.jupyter-widget-radio-box-vertical {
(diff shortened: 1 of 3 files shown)


martinRenou (approved):
Thanks! I think this approach looks good. Sorry for the delay in reviewing.

We can update the documentation separately

PR #3621, opened 2022-10-25
Update dependencies for doc build

jasongrout:
Two small updates to the documentation dependencies

With these updates, I can begin the build process for docs on a mac m1.


bollwyvl:
#3673 is up, and updates to `python-build`: I'm reasonably confident this will work from a mamba/conda environment on an M1, but pip's another story...

PR #3625, opened 2022-10-27
WidgetRenderer: Soften manager type requirement

martinRenou:
This way we can use a WidgetRenderer with a `KernelWidgetManager` as well


ibdafna (approved):
Thank you!

PR #3642, opened 2022-11-29
fix: do not hold sync during set_state

### LINKED ISSUE #3635: Hold sync during update from front-end is very un-ipywidgets like
## Description

In https://github.com/jupyter-widgets/ipywidgets/pull/3271 we added the behavior that updates from the frontend will wrap the update of traits in a `hold_sync()` (which is released in `>7, <=8.0.2`). This behavior feels to me un-ipywidget like (update a property, and the frontend reflects it).

As a response to an event from the front end, very often, some calculation occurs, which takes a bit of time. Before and after that calculation, we often want to give the user some indicator that work is being done.

Now with 8.x we cannot set such an indicator (by changing a property, like description or disabled) on the widget that triggered the original event due to the `hold_sync()` context manager. Note that it is possible to change any other widget, just not the one we received the event from. This feels (as a user) very inconsistent.

### Code
```python
import ipywidgets as widgets
import time

def do_work():
    time.sleep(3)

def on_value(change):
    buttons.description = "Working..."
    buttons.style
    buttons.disabled = True
    do_work()
    buttons.disabled = False
    buttons.description = "Item (restored)"

buttons = widgets.ToggleButtons(options=["foo", "bar"], description="Item")
buttons.observe(on_value, "value")
buttons
```

## 7.x behaviour

In 7.x, this works as expected:
hold-sync-issue-7 x

## 8.x behaviour

In 8.x, this does not work anymore:
hold-sync-issue-8 x

# Solution

I think we should no do a hold_sync, and do a new release that removes this, considering this a bug.

maartenbreddels:
If we do, we get very inconsistent behavior, see #3635 for more details and #3271 for the reasons to implement this.

Reverts https://github.com/jupyter-widgets/ipywidgets/pull/3271
Closes https://github.com/jupyter-widgets/ipywidgets/issues/3635

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_set_state.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_set_state.py
@@ -287,10 +287,13 @@ def test_hold_sync(echo):
     msg = {'method': 'echo_update', 'state': {'value': 42.0}, 'buffer_paths': []}
     call42 = mock.call(msg, buffers=[])
 
-    msg = {'method': 'update', 'state': {'value': 2.0, 'other': 11.0}, 'buffer_paths': []}
+    msg = {'method': 'update', 'state': {'value': 2.0}, 'buffer_paths': []}
     call2 = mock.call(msg, buffers=[])
 
-    calls = [call42, call2] if echo else [call2]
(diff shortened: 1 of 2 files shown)


jasongrout (approved):
Thanks!

Issue #3630, opened 2022-11-09
Unable to get ipywidget web3 example working


andrewphilipsmith:
## Description

Unable to get ipywidget web3 example working

(Other notes: I posted this on discourse here https://discourse.jupyter.org/t/unable-to-get-ipywidget-web3-example-working/16601, but have had no response so far). 

## Reproduce

1. create and activate a new conda environment
2. clone ipywidgets
3. follow the ]setup instructions for the example “web3”](https://github.com/jupyter-widgets/ipywidgets/tree/master/examples/web3#try-it)
4. try opening the webpage (http://localhost:8080) in firefox, chrome or safari.
5. Check that the default value of the "Notebook BASEURL" matches the value given when starting the notebook server and then press OK.
6. A blank page is shown in the browser. (See JS Console output from each browser below).

Additionally, I have tried `git checkout tags/8.0.2` immediately after cloning the repo (step 2) without effect.

I have tried recreating this several times and always get the same result. I do not have an opinion on whether this is an error is the web3 example itself or the documentation describing it.

## Expected behaviour

Some visual indication of a widget being successfully loaded in the webpage.

## Context

- ipywidgets version 
`8.0.2`

- Operating System and version:
macOS 12.6.1 (21G217)

- Browser and version:
Google Chrome Version 107.0.5304.87 (Official Build) (arm64)
Firefox 106.0.3 (64-bit)

- Others (if required)
yarn --version ==> 1.22.19
conda --version ==> 22.9.0

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.

$PATH:
	</my/absolute/home>/anaconda3/envs/ipywidgets-test/bin
	</my/absolute/home>/.cargo/bin
	/opt/homebrew/opt/tcl-tk/bin
	</my/absolute/home>/bin/google-cloud-sdk/bin
	~/.gem/ruby/2.6.0/bin
	/opt/local/bin
	/opt/local/sbin
	</my/absolute/home>/.poetry/bin
	/opt/homebrew/bin
	/opt/homebrew/sbin
	</my/absolute/home>/.pyenv/shims
	</my/absolute/home>/.pyenv/bin
	</my/absolute/home>/anaconda3/condabin
	/usr/local/bin
	/usr/bin
	/bin
	/usr/sbin
	/sbin
	/opt/X11/bin
	/usr/local/anaconda3/bin

[...]

PR #3643, opened 2022-12-02
define `process.env` for jlab dependencies

stevejpurves:
Addresses issue #3630 by using webpack to provide stubs for the expected globals `process` and `process.env`. Note defining `process` alone solved the first error but exposed a further issue in `bluebird` that required `process.env`. I maintained the `cwd` stub as per `jlab` example.

### DIFF
--- a/examples/web3/webpack.config.js
+++ b/examples/web3/webpack.config.js
@@ -18,4 +18,12 @@ module.exports = {
       { test: /\.svg$/i, type: 'asset' },
     ],
   },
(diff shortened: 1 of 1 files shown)


supriyakhandekar (approved):
LGTM

jasongrout:
It looks like the test failure is unrelated and common across all recent PRs, so I'm assuming it is not related to this. Thanks for testing this, Supriya!

PR #3673, opened 2023-01-22
Overhaul docs toolchain

bollwyvl:
## References

- fixes #3656

## Changes
- [x] replace `nbsphinx` with `myst-nb`
- [x] replace `sphinx-rtd-theme` with `pydata-sphinx-theme` 
- [x] use `jupyterlite` directly
- [x] update `prettier`
  - [x] apply to `.md`, `.ya?ml`
- [x] move `index.rst` to `.md`
- [x] update `lerna`
- [x] update `eslint`
- [x] update `typescript`
- [x] update CI action versions   
- [x] add `sphinx-autodoc-typehints` 
- [x] add `intersphinx`
- [x] add `sphinx-copy-button`  
- [x] add `typedoc`
- [x] add `jupyterlab-myst` 
- [x] reorganize docs structure around [4 kinds of docs](https://www.writethedocs.org/videos/eu/2017/the-four-kinds-of-documentation-and-why-you-need-to-understand-what-they-are-daniele-procida/)
- [x] clean up broken markdown links
- [x] fix galata tests

## Future Work
- [ ] add dark theme variables
- [ ] add docs for integrating with `myst-nb` (and potentially `pydata-sphinx-theme` and downstreams)
- [ ] hoist preferred CDN URL (ick) to API somewhere
- [ ] probably ship a known-good requirejs somewhere
- [ ] add anchor-level `pytest-check-links` (especially for internal URLs)
- [ ] squash more sphinx warnings, ideally getting to "fail on warning"
- [ ] update prettier to consistently apply `printWidth: 88` and `proseWrap: always`

PR #3674, opened 2023-01-24
More docs updates

bollwyvl:
## References

- follow-up on #3673 
  - thanks @jtpio re piplite path https://github.com/jupyter-widgets/ipywidgets/pull/3673#discussion_r1084660125

## Changes

- [x] fix jupyterlite wheel paths
- [x] update branch references: `master` -> `main`
- [x] try to fix binder workflow (haven't seen it re-run yet, because it only fires on `pull_request_opened`)
- [x] look at some more sphinx warnings
- [x] restore full index sidebar
- [x] rework demo links (less bs (ha!))

### DIFF
--- a/.binder/README
+++ b/.binder/README
@@ -1,10 +1,10 @@
 This directory holds configuration files for https://mybinder.org/.
 
 The interactive notebooks can be accessed with this link:
-https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/master?filepath=docs/source/examples
+https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/main?filepath=docs/source/examples
 
-To check out a different version, just replace "master" with the desired
+To check out a different version, just replace `main` with the desired
 branch/tag name or commit hash.
 
 To use JupyterLab, use:
-https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/master?urlpath=lab/tree/docs/source/examples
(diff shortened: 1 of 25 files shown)


bollwyvl:
Modulo not knowing if this will fix the binder badge workflow,  this is ready for review (and :jack_o_lantern: merge is again fine).

agoose77:
These changes are all fantastic.

Minor points:

- For me, the button icons are misaligned
image
- Personally I prefer `API Reference` to `Reference`
- I'd be tempted to move the try buttons to the index page only, maybe as a card?
- On Awkward Array, we have a quick-links card that might be similarly useful (maybe without the badges for ipywidgets) https://awkward-array.org/doc/main/

bollwyvl:
> button icons are misaligned

the first two are actually radios, was trying to "act like a local" and use bootstrap sugar. I'll try a treatment without.

>  move the try buttons to the index page only,

Could move to the bottom. Actually, it would be lovely to have a sidebar that was "open in lite," next to "edit in github".

> API Reference to Reference

I'm just working off the upstream language guidance. There might be things there _other_ than the API at some point, though can't tell what those would be.

bollwyvl:
image

agoose77:
> image

Much better!!

jasongrout:
Thanks!

bollwyvl:
Cool. I think that's all the high-value stuff that wouldn't require a new release of something.

jasongrout:
> Cool. I think that's all the high-value stuff that wouldn't require a new release of something.

Thanks again for the massive amount of overhauling you did!

PR #3677, opened 2023-01-24
Add pytest-check-links for internal URLs

bollwyvl:
## References
- contains #3674 
  - this could be an _unrelated_ PR, but is also testing that the binder badge workflow is correct 
- continues #3673

## Changes
- [x] adds `pytest-check-links`
- [x] adds docs about running locally
- [x] adds to docs CI

## Future work
- checking external links is a bit more fiddly, and benefits from a more precise use of caching to avoid making _way_ too many requests and getting rate-banned

### DIFF
--- a/.binder/README
+++ b/.binder/README
@@ -1,10 +1,10 @@
 This directory holds configuration files for https://mybinder.org/.
 
 The interactive notebooks can be accessed with this link:
-https://mybinder.org/v2/gh/jupyter-widgets/ipywidgets/master?filepath=docs/source/examples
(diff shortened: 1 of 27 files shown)


bollwyvl:
Note this is _still_ failing the binder badge, as I guess it uses the one from `main`?

bollwyvl:
Whew, the generated pages are: a) really slow and b) do have a lot of errors.

I'm not sure how to gate this properly to make it effective in CI: waiting for a merge to `main` seems _too late_, while blocking every PR seems worse.

I guess it could be split into multiple jobs:
- one for typedoc
- one for autodoc
- one for "the rest"

bollwyvl:
I don't think it's gonna finish :crying_cat_face:

PR #3689, opened 2023-02-03
fix: deepcopy with `structuredClone` over JSON.parse/stringify

manzt:
[`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) is now the preferred way to deeply copy objects in JS. It is widely supported across browsers, minus some edge cases like in Safari web-workers (see the table in the linked documentation).

 The old `JSON.parse` + `JSON.stringify` trick **deletes** any `DataView`s (without a custom serializer) within `WidgetModel.serialize`, leading to a serious bug when using raw "binary" traitlets. With the `{remove,put}_buffers` implementations, extra serializers are _not_ required for transferring binary data. The following example should work without custom serializers but immediately deletes `text` value when `this.model.save_changes()` is called for the first time.

```javascript
import { DOMWidgetModel, DOMWidgetView } from "@jupyter-widgets/base";
class CustomModel extends DOMWidgetModel { /* ... */ }
class CustomView extends DOMWidgetView {
  render() {
    let decoder = new TextDecoder(), encoder = new TextEncoder();
    let text = document.createElement("textarea");
    text.oninput = () => {
      let bytes = encoder.encode(text.value);
      this.model.set("text", new DataView(bytes.buffer));
      this.model.save_changes(); // internally calls JSON.parse(JSON.stringify(DataView)) -> {}
    };
    let on_change = () => text.value = decoder.decode(this.model.get("text"));
    on_change();
    this.model.on("change:text", on_change);
    this.el.appendChild(text);
  }
}
export { CustomModel, CustomView };
```

```python
class CustomWidget(ipywidgets.DOMWidget):
    # ...
    text = traitlets.Any(b'hello, world').tag(sync=True)
```

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -524,6 +524,8 @@ export class WidgetModel extends Backbone.Model {
    * binary array buffers.
    */
   serialize(state: Dict<any>): JSONObject {
+    const deepcopy = 
+      globalThis.structuredClone || (x: any) => JSON.parse(JSON.stringify(x));
     const serializers =
       (this.constructor as typeof WidgetModel).serializers || {};
     for (const k of Object.keys(state)) {
@@ -532,7 +534,7 @@ export class WidgetModel extends Backbone.Model {
           state[k] = serializers[k].serialize!(state[k], this);
         } else {
           // the default serializer just deep-copies the object
-          state[k] = JSON.parse(JSON.stringify(state[k]));
+          state[k] = deepcopy(state[k]);
         }
         if (state[k] && state[k].toJSON) {
           state[k] = state[k].toJSON();
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -524,8 +524,8 @@ export class WidgetModel extends Backbone.Model {
    * binary array buffers.
    */
   serialize(state: Dict<any>): JSONObject {
-    const deepcopy = 
-      globalThis.structuredClone || (x: any) => JSON.parse(JSON.stringify(x));
+    const deepcopy =
+      globalThis.structuredClone || ((x: any) => JSON.parse(JSON.stringify(x)));
     const serializers =
       (this.constructor as typeof WidgetModel).serializers || {};
     for (const k of Object.keys(state)) {


manzt:
You can see this bug with [anywidget](https://github.com/manzt/anywidget) in [Google Colab](https://colab.research.google.com/drive/1FcawqJ8zZXa9uFXk_ms3OvmqoADHlisz?usp=sharing).

```python
import anywidget
import traitlets

class BinaryTextArea(anywidget.AnyWidget):
    _esm = """
    let decoder = new TextDecoder(), encoder = new TextEncoder();
    export function render(view) {
        let text = document.createElement("textarea");
        text.oninput = () => {
          let bytes = encoder.encode(text.value);
          view.model.set("text", new DataView(bytes.buffer));
          view.model.save_changes();
        };
        function on_change() {
          text.value = decoder.decode(view.model.get("text"));
        }
        on_change();
        view.model.on("change:text", on_change);
        view.el.appendChild(text);
    }    
    """
    text = traitlets.Any(b'hello, world').tag(sync=True)


area = BinaryTextArea()
area
```

packages/base/src/widget.ts
```diff
@@ -524,6 +524,8 @@ export class WidgetModel extends Backbone.Model {
    * binary array buffers.
    */
   serialize(state: Dict<any>): JSONObject {
+    const deepcopy = 
+      globalThis.structuredClone || (x: any) => JSON.parse(JSON.stringify(x));
```
jasongrout, line 528: From the test failures, it looks like we need parentheses around the fallback function.

manzt: Good catch, thanks.

maartenbreddels:
Whow, this is good news. This makes it much better to implement binary support on the front end. I think it would be good to backport this to 7.
@jasongrout would you classify this as a fix or feature?

jasongrout:
> @jasongrout would you classify this as a fix or feature?

I think it's fine to backport to 7.x. It's definitely backwards compatible, right? Especially with the fallback.

jasongrout:
The visual test failure is unrelated - looks like jlab puts a notification in the background that we are capturing:

Before: 
5ce58ad328bfeb9101010a5869582d09cd9a3423

After: 
e8e28c1915c46f081dfa437328e00a0e2b83efc1

Diff:
53c0f077a9d975bdaf6162ddf4a69d6885b84e5f

jasongrout (approved):

martinRenou:
> The visual test failure is unrelated

Updating Galata seems to do the trick for getting rid of this new notification. We had the same in bqplot.

jasongrout:
Visual regression fixed in #3694. Thanks @martinRenou

jasongrout:
Thanks @manzt!

manzt:
I know this caused a lot of issues, and I'm very sorry for that. Since anywidget doesn't have the same backward compat requirements, I decided to use `structuredClone` there.

maartenbreddels:
No need to say sorry, thank you for contributing! The burden of breaking changes rests on the maintainers, we are all responsible.
I hope you continue to contribute, I love the work you are doing ❤️

PR #3696, opened 2023-02-06
fix: unset cursor of widgets

### LINKED ISSUE #3692: widgets do not compose well with `cursor: default`
Due to lumino:
https://github.com/jupyterlab/lumino/blob/dbb4f6aae04996abfbc5ee17d5fa8e693f3eb024/packages/widgets/style/widget.css#L19

All our widgets gets `cursor: default`. This means that when an `Image` widget is a child of an anchor tag (e.g. it's embedded in a container that turns the image into a link), we don't get a "hand" icon.

Not sure if lumino shouldn't do this, or if we should reset this.

maartenbreddels:
Fixes #3692

### DIFF
--- a/packages/controls/css/widgets-base.css
+++ b/packages/controls/css/widgets-base.css
@@ -1062,3 +1062,14 @@ actually support closable tabs, so it really doesn't matter */
     max-width: 100%;
     height: auto;
 }
+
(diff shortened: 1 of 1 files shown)


maartenbreddels:
if this looks good, I'm happy to port this to main/8.x

jasongrout (approved):

PR #3697, opened 2023-02-06
fix: deepcopy with structuredClone over JSON.parse/stringify

maartenbreddels:
Backport of #3689 from main/8.x to 7.x

cc @manzt

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -492,6 +492,8 @@ class WidgetModel extends Backbone.Model {
      * binary array buffers.
      */
     serialize(state: {[key: string]: any}) {
+        const deepcopy =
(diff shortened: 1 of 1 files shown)


bollwyvl:
Nice. Looks fine on [caniuse](https://caniuse.com/?search=structuredClone)... safari in a worker doesn't seem like a blocker.

Of note, lumino _also_ offers a [deepCopy](https://github.com/jupyterlab/lumino/blob/38757e291d84837b8dfdb43f5098951feaec3d40/packages/coreutils/src/json.ts#L228), which might be an intermediate fallback before the painful JSON roundtrip.

maartenbreddels:
I think that would be better, and we should do it for 7.x and main/8.x

PR #3699, opened 2023-02-09
Add potential fix for increasing map size bug

filipre:
*Untested, how can I easily test this?*

This PR fixes https://github.com/jupyter-widgets/ipywidgets/issues/3448 and https://github.com/jupyter-widgets/ipywidgets/issues/3448

Solution proposed by @maartenbreddels here: https://github.com/jupyter-widgets/ipywidgets/issues/3448#issuecomment-1423215233

### DIFF
--- a/python/ipywidgets/ipywidgets/embed.py
+++ b/python/ipywidgets/ipywidgets/embed.py
@@ -12,10 +12,12 @@ Functions for generating embeddable HTML/javascript of a widget.
 
 import json
 import re
-from .widgets import Widget, DOMWidget, widget as widget_module
-from .widgets.widget_link import Link
-from .widgets.docutils import doc_subst
+
 from ._version import __html_manager_version__
+from .widgets import DOMWidget, Widget
+from .widgets import widget as widget_module
+from .widgets.docutils import doc_subst
+from .widgets.widget_link import Link
 
 snippet_template = """
 {load}
@@ -299,7 +301,21 @@ def embed_minimal_html(fp, views, title='IPyWidget export', template=None, **kwa
         will be replaced by all the widgets.
     {embed_kwargs}
     """
-    snippet = embed_snippet(views, **kwargs)
+
+    if kwargs.get("state") is None:
+        state = dependency_state(views, drop_defaults=kwargs.get("drop_defaults"))
+    else:
+        state = kwargs["state"]
+        
+    snippet = embed_snippet(
+        views,
+        drop_defaults=kwargs.get("drop_defaults", True),
+        state=state,
+        indent=kwargs.get("indent", 2),
+        embed_url=kwargs.get("embed_url", None),
+        requirejs=kwargs.get("requirejs", True),
+        cors=kwargs.get("cors", True)
+    )
 
     values = {
         'title': title,


python/ipywidgets/ipywidgets/embed.py
```diff
@@ -299,7 +301,21 @@ def embed_minimal_html(fp, views, title='IPyWidget export', template=None, **kwa
...
+        
+    snippet = embed_snippet(
+        views,
+        drop_defaults=kwargs.get("drop_defaults", True),
+        state=state,
+        indent=kwargs.get("indent", 2),
+        embed_url=kwargs.get("embed_url", None),
+        requirejs=kwargs.get("requirejs", True),
+        cors=kwargs.get("cors", True)
+    )
```
maartenbreddels, line 318: I think it would be good to explicitly add the arguments (just copy-paste them from embed_snippet).
Would also be good to update the docstring, which is not incorrect (the behaviour of state=None is different).
Maybe this should be an opt-in for backward compatibility. (`state_tree_shake=False` maybe?)

maartenbreddels (changes requested):
Awesome, for taking this on. Left a comment for you, let me know what you think.

python/ipywidgets/ipywidgets/embed.py
filipre: If the correct behaviour is `dependency_state` and not `get_manager_state`, then I think the backwards compatibility relies on a bug 🤔 What do you think about passing a string instead, e.g. `state="all"` to include *everything* instead of having an argument toggle that switches between dependencies and full state (and only if `state` is `None`)

filipre: I updated the documentation and also used a string to toggle between the modes:

`state` is `None` or equals `"dependent"`: function will use `dependency_state`:
```
Current: 0.html: 0.0067501068115234375 MB
Current: 1.html: 0.0067501068115234375 MB
Current: 2.html: 0.0067501068115234375 MB
Current: 3.html: 0.0067501068115234375 MB
Current: 4.html: 0.0067501068115234375 MB
...
```

`state` equals `"complete"`: function will use `get_manager_state`:
```
Current: 0.html: 0.006749153137207031 MB
Current: 1.html: 0.012719154357910156 MB
Current: 2.html: 0.01868915557861328 MB
Current: 3.html: 0.024659156799316406 MB
Current: 4.html: 0.03062915802001953 MB
...
```

`state` is a dictionary: function will use passed state direclty
```
Current: 0.html: 0.0007772445678710938 MB
Current: 1.html: 0.0007772445678710938 MB
Current: 2.html: 0.0007772445678710938 MB
Current: 3.html: 0.0007772445678710938 MB
Current: 4.html: 0.0007772445678710938 MB
...
```

Could you take another look on the code?

python/ipywidgets/ipywidgets/embed.py
```diff
@@ -297,9 +310,44 @@ def embed_minimal_html(fp, views, title='IPyWidget export', template=None, **kwa
...
+        If True avoids sending user credentials while requesting the scripts.
+        When opening an HTML file from disk, some browsers may refuse to load
+        the scripts.
     """
-    snippet = embed_snippet(views, **kwargs)
+
+    if state is None or state == "dependent":
+        state = dependency_state(views, drop_defaults=drop_defaults)
+    elif state == "complete":
+        state = Widget.get_manager_state(drop_defaults=drop_defaults, widgets=None)['state']
```
vidartf, line 340: Would it make sense to move this logic into `embed_data` instead? Then we get support for these enums in more functions (and we also get to reuse the docstrings 😄).

python/ipywidgets/ipywidgets/embed.py
```diff
@@ -297,9 +310,44 @@ def embed_minimal_html(fp, views, title='IPyWidget export', template=None, **kwa
...
+        Enables the requirejs-based embedding, which allows for custom widgets.
+        If True, the embed_url should point to an AMD module.
+    cors: boolean (True)
+        If True avoids sending user credentials while requesting the scripts.
+        When opening an HTML file from disk, some browsers may refuse to load
+        the scripts.
     """
-    snippet = embed_snippet(views, **kwargs)
+
+    if state is None or state == "dependent":
```
vidartf, line 337: Changing the default (`None` case) is the most controversial here, as it is strictly speaking backwards incompatible. There might be a case for calling it a workaround for the poor GC of ipywidgets, but making this change should at least have some more eyes on it before merging.

filipre: I understand your argument and it's difficult for me to estimate how much of an impact this change would have to other projects. Coming from [ipyleaflet](https://github.com/jupyter-widgets/ipyleaflet/issues/1098), it just looks like a bug to me and then I'd argue, that backwards compatibility should not rely on bugs. However, if other people really need the whole state and really have that implementation assumption as well, then I agree, we should stay backwards compatible as well. What do you think?

Your [solution](https://github.com/jupyter-widgets/ipywidgets/issues/3448#issuecomment-1459919374) definelty will work in the end but as a user, it's not so clear to me when I'd use the dependent state and when the complete.

maartenbreddels: I kind of agree it is a bug, we didn't implement this feature for nothing. I think this was the intended behavior, but yeah, fixing a bug can be considered a breaking change :)

vidartf: See original discussion here: https://github.com/jupyter-widgets/ipywidgets/pull/1387#discussion_r119173856

PR #3720, opened 2023-03-17
Fix CI

martinRenou:


martinRenou:
I'm not sure what is the exact origin of the issue, but there is a mismatch in the mimebundle between Python 3.7 tests and higher versions (maybe a mismatch in IPython/ipykernel versions?)

maartenbreddels:
thanks!

martinRenou:

PR #3721, opened 2023-03-17
Handle `null` _view_count

bollwyvl:
This ensures the `_view_count` doesn't stay `null` once displayed.

c/f https://discourse.jupyter.org/t/how-do-i-trigger-an-event-after-a-widget-is-displayed/18506

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -759,7 +759,7 @@ export class WidgetView extends NativeView<WidgetModel> {
 
     this.once('displayed', () => {
       if (typeof this.model.get('_view_count') === 'number') {
-        this.model.set('_view_count', this.model.get('_view_count') + 1);
+        this.model.set('_view_count', (this.model.get('_view_count') || 0) + 1);
         this.model.save_changes();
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I'm not sure I understand this change. If `_view_count` is a number, we can do a +1 right? Why is the `|| 0` part needed?

bollwyvl:
It seems like the intent is to embiggen the count, but it initializes to
null, so adding to it always results in null, and therefore never changes,
and never sends a sync back to the kernel.

Unless initialized by the user to 0.

Unless it is not supposed to be used at all. In which case I don't know how
one is supposed to do what the linked forum post asks anymore (do something
in kernel when drawn in client for the first time).

maartenbreddels:
but the `if (typeof this.model.get('_view_count') === 'number') {` which handles that right?

The idea is that it is disabled by default, when the value is null/None. It can be enabled by initializing it to 0. This was done to reduce the amount of traffic.

bollwyvl:
Happened as part of this big boy: #2021

PR #3738, opened 2023-03-24
Add default `pack_models`

bollwyvl:
## References
- naive, untested sketch of proposal from https://github.com/jupyter-widgets/ipywidgets/issues/3735#issuecomment-1481647590

## Changes
- [x] hoists critical `IPY_MODEL_` magic string to a constant
- [x] removes some anonymous function definitions in (recursive) loops in `unpack_models`
- [x] adds `pack_models`
  - [x] uses for any unset-`serializer` that's a neighbor of `deserialize: unpack_models`
- [x] makes use of the frozen `JSONExt.emptyObject` in some places to avoid allocating new objects

## Questions
- does it need to be a promise?
- throw a warning?
  - ick. these get called _a lot_. maybe call once per class?
- hoist `widget_serialization` to mirror the backend API?

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -70,8 +70,12 @@ export function unpack_models(
 
 /** Replace models with ids recursively.
  *
- * If the commonly-used `unpack_models` is given as the `seralize` method,
- * and no `deserialize` is given, this will be used as a default.
+ * If the commonly-used `unpack_models` is given as the `deseralize` method,
+ * pack_models would be the appropriate `serialize`.
+ * However, the default serialize method will have the same effect, when
+ * `unpack_models` is used as the deserialize method.
+ * This is to ensure backwards compatibility, see:
+ *   https://github.com/jupyter-widgets/ipywidgets/pull/3738/commits/f9e27328bb631eb5247a7a6563595d3e655492c7#diff-efb19099381ae8911dd7f69b015a0138d08da7164512c1ee112aa75100bc9be2
  */
 export function pack_models(
   value: WidgetModel | Dict<WidgetModel> | WidgetModel[] | any,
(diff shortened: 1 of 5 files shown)


bollwyvl:
I've taken this out of draft and updated the notes a bit: probably don't have time to dig into the packaging file... more `node`, more problems.

.binder/requirements.txt
```diff
@@ -9,10 +9,3 @@ pandas
 scikit-image
 scikit-learn
 sympy
-
-# checking https://github.com/jupyter-widgets/ipywidgets/issues/3735
```
bollwyvl: I've removed these here, and opened this issue: https://github.com/jupyter-widgets/team-compass/issues/20

maartenbreddels:
@bollwyvl thank you for taking a look at this!

I wonder if instead of having a `pack_models` implementation, we should use the "old" `JSON.parse(JSON.stringify(x)))` method instead. We may be breaking code that relies on the `toJSON` option, what do you think of that?

I'll see if I can at least add a test to make this not happen again.

maartenbreddels:
In code, what I propose:
```diff
-const deepcopy =
-  globalThis.structuredClone || ((x: any) => JSON.parse(JSON.stringify(x)));
+const deepcopyJSON = (x: JSONValue) => JSON.parse(JSON.stringify(x));
+
+const deepcopy = globalThis.structuredClone || deepcopyJSON;
```

```diff
@@ -572,7 +574,7 @@ export class WidgetModel extends Backbone.Model {
 
         if (serialize == null && keySerializers.deserialize === unpack_models) {
           // handle https://github.com/jupyter-widgets/ipywidgets/issues/3735
-          serialize = pack_models;
+          serialize = deepcopyJSON;
```

I do like you pack_models and I think we should ship it, and use it, so we don't rely on the JSON trick in the ipywidget code base. This asymmetry has confused me many many times.

maartenbreddels:

PR #3749, opened 2023-03-28
fix: revert ipykernel dependency removal

maartenbreddels:
This change was too large for a fix release and hurt many people.
Let's move this to an 8.1 release, add comm as a dependency, and not have it break anything
see https://github.com/jupyter-widgets/ipywidgets/pull/3748 for the 7.x fix

### DIFF
--- a/python/ipywidgets/setup.cfg
+++ b/python/ipywidgets/setup.cfg
@@ -34,6 +34,7 @@ zip_safe = False
 packages = find:
 
 install_requires =
+  ipykernel>=4.5.1
   ipython>=6.1.0
   traitlets>=4.3.1
   widgetsnbextension~=4.0


martinRenou (approved):

SylvainCorlay:
I think it is very important that we remove this dependency. Pulling ipykernel (which depends on e.g. Tornado) also makes the case of WebAssembly more cumbersome, as well as the usage of ipywidgets in xeus-python.

I wish we removed it for 8.0. Could we move forward with removing the ipykernel dependency in ipywidgets?

martinRenou:
> e.g. ipywidgets-core that only includes the python code and has less dependencies (just traitlets and ipython?). Move ipywidgets into a meta package (no code, just dependencies) that depends on ipywidgets-core, ipykernel, widgetsnbextension and jupyterlab_widgets.

I don't think we should take such a complicated path. The initial issue that lead to reverting this change was merely test setups that were relying on ipykernel to be around through the ipywidgets dependency.

I think having the comm package around, providing a default comm that fits the requirements for testing with ipywidgets, should be enough.

maartenbreddels:
So, if we include the comm dependency, no CI should break? Do you remember why we didn't do that before?

martinRenou:
Adding the comm dependency may be problematic because of the check here: https://github.com/jupyter-widgets/ipywidgets/blob/39d3c5df684539830e2232ac9265abf4331555f1/python/ipywidgets/ipywidgets/__init__.py#L28
This check acts more or less as an ipykernel version check, without importing ipykernel. If we add the comm dependency we may break old ipykernel versions by going into this condition.

The CI/tests break was due to having no-one registering a proper comm implementation in tests, because pytest runs are not done inside an ipykernel session, nor a xeus-python session. But you actually fixed this issue in https://github.com/ipython/comm/commit/309b8295ca950a9ca9bdc0daa796215d72a7cb09 by providing a default comm implementation that does nothing (you basically implemented the "mock comms" I was suggesting above).

So we may be able to remove again the ipykernel with no other change.

martinRenou:
Prior to your change in the comm package, another way to workaround those CI/tests failure was by importing ipykernel in tests (what I suggest here https://github.com/jupyter-widgets/ipywidgets/issues/3729#issuecomment-1478152137), that way it would actually register a comm implementation that ipywidgets can use.

maartenbreddels:
If we make comm a dependency, we can remove the try/except. We could, however, see if ipykernel is in sys.modules, and if below a specific version, use ipykernel's Comm (basically the reverse of what we do now?)

martinRenou:
Yes that could be a good approach indeed. I'll come up with a PR.

PR #3707, opened 2023-02-23
Update to jupyterlab 4 and lumino 2

echarles:
Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3700

PR #3752, opened 2023-04-04
Add support for building against JupyterLab 4 and Lumino 2

### LINKED ISSUE #3781: Upgrading extension for JupyterLab 4 compatibility
### Problem

Last week, JupyterLab release [v4.0.0](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.0) that need upgrade existing extensions to use the new extesion systen and packaging.

Using current ipywidgets extesions in JupyterLab 4 that would throw `Error displaying widget`.

### Proposed Solution

Check JupyterLab [Extension Migration Guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#upgrading-extension-using-the-upgrade-script).

### LINKED ISSUE #3790: In JupyterLab 4.02, ipywidgets throws error in console on saving widget state.
In JupyterLab 4.02 with the setting  "save widget state automatically" enabled, I get the following message in the console when saving the notebook (even if there's no widget present in the notebook):
```
TypeError: this._context.model.metadata.set is not a function at f._saveState
```
The error does not come when the "save widget state automatically" setting is not activated.

here is the environment that I've used:
```
python3.11 -m venv .venv
source .venv/bin/activate   
pip install jupyterlab   
pip install ipywidgets 
python -m jupyterlab
```
image

I've also tested this in an older version (pip install jupyterlab==3.6), and there it worked fine without the error.

Context: I've observed this error while working on this issue https://github.com/widgetti/ipyreact/issues/40

martinRenou:
Follow up #3707 

Fix #3790


echarles:
@martinRenou  I was just going to take over https://github.com/jupyter-widgets/ipywidgets/pull/3707 which was working apart the galata test (now has conflicts).

Question around dependencies: You are giving choice to javascript in package.json (e.g. "@jupyterlab/services": "^6.0.0 || ^7.0.0-beta.0" - "@lumino/coreutils": "^1.11.1 || ^2"), however the python version is pinned to verison 4 of jupyterlab (jupyterlab ==4.0.0b0). We already discussed that a bit, could you bring more details around that as my first feeling is that there is a risk on inconsistency, but I may miss background here.

martinRenou:
Thanks for commenting. I wanted to debug a bit on this.

> however the python version is pinned to verison 4 of jupyterlab (jupyterlab ==4.0.0b0)

In tests yes. To make sure we test against the latest jlab in the ui-tests.

> my first feeling is that there is a risk on inconsistency

There shouldn't be any issue coming from the the choice we give in the JavaScript packages versions. This is just giving more freedom on the downstream packages that depend on `@jupyter-widgets` packages (like Voila, which is currently stuck with `@lumino` 1 and `@jupyterlab` 3 due to ipywidgets). But the ipywidgets package *should* be built with Lumino 2 and Lab 4 and the built output published.

However, yarn is being annoying (again) here and pulls both Lumino 1 and Lumino 2 packages in the dev installation for no obvious reason. I'm a bit stuck now because of this.

python/jupyterlab_widgets/src/plugin.ts
```diff
@@ -86,9 +85,9 @@ function* outputViews(
     app.shell.widgets(),
     (w) => w.id.startsWith('LinkedOutputView-') && (w as any).path === path
   );
-  for (const view of toArray(linkedViews)) {
-    for (const outputs of toArray(view.children())) {
-      for (const output of toArray(outputs.children())) {
+  for (const view of linkedViews) {
+    for (const outputs of view.children()) {
+      for (const output of outputs.children()) {
```
martinRenou, line 90: We should check that this works in JupyterLab 3

python/jupyterlab_widgets/src/manager.ts
```diff
@@ -455,7 +455,7 @@ export class WidgetManager extends LabWidgetManager {
    */
   private _saveState(): void {
     const state = this.get_state_sync({ drop_defaults: true });
-    this._context.model.metadata.set('widgets', {
+    this._context.contentsModel?.content.metadata.set('widgets', {
```
fcollonval, line 458: You should never touch the `contentsModel`. In this case you should use the `model.setMetadata`

```suggestion
    this._context.model.setMetadata('widgets', {
```

martinRenou:
TODO:

Check that https://github.com/jupyter-widgets/ipywidgets/issues/3785 is fixed with this PR

ndmlny-qs:
@martinRenou I just checked and these changes do not fix #3785 or #3781. You mentioned above

> However, yarn is being annoying (again) here and pulls both Lumino 1 and Lumino 2 packages in the dev installation for no obvious reason. I'm a bit stuck now because of this.

I think I've figured out what is happening here. There are several package.json files that request lumino pacakages with versions that do not exist. For instance, `python/widgetsnbextension/package.json` is requesting `@lumino/messaging: "^1.10.1 || 2.1`. However, the latest published version for `@lumino/messaging` is 2.0.0. I think `yarn` is searching for `2.1`, doesn't find it, and reverts to the `1.x` version. Changing the places where a lumino package is not yet at version 2.1, but is at 2.0, prevents yarn from pulling in 1.x versions. I tried this locally and it works.

Unfortunately it still does not solve #3785 or #3781(using my local build) and jlab4, which I suspect are CSS issues needing to be resolved.

martinRenou:
We really need a bot to update snapshots 😓

python/jupyterlab_widgets/src/manager.ts
```diff
@@ -455,7 +455,7 @@ export class WidgetManager extends LabWidgetManager {
    */
   private _saveState(): void {
     const state = this.get_state_sync({ drop_defaults: true });
-    this._context.model.metadata.set('widgets', {
+    this._context.model.setMetadata('widgets', {
```
fcollonval, line 458: This is Lab 4 specific. So it should be tested like

```
if ( this._context.model.setMetadata ) {
} else {
}
```

python/jupyterlab_widgets/src/manager.ts
```diff
@@ -512,7 +512,7 @@ export class WidgetManager extends LabWidgetManager {
    * Load widget state from notebook metadata
    */
   async _loadFromNotebook(notebook: INotebookModel): Promise<void> {
-    const widget_md = notebook.metadata.get('widgets') as any;
+    const widget_md = notebook.getMetadata('widgets') as any;
```
fcollonval, line 515: This is also Lab 4 specific. So it should be tested like

if ( this._context.model.getMetadata ) {
} else {
}

fcollonval (changes requested):
Thanks a lot @martinRenou. I spotted two places where Lab4 specific code is used.

python/jupyterlab_widgets/src/manager.ts
martinRenou: Thanks!

martinRenou:
Thanks a lot for the review @fcollonval !

ui-tests/tests/widgets.test.ts-snapshots/widgets-cell-10-linux.png
paddymul, line 1: It looks like all of these image changes added 1 px vertically.  Was that due to something inherent in lumino2/JL4 ?  If it is, then this change is just a downstream effect from the lumino2/JL4 change.  If not, then this PR shouldn't cause updated images.

trungleduc (approved):

trungleduc:
Hi all, is there anything else that needs to be fixed or this PR is ready?

ui-tests/tests/widgets.test.ts-snapshots/widgets-cell-10-linux.png
martinRenou: I suspect this comes with the galata and JupyterLab 4 update.

Since this PR is not touching CSS code nor visual JavaScript code I guess it's fine to just update the references to match what the UI-tests are getting.

martinRenou:
All good on my side! Thank you for the review

SylvainCorlay:
Hey @afshin do you have a bit of time to review the lumino bits of this PR?

ui-tests/tests/widgets.test.ts-snapshots/widgets-cell-10-linux.png
paddymul: thinking about this further. Do the unit tests also test against JL 3.6?  will there be an oscillation of failures depending on which is tested against?

martinRenou: I updated galata which comes from JupyterLab 4, so it's only testing JupyterLab 4 from now on

afshin (approved):

paddymul:
What is the ongoing ipywidgets story for 3.6?  Since we removed integration testing against 3.6, how can library authors have confidence their code will work against 3.6 and 4.0?Sent from my iPhoneOn Jun 23, 2023, at 3:41 PM, Afshin Taylor Darian ***@***.***> wrote:﻿
@afshin approved this pull request.

SylvainCorlay (approved):

martinRenou:
> What is the ongoing ipywidgets story for 3.6?  Since we removed integration testing against 3.6, how can library authors have confidence their code will work against 3.6 and 4.0?

We could indeed add another CI workflow for testing against JupyterLab 3.x.

trungleduc:
🚀 🎉 💯

PR #3759, opened 2023-04-14
feat: Thread safe kernel side Output widget

maartenbreddels:
# Background
Working on support the output widget in solara:
 * https://github.com/widgetti/solara/pull/68
 * More specifically https://github.com/widgetti/solara/pull/68/commits/6802aa53a307eeb15ca0b57c87e9b414f97d9149

I found out about the hooks in ipykernel https://github.com/ipython/ipykernel/pull/115

This approach will probably be better way forward.

# The issue
Without a frontend, the output widget does not work. We had to work around that for Voila https://github.com/voila-dashboards/voila/pull/91 (now ported to nbclient https://github.com/jupyter/nbclient/pull/68 ).

Futhermore, the frontend is not aware of threads, causing all kinds of issues, e.g.:
 * https://github.com/jupyter-widgets/ipywidgets/issues/3260
 * https://github.com/jupyter-widgets/ipywidgets/issues/2358

This does require https://github.com/ipython/ipykernel/pull/1110, so even if we go forward with this, the old code will have to stay in.

# Testing and ipykernel

This solution only works with ipykernel/ipython, and I am not sure how to test this. Also, this goes against the direction of begin independent of a particular implementation. For instance, I am not sure xeus-kernel could support this (cc @martinRenou ).

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_widget_output.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_widget_output.py
@@ -237,3 +237,16 @@ def test_append_display_data():
         },
     )
     assert widget.outputs == expected1 or widget.outputs == expected2
+
+
+def test_kernel_side_output():
+    output = widget_output.Output()
+    with output:
+        print("snakes!")
+    assert output.outputs == (
+        {
+            'output_type': 'stream',
+            'name': 'stdout',
+            'text': 'snakes!
+        }
+    )
--- a/python/ipywidgets/ipywidgets/widgets/widget_output.py
+++ b/python/ipywidgets/ipywidgets/widgets/widget_output.py
@@ -62,6 +62,10 @@ class Output(DOMWidget):
 
     __counter = 0
 
+    def __init__(self, *args, **kwargs):
+        super(Output, self).__init__(*args, **kwargs)
+        self._hooks = []
+
     def clear_output(self, *pargs, **kwargs):
         """
         Clear the content of the output widget.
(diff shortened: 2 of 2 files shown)


maartenbreddels:
Example code that works in the notebook:
```python
import threading
import time
import ipywidgets as widgets


def run1():
    with out1:
        for i in range(10):
            display("output to 1")
            time.sleep(0.2)

def run2():
    with out2:
        for i in range(10):
            display("output to 2")
            time.sleep(0.2)
        

out1 = widgets.Output()
out2 = widgets.Output()
threading.Thread(target=run1).start()
threading.Thread(target=run2).start()
widgets.HBox([out1, out2])
```

This would totally break before this PR, all output seems to go to 1 of the widgets.

jasongrout:
> This solution only works with ipykernel/ipython, and I am not sure how to test this. Also, this goes against the direction of begin independent of a particular implementation. For instance, I am not sure xeus-kernel could support this (cc @martinRenou ).

This hits on the fundamental decision we had to make when originally implementing output widgets and evaluating the experiments at https://github.com/ipython/ipykernel/pull/115: Do we make the kernel more complicated (with an output redirection mechanism) and the frontend simple (which arguably is a better architecture), or do we make the frontend more complicated and need no changes to any kernel? We decided there were likely to be far more kernels than frontends, so the balance shifted to the somewhat awkward architecture we have today, which is kernel agnostic.

In practice, though, there have been fewer kernels implementing ipywidgets support and more frontends implementing ipywidgets support than (at least I) anticipated. So I'm glad you are reevaluating the original tradeoff.

maartenbreddels:
Thank you for the context Jason!

Note that this also requires https://github.com/ipython/ipykernel/pull/1135 to be able to intercept the clear_output

PR #3768, opened 2023-05-11
fix: fallback to lumino deepCopy rather than JSON.parse

naterush:
This PR fixes issue: https://github.com/jupyter-widgets/ipywidgets/issues/3701

You can test this by:
1. Rendering a widget
2. Using it as normal
3. Opening your console and writing `delete window.structuredClone`
4. Continuing to use the widget and seeing it still works as normal

I think this is a proper test, as I have verified that the `serialize` code is being called, and that it defaults to the correct function. Let me know if there is more to test though!


maartenbreddels:
hmm, do you understand the failure?

PR #3779, opened 2023-05-23
DOC: added instruction to "Installing with pip" section

StefanieSenger:
I've added an instruction to navigate to a sub-folder in order to install the project with pip. Hope this is valuable to future contributors.


StefanieSenger:
Adding some context: The `setup.py` file is contained in the sub-folder. So before installing the package, the contributing developer needs to navigate there.

jasongrout:
Thanks!

PR #3792, opened 2023-06-27
Update yarn version in dev installation documentation

trungleduc:


maartenbreddels:
I thought we needed <2 everywhere, and this is also what is in https://github.com/jupyter-widgets/ipywidgets/blob/main/docs/source/dev_release.md and what I used for releasing.

trungleduc:
I saw It is bumped to version 3 in the readme in the last PR. The yarn 1 config file is also replaced with the yarn 3 one.

ferdnyc:
> I saw It is bumped to version 3 in the readme in the last PR. The yarn 1 config file is also replaced with the yarn 3 one.

Yeah, actually the bump to requiring Yarn 3 was done in 8c08faf4a1b634503255b66f6e5a550685f0423b — I just updated the README and the Binder scripting to match, plus shook out some of the consequences of the change (like dropping a yarn plugin that's only compatible with version 1).

Looks like https://github.com/jupyter-widgets/ipywidgets/blob/main/docs/source/dev_release.md should be updated as well. Running `yarn set version stable` from the ipywidgets repo root directory will switch to using Yarn 3 locally, even on systems where the `yarn` command is otherwise version 1.

martinRenou:
It seems this PR is outdated and the yarn version was fixed in another PR, sorry @trungleduc

PR #3793, opened 2023-06-29
Docs: Remove the note about mappings not working with selection widgets.

ferdnyc:
The documentation for Selection widgets still contained a prominent admonishment that `options=` could no longer be a Mapping as of ipywidgets 8.0. But #3557 undid that change, so it's no longer an issue.

This reverts commit 5a7307413a0d432af0b6aaa773ee3372178bf504.


ferdnyc:
#3794 has the fix for CI doc builds.

martinRenou (approved):
Thank you!

PR #3795, opened 2023-06-30
Binder and yarn tweaks

ferdnyc:
(Separate) commits in this PR:

1. Upgrade jupyterlab in Binder from 3.x to 4.0.2
2. Run 'yarn set version stable' in Binder, to use yarn 3
3. Remove `yarn-deduplicate`, which is only compatible with yarn 1. (It's unnecessary with yarn 2+, anyway.)
4. Adds `packageExtensions:` configs to the `.yarnrc.yml`, to silence warnings about undeclared dependencies in imported packages. (Except for the fundamental incompatibility of `istanbul-instrumenter-loader` with webpack 5.) See #3797 for more info.

    (Note: There's also a remaining undeclared `react` dependency in `jupyterlab/settingregistry`, but adding that to the `packageExtensions` makes `react` a dependency of the project as well, which was a road I decided not to go down.)

5. <s>Create a `.prettierignore` file in the repo root, and add `.yarnrc.yml` to it.</s> 
    Add a comment, 
    ```yml
    # prettier-ignore
    ```
    to the section of `.yarnrc.yml` that Yarn will edit on its own. (Using a `.prettierignore` file didn't work, because the scripting is already passing `--ignore-file .gitignore` on the command line, and Prettier doesn't support multiple ignore files.)

    Yarn edits the `.yarnrc.yml` file automatically, and its own style rules conflict with Prettier's. The two are in constant tension if Prettier is also making changes. It's best to consider the file's contents as managed by Yarn.

    Additional `# prettier-ignore` comments may be needed as the file grows. The comment only disables Prettier for the single block of code immediately following it.

### DIFF
--- a/.binder/postBuild
+++ b/.binder/postBuild
@@ -3,5 +3,6 @@
 set -e
 
 npm install -g yarn
+yarn set version berry
 
 ./dev-install.sh
--- a/.binder/postBuild
+++ b/.binder/postBuild
@@ -3,6 +3,6 @@
 set -e
 
 npm install -g yarn
-yarn set version berry
+yarn set version stable
 
 ./dev-install.sh
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,7 +1,7 @@
 bqplot
 ipyleaflet
 jupyterlab-myst
-jupyterlab
+jupyterlab==4.0.2
 matplotlib
 networkx
 numpy
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,7 +1,7 @@
 bqplot
 ipyleaflet
 jupyterlab-myst
-jupyterlab~=3.0
+jupyterlab
 matplotlib
 networkx
 numpy
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,10 +1,11 @@
 nodeLinker: node-modules
 
+# prettier-ignore
 packageExtensions:
-  '@jupyterlab/settingregistry@^4':
+  "@jupyterlab/settingregistry@^4":
     dependencies:
       react: ^18
-  '@nrwl/devkit@^15':
+  "@nrwl/devkit@^15":
     dependencies:
       nx: ^15
   postcss-cssnext@^3.1.0:
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -2,9 +2,6 @@ nodeLinker: node-modules
 
 # prettier-ignore
 packageExtensions:
-  "@jupyterlab/settingregistry@^4":
-    dependencies:
-      react: ^18
   "@nrwl/devkit@^15":
     dependencies:
       nx: ^15
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,15 @@
 nodeLinker: node-modules
+
+packageExtensions:
+  '@jupyterlab/settingregistry@^4':
+    dependencies:
+      react: ^18
+  '@nrwl/devkit@^15':
+    dependencies:
+      nx: ^15
+  postcss-cssnext@^3.1.0:
+    dependencies:
+      caniuse-lite: ^1
+  source-map-loader@^4:
+    dependencies:
+      webpack: ^5
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "build:nbextension": "lerna run --stream build:nbextension",
     "build:test": "lerna run build:test --stream --ignore \"@jupyter-widgets/example-*\"",
     "clean": "lerna run --stream clean",
-    "deduplicate": "yarn && yarn-deduplicate -s fewer --fail",
     "docs": "typedoc",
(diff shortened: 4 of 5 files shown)


ferdnyc:
Well, that did nothing. 3.6.4 is already preinstalled, so it's kept as-is.

ferdnyc:
@martinRenou 

Ah! Well, the explicitly not-matching versions of both jupyterlab and yarn seem bad, here...

```text
Checking JupyterLab (assuming JupyterLab >=4)... 3.6.4
Installing and building all yarn packagesyarn install v1.21.1
[1/5] Validating package.json...
error @: The engine "yarn" is incompatible with this module. Expected version ">=3". Got "1.21.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

martinRenou:
Oh yes. It is indeed more subtle and binder may need JupyterLab 4 if it uses a **dev version** of ipywidgets.

ipywidgets now requires to **be built** against JupyterLab 4. Once built and published on PyPi/conda-forge, it will work with both JupyterLab 3 and JupyterLab 4.

Before https://github.com/jupyter-widgets/ipywidgets/pull/3752, ipywidgets required to be built against JupyterLab 3, but that does not mean it did not work with JupyterLab 4 (apart from a couple small issues it is working).

So you're right, we should update the binder setup to use JupyterLab 4 if binder builds its own version from dev.

martinRenou:
I would be in favor of not using a dev version of ipywidgets on Binder though, and instead use the latest published version on either PyPi or conda-forge. 

Because it may be surprising to users to not get the same features in their local setup as in the binder.

ferdnyc:
@martinRenou 

Binder _from pull requests_ should probably build ipywidgets from the PR's source branch, though, no? Otherwise, that automatic comment with a link to a Binder notebook... running ipywidgets without any of the code from the PR... seems not entirely helpful/useful.

.binder/requirements.txt
```diff
@@ -1,7 +1,7 @@
 bqplot
 ipyleaflet
 jupyterlab-myst
-jupyterlab~=3.0
+jupyterlab
```
martinRenou, line 4: Does it work if you specifically pin jupyterlab?

```suggestion
jupyterlab==4.0.2
```

ferdnyc: @martinRenou Honestly, I'm having trouble getting past the yarn 3 requirement.

Even trying _locally_ to build ipywidgets with yarn 3, I can't get past the `yarn install` stage. ...Actually, the whole thing is complicated enough that I'll open a new Issue for it, rather than cluttering this PR.

ferdnyc: I haven't been able to get past the yarn 3 requirement to even _try_, TBH. Rather than clutter this PR with it, I opened #3797 documenting that issue.

ferdnyc: (ISTR I tried something like that, and it _did_ work to set the jupyterlab version, but still complained that yarn wasn't version 3. And upgrading yarn to version 3 causes all of #3797 to happen.)

ferdnyc: So, long story short it turned out my yarn issues were due to my misinterpreting dependency warnings, followed by a silent segfault during the download (Fetch) stage, as dependency **errors**.

I switched to running `yarn` in a Docker container and now it no longer segfaults, so yay for that. Now I'm fighting a bit with `conda` not liking being "activated" from a Dockerfile run, but at least running interactively inside the container all seems well.

ferdnyc: That, along with a `yarn set version berry` line added to `postBuild` to ensure that yarn is at version 3, seems to have fixed things.

.binder/postBuild
```diff
@@ -3,5 +3,6 @@
 set -e
 
 npm install -g yarn
+yarn set version berry
```
ferdnyc, line 6: Actually, it [seems](https://yarnpkg.com/getting-started/install#updating-to-the-latest-versions) this is the more correct invocation.

```suggestion
yarn set version stable
```

ferdnyc (commented):
Speaking of yarn invocations, [apparently](https://yarnpkg.com/features/zero-installs) with Yarn 2+ it's... well, _debatably_ a good idea, but at the very least "not wrong" (according to **them**, anyway) to check into the repo either:

1. The Yarn version in use, so that everyone is running the same tools, or
2. The _entire_ dependency cache, iff it's in the form of compressed package archives rather than a giant, exploded `node_modules` hierarchy. That requires switching to using [their `pnp` linker](https://yarnpkg.com/features/pnp), instead of the `node-modules` one that's the default for projects coming from yarn 1.

Not necessities by any stretch, but apparently options possibly worth considering.

ferdnyc:
The test failure is just a change in font selection, here's the expected:
8748451373021fa6d781e7340c226aa3b04b8212

And the actual:
13803ac72b8a2a5a1ff6f4a05e97b811f37659da

martinRenou (approved):
Thanks! Sorry for the late review

PR #3801, opened 2023-07-10
Fix link to "Output widget examples"

warrickball:
The space in the URL disrupts the Markdown syntax so that the hyperlink isn't rendered properly. This replaces the space with `%20` to fix the link.


martinRenou (approved):
Thanks!

PR #3803, opened 2023-07-15
Gf drag drop

gregfreeman:
This is a revival of  [PR 2720](https://github.com/jupyter-widgets/ipywidgets/pull/2720)..  PR 2720


gregfreeman:
There is a very subtle visual difference in one test case.  

image
  
image

image

PR #3807, opened 2023-07-17
Remove shebang lines from test_* files

frenzymadness:
The files are not executable and there is no code that would make sense to run directly so the shebang lines are not needed.


martinRenou (approved):
Thanks

PR #3811, opened 2023-07-26
Replace ipykernel dependency by the comm dependency

martinRenou:
As discussed in https://github.com/jupyter-widgets/ipywidgets/pull/3749

This will:
- Use the comm implementation from ipykernel if it's available
- Otherwise import the comm implementation from the new comm package (typically the case of xeus-python, or in a test runtime where there is no kernel)

### DIFF
--- a/python/ipywidgets/ipywidgets/comm.py
+++ b/python/ipywidgets/ipywidgets/comm.py
@@ -1,0 +1,33 @@
+# compatibility shim for ipykernel < 6.18
+import sys
+from IPython import get_ipython
+import comm
+
+
+def requires_ipykernel_shim():
+    if "ipykernel" in sys.modules:
+        import ipykernel
+
+        version = ipykernel.version_info
+        return version < (6, 18)
+    else:
+        return False
+
+
+def get_comm_manager():
+    if requires_ipykernel_shim():
+        ip = get_ipython()
+
+        if ip is not None and getattr(ip, "kernel", None) is not None:
+            return get_ipython().kernel.comm_manager
+    else:
+        return comm.get_comm_manager()
+
+
+def create_comm(*args, **kwargs):
+    if requires_ipykernel_shim():
+        from ipykernel.comm import Comm
+
+        return Comm(*args, **kwargs)
+    else:
+        return comm.create_comm(*args, **kwargs)
--- a/python/ipywidgets/ipywidgets/widgets/widget.py
+++ b/python/ipywidgets/ipywidgets/widgets/widget.py
@@ -15,6 +15,7 @@ from traitlets import (
     Any, HasTraits, Unicode, Dict, Instance, List, Int, Set, Bytes, observe, default, Container,
     Undefined)
 from json import loads as jsonloads, dumps as jsondumps
+from .. import comm
 
 from base64 import standard_b64encode
 
@@ -525,16 +526,7 @@ class Widget(LoggingHasTraits):
             if self._model_id is not None:
                 args['comm_id'] = self._model_id
 
-            # ipykernel <6.18 when the comm package did not exist
-            if "ipykernel" in sys.modules and "comm" not in sys.modules:
-                def create_comm(**kwargs):
-                    from ipykernel.comm import Comm
-
-                    return Comm(**kwargs)
-            else:
-                from comm import create_comm
-
-            self.comm = create_comm(**args)
+            self.comm = comm.create_comm(**args)
 
     @observe('comm')
     def _comm_changed(self, change):
--- a/python/ipywidgets/ipywidgets/widgets/widget.py
+++ b/python/ipywidgets/ipywidgets/widgets/widget.py
@@ -6,6 +6,7 @@
 in the Jupyter notebook front-end.
 """
 import os
+import sys
 import typing
 from contextlib import contextmanager
 from collections.abc import Iterable
@@ -524,13 +525,14 @@ class Widget(LoggingHasTraits):
             if self._model_id is not None:
                 args['comm_id'] = self._model_id
 
-            try:
-                from comm import create_comm
-            except ImportError:
+            # ipykernel <6.18 when the comm package did not exist
+            if "ipykernel" in sys.modules and "comm" not in sys.modules:
                 def create_comm(**kwargs):
                     from ipykernel.comm import Comm
 
                     return Comm(**kwargs)
+            else:
+                from comm import create_comm
 
             self.comm = create_comm(**args)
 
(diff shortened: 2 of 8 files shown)


martinRenou:
CI failure is due to the Notebook 7 release

python/ipywidgets/ipywidgets/widgets/widget.py
```diff
@@ -524,13 +525,14 @@ def open(self):
...
-                from comm import create_comm
-            except ImportError:
+            # ipykernel <6.18 when the comm package did not exist
+            if "ipykernel" in sys.modules and "comm" not in sys.modules:
                 def create_comm(**kwargs):
                     from ipykernel.comm import Comm
 
                     return Comm(**kwargs)
+            else:
+                from comm import create_comm
```
SylvainCorlay, line 535: 👍

SylvainCorlay (approved):

maartenbreddels:
I'll test this today!

maartenbreddels:
@martinRenou this was more difficult to get right, the comm module could be imported, and ipykernel count not (yet) be imported (as is the case for solara).

Also, by putting the same API in a shim module, I think it cleaned up the code a bit. What do you think?

python/ipywidgets/ipywidgets/comm.py
```diff
@@ -0,0 +1,33 @@
...
+            return get_ipython().kernel.comm_manager
+    else:
+        return comm.get_comm_manager()
+
+
+def create_comm(*args, **kwargs):
+    if requires_ipykernel_shim():
+        from ipykernel.comm import Comm
+
+        return Comm(*args, **kwargs)
```
martinRenou, line 31: I'd really like to get rid of this some day.

That way we can remove the nasty patches of ipykernel.comm in xeus-python.

martinRenou:
Yes putting it all in a separate file makes sense. Thanks for looking into it!

I need to check that this works nicely with xeus-python.

python/ipywidgets/ipywidgets/comm.py
maartenbreddels: only for old ipykernels right? Do you think that will still happen?

maartenbreddels: I'd like this piece of code to basically stay around forever, so we better get it right! :)
We can also skip this if we can detect xeus-python, or some other condition to make this more robust? Do you make use of a 'fake' ipykernel module in xeus-python?

martinRenou: Only for old ipykernel yes.

> I'd like this piece of code to basically stay around forever

Let's talk again about it in ten years? 😄 

> Do you make use of a 'fake' ipykernel module in xeus-python?

Yeah we mock the ipykernel module https://github.com/jupyter-xeus/xeus-python/blob/main/src/xinterpreter.cpp#L77

martinRenou (commented):
Tested with xeus-python, it works nicely! Thanks @maartenbreddels

python/ipywidgets/ipywidgets/comm.py
maartenbreddels: Happy to discuss this in ten years from now :)

martinRenou:

vincent-grosbois:
hello, with the backport of this to 7.X, it means now that ipywidgets 7.X is not working anymore on python 3.6 (as the comm module will require traitlet >= 5, which require python >= 3.7). Is it intended that ipywidgets 7.X won't support python 3.6 ? thanks

vincent-grosbois:
I don't think I'll have the time soon, but here's another proposal: https://github.com/ipython/comm/issues/15

bollwyvl:
This broke `import ipywidgets` in `jupyterlite-pyodide-kernel` as can be seen in e.g.

https://ipywidgets.readthedocs.io/en/stable/try/lab/index.html?path=Widget%20List.ipynb

```py
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 import ipywidgets as widgets

File /lib/python3.10/site-packages/ipywidgets/__init__.py:54
     51         return
     52     register_comm_target()
---> 54 _handle_ipython()

File /lib/python3.10/site-packages/ipywidgets/__init__.py:52, in _handle_ipython()
     50 if ip is None:
     51     return
---> 52 register_comm_target()

File /lib/python3.10/site-packages/ipywidgets/__init__.py:41, in register_comm_target(kernel)
     39 """Register the jupyter.widget comm target"""
     40 from . import comm
---> 41 comm_manager = comm.get_comm_manager()
     42 if comm_manager is None:
     43     return

File /lib/python3.10/site-packages/ipywidgets/comm.py:18, in get_comm_manager()
     17 def get_comm_manager():
---> 18     if requires_ipykernel_shim():
     19         ip = get_ipython()
     21         if ip is not None and getattr(ip, "kernel", None) is not None:

File /lib/python3.10/site-packages/ipywidgets/comm.py:11, in requires_ipykernel_shim()
      8 if "ipykernel" in sys.modules:
      9     import ipykernel
---> 11     version = ipykernel.version_info
     12     return version < (6, 18)
     13 else:

AttributeError: module 'ipykernel' has no attribute 'version_info'
```

Some ways forward:
- add `version_info` to the [shim](https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/py/ipykernel/ipykernel/__init__.py)
  - last i checked we _couldn't_ use `comm` directly for... some reason, and it has its own comm implementation
  - but less duplication is good
- add some other attribute to check
- add a special case for `'pyodide' in sys.modules` (wouldn't recommend)

As this isn't a part I've monkied with much, would appreciate some insight from those that have, but would happily do the PR.

Also, worrying about python 3.6, and frankly, 3.7, is getting a bit wearisome if it can lead to _any_ problems for folks that are on currently supported pythons.

martinRenou:
We've been discussing this in https://github.com/jupyter-widgets/ipywidgets/issues/3819

> add version_info to the [shim](https://github.com/jupyterlite/pyodide-kernel/blob/main/packages/pyodide-kernel/py/ipykernel/ipykernel/__init__.py)

I think that' the way to go for now, longer term we get rid of the shim and use the comm module only.

> last i checked we couldn't use comm directly for... some reason

I'm surprised, using the comm module to expose the pyolite's implementation of the comms should be almost straightforward.

> Also, worrying about python 3.6, and frankly, 3.7

What do you mean about this? Related, it has been raised that [comm could not be installed on Python 3.6 due to a strict traitlets dependency](https://github.com/ipython/comm/issues/15), it has been fixed in comm 0.1.4 by loosening the traitlets dependency.

martinRenou:
I'll work on a PR to update the shims AND make use of the comm module right now.

PR #3820, opened 2023-08-03
Update to JupyterLite 0.1.0

jtpio:
Changes:

- [x] Update to `jupyterlite >=0.1.0,<0.2.0`
- [x] Install `jupyterlite` from conda forge in the docs environment

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,14 +28,9 @@ dependencies:
   - bqplot
   - ipykernel
   - ipyleaflet
+  - jupyterlite >=0.1.1,<0.2.0
   - matplotlib-base
   - numpy
   - scikit-image
   - scikit-learn
   - sympy
-  # jupyterlite dependencies
-  - doit
-  - pkginfo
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Thanks! Now that xeus-python has stabilized a bit, splitting the huge kernel file into multiple small files easy to cache, performing better at startup, I would be in favor of replacing pyodide-kernel by jupyterlite-xeus-python in the docs. What do you think?

It can also now use a dev version of ipywidgets since you can now pip install stuffs in the emscripten environment. Allowing us to test ipywidgets directly in PRs without spawning binder.

jtpio:
Good point, we could indeed give `jupyterlite-xeus-python` another try, especially since now it's also a bit faster to start.

Out of curiosity I quickly compared the Pyodide and Xeus Python kernels locally and here are some numbers:

**jupyterlite-pyodide-kernel==0.1.0**

image

**jupyterlite-xeus-python==0.9.2**

image

PR #3830, opened 2023-09-11
Reverts #3689 and #3738 back to the original working codebase

martinRenou:
Reverts jupyter-widgets/ipywidgets#3689 and jupyter-widgets/ipywidgets#3738

As discussed in https://github.com/jupyter-widgets/ipywidgets/issues/3735. We should probably revert those in a patch release, then be back with a solution that works nicely later.


ibdafna (approved):

martinRenou:

PR #3862, opened 2023-11-17
Fix wrong tag index in TagsInput when updated via python

s-weigand:
This change fixes a bug in the `TagsInput` when the `value` attribute is updated from `python` with a list of different length than the widget currently has and the `inputIndex` in the frontend not being updated.

This leads to wrong insertions of new values
TagsInput-bug


martinRenou (approved):
Thanks!

Issue #3823, opened 2023-08-12
Widget rendering in JupyterLab4 and JupyterNotebook7


kolibril13:
Today I was investigating how I can store the state of a widget with the latest Jupyter setups.

I wanted to know:
* Can I **save** the widget state?
* Is there an **indicator** that shows that the widget-kernel connection is lost?
* Does the widget render correctly **after restarting** the Jupyter server?
* Does the widget render correctly **on sphinx** with myst-nb?

And here is a summary of my findings:

<img width="2380" alt="image" src="https://github.com/jupyter-widgets/ipywidgets/assets/44469195/e29fa3ef-6b9d-44e1-893e-84454a863e64">

used code:
```py
import ipywidgets
ipywidgets.__version__

ipywidgets.widgets.Button(description="Hello World!")
```

Key findings from the above chart:

1. Jupyter Notebook 6.5.4 is a setup where everything works great and there's the option "*Save Notebook Widget State*".
2.  jlab4 and jn7 have the new option "*Save Widget State Automatically*". I couldn't find out if that actually works or no.
3. Jupyter VS Code cannot save widget states, and that's also not planned https://github.com/microsoft/vscode-jupyter/issues/4404 . 
4. When I shut down the jupyter server and restart it, the jlab4 and nb7 console says **"Error: widget model not found"**.  
I think making widget rendering possible after a notebook restart in the latest Jupyter environments would be a fantastic improvement. This could impact the widget ecosystem in a very positive way, especially since numerous people in the community have reported similar widget model not found errors: 
https://github.com/jupyter-widgets/ipywidgets/issues/3758,
https://github.com/jupyter-widgets/ipywidgets/issues/3773, 
https://github.com/jupyter-widgets/ipywidgets/issues/2628, 
https://github.com/jupyter-widgets/ipywidgets/issues/3183, 
https://github.com/jupyter-widgets/ipywidgets/issues/3790

https://discourse.jupyter.org/t/widgets-not-rendering-after-the-second-time-a-notebook-is-reopened/10631
https://discourse.jupyter.org/t/error-displaying-widget-model-not-found-geemap/11473

jasongrout:
> jlab4 and jn7 have the new option "_Save Widget State Automatically_". I couldn't find out if that actually works or no.

FYI, turning this option on saves the widget state when you save the notebook. Turning the option off does not save the widget state when you save the notebook (and as a consequence, the notebook files are much smaller).

[...]

PR #3866, opened 2023-11-27
fix: restoring widgets from saved state is broken since v8 in Lab

### LINKED ISSUE #15361: Feature Request: Enable ipywidget rendering after notebook restart
Hi there!👋

In Jupyter Notebook 6.5.4, it's possible to render an ipywidget after notebook restart.
It would be great to bring that behavior to JupyterLab and JupyterNotebook 7 as well!

When I currently have a notebook with widget state saved, and try to open it in JupyterLab, the widget is not rendered. 
Instead, it will show the error message:
`Error displaying widget: model not found`
image

# Reproducing this error:
```
python3.11 -m venv .venv && source .venv/bin/activate
pip install jupyterlab==4.0.8
pip install ipywidgets==8.1.1
```
Next, start JupyterLab and create a notebook with this cell:

```py
from ipywidgets import Button
Button(description="Hello World!")
```
Next, save the notebook with the setting "Save Widget State Automatically" enabled.
Next, restart the kernel. Then close the notebook, then reopen it.
Now there will be the message `Error displaying widget`.

I'm on a Mac M1 and I'm using a Python venv. I've also tried some older JupyterLab versions and got the same behavoiur.

For reference: I've reported this issue previously at https://github.com/jupyter-widgets/ipywidgets/issues/3823.

mariobuikhuizen:
In 9d999d7b waiting for sessionContext.ready was removed, this caused _loadNotebook() not to work, because it was called before the widget metadata was available.

This commit restores waiting for sessionContext.ready. This has to be before _loadFromKernel(), otherwise this.kernel is not present, causing an exeption, skipping _loadNotebook().

Fixes: https://github.com/jupyterlab/jupyterlab/issues/15361


mariobuikhuizen:
Also related to #3823

maartenbreddels:
@vidartf and @jasongrout do you remember why this line was removed in https://github.com/jupyter-widgets/ipywidgets/pull/2532 ?

Also, the failure on the visual test is odd, since it's a font that was already used. Could it be that we don't pin a particular library? I do see we use 'ubuntu-latest' which can lead to changes in visual results.

@mariobuikhuizen can you first open an 'empty' PR to see if this is unrelated to this PR?

ibdafna:
I'm re-running the tests to see if we can figure out the visual regression test failure. Once we get to the bottom of that, we should merge this as the saved widget state functionality is important.

ibdafna:
Looking at the failed regression tests, I'm not seeing any material difference between the two rendered items. The failure is triggered by a few changed pixels, but not a real regression.

gutow:
> Looking at the failed regression tests, I'm not seeing any material difference between the two rendered items. The failure is triggered by a few changed pixels, but not a real regression.

If this really is the only problem, I strongly urge this to be pushed ahead quickly. The broken saved widget state is a big deal that prevents me from being willing to work on moving my projects to jupyter lab from nbclassic.

Thanks.

maartenbreddels:
Indeed, this seems unrelated, and it also happens for cell 41, so it's not likely to be a timing issue due to loading CSS slowly. 
Let's do a few restarts to see if it's flakey or consistently failing.

pplonski:
Can I help somehow with this issue, to have it merged?

maartenbreddels:
We also really want to have this released soon. I'm gonna push the visual changes as a separate commit, with an explanation that we do not fully understand why this visual diff happens.

maartenbreddels:
sidenote: images 40 to 44 have changes, but the visual report only shows 1 image at a time. To make sure we really commit small changes if have to push 1 image at a time. I'm not sure if this is a galata setting, since it will speed things up in the future?

krassowski:
> I'm not sure if this is a galata setting, since it will speed things up in the future?

Since you have a number of `expect`s in a loop, you probably want to switch to [`expect.soft`](https://playwright.dev/docs/test-assertions#soft-assertions) here:

https://github.com/jupyter-widgets/ipywidgets/blob/b2531796d414b0970f18050d6819d932417b9953/ui-tests/tests/widgets.test.ts#L41-L44

this will still mark the test as failed, but will run through all of them (and generate images for all).

maartenbreddels:
Leaving this for future reference, in case we want to review this again, image 42:


Actual 42:
image

Expected 42:
image


Diff 42:
image

maartenbreddels:
> this will still mark the test as failed, but will run through all of them (and generate images for all).

Many thanks 🤗
Pushed a commit to this PR, hope it works!

maartenbreddels:
Looking at a recent run of #3880 it seems the visual test are also failing in that PR (I've checked before that other PR's were green, maybe a caching 'luck'). Also, the docs fail in #3880 so I thin we can merge now.

Many thanks to the people who helped with this and the patience of our users; I'll work on a release now!

kolibril13:
Great to hear! Looking forward to testing this new feature!

martinRenou:
Thanks!!

maartenbreddels:
We have the first 2024 release 🥳 
[Release: ipywidgets 8.1.2, widgetsnbextension 4.0.10, jupyterlab_widgets 3.0.10](https://github.com/jupyter-widgets/ipywidgets/commit/4d93d8eff1bd88faf1e8906c50a80718fd57f867)

kolibril13:
Awesome!
I've just tested this with
```
python3.11 -m venv .venv && source .venv/bin/activate
pip install jupyterlab==4.1.0
pip install ipywidgets==8.1.2
```

and Save -> Restart -> Close -> Reopen Works now! 🎉 

https://github.com/jupyter-widgets/ipywidgets/assets/44469195/79c94bf8-4b16-4d49-8b48-c9c2e3d4dedc

I've also tried another order (Restart before save)
Restart -> Save -> Close -> Reopen

That one still gives the error:


https://github.com/jupyter-widgets/ipywidgets/assets/44469195/96ca6a35-4a66-4007-9eaf-797f69709f49

kolibril13:
and here's another observation when I do the following steps:

1. Reopening a notebook with a saved widget state 
2. Not changing anything in the notebook
3. Closing the notebook

it will ask me if I want to save changes before closing. That's unexpected, because I did not make changes.
When I click on "Discard changes", I can simply reopen the notebook again and see the widget.

However, when I click on "Save changes" and reopen the notebook, the error of not rendering the widget occurs again: 



https://github.com/jupyter-widgets/ipywidgets/assets/44469195/8b188aad-545c-49e0-ab93-36f12dc8e678

mariobuikhuizen:
> I've also tried another order (Restart before save)
Restart -> Save -> Close -> Reopen

> That one still gives the error:


Ipywidgets v7 has the same behaviour and this also happens in the classic notebook, so this might be a different issue.

PR #3880, opened 2024-01-19
Docs: specify Jupyterlab (version 3.x or above)

danmarshall:
Current Jupyterlab is 4.x, added clarity to version compat.


martinRenou (approved):
Thanks!

PR #3881, opened 2024-01-30
Update Output Widget.ipynb

tomichec:
Fix: the `itertools` library has not been loaded before being used.


martinRenou (approved):
Thanks!

PR #3893, opened 2024-03-10
weakref to enable gc + setting children with iterable + fix loading state via jupyter.widget.control + fix various tooltips

fleming79:
This PR consists of several enhancements/fixes:

* Use weak reference to enable garbage collection fixing https://github.com/jupyter-widgets/ipywidgets/issues/1345
* A new class `Children` for validating box.children.
* Fixed loading state via jupyter.widget.control. 
* Fixes tooltips: 
    * https://github.com/jupyter-widgets/ipywidgets/issues/3703
    * https://github.com/jupyter-widgets/ipywidgets/issues/3440
    * https://github.com/jupyter-widgets/ipywidgets/issues/3860

No breaking changes are foreseen.

### Garbage collection 

A weak reference is used for the comm message removing internal strong references.

A new function `enable_weakreference` replaces the global `_instances` dict with a weak value dictionary moving any existing mappings. This removes the last strong reference to objects in the `ipywidgets` module enabling automatic garbage collection once the user generated strong references are removed.

The inverse function `disable_weakreference` does the opposite, restoring a normal dictionary and copying existing mappings.

## Box.children

Box.children has new features:
* Can be set with any iterable
* All invalid items and closed widgets are identified. The behavior on detection of invalid items is configurable to either:
    * Raise an error (default behavior), or 
    * log an error and omit invalid widgets.

### Other changes

* Representation of a closed widget is changed to `f'<closed:{self.__class__.__name__}>'`
* Auto loading of comms is improved

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ ui-tests/playwright-report
 **/lite/.cache
 **/*.doit.*
 **/docs/typedoc/
+.yarn/*
--- a/docs/source/dev_release.md
+++ b/docs/source/dev_release.md
@@ -58,11 +58,14 @@ Lerna will prompt you for version numbers for each of the changed npm packages i
 ## Configure twine username
 
 If you have 2FA on, make sure you have your `~/.pypirc` file set to:
+
 ```
 [pypi]
 username = __token__
 ```
+
 Or set the environment variable
+
 ```
 export TWINE_USERNAME=__token__
 ```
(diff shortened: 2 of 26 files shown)


fleming79:
@martinRenou  - thank  you for approving the workflows. I'm a bit unsure about workflows, but it looks like I the lockfile should have been updated? I've recently pushed the lockfile. Any advice would be appreciated.
Regards,
Alan

PR #3894, opened 2024-03-15
fix: get_metadata deprecated since traitlets 4.1

ManonMarchand:
Hello, 

This is a very tiny PR, but I get a deprecation warning raised by this line from the traitlets library when using a custom ipywidget:

 https://github.com/ipython/traitlets/blob/cb672eb5ba235fdfee6407bd7f1b1985de934e7c/traitlets/traitlets.py#L833
 
 Here is a fix that should remove it?
 
 Edit: I explicitly allow anyone to edit anything I might have done wrong, I have no idea why the tests are failing

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/widget.py
+++ b/python/ipywidgets/ipywidgets/widgets/widget.py
@@ -684,7 +684,7 @@ class Widget(LoggingHasTraits):
         """Dynamically add trait attributes to the Widget."""
         super().add_traits(**traits)
         for name, trait in traits.items():
-            if trait.get_metadata('sync'):
+            if 'sync' in trait.metadata:
                 self.keys.append(name)
                 self.send_state(name)
 


dangotbanned:
@ManonMarchand could you push this again please?

I can try to help with CI failures, but the original logs aren't visible anymore. 

Arrived here after planning to submit an identical PR for `altair`

ManonMarchand:
Sure, also rebased on main in case the ci wasn't due to the tiny fix

dangotbanned:
@ManonMarchand looking at the last 4 [merged commits](https://github.com/jupyter-widgets/ipywidgets/commits/main/), I don't think you should expect anything other than the last 4 tests passing 

image

I've tried using that auto binder link https://mybinder.org/v2/gh/ManonMarchand/ipywidgets/fix-get_metadata?urlpath=lab/tree/docs/source/examples/
No idea what the expectation there is, but it spent 45 mins building and then printed "Service Unavailable".

---

I think a maintainer will just need to merge as-is.
The failures are already on main and seems very unlikely to me that your change has anything to do with it.

There is also https://github.com/jupyter-widgets/ipywidgets/pull/3932 which seems to be passing CI, but is yet to be reviewed

martinRenou:
I will try a patch release very soon-ish with this and https://github.com/jupyter-widgets/ipywidgets/pull/3932

martinRenou (approved):
Thanks!

PR #3896, opened 2024-03-28
update image processing example notebok imports and function call

RRosio:
I was running into the error:
```
AttributeError: module 'matplotlib' has no attribute 'image'
```
and I found two workarounds, using `import matplotlib.pyplot as plt` to use the pyplot submodule's `imsave()` or to import the image submodule with `import matplotlib.image as mpimg` to use its `imsave()` function. After some brief research I could not discern any major differences between the two functions so I have gone with the pyplot submodule's function.


martinRenou (approved):
Thanks!

PR #3897, opened 2024-04-01
Update Widget List.ipynb

liamb27:
Example for FloatLogSlider labeled min and max in reverse order; changed it so it is correct.


jtpio (approved):
Thanks!

PR #3903, opened 2024-04-11
Fix compatibility with pytest 8

### LINKED ISSUE #3883: pytest 8 deprecated nose support
With pytest 8 we now get the warning:
```
ERROR ipywidgets/tests/test_embed.py::TestEmbed::test_minimal_html_filename - pytest.PytestRemovedIn8Warning: Support for nose tests is deprecated and will be removed in a future release.
ipywidgets/tests/test_embed.py::TestEmbed::test_minimal_html_filename is using nose-specific method: `teardown(self)`
To remove this warning, rename it to `teardown_method(self)`
See docs: [https://docs.pytest.org/en/stable/deprecations.html#suppor](https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose)
```

We currently fail on warnings. We should probably modify our test and unpin pytest (pinning of < 8 originally done in #3866 )

I think this is a great project for a (new) contributor

frenzymadness:
Resolves: https://github.com/jupyter-widgets/ipywidgets/issues/3883

### DIFF
--- a/python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
+++ b/python/ipywidgets/ipywidgets/widgets/tests/test_interaction.py
@@ -19,8 +19,6 @@ from ipywidgets import (interact, interact_manual, interactive,
 # Utility stuff
 #-----------------------------------------------------------------------------
 
-from .utils import setup, teardown
-
 def f(**kwargs):
     pass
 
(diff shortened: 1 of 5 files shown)


bnavigator:
Thanks, this works for me!

martinRenou (approved):
Thanks!

PR #3918, opened 2024-05-23
Docs: Update jupyterlite

martinRenou:


martinRenou:
Merging as it's a minor docs patch

docs/source/conf.py
```diff
@@ -131,7 +131,7 @@ def setup(app):
 rediraffe_redirects = {
     "typedoc/index": "_static/typedoc/index",
     "try/lab/index": "_static/lab/index",
-    "try/retro/index": "_static/retro/tree/index",
+    "try/notebook/index": "_static/tree/index",
```
jtpio, line 134: Should this be `tree` instead, so it matches the value set above?

martinRenou: Indeed!

docs/environment.yml
```diff
@@ -28,7 +28,7 @@ dependencies:
   - bqplot
   - ipykernel
   - ipyleaflet
-  - jupyterlite >=0.1.1,<0.2.0
+  - jupyterlite >=0.3.0,<0.4.0
```
jtpio, line 31: This should probably be `jupyterlite-core`, since the `jupyterlite` metapackage does not include any kernel anymore: https://jupyterlite.readthedocs.io/en/latest/migration.html#id3

jtpio: And we would need to add a kernel, for example `jupyterlite-pyodide-kernel` or `jupyterlite-xeus`.

PR #3920, opened 2024-05-25
DRAFT: Single widget manager per kernel

fleming79:
This PR is exploring the possibility of a per-kernel widget manager as suggested  by @jasongrout in the comments of the recently merged PR: https://github.com/jupyter-widgets/ipywidgets/pull/3004.

This PR follows on from a recent PR https://github.com/jupyter-widgets/ipywidgets/pull/3893, but could be easily re-based independently.

In it's current form, it can render the same widget in notebook[s] and consoles connected to the same kernel. The only requirement being that the KernelWidgetManager is registered to (instantiated with) the kernel.

image

### How it works

`KernelWidgetManager` was modified to create only one instance per kernel.id.
`WidgetManager` was changed to be an interface for notebooks that swaps the notebook (panel) rendermime factory to use the KernelWidgetManager corresponding to the new kernel as the kernel is changed.

### Other changes

* `WidgetRenderer` was modified to search for the widget manager of a model_id if the widgetManager wasn't specified.
* A timeout was added to ManagerBase.get_model which help avoid 'model not found' errors that occur when the view tries to render before the model promise has been created. 
* The plugin system was also modified to make it easier to create a KernelWidgetManager for a kernel that doesn't have a notebook attached (downstream use).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ ui-tests/playwright-report
 **/lite/.cache
 **/*.doit.*
 **/docs/typedoc/
+.yarn/*
--- a/docs/source/dev_release.md
+++ b/docs/source/dev_release.md
@@ -58,11 +58,14 @@ Lerna will prompt you for version numbers for each of the changed npm packages i
 ## Configure twine username
 
 If you have 2FA on, make sure you have your `~/.pypirc` file set to:
+
 ```
 [pypi]
 username = __token__
 ```
+
 Or set the environment variable
+
 ```
 export TWINE_USERNAME=__token__
 ```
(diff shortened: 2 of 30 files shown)


martinRenou:
Thanks for opening a PR for this! Was this closed on purpose?

fleming79:
Yes and no. 

If there is still the possibility of a per-kernel widget manager I can re-open  or make a new PR with updates.

martinRenou:
I believe we're still open to refactor the widget manager logic to be more "general" and less tight to the Notebook (or the console). #3004 was merged as-is because it's already an improvement of the current state, but Jason's comments in the PR are still relevant.

PR #3921, opened 2024-05-26
weakref to enable gc + setting children with iterable + fix loading state via jupyter.widget.control + fix various tooltips

fleming79:
Replaces https://github.com/jupyter-widgets/ipywidgets/pull/3893.
Commits have been tidied up.


packages/base/src/widget.ts
```diff
@@ -1101,14 +1101,18 @@ export class DOMWidgetView extends WidgetView {
   }
 
   updateTooltip(): void {
-    const title = this.model.get('tooltip');
+    const title = this.tooltip;
     if (!title) {
       this.el.removeAttribute('title');
-    } else if (this.model.get('description').length === 0) {
+    } else if (!this.model.get('description')) {
```
vijayi1, line 1107: currently the widget.tooltip is displayed only if the widget.description is empty. 
(tried various widgets such as text, checkbox, dropdown, on ipywidgets v8.1.5)

can we display the widget.tooltip even if the widget.description is not-empty?

fleming79:
@vijayi1 Thanks for pointing that out. Attached are files for [the latest ](https://github.com/jupyter-widgets/ipywidgets/pull/3921/commits/7fa9cba62169062eb4ef1e1856c55742a3a45c1c).


[files.zip](https://github.com/user-attachments/files/16942774/files.zip)

vijayi1:
@fleming79 , the fix works for me, thanks.
the table below summarizes the behavior with and without the fix for the text, dropdown, checkbox and button widgets.
(note line #3, the behavior was inconsistent across widgets previously. if someone expects nothing to be displayed, with the fix they would have to set the tooltip to an empty string)

image

PR #3923, opened 2024-05-28
Add return to pack_models

gjmooney:
Add a return for `packed` in `pack_models` . @martinRenou


martinRenou (approved):
This fixes a longstanding issue on the widget serialization logic front-end side. It looks like it's not well tasted nor used in custom widgets.

martinRenou:
Thanks @gjmooney !!

PR #3924, opened 2024-06-18
Make tests compatible with Python 3.13

frenzymadness:
Python compiler newly removes indent from docstrings https://github.com/python/cpython/issues/81283


martinRenou (approved):
Thanks!

PR #3932, opened 2024-07-30
HTMLManager: Support both ipywidgets 7 and 8 models

martinRenou:
Currently, the HTML manager brings the jupyter-widgets front-end for ipywidgets 8, disregarding the version stored in the widget model.
This works fine in most cases, as long as the user did not store an ipywidgets 7 model that was making use of removed APIs.

This PR resolves this by lazily loading either the jupyter-widgets front-end for 7 or 8 depending on the version stored in the model. Similar to the approach taken in https://github.com/voila-dashboards/voila/pull/1482

Testing nbconvert output with ipywidgets 7:
Screenshot from 2024-07-30 15-43-30

Testing nbconvert output with ipywidgets 8:
Screenshot from 2024-07-30 15-43-36


trungleduc (approved):

packages/html-manager/package.json
```diff
@@ -38,7 +38,9 @@
     "@fortawesome/fontawesome-free": "^5.12.0",
     "@jupyter-widgets/base": "^6.0.8",
     "@jupyter-widgets/base-manager": "^1.0.9",
+    "@jupyter-widgets/base7": "npm:@jupyter-widgets/base@4.1.6",
     "@jupyter-widgets/controls": "^5.0.9",
+    "@jupyter-widgets/controls7": "npm:@jupyter-widgets/controls@3.1.6",
```
SylvainCorlay, line 43: 👍

SylvainCorlay:
I like this solution. It will solve many user's headaches.

SylvainCorlay (approved):

maartenbreddels:
Nice work! In solara we use a similar trick with the npm alias, but we actually have to build 2 separate managers. I think this is a cleaner solution.

martinRenou:
> Nice work! In solara we use a similar trick with the npm alias, but we actually have to build 2 separate managers. I think this is a cleaner solution.

Oh nice I did not know you tried that in Solara too! Note that we've done something similar in Voila 0.4.x but with just one manager: https://github.com/voila-dashboards/voila/commit/5934c53f5b8dadeb5a7cc4fd82f23f31591f0654, it's a bit different from the approach in this PR as it makes use of the lazy loading approach in the widgets register, so the widgets assets will load only upon request.

I [tried with latest Voila](https://github.com/voila-dashboards/voila/pull/1487) but things are more complicated there for some reason.

maartenbreddels:
In solara (because we run in the same process), we know beforehand at the server what the version of ipywidgets is, which allows for a different strategy.
We also build the final bundle twice using the webpack alias trick:
https://github.com/widgetti/solara/blob/20b2e8012e7683eb38ff62a74a371eb1a052bf41/packages/solara-vuetify-app/webpack.config.js#L95

I don't think you need it here, but both these tricks allow for some fancy building/bundling options.

martinRenou:
> We also build the final bundle twice using the webpack alias trick:
https://github.com/widgetti/solara/blob/20b2e8012e7683eb38ff62a74a371eb1a052bf41/packages/solara-vuetify-app/webpack.config.js#L95

Ah, that trick could probably be useful to me for Voila 0.5.x, I'll have to give it a try. Thanks for sharing!

PR #3958, opened 2024-10-22
Fix lumino and lab packages pinning

martinRenou:
There was a duplication of lumino and lab packages due to:
- some wrong pinning e.g. `"@lumino/messaging": "^1.10.1 || ^2.1",` with `^2.1` that does not exist yet so this was pulling lumino 1 packages
- updating ipywidgets 7 packages thanks to https://github.com/jupyter-widgets/ipywidgets/pull/3955 that allows pulling lumino 2 and lab 4 packages


trungleduc (approved):

fcollonval:
@martinRenou would it be possible to publish a new version with this fix?

PR #3962, opened 2024-10-29
Update lables even without MatJax/TypeSetter

### LINKED ISSUE #3961: Label text content not initialized if MathJax and latexTypesetter not initialized
## Description

In previous versions `5.0.3`, when rendering a Label widget, the text would be renderer.
However in the latest version it is not renderer.

## Reproduce

1. You need to install VS Code
2. Run the following cell in a notebook
```python
import ipywidgets as widgets


widgets.Label(value='Label Widget')
```

## Expected behavior

The label should be rendered

## Context

Sorry, but I do not think the rest is necessary, hence skipping that.
Its pretty easy to see whats going on here, see below

* In VS Code, `window.MathJax` is `undefined`
* In VS Code `latexTypesetter` is also `undefined`
* As a result, the element content is never set.
* 
```typescript
  typeset(element: HTMLElement, text?: string): void {
    this.displayed.then(() => {
      if ((window as any).MathJax?.Hub?.Queue) {
        return typeset(element, text);
      }
      const widget_manager: any = this.model.widget_manager;
      const latexTypesetter = widget_manager._rendermime?.latexTypesetter;
      if (latexTypesetter) {
        if (text !== void 0) {
          element.textContent = text;
        }
        latexTypesetter.typeset(element);
      }
    });
  }
```
**Note:**
The `typeset` function in `utils.ts` will already check whether MathJax is available.
Hence there's no need to test the availability.

```typescript
export function typeset(element: HTMLElement, text?: string): void {
  if (text !== void 0) {
    element.textContent = text;
  }
  if ((window as any).MathJax !== void 0) {
    MathJax!.Hub!.Queue(['Typeset', MathJax.Hub, element]);
  }
}
```

I think the right fix is

```typescript
  typeset(element: HTMLElement, text?: string): void {
    this.displayed.then(() => {
      const widget_manager: any = this.model.widget_manager;
      const latexTypesetter = widget_manager._rendermime?.latexTypesetter;
      if (latexTypesetter) {
        if (text !== void 0) {
          element.textContent = text;
        }
        latexTypesetter.typeset(element);
      } else {
        return typeset(element, text);
      }
    });
  }
```

Previously it used to be `this.displayed.then(() => typeset(element, text));` 
The change was introduced in this commit (basically there's never a fallback if either one is not present)
 https://github.com/jupyter-widgets/ipywidgets/commit/388c782cbb1fe396ddf1b531fbccc315dbc9dacf#diff-5cc2749aa43c522b3c6b9485d1c9f7b93df1536d43c18ca94553f3a55f70b03fL69

DonJayamanne:
Fixes https://github.com/jupyter-widgets/ipywidgets/issues/3961

### DIFF
--- a/packages/controls/src/widget_description.ts
+++ b/packages/controls/src/widget_description.ts
@@ -67,9 +67,6 @@ export class DescriptionView extends DOMWidgetView {
 
   typeset(element: HTMLElement, text?: string): void {
     this.displayed.then(() => {
-      if ((window as any).MathJax?.Hub?.Queue) {
-        return typeset(element, text);
-      }
       const widget_manager: any = this.model.widget_manager;
       const latexTypesetter = widget_manager._rendermime?.latexTypesetter;
       if (latexTypesetter) {
@@ -77,6 +74,8 @@ export class DescriptionView extends DOMWidgetView {
           element.textContent = text;
         }
         latexTypesetter.typeset(element);
+      } else {
+        return typeset(element, text);
       }
     });
   }


martinRenou (approved):
Thanks!

PR #3983, opened 2025-04-07
Update github actions and fix readthedocs

brichet:
The `upload-artifact@v3` and `download-artifact@v3` actions are deprecated and prevents the workflow to run.

This PR updates it to `v4`, and updates also `checkout` action to `v4`.

It also fixes the documentation build by adding the sphinx configuration file in `.readthedocs.yaml` conf.


martinRenou:
Thanks! Maybe, let's drop the Python 3.7 workflows?

martinRenou:
It looks like the snapshots drifted a tiny bit but CI passed. There is no bot to update the snapshots though. We can either increase the comparison threshold to make the tests more reliable or update the snapshots manually.

brichet:
I'll update the snapshots, they are in the artifacts

martinRenou (approved):
Thanks a lot for this!

PR #3987, opened 2025-04-25
add an audio player and recorder custom widget to README.md

pengzhendong:


martinRenou (changes requested):
Thanks for opening a PR 😃 

ipyaudio looks really cool!

Sorry to decline the PR though, but I feel like adding ipyaudio to the "popular widgets libraries" section of the official ipywidgets README is a bit early. There are many widgets libraries out there. There may be a better place for listing ipyaudio, like https://github.com/nicole-brewer/awesome-jupyter-widgets or similar

PR #3988, opened 2025-04-29
Dynamic widgets registry

martinRenou:
Make the widget manager support handling dynamic register of widgets.

This doesn't change any public API, and is backward compatible.

### DIFF
--- a/python/jupyterlab_widgets/src/plugin.ts
+++ b/python/jupyterlab_widgets/src/plugin.ts
@@ -58,8 +58,29 @@ import '@jupyter-widgets/controls/css/widgets-base.css';
 import { KernelMessage } from '@jupyterlab/services';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
 import { ISessionContext } from '@jupyterlab/apputils';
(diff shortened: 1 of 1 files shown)


python/jupyterlab_widgets/src/plugin.ts
```diff
@@ -58,8 +58,31 @@ import '@jupyter-widgets/controls/css/widgets-base.css';
...
+  get widgets(): base.IWidgetRegistryData[] {
+    return [...this._registry];
+  }
+
+  push(data: base.IWidgetRegistryData) {
+    this._registry.push(data);
+    this._registeredWidget.emit(data);
+  }
+
+  get registeredWidget(): ISignal<WidgetRegistry, base.IWidgetRegistryData> {
```
trungleduc, line 73: `widgetRegistered` sounds more like a signal name.

trungleduc (approved):

martinRenou:
Thanks for the review!

jgunstone:
This sound interesting - can you describe the use case?

PR #3989, opened 2025-04-30
Fix CI + remove Python 3.8

martinRenou:
- Remove Python 3.7 and 3.8 classifiers
- Add Python 3.12 and 3.13 classifiers
- Remove Python 3.8 CI run
- Add Python 3.13 CI run
- Modernize jupyterlab_widgets's packaging, removing wheel generation error
- Update visual regression snapshots, now runs with latest jupyterlab


python/jupyterlab_widgets/jupyterlab_widgets/__init__.py
```diff
@@ -5,16 +5,11 @@
...
-    if not src_prefix.exists():
-        src_prefix = Path(sys.prefix) / f'share/jupyter/labextensions/{labext_name}'
-
-    return [{'src': str(src_prefix), 'dest': labext_name}]
+    return [
+        {
+            'src': 'labextension',
+            'dest': '@jupyter-widgets/jupyterlab-manager'
+        }
+    ]
```
martinRenou, line 13: Simplifying this way is closer [to the cookiecutter](https://github.com/jupyter-widgets/widget-ts-cookiecutter/blob/master/%7B%7Bcookiecutter.github_project_name%7D%7D/%7B%7Bcookiecutter.python_package_name%7D%7D/__init__.py#L23) and works well with editable AND non-editable installs.

The old approach was not working well with a non-editable install.

maartenbreddels:
Shall we also remove support of it, we might introduce breaking changes for 3.8 if we don't test it.
I suggest updating pyproject.toml for it.

PR #3998, opened 2025-09-20
Add JupyterCon banner and jupyter colors

choldgraf:
This PR does a couple of things:

- Adds an announcement banner for JupyterCon
- Links a CSS file from the Jupyter documentation that we can use for jupyter colors in our docs


choldgraf:
the build problems seem to be unrelated to this, I think they are redirect errors?

PR #4015, opened 2026-02-16
Fix saving notebook with disconnected widgets losing widget model state

### LINKED ISSUE #3270: Saving a Notebook With Disconnected Widgets Loses Widget Model in ipynb
## Description

Here’s the scenario: I’m in JupyterLab. I create a notebook that displays a widget. I save the notebook, close it, shut down the kernel.

Sometime later I come back and reopen the notebook. The widget reconstitutes itself from the model and displays (albeit with the “disconnected from kernel” icon). So far everything is good. But here’s where the problem starts…

I save the notebook again, close it and shut the kernel down. I don’t even need to change anything. The next time I open it up, the widget tries to reconstitute itself from the model, but instead I get an “Error displaying widget: model not found” message.

Sure enough, I open up the ipynb in a json editor and can see that the cell has an application/vnd.jupyter.widget-view+json output with a model_id, but no corresponding model with that id is to be found in the metadata.

## Reproduce

[Here is a binder link that demonstrates the issue.](https://mybinder.org/v2/gh/genepattern/example-notebooks/HEAD?urlpath=lab/tree/WidgetTest.ipynb)

1. Run the two cells below. Go ahead and make sure that the notebook is trusted and that "Save Widget State Automatically" is checked in the Settings menu.
2. Save this notebook, close it and shut down the kernel.
3. Open this notebook a second time. Don't change a thing. Just make sure that the Text widget renders from the model appropriately. Great!
4. Save the notebook again. Close it and shut down the kernel.
5. Open the notebook a third time. Watch as the cell with the widget displays: "Error displaying widget: model not found".

## Expected behavior

Since the notebook was not edited, I would expect the widget model to still be in the metadata and the widget to still be able to reconstitute itself from the model.

## Context

- ipywidgets version: 7.6.3
- Operating System and version: Linux & Mac
- Browser and version: Firefox & Chrome
- JupyterLab: 3.1.1

pablogsal:
When saving a notebook containing disconnected widgets (restored from saved state but without a live kernel), the widget state was being dropped from notebook metadata. This happened because get_state_sync() only serialized models with comm_live=true, filtering out any widgets that had been restored from saved state without a kernel connection.

On subsequent reopens, the widget output cells still referenced model IDs that no longer existed in the metadata, causing "Error displaying widget: model not found" errors.

The fix removes the comm_live filter from get_state_sync(), so all registered widget models are serialized when saving. This aligns with the base manager's async get_state() method, which already includes all models regardless of comm_live status.

Fixes #3270

### DIFF
--- a/python/jupyterlab_widgets/src/manager.ts
+++ b/python/jupyterlab_widgets/src/manager.ts
@@ -287,9 +287,9 @@ export abstract class LabWidgetManager
   }
 
   /**
-   * Synchronously get the state of the live widgets in the widget manager.
+   * Synchronously get the state of the widget models in the widget manager.
    *
-   * This includes all of the live widget models, and follows the format given in
+   * This includes all of the widget models, and follows the format given in
    * the @jupyter-widgets/schema package.
    *
    * @param options - The options for what state to return.
@@ -298,9 +298,7 @@ export abstract class LabWidgetManager
   get_state_sync(options: IStateOptions = {}): ReadonlyPartialJSONValue {
     const models = [];
     for (const model of this._modelsSync.values()) {
-      if (model.comm_live) {
-        models.push(model);
-      }
+      models.push(model);
     }
     return serialize_state(models, options);
   }


pablogsal:
@krassowski gentle ping.

krassowski:
Thanks for the PR. @jasongrout does it need any other changes around state restoration?

PR #4019, opened 2026-05-28
fix(deps): bump lodash to ^4.18.1 (CVE-2026-4800)

rodrigosf672:
## Summary

- Bumps `lodash` from `^4.17.4` to `^4.18.1` in `packages/base/package.json`
- Addresses High severity arbitrary code injection vulnerability in `_.template` (CVSS 8.6)
- Updates yarn.lock accordingly

## Context

CVE-2026-4800 (SNYK-JS-LODASH-15869625) affects all lodash versions below 4.18.1. While the codebase only uses `lodash/isEqual` and does not call the vulnerable `_.template` function, the outdated resolution triggers High severity findings in downstream security scanners, blocking release gating and compliance checks.

This is a minor version bump with no breaking changes.

## Test plan

- [x] `yarn install` resolves without errors
- [x] `lodash` resolves to ≥4.18.1 in lockfile
- [x] Existing test suite passes


ktaletsk (approved):

packages/base/package.json
```diff
@@ -43,7 +43,7 @@
     "@types/lodash": "^4.14.134",
     "backbone": "1.4.0",
     "jquery": "^3.1.1",
-    "lodash": "^4.17.4"
+    "lodash": "^4.18.1"
```
goelakash, line 46: nit: the CVE fix is applicable to 4.18.0 - any particular reason to specify the patch 4.18.1 ?

goelakash (commented):
Minor comment, but LGTM otherwise

packages/base/package.json
rodrigosf672: Thanks for asking, @goelakash!

`4.18.1` is preferrable instead of `4.18.0` because `4.18.1` is the latest patch release and includes the CVE-2026-4800 fix. 

According to [SNYK-JS-LODASH-15869625](https://security.snyk.io/vuln/SNYK-JS-LODASH-15869625), `4.18.0` was subsequently deprecated after introducing a breaking functionality issue, and their recommendation is to upgrade to `4.18.1` or later. So, additionally, the latter fixed some additional issues on top of the vulnerability fix above, hence why it is prefferrable.

goelakash (approved):

rodrigosf672:
@jasongrout, the failed CI here is likely unrelated to this upgrade. Could you (or another maintainer) please re-trigger it and help me getting this merged? Thank you in advance!

martinRenou:
Thanks for doing this!

> Addresses High severity arbitrary code injection vulnerability in _.template (CVSS 8.6)

We are not using this method here. I guess it's fair to say we're not impacted by this vulnerability then?

(I'm more or less asking if I should make it my top priority to help getting this merged and released.)

rodrigosf672:
> Thanks for doing this!
> 
> > Addresses High severity arbitrary code injection vulnerability in _.template (CVSS 8.6)
> 
> We are not using this method here. I guess it's fair to say we're not impacted by this vulnerability then?
> 
> (I'm more or less asking if I should make it my top priority to help getting this merged and released.)

Since `_.template` isn't used, I'd agree this is _not_ an urgent exploitable issue for `ipywidgets`.

That said, many users rely on the automated security scanning, and keeping the dependencies on patched versions helps avoid unnecessary vulnerability findings downstream.

So if you can help getting this merged and released, that would be really helpful. Thank you!

krassowski:
https://github.com/jupyter-widgets/ipywidgets/pull/4022 fixes CI

yarn.lock
martinRenou: The old lodash version is still there, can you try deduping this?

rodrigosf672: Yes, `yarn dedupe lodash` worked! Just a note that there is [one entry](https://github.com/jupyter-widgets/ipywidgets/pull/4019/changes#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR10732-R10733) that can't be deduped without bumping a dev-only dependency (verdaccio), which is pinned hence why it remains as 4.17...

PR #4021, opened 2026-06-08
Fix Output capture from background threads

MUFFANUJ:
Fixes #4020

Set the current `IPython` parent when entering an `Output` context so output generated in background threads is routed to the widget with ipykernel 7.

PR #4024, opened 2026-07-09
Add Juyter-wide banner to ipywidgets docs

jasongrout:
Add the jupyter.org banner automatically to the docs. See https://github.com/jupyter/jupyter.github.io/issues/868


jasongrout:
Right now on stable we still have a banner advertising JupyterCon from last year. I'll merge this as a very low-risk change, so it can go out with a patch release containing #4021.

PR #4027, opened 2026-08-08
Drop support for JupyterLab 3 and update build to use Jupyter Builder

jasongrout:
A few changes:

* JupyterLab 3 was end-of-life in December 2024, more than 1.5 years ago, so we drop support for it.
* Move to the new Jupyter Builder package, which was recently factored out of JupyterLab
* Update playwright, which seems to render the fonts slightly differently, leading to image snapshot updates
* Add a test that the extension loads in JupyterLab 4.0.x and latest 

Claude Fable 5 helped a lot in this PR. 

With the drop for JupyterLab 3, here are proposed version bumps for the next release:

| Package                               | Current | Proposed  | Bump              | Reason                                                        |
|---------------------------------------|---------|-----------|-------------------|------------------------------------------------------------------|
| `@jupyter-widgets/jupyterlab-manager` | 5.0.15  | **6.0.0** | major             | Drops JupyterLab 3     |
| `jupyterlab_widgets` (PyPI)           | 3.0.16  | **4.0.0** | major             | Tracks the npm package major version bump, JLab 3 users aren't surprised by a minor update dropping support              |
| `ipywidgets` (PyPI)                   | 8.1.8   | **8.2.0** | minor             | No API/protocol change; dependency moves to `jupyterlab_widgets~=4.0`   |
| `widgetsnbextension` (PyPI)           | 4.0.15  | 4.0.16    | patch  | Notebook shim; unaffected by the JLab 3 drop                     |
| `@jupyter-widgets/base`               | 6.0.11  | 6.1.0     | minor             | API unchanged, but build system did; stay in major 6 version so third-party widgets with dependency `^6` singleton declared keep loading |
| `@jupyter-widgets/controls`           | 5.0.12  | 5.1.0     | minor             | Same  |
| `@jupyter-widgets/base-manager`       | 1.0.12  | 1.1.0     | minor             | Same                                                             |
| `@jupyter-widgets/html-manager`       | 1.0.14  | 1.1.0     | minor             | Same                                                             |
| `@jupyter-widgets/output`             | 6.0.11  | 6.1.0     | minor  | Same                                       |
| `@jupyter-widgets/schema`             | 0.5.6   | unchanged | —                 | Widget protocol untouched                                        |


yarn.lock
```diff
@@ -2845,13 +3238,31 @@ __metadata:
...
+
 "@lumino/properties@npm:^2.0.1, @lumino/properties@npm:^2.0.2":
   version: 2.0.2
   resolution: "@lumino/properties@npm:2.0.2"
   checksum: cbe802bd49ced7e13e50b1d89b82e0f03fb44a590c704e6b9343226498b21d8abfe119b024209e79876b4fc0938dbf85e964c6c4cd5bbdd4d7ba41ce0fb69f3f
   languageName: node
   linkType: hard
 
+"@lumino/properties@npm:^2.0.4, @lumino/properties@npm:^2.0.5":
+  version: 2.0.5
```
martinRenou: There seems to be non-conflicting duplications in the yarn.lock. Probably not a big deal, but could you try `yarn dedupe`?

martinRenou (commented):
Thanks @jasongrout for doing this! Looks great. I just have a minor comment

yarn.lock
jasongrout: Agreed! I ran `jlpm dedupe` and pushed the result.

PR #4033, opened 2026-08-17
Have get_model retry multiple times until we get the model from the backend

martinRenou:
Should mitigate https://github.com/jupyter-widgets/ipywidgets/issues/4026

I am not able to reproduce the issue anymore with this, so it seems to fix it


SylvainCorlay (approved):

SylvainCorlay:
This looks good to me, and seems to fix the issue.

I will merge in a few hours unless concerns are raised.

PR #4034, opened 2026-08-18
Only run interact_manual on submit, not on text widget blur

### LINKED ISSUE #3997: Text widgets do not respect on-demand execution
## Description

Clicking away from a text-based widget in an `interact_manual` function triggers an execute event without the button being clicked.

This also occurs if one tabs away from it (or shift-tabs, etc.)

## Reproduce

```python
import ipywidgets as widgets

@widgets.interact_manual(x='y')
def f(x):
    return x
```

## Context

- Clean environment with ipywidgets version 8.1.7 
- Windows 11
- VSCode Jupyter extension

ghostiee-11:
Fixes #3997. In manual mode `interactive` observed `Text.value`, which the front end also commits on blur, so clicking or tabbing out of a text box ran the function; it now listens for the submit event (enter) instead.

`TextView.handleKeypress` also sent the submit event before the DOM `change` event that commits the value, so the kernel could see a submission with a stale value; the value is now committed first.


ghostiee-11:
@martinRenou / @jasongrout, Can you please review this :)

PR #4035, opened 2026-08-18
Revamp lint infrastructure

jasongrout:
We replace husky with pre-commit and centralize the lint config, which expands coverage slightly to fix a few more files. We also make sure the lint CI runs the same pre-commit hooks so the CI and local linting agree.

jupyterlab-widgets had a few more rules in its lint config, so we special-case those in our central config. We don't apply them everywhere right now since it would introduce a lot of fixes in other packages.

I guided Claude Fable 5 through this work and tested it.


martinRenou (approved):
Thanks! LGTM

